PR #2, opened 2023-10-25
Try out making surface codes in plaquette

vtomole:


surface_codes.ipynb
```diff
@@ -0,0 +1,241 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "5c23fb7b-4057-4de4-b907-a8e9b778a7e1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Is the size of the code the distance? Looks like \"Yes\"\n",
```
perlinm, line 96: This is true for the surface code, but not true in general.

vtomole:
Stale.

PR #4, opened 2023-10-30
TQEC class with Stim code ( Test version )

younghunk1995:
Uploading QEC examples ( Repetition code, Surface code )


vtomole:
I think it would be good if others are able to see this. I'll re-open and turn it into draft.

vtomole:
Stale.

PR #7, opened 2023-11-30
Building templates

nelimee:
The solution using templates, still a **work in progress** for the moment.

Multiple improvements can be done, among which:

1. better documentation,
2. better plaquette index input,
3. small refactor of the internals to only use corner positioning,
4. validity verification,
5. [maybe] move the code one directory up to avoid the indirection with `python_prototype`,
6. more...


nelimee:
Currently there is:
- [main] the whole repository is a python package. 
- [this branch] that hosts the main python package in a folder `python_directory`.

What is the best way going forward? I am asking this because I do not know how the file organisation will go with the javascript files from UI. It might be easier to split the Python project and the Javascript one, I let you tell me which option is the best.

Also, my approach is not the "non_template", so I will likely create a new package.

smburdick:
Is this still going to be merged? Is it superseded by #14?

nelimee:
It is superseded by #14, closing.

PR #9, opened 2023-12-06
[WIP] Plaquette service API

Gistbatch:
I've created an initial draft of how I imagine the API between UI and our business logic.

The idea is the following:
- The user(interface) only interacts with the API when creating plaquettes
- The system itself doesn't need to know about the plaquette details; the `index` is enough
- We can easily exchange the internals of the plaquettes once we know what the data structure looks like.

I've added some "validation" logic, i.e., checking if the stabilizers commute (this does not guarantee that they alternate), and a draft to check the circuits.

Let me know if you think this approach is sufficient or what is lacking.

### DIFF
--- a/python_prototype/src/tqec/plaquettes/__init__.py
+++ b/python_prototype/src/tqec/plaquettes/__init__.py
@@ -1,0 +1,4 @@
+"""Functionality for plaquette management."""
+from .api import PlaquetteAPI
+from .plaquettes import Plaquette, Stabilizer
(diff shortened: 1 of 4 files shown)


smburdick:
I think this depends on the decision of how to structure the backend service/communication -- sounds like we're going to use Flask

vtomole:
>  sounds like we're going to use Flask

Yep. This PR is out of date.

PR #11, opened 2023-12-08
[WIP] Vue draft

Gistbatch:
Vue.js based prototype for the TQEC UI


rryoung98:
Thanks @vtomole, 

First off, using a canvas is correct. I like how you added a circuit! I'm it'll be helpful later done the road for the local quantum circuits. 

I think the way we should do it is with react + pixi, given that Vue requeres a mix of html + javascript understanding which can be a confusing for the python community. FYI, I've tweaked the typical software architecture when using react and pixi. The typical approach is to use [`react-pixi`](https://pixijs.io/pixi-react/) and pixi containers and graphics as functional components, but this would be a high learning curve for OOP from python, so I have just attached a pixi.js application. 
With the pixi.js application someone contributing should be able to create draggable highly interactive shapes, and interactive buttons, grids, sliders etc with just pure javascript which should be pretty manageable.

In my [pull request](https://github.com/QCHackers/tqec/pull/12) i have already developed some drag and drop plaquettes (more like polygons) and it's not that difficult to make them with pixi. I hope to make them snap and check the qubit orderings in the next week or so!

Help appreciated. Please let me know your guys's thoughts as well.

Gistbatch:
@rryoung98 ok, then let's switch to react + pixie.
I used Vue because I have prior experience with it, but I can't argue why it's better. I'll switch to your repro and try to recreate the circuit component there.

vtomole:
> ok, then let's switch to react + pixie.

I'm going to close based on the above comment :+1:

PR #12, opened 2023-12-11
[WIP] first try tqec visualizer 2023.12.10

rryoung98:
This is still a WIP. 

I have tried to attempted steps 1 ~ 3; I will keep working on it as parts of the code still need to be fixed. Please note that there are some issues as of 12/11/2023:

1. I don't know what the infinite qubit tile/array structure should be. So i have just recursively created black dots for each of the vertices after the user clicks the button to `create the qubit Array`.
2. There are some issues with the state updates for building a plaquette. The plaquettes do not show up when different vertices are clicked, but rather when a user clicks the same vertices. Should be able to fix this by wednesday. 
3. I would like to understand if tasks 6 ~ 9 help with building the logic for the plauqette.

TODO:
Refactor plaquette.
Add colors
Add rotating the plaquette (with qubit ordering logic included)

https://github.com/QCHackers/tqec/assets/32727721/bbe93a5b-3e36-4960-b981-bed3062b71d0

Those are my thoughts for now.


nelimee:
Hi @rryoung98 !

I very much like what you did :) My thoughts:

1. I agree with your 1., doing it this way should be enough for the moment. The need for the user to click on a "create the qubit array" is not an issue in my opinion.
2. Your plaquettes are not large enough. In theory, the round plaquette should have 1 qubit at both of its corner and 1 qubit at the tip of the round-shaped part. The square plaquette should have 1 qubit at each corner and 1 qubit at middle.
3. For future developments, it would be nice to have a "stick to qubit" behaviour. Basically, the plaquette sticks to places it can be on. This requires to know, for each plaquette, where the qubits are supposed to be and to perform more checks, but this would also greatly improve the user experience in my opinion.
4. I can explain you steps 6-9 on a call. Contact me directly by mail if you are interested, and we will schedule a meeting.

rryoung98:
@nelimee  Thanks for the points!

2. Absolutely, what i have is a little glitchy, and to get the plaquettes to show up in the dashboard. I have the user click the same qubit right now which is a bug. 

in this PR I intend to adjust the data structures and code to accommodate for the plaquetttes to:
**MVP (mostly foundational UI/UX):**
    a. be created at the right size as you mentioned
    b. show up only as the user clicks on different qubits and stores the correct qubit ordering. (intended to do so in this PR, and will make adjustments)
    c. be rotatable, as a shape for now, later with qubit orderings in mind.
    d. adjust colors
    e. love the 3. idea of "snapping to qubits" will try and make that happen.
    
    
**ALPHA (if i have time, more quantum required):**
    a. check if the qubits are hierarchical with adjacent plaquettes
    b. be able to "fill" with a drag and extend  action with a certain plaquette (similar to filling in a column or row in an  excel sheet)



Please let me know your thoughts on the above. 

4. Happy to hop on a call! @nelimee

afowler:
"check if the qubits are hierarchical with adjacent plaquettes"

Could you elaborate on what the above means?

Best,
Austin.

rryoung98:
Yes @afowler , sorry i think hierarchical was the wrong term. I'll adjust it after your response. 

What i meant to say was preserving the strict ordering with neighboring stabilizers to avoid entanglement. Currently the above shapes don't have a notion of keeping track of which vertex is associated with which data qubit. They are just shapes.

afowler:
Thanks for clarifying. If each corner of the shape is associated with a
specific qubit and contains information on when that qubit is touched, then
the right term is local ordering to ensure that every pair of plaquettes
that touch have the property that one or the other touches all of its
common qubits before the other.

The only exception to this I am aware of is if neighboring plaquettes are
permitted to contain copies of exactly the same gate, which may then
overlap. In a sea of measurements, for example, one could imagine building
a plaquette with measurements in all four corners, but when tiled you would
then have overlapping identical gates. It's a design choice as to whether
you would permit this.

Best,
Austin.

rryoung98:
The features have been stripped back and I've cleaned the code base up to the point where I feel that individuals can contribute in various parts. 

1. A qubitclass has been created. The qubit class will store information such as the qubits type (measurement, data). It extends the `Graphics` class of Pixi.graphics allowing it to be a circular visual element that is editable if necessary.
2. The plaquette "control panel" has been tabled for now. I want to get a sense of how someone would use this on a day to day basis before I build additional features.
スクリーンショット 2023-12-13 午前4 09 32
3. The `PlaquetteClass` has been developed. Currently the plaquette class will take in a list of qubits and create a Plaquette that is draggable and rotatable. Size has been adjusted as well.  It does not snap to the grid yet.
4. The code should be overall much cleaner and easier to understand, even if coming from a python background.

Gistbatch:
Quick update on my progress: I've tried adding the circuit component in your style, but React is giving me headaches.

https://github.com/QCHackers/tqec/assets/18579069/d782ec7f-de30-4f17-bc5e-6f3255118fb1

I'll try to see if I can get the cx gates working and snapping to the lines is also something I'll try

rryoung98:
ah @Gistbatch I am curious as to what issues you encountered. 

If it helps I will be at the meeting today show that the visualizer should require little to no react code.

afowler:
Hi Ricky,

Do you absolutely need to label qubits as data or measurement? The roles
can change, so it would be good not to denote them as such if possible.
Numbered qubits with relative coordinates, all subjected to a circuit, with
a detector constructed out of any subset of local and non-local
measurements, each specified with relative coordinates that could include a
negative time offset to access older measurements in previous layers.

Best,
Austin.

giangiac:
@rryoung98 I would like to check out and test this project.
First time using JS, React, ...
Could you share the basic steps to install the dependencies and this project? How do I run it?

Gistbatch:
@giangiac do the following:

1. Install [nodejs ](https://nodejs.org/en) which includes a package manager `npm`
2. navigate to the tqecjs directory. In there is the `package.json`, which acts similarly to a `requirements.txt`
3. Install project dependencies with `npm install`
4. Start the nodjs development server with `npm start`
5. Open the web app in a WebGL-compatible browser at `localhost:3000`

rryoung98:
> Hi Ricky, Do you absolutely need to label qubits as data or measurement? The roles can change, so it would be good not to denote them as such if possible. Numbered qubits with relative coordinates, all subjected to a circuit, with a detector constructed out of any subset of local and non-local measurements, each specified with relative coordinates that could include a negative time offset to access older measurements in previous layers. Best, Austin.
> […](#)
It's very flexible. Can make the changes, thank you for clarifying

Gistbatch:
> It's very flexible. Can make the changes, thank you for clarifying
This should also be part of what I'm doing. My idea was that once you define a circuit for a plaquette, all the qubits with a measurement are automatically `measurement` qubits.

afowler:
My question is why a measurement, which is an operation occurring on a
qubit, needs to label the qubit as a measurement qubit. If the label is
necessary for some reason, it should be temporary as the role can be
different in subsequent rounds. Also, it's not obvious that measuring all
of the data qubits in a surface code should trigger labelling them as
measurement qubits.

This is a minor point so long as we are all clear that any qubit can be
measured and the set of qubits measured can differ from round to round.

rryoung98:
@afowler  The qubitType can be changed to be any string and used for to help determine which qubit should not be included as a vertex when creating the convex hull. However, it's not difficult to change and I don't think I used it if at all.

rryoung98:
Hi all, ive made some edits to the UI side which is now able to make a convex hull out of any qubits on the grid. 

Further the plaquette class has been updated to change colors, rotate, and to clear itself (though the clearing functionality has some issues). 

It has not quantum components yet, but i believe that the UI parts are stabilizing enough to start discussing the interconnect between the python code and javascript code. I've messaged Adrien about to discuss meeting and discussing this aspect.

https://drive.google.com/file/d/1SbnnR_d1aQHjE0o7HcjuajHxt_aLOm8Q/view?usp=sharing

smburdick:
Looks good so far, Ricky. Do you need a code reviewer?

rryoung98:
yea that would be helpful @smburdick :)

tqecjs/src/workspace/components/button.js
```diff
@@ -0,0 +1,50 @@
...
+		buttonText.x - buttonText.getBounds().width / 2 - 15,
+		buttonText.y - buttonText.getBounds().height / 2 - 15,
+		buttonText.getBounds().width + 30,
+		buttonText.getBounds().height + 30
+	);
+
+	buttonBackground.endFill();
+
+	// Add effects
+	button.interactive = true;
```
smburdick, line 31: This causes a deprecation warning, I think it is safe to remove.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+		if (val === 0) return 0; // Collinear
+		return val; // Clockwise  < 0 or counterclockwise >0
+	};
+
+	createPlaquette = () => {
+		// The graphic should connect the points from the qubits
+		const nQubits = this.qubits.length;
+		if (nQubits < 3) {
+			console.log('Plaquette must have at least 3 qubits');
+			this.plaquetteMade = false;
```
smburdick, line 181: Would it make sense to throw an exception here?

tqecjs/src/workspace/QubitClass.js
```diff
@@ -0,0 +1,130 @@
...
+		this.cursor = 'pointer';
+		this.globalX = x;
+		this.globalY = y;
+		this._createCircle(x, y, radius, color);
+		this.maxNeighborDist = 2 * this.gridSize;
+		this.neighbors = [];
+		this.gridSize = gridSize;
+		// Adjacent (degree 1) qubits
+		// QC properties
+		this.quantumState = '0'; // NOTE: In javascript, becareful updating values such as "this", "state", etc.
```
smburdick, line 25: Can `quantumState` be anything other than 0 or 1? My intuition is yes, but I'm unsure of the representation here

tqecjs/src/workspace/QubitClass.js
```diff
@@ -0,0 +1,130 @@
...
+
+		// Add hover event
+		this.on('pointerover', this._onPointerOver);
+		this.on('pointerout', this._onPointerOut);
+	}
+
+	checkHitArea(eventX, eventY, threshold = 5) {
+		// Calculate the distance between event coordinates and qubit's global position
+		const distance = Math.sqrt(
+			Math.pow(eventX - this.globalX, 2) + Math.pow(eventY - this.globalY, 2)
```
smburdick, line 56: Might be good to have a general `distance()` function

tqecjs/src/workspace/QubitClass.js
```diff
@@ -0,0 +1,130 @@
...
+		);
+		// Define a threshold to determine the hit area
+		if (distance <= threshold) {
+			// Create a text element
+			if (this.children.length > 0) return true; // If there is already a text element, don't create another one
+			const text = new Text(`Qubit:(${this.globalX},${this.globalY})`, {
```
smburdick, line 62: Since we are in the qubit class, might we benefit from a `toString()` function? I thought I saw something along those lines discussed earlier

tqecjs/src/workspace/components/button.js
```diff
@@ -0,0 +1,50 @@
...
+	// Add hover event
+	button.on('pointerover', () => {
+		buttonBackground.alpha = 0.5;
+	});
+	button.on('pointerout', () => {
+		buttonBackground.alpha = 1;
+	});
+
+	// Add shadow
+	buttonBackground.filters = [new DropShadowFilter()];
```
smburdick, line 42: This also creates a warning, but I'm unsure of an alternative
```
deprecation.ts:36 PixiJS Deprecation Warning: settings.FILTER_RESOLUTION is deprecated, use Filter.defaultResolutionDeprecated since v7.1.0
deprecation.ts:44     at ./node_modules/@pixi/filter-drop-shadow/dist/filter-drop-shadow.mjs (http://localhost:3000/static/js/bundle.js:87183:64)
```

tqecjs/src/workspace/components/notifier.js
```diff
@@ -0,0 +1,54 @@
...
+	// Create a container for the notification
+	const notificationContainer = new Container();
+
+	// Create a text object for the notification text
+	const notificationText = new Text(text, {
+		fontSize: 15,
+		fill: 0xffffff,
+		fontWeight: 'bold',
+	});
+	notificationText.anchor.set(0.5);
```
smburdick, line 23: Will anchors be preferred over pivots in this project?

tqecjs/src/workspace/index.js
```diff
@@ -0,0 +1,93 @@
...
+	for (let x = 0; x <= app.renderer.width; x += gridSize) {
+		for (let y = 0; y <= app.renderer.height; y += gridSize) {
+			// Skip every other qubit
+			if (x % (gridSize * 2) === 0 && y % (gridSize * 2) === 0) continue;
+			if (x % (gridSize * 2) === gridSize && y % (gridSize * 2) === gridSize)
+				continue;
+			// Create a qubit
+			const qubit = new Qubit(x, y, 5, gridSize);
+			// Name the qubit according to its position
+			qubit.name = `${x}_${y}`;
```
smburdick, line 31: I think this was discussed earlier. We could have a name() function (like toString) that can return an absolute grid position, or relative to its neighbors in the plaquette. Might be worth discussing in our next meeting.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+		this.colorButton = button('Change color', 200, 350, 'green');
+		this.makeRotatable(); // Add the rotate button to screen
+		this.clearPlaquette(); // Add the clearbutton to the screen
+		this.changeColorButton(); // Add the change color button to the screen
+		//this.changeColorButton(); // Addh the color change button
+		// QC properties
+		this.qubits = qubits || []; // We assume that the list is the order of the qubits
+		this.qubitStack = []; // The stack of qubits that form the convex hull
+		this.createPlaquette(); // Create the plaquette
+		this.quantumCircuit = null; // Might be better to have a quantum circuit class
```
smburdick, line 30: Would it make sense to use something like Q.js?

https://quantumjavascript.app/Q-Circuit.html

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+		this.on('pointermove', this.onDragMove);
+		this.on('pointerup', this.onDragEnd);
+	};
+	onDragStart(event) {
+		this.isDragging = true;
+		this.initialPosition = event.data.getLocalPosition(this.parent);
+		console.log(this.initialPosition);
+	}
+
+	onDragMove = (event) => {
```
smburdick, line 213: Is this supposed to copy one plaquette at a time?

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+		this.isDragging = false;
+		this.initialPosition = null;
+	}
+
+	toggleCtrlButtons = () => {
+		let lastClickTime = 0;
+		this.on('click', (e) => {
+			const currentTime = Date.now();
+			if (currentTime - lastClickTime < 300) {
+				// Set the alpha of the plaquette to 0.8 to indicate
```
smburdick, line 327: Indicate that it's selected?

tqecjs/src/workspace/components/button.js
rryoung98: yea good point can do

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: I think it would make more sense to use the notifier and tell the user to select another qubit

tqecjs/src/workspace/QubitClass.js
rryoung98: unused at the moment. it's just a dummy variable

tqecjs/src/workspace/QubitClass.js
rryoung98: sg

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: yea i still need to implement an alpha change and definitely will. just am still thinking about where to put new plaquettes in with respect to the pixi stage.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: yes

tqecjs/src/workspace/index.js
rryoung98: it would have to be a different name than name() since name is already a built in graphic property. ill think of something that works with your suggestion :)

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
smburdick: I have been able to drag a single plaquette and copy two qubits. I'll play around with it more, maybe capture a video if possible.

smburdick: https://clipchamp.com/watch/NNNNwXJl3lF

This video shows what I mean. Is that intentional?

tqecjs/src/workspace/components/button.js
```diff
@@ -0,0 +1,50 @@
+import { Text, Container, Graphics } from 'pixi.js';
+import { DropShadowFilter } from '@pixi/filter-drop-shadow';
+
+export const button = (text, x, y, color = 0x000000) => {
+	// Create the button container
+	const button = new Container();
+	// Create the button text
+	const buttonText = new Text(text, {
+		fontFamily: 'Arial',
+		fontSize: 15,
```
smburdick, line 10: 15 looks like it should be a constant (or variable, for button and label scaling)

tqecjs/src/workspace/components/notifier.js
```diff
@@ -0,0 +1,54 @@
...
+		notificationText.getBounds().width + 30,
+		notificationText.getBounds().height + 30
+	);
+	background.endFill();
+	notificationContainer.addChild(background);
+
+	// Position the notification at the top center of the stage
+	notificationContainer.position.set(
+		app.screen.width / 2 - notificationContainer.width / 2,
+		150
```
smburdick, line 42: I noticed more multiples of 15 in this file as well.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+		this.gridSize = gridSize;
+		this.gridOffsetX = this.gridSize;
+		this.gridOffsetY = this.gridSize;
+		this.isDragging = false;
+		this.plaquetteMade = false;
+		this.showButton = false;
+		this.controlPanel = new Container();
+		this.controlPanel.name = 'control_panel';
+		this.controlPanel.visible = false;
+		this.rotateButton = button('Rotate', 200, 200, 0x000000);
```
smburdick, line 19: or `'black'` but I am bad at remembering hex colors ;-)

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+					const newlq = shiftQ.neighbors.find(
+						(qubit) =>
+							qubit.globalX < shiftQ.globalX && qubit.globalY === shiftQ.globalY
+					);
+					diff = newlq.globalX - shiftQ.globalX;
+				}
+				// Shift the qubits by the difference
+				for (const qubit of this.qubits) {
+					const q = qubit.neighbors.find(
+						(q) =>
```
smburdick, line 253: You can add `// eslint-disable-next-line` here to suppress a warning about `diff`

Apparently ESlint dislikes this pattern -- https://stackoverflow.com/questions/58816244/debugging-eslint-warning-function-declared-in-a-loop-contains-unsafe-reference

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+			y: qubit.globalY,
+		}));
+
+		this.beginFill(this.color);
+		// Fill the convex hull
+		this.drawPolygon(qubitPos);
+		this.cursor = 'pointer';
+		this.endFill();
+		// Assign the qubit stack to the plaquette
+		this.qubitStack = qubitStack;
```
smburdick, line 163: Why don't we perform a sort here, so we don't have to in the getLeftmost, rightmost, etc. functions above every time they are called.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+	};
+
+	_ccw = (p, q, r) => {
+		// Check if the points are counterclockwise
```
smburdick, line 167: This is if q and r are counterclockwise with p as the origin, correct?

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+		const v1 = { x: q.globalX - p.globalX, y: q.globalY - p.globalY };
+		const v2 = { x: r.globalX - q.globalX, y: r.globalY - q.globalY };
+		// Take the cross product
+		const val = -(v1.x * v2.y - v1.y * v2.x); // If val > 0, then counterclockwise, else clockwise
+		if (val === 0) return 0; // Collinear
+		return val; // Clockwise  < 0 or counterclockwise >0
```
smburdick, line 173: return a boolean instead?

tqecjs/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
smburdick, line 3: "TQEC Visualizer"

tqecjs/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>React App</title>
```
smburdick, line 27: TQEC Visualizer?

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
```diff
@@ -0,0 +1,370 @@
...
+	};
+
+	createPlaquette = () => {
+		// The graphic should connect the points from the qubits
+		const nQubits = this.qubits.length;
+		if (nQubits < 3) {
+			console.log('Plaquette must have at least 3 qubits');
```
smburdick, line 180: Use `alert()` instead? Or is that the job of the Notification class?

tqecjs/src/workspace/TileClass.js
```diff
@@ -0,0 +1,46 @@
...
+	makeTile() {
+		// Get the plaquette
+		if (this.plaquette) {
+			this.container.addChild(this.plaquette.onDragMove());
+		} else {
+			notification(this.container, this.app);
+		}
+	}
+
+	// Create the plaquettes that are
```
smburdick, line 32: ?

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
smburdick: I've also noticed that when I click and drag to create a new tile, a new tile will pop up elsewhere on the tileset.

smburdick (approved):
I think this is a great foundation for us to build upon.

Once this is merged, I can submit my own PR implementing some of the suggestions I've made.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: Yes it's the job of the notification class, thanks for catching that

tqecjs/src/workspace/components/button.js
rryoung98: Done

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: Done, may need some testing and formatting

tqecjs/src/workspace/QubitClass.js
rryoung98: removed

tqecjs/src/workspace/QubitClass.js
rryoung98: the code will probably require another round of refactoring and ill leave it for it to be done then

tqecjs/src/workspace/QubitClass.js
rryoung98: toString() is a built in js function, do you perhaps want to use the name property of the qubit?

tqecjs/src/workspace/components/button.js
rryoung98: let's just keep this for now

tqecjs/src/workspace/components/notifier.js
rryoung98: it's just a notification text so I wouldn't think too hard at the moment.

rryoung98: we can discuss this once other aspects of the code base stabilizes :)

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: Updated all hexes to use literal name

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: It's the cross product of the three points

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: done

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: can you elaborate on what you would be sorting the qubitStack array? :) thanks

tqecjs/src/workspace/components/notifier.js
rryoung98: This needs to be generalized to any screen width and will probably need to be refactored

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
rryoung98: yea that's a bug that needs to be fixed

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
smburdick: If the order of `qubitStack` is irrelevant, outside of those methods, we can keep it sorted so that we don't have to sort again when we want the leftmost/rightmost/etc. qubits. May provide a bit of a performance/simplicity boost.

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
smburdick: I'll go ahead and create an issue in this repo and mark this convo resolved

tqecjs/src/workspace/QubitClass.js
smburdick: Sure

tqecjs/src/workspace/plaquettes/PlaquetteClass.js
smburdick: I believe that specific annotation needs to be on the line with `(q) =>`

Issue #13, opened 2024-01-02
Creating new plaquettes by copy-drag spawns extra plaquettes


smburdick:
When I copy a plaquette by dragging, sometimes an extra plaquette appears, usually next to the new one, but sometimes on a different part of the tileset. [Example video](https://clipchamp.com/watch/NNNNwXJl3lF)

Mentioned in #12.

smburdick:
I've been testing the latest commit (macOS / trackpad) and cannot copy-drag. Will post a video when that uploads to GDrive.

smburdick:
For the above, need to set `this.eventMode = 'dynamic';` in the plaquette class. Will submit PR

smburdick:
Rather than dragging by copy, it might be easier (both to use, and debug/program) to copy through a menu option.

1. Double click and select "Copy"
2. Mouse the cursor to where you want to place the plaquette. Placing can be disallowed, until a legal spot is selected
3. Click to copy the plaquette

rryoung98:
Yes I was thinking of adding arrow keys @smburdick  looking forward to take a look at your PR :)

smburdick:
<img width="859" alt="image" src="https://github.com/QCHackers/tqec/assets/16883876/3bf2c59d-b134-4376-b21a-238e128ae62f">
There are many additional logs statements that shouldn't be there just by mousing over the plaquettes after copying them.

smburdick:
I believe this may be the culprit (in `makeExtensible()`)
```
		this.on('pointermove', this.onDragMove);
```

PR #17, opened 2024-01-04
Make plaquettes interactive, minor cleanups

smburdick:
Per my comment #13, that plaquettes cannot be copied. Adding in interactivity.

Adding additional minor improvements, particularly around colors.

### DIFF
--- a/tqecjs/src/App.js
+++ b/tqecjs/src/App.js
@@ -2,7 +2,9 @@ import './App.css';
 import { Stage } from '@pixi/react';
 import PixiApp from './workspace';
 
-const tqecBlue = 0x1099bb;
+var stageOptions = {
+	backgroundColor:0x1099bb
+}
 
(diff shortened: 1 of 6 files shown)


tqecjs/src/App.js
```diff
@@ -2,6 +2,8 @@ import './App.css';
 import { Stage } from '@pixi/react';
 import PixiApp from './workspace';
 
+const tqecBlue = 0x1099bb;
```
rryoung98, line 5: Maybe we can make css files for this? I

rryoung98 (approved):

tqecjs/src/App.js
smburdick: Agreed. The only real snag is that the React Stage expects the `options` element as a JavaScript object, not a CSS file.

Opening an issue to implement this later on.

PR #26, opened 2024-01-05
Create webpack.yml

smburdick:
This can be the start for deploying the frontend.


rryoung98:
Thanks @smburdick, we use create-react-app which uses webpack under the hood, so it's actually not necessary to add any additional webpack configuration.

Issue #24, opened 2024-01-05
Cannot delete cloned plaquette, _geometry becomes null pointer


smburdick:
https://github.com/QCHackers/tqec/assets/16883876/814b91fc-2f7f-4cd7-8941-85fb9c377238

rryoung98:
#29 is a duplicate so closing

rryoung98:
In the initalizeClearButton() this.destroy() is called which is dangerous.

If the subchild plaquette does not have graphic associated with itself, there is a possibility that the clearing functionality will crash the app.

TypeError: this._geometry is null

smburdick:
I saw that error too when testing the app. I will self assign this and start looking for reworks.

smburdick:
Working in this branch: https://github.com/smburdick/tqec/tree/delete-cloned-plaquettes

I am able to delete plaquettes that I want. However, when I deselect and re-select a plaquette, and try to delete it, I get the error (paste -- https://pastebin.com/8eiL5tXz)

Here is a video showing what I mean.


https://github.com/QCHackers/tqec/assets/16883876/48c23735-cc58-4767-be23-8eb6d0c539aa

rryoung98:
Yea I can speak more to the hunches I have for this issue. 

The issue we are facing is that calling destroy on the graphic is a code smell and a big no no. 

Here are some suggestions to fix it:

1. we can spin out the control panel for each plaquette into a plaquettecontrolpanel that is a child of the `workspace` container. When clicked on, the panel will find the plaquette that has been clicked on by using `getChildByName()`  builtin function. It may be necessary to traverse the entire child tree.  
2. We can also check that the cloning of plaquettes also does not add the plaquette graphic to another plaquette. Instead it should add the plaquette to a template.
3. when the current control panel is opened, another issue is that you can still double click on another plaquette and open the control panel for both plauqettes. This may be causing the .geometry issue as well since if we destroyed  a parent plaquette then the children will also be destroyed.

Let me know if that makes sense

smburdick:
@rryoung98 I think point (1) resonated with me most. Because each plaquette can be manipulated individually, we can make each plaquette a child of the workspace container. I am concerned about how plaquettes are structured together into a larger surface code, which could mean that if a single plaquette was deleted, the structure representing the surface code would need to be updated as well.

>  It may be necessary to traverse the entire child tree.

JS objects (being hash tables) imply this can be done in O(1) time, amortized, so I don't think that's something to worry about.

[...]

PR #31, opened 2024-01-09
Fix bugs when deleting cloned plaquettes; restructure workspace and control panel; add basic UI for plaquette cloning

smburdick:
Address #24 

Conversation in that issue needs to be spun off into another issue -- in particular, address why `this._geometry` is null when selecting a plaquette and deleting it.


rryoung98 (changes requested):
Thanks for the PR @smburdick  

The buttons are really helpful!

I believe after todays conversation, that it's important to think about the basic UI more. 

First I'll outline the pros and cons of the two methods proposed here I think having both is vital for now to make sure we can work quickly:

1. Drag and create
**Pro:** 
- It's considerably faster when the graphics are repetitive.
**Con:**
- The drag is buggy (Though I believe it can be fixed and added additional notification for the user).
- It's not immediately intuitive (especially in its current state where there are not enough instructions).
- Deleting graphics issue (should be spun out into a separate issue outside of the UI since both implementations have this issue).

2. Arrows
**Pro:** 
- Much more explicit
- Great for testing
**Con:**
- It takes a while to click on individual shapes over and over. (this can be improved too)
- Still has the ._geometry null bug. 


Can we keep both for now and try and fix the above Cons? This is just a function of putting in time to make the UI perfect for both methods.

smburdick:
@rryoung98 I agree with your assessments, and have restored the ability to drag-copy.

Do we want to address the `._geometry` null bug in this PR as well?

I've had some success deleting copied plaquettes, when they were spawned through the control panel, but not so when copied by dragging.

There are some lingering bugs with copying (when both styles are mixed). For example: sometimes, I cannot make a copy of a copied plaquette.

I also agree that several issues here should be tracked separately.

rryoung98:
Hmmm. That's up to you, but fixing .geometry takes priority. I've added some ideas on how to fix it to the issue. I will tackle it this weekend if not finished.

smburdick:
@rryoung98 I'll take a look at the geometry issue as well (starting tomorrow) and will keep this PR open.

smburdick:
Addressing #24 in latest commit here.

smburdick:
Notably, plaquettes spawned unintentionally still cannot be deleted, but can if done through the control panel.

rryoung98 (commented):
This is a pretty good start nice adds still some work to do to make sure the plaquettes are deleted properly and the dragging functionality is fixed.

Nice work :)

rryoung98 (approved):

Issue #39, opened 2024-01-11
Grid should fit entire window when window is resized


smburdick:

smburdick:
Seems to work on Windows but not macOS ...

smburdick:
The workspace will have to be re-rendered upon the viewport size change, should be fairly trivial

nelimee:
This issue is not relevant anymore. Closing it. See #323 for more details.

PR #42, opened 2024-01-14
Add font color to button class, reposition and recolor download button

smburdick:
Going to lock the download button in the top-right, so it doesn't vanish when the window is resized.

Related issue: #39


rryoung98 (approved):
Lgtm for now, as this is a small change

PR #43, opened 2024-01-15
Update JSON specification

nelimee:
This pull request updates the JSON schemes provided a few weeks ago to the current status of the code.

It also opens a new space to discuss about the specification.

### DIFF
--- a/specification/schemas/circuit.schema.json
+++ b/specification/schemas/circuit.schema.json
@@ -1,9 +1,12 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "/tqec/specification/schemas/circuit.schema.json",
-  "$comment": "For the moment, this field is not fixed. Anything compatible with cirq is a good input. If this is something compatible with criq, we will also need a way to get information on detectors, observables and shift_coords that are not part of cirq official gates.",
   "title": "Quantum circuit",
   "description": "A quantum circuit representing a quantum computation.",
   "type": "object",
-  "properties": {}
+  "properties": {
+    "openqasm": {
+      "type": "string"
+    }
+  }
 }
--- a/specification/schemas/plaquette.schema.json
+++ b/specification/schemas/plaquette.schema.json
@@ -13,22 +13,9 @@
         "minItems": 1
       }
     },
-    "layers": {
-      "type": "array",
-      "description": "A list of scheduled circuits, each one representing a layer of the QEC code.",
-      "items": {
-        "description": "The scheduled quantum circuit representing",
-        "$ref": "/tqec/specification/schemas/scheduled_circuit.schema.json"
-      }
-    },
-    "color": {
-      "type": "array",
-      "description": "An RGB(A) color for the plaquette.",
-      "prefixItems": [
-        {"type": "integer", "minimum": 0, "maximum": 255},
-        {"type": "integer", "minimum": 0, "maximum": 255},
-        {"type": "integer", "minimum": 0, "maximum": 255}
-      ]
+    "circuit": {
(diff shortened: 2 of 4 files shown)


specification/schemas/circuit.schema.json
```diff
@@ -1,9 +1,12 @@
...
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "/tqec/specification/schemas/circuit.schema.json",
-  "$comment": "For the moment, this field is not fixed. Anything compatible with cirq is a good input. If this is something compatible with criq, we will also need a way to get information on detectors, observables and shift_coords that are not part of cirq official gates.",
   "title": "Quantum circuit",
   "description": "A quantum circuit representing a quantum computation.",
   "type": "object",
-  "properties": {}
-}
+  "properties": {
+    "openqasm": {
```
smburdick, line 8: Is this the qasm representation of the circuit?

https://quantumai.google/cirq/build/interop#importing_from_openqasm

smburdick:
@nelimee What would you like help with in particular?

nelimee:
> @nelimee What would you like help with in particular?

I think advancing this PR will amount to making round trips between the frontend and backend. I expect that at least some amount of information in the JSON schema in this PR will be inconvenient to provide for the frontend, in which case we should adapt the JSON schema to something that the frontend can provide.

Right now, it would help to know if there are any part of the JSON schema in this PR that you *think* will be hard to provide by the frontend, so that I can implement these parts last, when they will hopefully be resolved on the frontend side.

smburdick:
I have a different viewpoint. Can we specify what we need on the backend through these JSON specs, but for now make some fields optional?  What the frontend can do is changing all the time (eg, templates in the near future), but specifying what is expected is important for prioritizing new features on the frontend.

nelimee:
> Can we specify what we need on the backend through these JSON specs

This is what is done currently. I expect that it might change alongside the developments you are performing on the frontend, but right now the JSON encodes everything that the backend needs to operate and generate a Stim file.

To write it differently, the current JSON schemes provided are a good view of what is expected. If you encounter any issue on the frontend side to provide that, we should discuss together to try to change that "base" into something that will be convenient for both the frontend and the backend.

smburdick:
> right now the JSON encodes everything that the backend needs to operate and generate a Stim file.

> To write it differently, the current JSON schemes provided are a good view of what is expected. 

I think that's what we need; we're on the same page. I'm also more inclined to keep the backend specs as our source of truth  -- the frontend is its client, after all.

specification/schemas/circuit.schema.json
nelimee: This is, for the moment, the OpenQASM representation of the circuit. As noted in https://github.com/QCHackers/tqec/issues/38#issuecomment-1885885099:
> OpenQASM (2.0 I guess?) is fine on my side. I have no strong feeling about that part, anything that fits the needs (easy to generate on the frontend, easy to parse on the backend and able to represent everything that we need) is fine for me.

So anything in the following list should be fine:
- OpenQASM 2.0 (I just saw that the `cirq` importer seems to not support `reset`, this might be an issue)
- Quirk visual format
- Any `stim`-compatible format (textual, json, etc.)

Anything **not** in the previous list is up to debate.

smburdick:
Ping - are we proceeding with this?

nelimee:
> Ping - are we proceeding with this?

In my opinion this issue is a good placeholder that will be replaced by more specific issues when we will really start looking at the interaction between the frontend and the backend.

smburdick:
Since the JSON spec is being determined dynamically on the frontend, we can close this for now.

Issue #53, opened 2024-01-17
Implement Z and X extensions without rotating


nelimee:
The last step before the full "corner" is to be able to perform the following steps:

1. Initialize a logical qubit in the $\vert 0_L\rangle$ state.
2. Perform a few rounds of memory.
3. Move the logical qubit in the $Z$ direction (exactly #50).
4. Perform a few rounds of memory on the newly placed logical qubit.
5. Move the logical qubit in the $X$ direction (exactly #51).
6. Perform a few rounds of memory on the newly placed logical qubit.
7. Measure the observable.

Basically, a corner without the complexity of changing plaquette orientation midway and changing the orientation of the observables.

Idea provided by Austin.

Issue #52, opened 2024-01-17
Implement 90° rotated experiments


nelimee:
Implement the experiments described in #50 and #51, rotating everything by 90°. 

In particular this means that:
- $X$ and $Z$ plaquettes will be placed differently,
- $X$ and $Z$ observables will change their position and orientation (horizontal $\leftrightarrow$ vertical).

Idea provided by Austin.

Issue #51, opened 2024-01-17
Implement a vertical X movement QEC


nelimee:
Implement the exact same experiment as #50 but:

- physical qubits are initialised in the $\vert + \rangle$ state,
- the logical qubit is extended vertically, meaning that the $X$ observable is extended and the $Z$ observable is moved 2 plaquettes up.

Idea provided by Austin.

Issue #50, opened 2024-01-17
Implement a horizontal Z movement QEC 


nelimee:
Implement the following TQEC:

![experiment1](https://github.com/QCHackers/tqec/assets/12374487/a57031f8-82e1-4521-ab7c-5ea7474d2e39)

Needs at least #45 fixed.

Idea provided by Austin.

nelimee:
Note that I received a mail yesterday evening (just after my message on the Google group) from someone (I do not know if he/she want his/her name publicly here) that will also likely be interested in that. 
Go ahead, you might be joined by at least someone else.

inmzhang:
That's totally fine. It's no problem anyone else would like to work on this. Regardless of whether my own PR is finally accepted, it should still be an experience of learning in qec :)

PR #55, opened 2024-01-17
Improve the MeasurementMap structure

nelimee:
This pull-request improves the way measurements are recorded by the MeasurementMap to compute the record offsets needed by stim.

MeasurementMap now flattens recursively all the `cirq.CircuitOperation` instances in the given cirq.AbstractCircuit. The queries to the MeasurementMap structure have also been adapted to take into account the fact that repetitions of a cirq.CircuitOperation are now taken into account to number the Moment instances.

### DIFF
--- a/tqec/detectors/measurement_map.py
+++ b/tqec/detectors/measurement_map.py
@@ -1,6 +1,19 @@
 import cirq
 
 
+def flatten(obj: cirq.Moment | cirq.AbstractCircuit) -> cirq.Circuit:
+    # From https://github.com/quantumlib/Cirq/issues/3783
+    if isinstance(obj, cirq.Moment):
+        flattenable = [op for op in obj if isinstance(op, cirq.CircuitOperation)]
+        if not flattenable:
+            return cirq.Circuit(obj)
+        return cirq.Circuit.zip(
+            cirq.Circuit(obj - flattenable),
+            *[flatten(e.mapped_circuit(True)) for e in flattenable],
+        )
+    return cirq.Circuit([moment for e in obj for moment in flatten(e)])
+
+
 class CircuitMeasurementMap:
     def __init__(self, circuit: cirq.AbstractCircuit) -> None:
         """Stores information about all the measurements found in the provided circuit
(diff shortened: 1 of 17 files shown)


tqec/detectors/measurement_map.py
```diff
@@ -1,23 +1,33 @@
 import cirq
 
 
+def flatten(obj: cirq.Moment | cirq.AbstractCircuit) -> cirq.Circuit:
+    # From https://github.com/quantumlib/Cirq/issues/3783
+    if isinstance(obj, cirq.Moment):
```
smburdick, line 6: I would suggest writing a separate function for `Moment`s

tqec/detectors/measurement_map.py
```diff
@@ -1,23 +1,33 @@
...
+        flattenable = [op for op in obj if isinstance(op, cirq.CircuitOperation)]
+        if not flattenable:
+            return cirq.Circuit(obj)
+        return cirq.Circuit.zip(
+            cirq.Circuit(obj - flattenable),
+            *[flatten(e.mapped_circuit(True)) for e in flattenable],
+        )
+    return cirq.Circuit([moment for e in obj for moment in flatten(e)])
```
smburdick, line 14: Ah, this is a recursive function! I still think separating them out makes this a little more comprehensible.

tqec/detectors/measurement_map.py
```diff
@@ -122,50 +109,25 @@ def _get_global_measurement_index(
...
         for moment in circuit:
             global_measurement_indices.append(dict())
-            moment_contains_measurement: bool = False
-            moment_contains_CircuitOperation: bool = False
             for op in moment.operations:
+                # Note that cirq.CircuitOperation instances should never be encountered here because
+                # they should all have been flattened at the MeasurementMap creation.
+                assert not isinstance(
+                    op.untagged, cirq.CircuitOperation
+                ), "cirq.CircuitOperation should be flattened before calling this method."
```
smburdick, line 124: To make this a little more explicit, you could create a `FlattenedCircuitOperation` class that is the desired input type.

tqec/detectors/transformer.py
```diff
@@ -19,6 +21,19 @@ def _fill_in_detectors_global_record_indices_impl(
         number_of_moments_explored: int = 1
         for operation in moment.operations:
             if isinstance(operation, cirq.CircuitOperation):
```
smburdick, line 23: Again, I think using different functions instead of `isinstance` makes it less brittle and more comprehensible, but that may not be possible in this case if the function is already being invoked this way.

smburdick (approved):
Looking forward to using this library on the server :-). I am making some non-functional recommendations based on what I consider to be good practices, since I can tell you are writing this in a clear and correct way. Since we are still in the MVP phase (or are we?), things can be cleaned up/refactored as time goes on, and this can be safely merged.

nelimee:
I agree with most of your comments (the two first ones are up to debate for me). I think it would be worth factorising out everything uniquely related to `cirq` (like the `flatten` function) to its own module. 

In terms of merging, I would prefer to merge this PR as-is for several reasons:
1. it is quite succinct and (seems to) achieve the desired goal,
2. it is blocking most of the other non-trivial issues #50, #51, #52 and #53,
3. I would feel better knowing that it works correctly on `main` now and fixing the few non-functional things in a follow-up PR

As you seem to agree on the fact that this can be merged, I will merge it, open a new PR and re-copy your comments to keep a track on that.

PR #65, opened 2024-01-23
Fix the rectangle scaling problem

### LINKED ISSUE #63: Incorrect scaling result for `ScalableRectangle`
When running the example in [doubling_distance.ipynb](https://github.com/QCHackers/tqec/blob/main/notebooks/doubling_distance.ipynb), the template's scaling will not work correctly after scaled with `k=2`.

The result can be reproduced with the following code:
```shell
>>> display_template(doubling_template.scale_to(6))
 .  .  1  .  1  .  1  .  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  1  .  1  .  1  .  .
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
.  1  .  1  .  1  .  1  .  1  .  1  .  1  .  .
>>> doubling_template.scale_to(2)
>>> display_template(doubling_template.scale_to(6))
.  .  1  .  1  .  1  .  .  .  .  .  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .
1  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  .  .  .  .  .  .  .  .  .  .  .  .  .
.  1  1  1  1  1  1  1  .  .  .  .  .  1  .  1  .  1  .  .
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  .  .  .  .  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  .  .  .  .  1  1  1  1  1  1  .
.  1  1  1  1  1  1  .  .  .  .  .  .  1  1  1  1  1  1  .
.  1  1  1  1  1  1  .  .  .  .  .  .  1  1  1  1  1  1  .
.  1  1  1  1  1  1  .  .  .  .  .  .  1  1  1  1  1  1  .
.  1  1  1  1  1  1  .  .  .  .  .  .  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  .
.  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
.  1  .  1  .  1  .  1  .  .  .  .  .  1  .  1  .  1  .  .
```

The problem is with the scaling implementation of `ScalableRectangle`. When the `width == height`, both of the width and height will be scaled. However, in the case of "doubling distance"(or most of other cases), there should be some rectangles that only need to scale a single dimension.

To solve the problem, my idea is to only allow `ScalableRectangle` to change one of the dimension. I will submit a PR once #62 is merged.

inmzhang:
Close #63

Now the `ScalableRectangle` accepts `scale_width` to specify whether to scale the width or height, which will be set to the dimension with larger even value when not specified.

### DIFF
--- a/src/tqec/constructions/qubit.py
+++ b/src/tqec/constructions/qubit.py
@@ -45,6 +45,16 @@ class ScalableQubitSquare(TemplateOrchestrator):
 
 class ScalableQubitRectangle(TemplateOrchestrator):
     def __init__(self, width: int, height: int, scale_width: bool | None = None) -> None:
+        """A scalable rectangle error-corrected qubit.
+        
+        A scalable rectangle qubit can only scale its width **or** height, but not both.
+        
+        :param width: width of the qubit.   
+        :param height: height of the qubit.
+        :param scale_width: whether to scale the width or height. If None, the dimension
+            with the even value or the larger value will be scaled. If both dimensions
+            are even and equal, the width will be scaled by default.
+        """
         _templates = [
             # Central square, containing plaquettes of types 3 and 4
             TemplateWithIndices(ScalableRectangle(width, height, scale_width), [3, 4]),
--- a/src/tqec/constructions/qubit.py
+++ b/src/tqec/constructions/qubit.py
@@ -44,10 +44,10 @@ class ScalableQubitSquare(TemplateOrchestrator):
 
 
 class ScalableQubitRectangle(TemplateOrchestrator):
-    def __init__(self, width: int, height: int) -> None:
+    def __init__(self, width: int, height: int, scale_width: bool | None = None) -> None:
         _templates = [
             # Central square, containing plaquettes of types 3 and 4
-            TemplateWithIndices(ScalableRectangle(width, height), [3, 4]),
+            TemplateWithIndices(ScalableRectangle(width, height, scale_width), [3, 4]),
             # Top rectangle, containing plaquettes of type 1 only
             TemplateWithIndices(ScalableRectangle(width, 1), [0, 1]),
             # Left rectangle, containing plaquettes of type 2 only
--- a/src/tqec/templates/scalable/rectangle.py
+++ b/src/tqec/templates/scalable/rectangle.py
@@ -4,14 +4,13 @@ from tqec.templates.shapes.rectangle import Rectangle
 
 
 class ScalableRectangle(ScalableTemplate):
-    """Rectangle with scalable width **or** height."""
     def __init__(
         self, 
         width: int, 
         height: int, 
         scale_width: bool | None = None,
     ) -> None:
-        """Create an instance of ScalableRectangle.
+        """Rectangle with scalable width **or** height.
 
         A scalable rectangle can only scale its width **or** height, but not both.
 
--- a/src/tqec/templates/scalable/rectangle.py
+++ b/src/tqec/templates/scalable/rectangle.py
@@ -4,19 +4,37 @@ from tqec.templates.shapes.rectangle import Rectangle
 
 
 class ScalableRectangle(ScalableTemplate):
-    def __init__(self, width: int, height: int) -> None:
+    """Rectangle with scalable width **or** height."""
+    def __init__(
+        self, 
+        width: int, 
+        height: int, 
+        scale_width: bool | None = None,
+    ) -> None:
+        """Create an instance of ScalableRectangle.
+
+        A scalable rectangle can only scale its width **or** height, but not both.
+
+        :param width: width of the rectangle.   
+        :param height: height of the rectangle.
+        :param scale_width: whether to scale the width or height. If None, the dimension
+            with the even value or the larger value will be scaled. If both dimensions
+            are even and equal, the width will be scaled by default.
+        """
+        assert (width % 2 == 0) or (height % 2 == 0), "At least one dimension must be even to be scalable!"
+        # Determine which dimension to scale
+        if scale_width is None:
+            scale_width = height % 2 != 0 or (width % 2 == 0 and width >= height)
         super().__init__(Rectangle(width, height))
+        self._scale_width = scale_width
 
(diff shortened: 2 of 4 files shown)


src/tqec/templates/scalable/rectangle.py
```diff
@@ -4,19 +4,37 @@
 
 
 class ScalableRectangle(ScalableTemplate):
-    def __init__(self, width: int, height: int) -> None:
+    """Rectangle with scalable width **or** height."""
```
nelimee, line 7: I would remove that comment and include it as the first line of the `__init__` docstring. 
In general, I find it more meaningful to document the class behaviour in the `__init__` method, as it helps in documenting everything (i.e., also the `__init__` parameters) **and** avoid meaningless docstrings in the `__init__` that simply refer to the main class documentation. Also, as far as I am aware, most of the documentation-generation tools allow to replace the class documentation by the `__init__` one.

src/tqec/constructions/qubit.py
```diff
@@ -44,10 +44,10 @@ def __init__(self, dim: int) -> None:
 
 
 class ScalableQubitRectangle(TemplateOrchestrator):
-    def __init__(self, width: int, height: int) -> None:
+    def __init__(self, width: int, height: int, scale_width: bool | None = None) -> None:
```
nelimee, line 47: I would copy-paste the documentation of `ScalableRectangle` here also, to explain clearly what `width`, `height` and more importantly `scale_width` are doing.

src/tqec/templates/scalable/rectangle.py
```diff
@@ -4,19 +4,36 @@
...
+        :param scale_width: whether to scale the width or height. If None, the dimension
+            with the even value or the larger value will be scaled. If both dimensions
+            are even and equal, the width will be scaled by default.
+        """
+        assert (width % 2 == 0) or (height % 2 == 0), "At least one dimension must be even to be scalable!"
+        # Determine which dimension to scale
+        if scale_width is None:
+            scale_width = height % 2 != 0 or (width % 2 == 0 and width >= height)
         super().__init__(Rectangle(width, height))
+        self._scale_width = scale_width
```
nelimee, line 28: The type annotation is missing here. Most type checkers will be able to infer, but some might incorrectly infer `bool | None` instead of `bool`, so let's be explicit.
```suggestion
        self._scale_width: bool = scale_width
```

nelimee (changes requested):

nelimee (approved):

PR #66, opened 2024-01-23
Pytest setup

### LINKED ISSUE #64: What framework to implement tests for the `tqec` library?
The `tqec` Python library is missing tests. 

Currently, the only tests are functional ones, implemented in [`notebooks/`](https://github.com/QCHackers/tqec/tree/main/notebooks) and only run locally.

This is the first issue about adding tests to the `tqec` Python package. The question that I want to discuss here is:

> What package/framework should we use to test the `tqec` library?

Here is a list of features that I would like to have:

- easy to run in CI,
- well recognised in the Python ecosystem and well maintained,
- ability to filter out some tests (for example with a tagging system).

The last point would be nice to include for tests that need a lot of computing time / computing power to run. These could be filtered out in the CI but still included in the tests when run locally.

I am in favour of [`pytest`](https://docs.pytest.org/en/7.4.x/), but I am also really open to any other suggestion that makes sense.

nelimee:
Fixes #64 
Introduce the `pytest` dependency and implement a first test to check that everything is working as expected.


nelimee:
@inmzhang for a reason I ignore I cannot register you as a reviewer. Any idea?

inmzhang:
> @inmzhang for a reason I ignore I cannot register you as a reviewer. Any idea?

Is it because I don't have write access to the repo?

nelimee:
> @inmzhang for a reason I ignore I cannot register you as a reviewer. Any idea?

@vtomole any hints? I do not have the rights to manage the `tqec` repository.

tests/test_position.py
vtomole, line 1: Can we put the tests in the same directory as the source like https://github.com/quantumlib/Cirq?

vtomole:
@inmzhang I've sent you an invite to the repo.

tests/test_position.py
vtomole: How would it pollute the mainspace? All `quantumlib` repos do it like this and it makes tests easier to find cause they are right next to the source files.

nelimee: Both styles are in the [`pytest` best practices](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#choosing-a-test-layout-import-rules).

nelimee: > How would it pollute the mainspace? All `quantumlib` repos do it like this and it makes tests easier to find cause they are right next to the source files.

Aren't the test files also considered importable? They are regular Python files, so I do not see why tests files that are next to code files would not be considered importable. I might miss something though.

nelimee: Reading https://www.reddit.com/r/Python/comments/11zjjzy/arguments_against_separating_test_from_src_in_a/, I am also leaning to the in-package test files. I'll edit the PR and test my "doesn't it pollute the namespace" assumption.

nelimee: Moving the file `tests/test_position` to `src/tqec/test_position.py` works for the tests. But as I said, it pollutes the `tqec` namespace. The `test_position.py` is proposed as an auto-completion on my IDE at least.
It might be solved by 
1. excluding the test files from the package distribution (should be done eventually) for non-editable installations,
2. adding `__init__.py` files with appropriate `__all__` attribute to only include non-test files in the auto-completions for editable installations.

I am not a fan of 2. as it adds a lot of boilerplate code and also adds a required step whenever a new files is created (think about adding it to the `__all__` attribute).

Is there any obvious method I missed to avoid having `test_*.py` files appear as importable in editable installs?

nelimee: Note: just tested on `cirq`, the test packages are accessible and appear in the auto-completion even for a non-editable installation. It seems odd to me to include test files in release builds, but I might be the only one with this opinion.

Gistbatch: I agree with you, @nelimee, in that I'd rather have a a separate `tests` directory.
The python docs for the built-in [unittest](https://docs.python.org/3/library/unittest.html#organizing-test-code) library advocates for separating code and tests.

vtomole: I think importable tests are a good thing. Tests are great documentation and i usually use them to figure out how to call a module/function/class/method in my ide.

nelimee: I agree with the documentation purpose, but what is the point of importing tests from the main library?
In my opinion, the best of both worlds would be to find a solution to put all the tests besides their respective source file (i.e. no tests/ directory) but not being able to import them (at least from the non-editable installation, best would be for tests files to be also excluded from editable installation).

vtomole: What's the harm in having the tests be importable and be in the editable installation? Anyway it looks like I've been outvoted so having the `tests` directory is fine.

nelimee: It is annoying and pretty useless for the user of the library. Being able to find the tests quickly and easily is a good thing (for documentation purpose as you mention), but importable test makes no sense to me.
Also, you are not outnumbered, as I said earlier I am more on the side of having tests alongside the code now. I am simply annoyed by the presence of these tests in the importable modules under the main library package.

nelimee: I propose to put the tests alongside the source files (no tests/ directory) and I'll sit on the "importability of tests" for now, fixing it later.
@gistbatch would that be fine for you?

inmzhang: For me, both methods are fine. The importability problem seems can be solved under non-editable installation by excluding the test files during packaging and distribution. Under editable(develop) mode, I think it's just ok to allow the tests to be importable.

nelimee:
Per the discussion in @vtomole review, tests are now in `src/tqec`. 
Also, it seems like `setuptools` defaults to include tests (see the presence of `tests/*.py` in the list at [this link](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html)), so exclusion of test files from the distribution is not done yet.

src/tqec/test_position.py
vtomole, line 1: Rename file to `position_test.py` in order to have the test file be next to the `position.py` file.

vtomole (approved):
LGTM

PR #70, opened 2024-01-24
Add function to display a `TemplateOrchestrator` as a svg file

inmzhang:
Currently we only have `display_template` to view a `TemplateOrchestrator` as ASCII texts. It is hard to tell the boundaries between the templates from the raw texts.

I add a simple implementation of `display_templates_svg` to get a svg view for the `TemplateOrchestrator`. The implementation is not carefully designed and use several private methods of `TemplateOrchestrator`, which definitely needs to be refactored in the future. However, I think it is ok to use it during development for debugging.

The code demo is as below:
```python
from tqec.constructions.corner import ScalableCorner
from tqec.display import display_templates_svg

corner = ScalableCorner(3)
display_templates_svg(corner, "corner.svg")
```

Here are the svg views for the examples in [corner.ipynb](https://github.com/QCHackers/tqec/blob/main/notebooks/corner.ipynb) and [doubling_distance.ipynb](https://github.com/QCHackers/tqec/blob/main/notebooks/doubling_distance.ipynb):

corner_k=3

doubling_distance_k=3

### DIFF
--- a/src/tqec/display.py
+++ b/src/tqec/display.py
@@ -1,3 +1,6 @@
+import pathlib
+import math
+
 from tqec.templates.base import Template
 from tqec.templates.orchestrator import TemplateOrchestrator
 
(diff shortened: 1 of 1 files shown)


nelimee (commented):
Seems fine to me, really great addition!
A few improvement ideas:
1. Also display the relationships between templates (corner pinning)
2. Find a way to display the "scalable" property of the outer outer edges. For example, outer edges that will scale when `scale_to` is called are depicted as double-ended arrows whereas those that does not scale are depicted with inward-arrows?

The fact that you are using private attributes of the `TemplateOrchestrator` is not an issue for the moment, we will just have to remember to also change this function. 
A way to solve that would be to move that implementation into a `TemplateOrchestrator` method, that would be called by the `display_template_svg`. What do you think of that?

inmzhang:
> A few improvement ideas:...

Good points! Will it be a good first issue to guide someone to start with the codebase? Or I can implement it myself.

> A way to solve that would be to move that implementation into a TemplateOrchestrator method, that would be called by the display_template_svg. What do you think of that?

Personally I prefer more functional way to do this, which will have some benefits for the later refactoring.

nelimee (approved):

PR #79, opened 2024-01-26
Revert some changes from earlier commit

### LINKED ISSUE #78: Commit introduced mistakes
The commit 6824a556c0d30ffdac8ab6dc2ee52fb16c5dc2da removed the output of some notebooks, which is not desirable (see below) and would like to revert these changes. Moreover, it changed the README.md, leading to two ways of installing the package without any explanations on which to pick.

Also note: the section
> nbstripout is a great option for clearing the output of the jupyter notebooks. It can be installed using pip install nbstripout. For more info see below in before you commit section.

should be removed in my opinion. The notebooks in the `notebooks/` directory are here to demonstrate the code, document it for a novice user by providing a working example with its outputs and to functionally test the overall code.
These 3 purposes benefit from the inclusion of the notebooks output to the repository.

We can discuss on whether or not the notebook outputs should be included in the repository, I know the standard way of doing is to strip, so if the 3 justifications above do not satisfy you please tell me.

nelimee:
This PR reverts some of the changes to the `README.md` file an `notebooks/` folder made by an earlier commit.

Fixes #78


smburdick (approved):
LGTM. You can also rollback problematic commits; https://stackoverflow.com/questions/4372435/how-can-i-rollback-a-git-repository-to-a-specific-commit

nelimee:
> LGTM. You can also rollback problematic commits; https://stackoverflow.com/questions/4372435/how-can-i-rollback-a-git-repository-to-a-specific-commit

Thanks! I know about rollback, but it was easy enough to simply rollback manually the parts that were not desirable and keep the changes that were a nice addition.

Issue #34, opened 2024-01-10
Lifting the "`Plaquette.shape` should all be the same" assumption


nelimee:
The `tqec` Python package currently works under the assumption that all the [`Plaquette`](https://github.com/QCHackers/tqec/blob/f6f29323ebf2544ae576c6866167026adbfa6359/tqec/plaquette/plaquette.py#L10) instances provided when calling the [`generate_circuit`](https://github.com/QCHackers/tqec/blob/f6f29323ebf2544ae576c6866167026adbfa6359/tqec/generation/circuit.py#L11) function are of the same shape.

The [`shape`](https://github.com/QCHackers/tqec/blob/f6f29323ebf2544ae576c6866167026adbfa6359/tqec/plaquette/plaquette.py#L31) of a  [`Plaquette`](https://github.com/QCHackers/tqec/blob/f6f29323ebf2544ae576c6866167026adbfa6359/tqec/plaquette/plaquette.py#L10) instance is currently defined as the coordinates of the lower-right-most qubit, each increased by one, which assumes that qubits are the closest possible to the origin `(0, 0)`. This definition (and how we compute the shape, if we still need to compute it) have to change.

Lifting this assumption will likely require non-trivial changes, I am open to any suggestion.

afowler:
Provided a plaquette has a nominated qubit that serves as its corner and
this corner can be pinned to a template, it is not clear to me why each
plaquette needs to be the same shape. Fundamentally, a plaquette is just a
human visualization of an underlying quantum circuit, which is the real
thing that is being loaded into the template. In principle, the underlying
quantum circuit could be rather random in structure reaching into qubits
outside any typical square, but provided these quantum gates are compatible
in the sense they are all applied in different non-overlapping time steps,
this shouldn't be an issue.

Thoughts?

nelimee:
This is an assumption I made when initially writing the code because it seemed simpler to me and allowed me to make some simplifications. The idea was to remove this assumption at one point, when the code was mature enough, and the time as come.
Initially, I made this assumption because it helped in computing the global position of the local coordinate system origin of each plaquette. The computation will likely be a little bit more involved now, but as long as we are able to get the global coordinate of each plaquette "origin", everything should work as expected.
[...]

PR #80, opened 2024-01-26
Lift the plaquette shape restriction in circuit construction

Gistbatch:
closes: #34

I tried to solve this issue under the assumption that plaquettes are at least rectangular.

**Summary**
So far I have done the following:

1. I created a notion of origin for the circuit in the plaquette. So far, this only works with the option `TOP_LEFT`, which assumes the origin is `(0,0)`. The circuit should be defined relative to this origin.
2. I removed the assertion for the shape.
3. I added two arguments to `generate_circuit`, which dictates the base distance between two plaquettes. At the moment I only use this for `0` plaquettes.
4. I replaced the starting_indices by an offset, which is calculated by pushing the origin of the current plaquette to the bottom of the plaquette above and the right to the plaquette on the left.

**Issues**
Some issues remain:

1. I've only tested this against the (partial) output of this [example](notebooks/logical_qubit_memory_experiment.ipynb), and I'm quite sure it works for the square case.
2. The assertion doesn't work in the example https://github.com/QCHackers/tqec/commit/9dfa23ea6b49565c469567587a5ed74b727c9677#diff-3b8c000f9c1d54d5514d934053adb8f82d6ac3cf170b9850419d209590772cb4R42-R44 

**Open questions**

1. Should we use a linter? I'm personally using `black` but I've seen some code that was not according to `pep-8`
2. Where should I put the tests? And how should we name them? I'd prefer `test_{file_name}`.

### DIFF
--- a/src/tqec/enums.py
+++ b/src/tqec/enums.py
@@ -48,7 +48,7 @@ class PlaquetteOrientation(Enum):
 class PlaquetteOrigin(Enum):
     """Defines the origin of a plaquette."""
 
-    BOTTOM_RIGHT = auto()
     CENTRAL = auto()
     FIRST = auto()
+    TOP_LEFT = auto()
     USER_DEFINED = auto()
--- a/src/tqec/enums.py
+++ b/src/tqec/enums.py
@@ -43,3 +43,12 @@ class PlaquetteOrientation(Enum):
     LEFT = auto()
     DOWN = auto()
     UP = auto()
+
+
+class PlaquetteOrigin(Enum):
+    """Defines the origin of a plaquette."""
+
+    BOTTOM_RIGHT = auto()
+    CENTRAL = auto()
+    FIRST = auto()
+    USER_DEFINED = auto()
--- a/src/tqec/generation/circuit.py
+++ b/src/tqec/generation/circuit.py
@@ -74,13 +74,84 @@ def generate_circuit(
             qubit_x = get_plaquette_starting_index(plaquette_shape.x, plaquette_x)
             qubit_y = get_plaquette_starting_index(plaquette_shape.y, plaquette_y)
             scheduled_circuit = deepcopy(plaquette_circuits[plaquette_index])
-            qubit_map = {
-                # GridQubit are indexed as (row, col), so (y, x)
-                qubit: qubit + (qubit_y, qubit_x)  # type: ignore
-                for qubit in scheduled_circuit.raw_circuit.all_qubits()
-            }
-            scheduled_circuit.map_to_qubits(qubit_map, inplace=True)
-            all_scheduled_circuits.append(scheduled_circuit)
 
-    # Merge everything!
-    return merge_scheduled_circuits(all_scheduled_circuits)
+
+def _create_mapping(
+    plaquette: Plaquette,
+    scheduled_circuit: ScheduledCircuit,
+    offset: Position,
+) -> dict[cirq.Qid, cirq.Qid]:
+    origin = plaquette.origin
+    assert origin == Position(0, 0), "Only origin (0,0) is supported for now"
+
+    qubit_map = {
+        # GridQubit are indexed as (row, col), so (y, x)
+        # Qubits are given relative to an origin, so we need to add the offset
+        qubit: qubit + (offset.y, offset.x) + (origin.y, origin.x)  # type: ignore
+        for qubit in scheduled_circuit.raw_circuit.all_qubits()
+    }
+    return qubit_map
+
+
+def _find_offset(
+    scheduled_circuits: list[ScheduledCircuit],
+    row_index: int,
+    column_index: int,
+    length: int,
+    default_x_increment: int,
+    default_y_increment: int,
+) -> Position:
+    """Computes the offset by looking at already placed circuits
+
+    We anchor the top left corner by one of the following ways:
+        - at the origin (0,0) (first circuit)
+        - the top right corner of the previous circuit (left)
+        - the bottom left corner of the previous circuit (above)
+        - the bottom ove the above and the right of the left (above and left)
+    """
+    # first circuit
+    if len(scheduled_circuits) == 0 or row_index == column_index == 0:
+        return Position(0, 0)
+    default_multiplier = 0
+    for circuit in scheduled_circuits[: -(column_index + 1) : -1]:
+        if len(circuit.raw_circuit.all_qubits()) == 0:
+            default_multiplier += 1
+        else:
+            position_left = _find_coordinate(circuit, above=False)
+            position_left.x += default_x_increment * default_multiplier
+            break
+    else:
+        position_left = Position(default_x_increment * default_multiplier, 0)
+
+    if row_index == 0:  # left
+        return Position(position_left.x, 0)
+
+    default_multiplier = 0
+    for circuit in reversed(scheduled_circuits[column_index::length]):
+        if len(circuit.raw_circuit.all_qubits()) == 0:
+            default_multiplier += 1
+        else:
+            position_above = _find_coordinate(circuit, above=True)
+            position_above.y += default_y_increment * default_multiplier
+            break
+    else:
+        position_above = Position(0, default_y_increment * default_multiplier)
+
+    if column_index == 0:  # above
+        return position_above
+    if position_left == position_above:
+        return position_left
+    return Position(position_left.x, position_above.y)
+
+
+def _find_coordinate(final_circuit: ScheduledCircuit, above: bool = False) -> Position:
+    if len(final_circuit.raw_circuit.all_qubits()) == 0:
+        return Position(0, 0)
+    position = max(
+        (
+            Position(qubit.col, qubit.row)
+            for qubit in final_circuit.raw_circuit.all_qubits()
+        ),
+        key=lambda p: (p.y, -p.x) if above else (p.x, -p.y),
+    )
+    return Position(position.x, position.y)
--- a/src/tqec/generation/circuit.py
+++ b/src/tqec/generation/circuit.py
@@ -2,16 +2,17 @@ from copy import deepcopy
 
 import cirq
 
-from tqec.generation.topology import get_plaquette_starting_index
 from tqec.plaquette.plaquette import Plaquette
 from tqec.plaquette.schedule import ScheduledCircuit, merge_scheduled_circuits
-from tqec.position import Shape2D
+from tqec.position import Position
 from tqec.templates.orchestrator import TemplateOrchestrator
 
 
 def generate_circuit(
     template: TemplateOrchestrator,
     plaquettes: list[Plaquette],
+    default_x_increment: int = 2,
+    default_y_increment: int = 2,
 ) -> cirq.Circuit:
     """Generate a quantum circuit from a template and its plaquettes
 
@@ -22,7 +23,8 @@ def generate_circuit(
     This function requires that a few pre-conditions on the inputs are met:
     1. the number of plaquettes provided should match the number of plaquettes required by
        the provided template.
-    2. all the provided plaquettes should have the same shape.
+    2. all the provided plaquettes are rectengular. For '0' plaquettes, the shape is asumed to be
+        (default_x_increment, default_y_increment).
     3. all the provided plaquettes should be implemented on cirq.GridQubit instances **only**.
 
     If any of the above pre-conditions is not met, the inputs are considered invalid, in which
@@ -32,6 +34,9 @@ def generate_circuit(
         to implement.
     :param plaquettes: description of the computation that should happen at different time-slices
         of the quantum error correction experiment (or at least part of it).
+    :param default_x_increment: default increment in the x direction between two plaquettes.
+    :param default_y_increment: default increment in the y direction between two plaquettes.
+
     :returns: a cirq.Circuit instance implementing the (part of) quantum error correction experiment
         represented by the provided inputs.
 
@@ -44,11 +49,6 @@ def generate_circuit(
         f"The given template requires {template.expected_plaquettes_number - 1} plaquettes "
         f"but only {len(plaquettes)} have been provided."
     )
-
-    # Check that all the given plaquettes have the same shape. If not, this is an issue.
-    # The shape limitation is an assumption to simplify the code and will have to be
-    # eventually lifted.
-    plaquette_shape: Shape2D = plaquettes[0].shape
 
     # Instanciate the template with the appropriate plaquette indices.
     # Index 0 is "no plaquette" by convention.
@@ -71,9 +71,25 @@ def generate_circuit(
     plaquette_index: int
     for plaquette_y, line in enumerate(template_plaquettes):
         for plaquette_x, plaquette_index in enumerate(line):
-            qubit_x = get_plaquette_starting_index(plaquette_shape.x, plaquette_x)
-            qubit_y = get_plaquette_starting_index(plaquette_shape.y, plaquette_y)
             scheduled_circuit = deepcopy(plaquette_circuits[plaquette_index])
+
+            offset: Position = _find_offset(
+                all_scheduled_circuits,
+                plaquette_y,
+                plaquette_x,
+                len(line),
+                default_x_increment,
+                default_y_increment,
+            )
+            print(offset)
+            if plaquette_index > 0:
+                plaquette = plaquettes[plaquette_index - 1]
+                qubit_map = _create_mapping(plaquette, scheduled_circuit, offset)
+                scheduled_circuit.map_to_qubits(qubit_map, inplace=True)
+            all_scheduled_circuits.append(scheduled_circuit)
+
+    # Merge everything!
+    return merge_scheduled_circuits(all_scheduled_circuits)
 
 
 def _create_mapping(
--- a/src/tqec/generation/circuit.py
+++ b/src/tqec/generation/circuit.py
@@ -49,9 +49,6 @@ def generate_circuit(
     # The shape limitation is an assumption to simplify the code and will have to be
     # eventually lifted.
     plaquette_shape: Shape2D = plaquettes[0].shape
-    assert all(
-        p.shape == plaquette_shape for p in plaquettes
-    ), "All plaquettes should have exactly the same shape for the moment."
 
     # Instanciate the template with the appropriate plaquette indices.
     # Index 0 is "no plaquette" by convention.
@@ -70,7 +67,7 @@ def generate_circuit(
     ), "Qubits used in plaquette layers should be instances of cirq.GridQubit."
 
     # Generate the ScheduledCircuit instances for each plaquette instanciation
-    all_scheduled_circuits: list[ScheduledCircuit] = list()
+    all_scheduled_circuits: list[ScheduledCircuit] = []
     plaquette_index: int
     for plaquette_y, line in enumerate(template_plaquettes):
         for plaquette_x, plaquette_index in enumerate(line):
--- a/src/tqec/plaquette/plaquette.py
+++ b/src/tqec/plaquette/plaquette.py
@@ -17,7 +17,7 @@ class Plaquette:
         self,
         qubits: list[PlaquetteQubit],
         circuit: ScheduledCircuit,
-        origin: PlaquetteOrigin = PlaquetteOrigin.BOTTOM_RIGHT,
+        origin: PlaquetteOrigin = PlaquetteOrigin.TOP_LEFT,
         origin_index: int = 0,
     ) -> None:
         """Represents a QEC plaquette
@@ -39,8 +39,8 @@ class Plaquette:
 
         :raises ValueError: if the number of qubits doesn't match the number of qubits
         """
-        if len(qubits) != len(circuit.qubits):
-            raise ValueError("Number of qubits doesn't match number of circuit qubits")
+        # if len(qubits) != len(circuit.raw_circuit.all_qubits()):
+        #     raise ValueError("Number of qubits doesn't match number of circuit qubits")
 
         self._qubits = qubits
         self._circuit = circuit
@@ -61,10 +61,8 @@ class Plaquette:
     def _caclulate_origin(self, origin: PlaquetteOrigin, origin_index: int) -> None:
         """Calcluates the origin of the plaquette given the user input."""
         match origin:
-            case PlaquetteOrigin.BOTTOM_RIGHT:
-                max_x = max(qubit.position.x for qubit in self.qubits)
-                max_y = max(qubit.position.y for qubit in self.qubits)
-                self._origin = Position(max_x + 1, max_y + 1)
+            case PlaquetteOrigin.TOP_LEFT:
+                self._origin = Position(0, 0)
             case PlaquetteOrigin.CENTRAL:
                 raise NotImplementedError()
             case PlaquetteOrigin.FIRST:
--- a/src/tqec/plaquette/plaquette.py
+++ b/src/tqec/plaquette/plaquette.py
@@ -1,9 +1,9 @@
 import cirq
 
-from tqec.enums import PlaquetteOrientation
+from tqec.enums import PlaquetteOrientation, PlaquetteOrigin
 from tqec.plaquette.qubit import PlaquetteQubit
 from tqec.plaquette.schedule import ScheduledCircuit
-from tqec.position import Position, Shape2D
+from tqec.position import Position
 
 
 class Plaquette:
@@ -13,7 +13,13 @@ class Plaquette:
     def get_mergeable_tag() -> str:
         return Plaquette._MERGEABLE_TAG
 
-    def __init__(self, qubits: list[PlaquetteQubit], circuit: ScheduledCircuit) -> None:
+    def __init__(
+        self,
+        qubits: list[PlaquetteQubit],
+        circuit: ScheduledCircuit,
+        origin: PlaquetteOrigin = PlaquetteOrigin.BOTTOM_RIGHT,
+        origin_index: int = 0,
+    ) -> None:
         """Represents a QEC plaquette
 
         This class stores qubits in the plaquette local coordinate system and a scheduled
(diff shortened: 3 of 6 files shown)


src/tqec/enums.py
```diff
@@ -43,3 +43,12 @@ class PlaquetteOrientation(Enum):
...
     UP = auto()
+
+
+class PlaquetteOrigin(Enum):
+    """Defines the origin of a plaquette."""
+
+    CENTRAL = auto()
+    FIRST = auto()
+    TOP_LEFT = auto()
+    USER_DEFINED = auto()
```
nelimee, line 54: Wouldn't it be simpler to let the user define its origin implicitly? Basically, do we really need to have that enumeration? 
The current behaviour is to accept qubit coordinates as the user provides them. If the user provides qubits like in [`SquarePlaquette`](https://github.com/QCHackers/tqec/blob/4f047700c59514d17a6248c821119179da9126b5/src/tqec/plaquette/plaquette.py#L48):
https://github.com/QCHackers/tqec/blob/4f047700c59514d17a6248c821119179da9126b5/src/tqec/plaquette/plaquette.py#L48-L93
then the origin is implicitly `TOP_LEFT`. If the user provides qubits like
```python
class CenteredSquarePlaquette(Plaquette):
    def __init__(self, circuit: ScheduledCircuit) -> None:
        """Represents a centered square QEC plaquette

        By convention, the qubits are sorted as depicted in the text below:
        ```text
        |---------|
        | 1     2 |
        |         |
        | 3     4 |
        |---------|
        ```
        Sub-classes of this class should take that into account to apply operations
        on the correct qubits.
        This ordering is not to be confused with the temporal ordering of multi-qubit
        gates, for example in surface codes. It is a **qubit** ordering and has **no
        relation** with a temporal ordering.

        :param circuit: scheduled quantum circuit implementing the computation that the
            plaquette should represent.
        """
        super().__init__(self.get_data_qubits() + self.get_syndrome_qubits(), circuit)

    _data_qubits: list[PlaquetteQubit] = [
        PlaquetteQubit(Position(-1, -1)),
        PlaquetteQubit(Position(1, -1)),
        PlaquetteQubit(Position(-1, 1)),
        PlaquetteQubit(Position(1, 1)),
    ]
    _syndrome_qubits = [PlaquetteQubit(Position(0, 0))]
    # Everything else the same
```
then the origin is implicitly `CENTER`. 

Do we care about / need the information provided by this enumeration?

src/tqec/templates/orchestrator.py
```diff
@@ -110,7 +110,7 @@ def add_template(
         self._templates.append(template_to_insert.template)
         self._relative_position_graph.add_node(template_id, plaquette_indices=indices)
         self._maximum_plaquette_mapping_index = max(
-            self._maximum_plaquette_mapping_index, max(indices)
+            [self._maximum_plaquette_mapping_index] + indices
```
nelimee, line 113: This is a style preference, but I am not sure creating a whole new list is the best way to compute the maximum. Also, I find the previous version easier to read, but again, personal preference.

src/tqec/plaquette/plaquette.py
```diff
@@ -44,6 +58,18 @@ def qubits(self) -> list[PlaquetteQubit]:
...
+        match origin:
+            case PlaquetteOrigin.TOP_LEFT:
+                self._origin = Position(0, 0)
+            case PlaquetteOrigin.CENTRAL:
+                raise NotImplementedError()
+            case PlaquetteOrigin.FIRST:
+                self._origin = self._qubits[0].position
+            case PlaquetteOrigin.USER_DEFINED:
+                self._origin = self._qubits[origin_index].position
+
```
nelimee, line 72: Following another comment, if the origin is implicitly defined by the user input (i.e., always `(0, 0)`), this method may not be needed.

src/tqec/plaquette/plaquette.py
```diff
@@ -26,15 +32,23 @@ def __init__(self, qubits: list[PlaquetteQubit], circuit: ScheduledCircuit) -> N
...
         self._circuit = circuit
+        self._caclulate_origin(origin, origin_index)
 
     @property
-    def shape(self) -> Shape2D:
-        maxx = max(qubit.position.x for qubit in self.qubits)
-        maxy = max(qubit.position.y for qubit in self.qubits)
-        return Shape2D(maxx + 1, maxy + 1)
+    def origin(self) -> Position:
+        return self._origin
```
nelimee, line 51: Following another comment, if the origin is implicitly defined by the user input (i.e., always (0, 0)), this method may be replaced by
```suggestion
    def origin(self) -> Position:
        return Position(0, 0)
```

src/tqec/enums.py
nelimee: Also, this method would simplify (in my opinion) user input.

src/tqec/generation/circuit.py
```diff
@@ -22,7 +23,8 @@ def generate_circuit(
     This function requires that a few pre-conditions on the inputs are met:
     1. the number of plaquettes provided should match the number of plaquettes required by
        the provided template.
-    2. all the provided plaquettes should have the same shape.
+    2. all the provided plaquettes are rectengular. For '0' plaquettes, the shape is asumed to be
```
nelimee, line 26:
```suggestion
    2. all the provided plaquettes are rectangular. For '0' plaquettes, the shape is assumed to be
```

src/tqec/generation/circuit.py
```diff
@@ -70,20 +67,107 @@ def generate_circuit(
...
+
+            offset: Position = _find_offset(
+                all_scheduled_circuits,
+                plaquette_y,
+                plaquette_x,
+                len(line),
+                default_x_increment,
+                default_y_increment,
+            )
+            print(offset)
```
nelimee, line 84:
```suggestion
```

src/tqec/generation/circuit.py
```diff
@@ -70,20 +67,107 @@ def generate_circuit(
...
+    if len(final_circuit.raw_circuit.all_qubits()) == 0:
+        return Position(0, 0)
+    position = max(
+        (
+            Position(qubit.col, qubit.row)
+            for qubit in final_circuit.raw_circuit.all_qubits()
+        ),
+        key=lambda p: (p.y, -p.x) if above else (p.x, -p.y),
+    )
+    return Position(position.x, position.y)
```
nelimee, line 173: I have hard times understanding the usefulness of these functions. In my mind, with the pre-conditions we imposed, the global position of any plaquette origin should be 
```
(column_index * default_x_increment, row_index * default_y_increment)
```
with potentially a small offset to avoid to top-most and left-most plaquettes to have qubits with negative coordinates if we want to avoid that. 
Am I missing something?

nelimee:
> Should we use a linter? I'm personally using `black` but I've seen some code that was not according to `pep-8`

I am using [ruff](https://github.com/astral-sh/ruff), and format on each save. We should probably agree on some linter though, I agree.

> Where should I put the tests? And how should we name them? I'd prefer `test_{file_name}`.

From the [discussion on tests](https://github.com/QCHackers/tqec/pull/66#discussion_r1465488155), the format is `[file_name]_test.py` to have the test close by the code file.

src/tqec/enums.py
Gistbatch: My idea was to have this somewhat flexible, but simple setting would be fine with me. I mainly did this to see what would be working the best in the beginning.

src/tqec/templates/orchestrator.py
Gistbatch: I just changed it, as pylint suggested not using nested max calls

src/tqec/enums.py
nelimee: Do you have a situation in mind where the simple setting above is not enough and we actually need the information provided by `PlaquetteOrigin`? I tried to find one, but could not.

src/tqec/templates/orchestrator.py
nelimee: Who am I to contradict pylint ^^ I do not have that warning, but I am 100% following pylint, so I agree with your version. Do you have a link to the rule? I am interested by the rationale behind this pylint check.

src/tqec/generation/circuit.py
Gistbatch: The functions push a circuit to the top left of its two neighboring circuits that have already been placed.
From my understanding, using
```
(column_index * default_x_increment, row_index * default_y_increment)
```
implicitly assumes equally shaped, but rectangular plaquettes; am I wrong?
Say we have a plaquette of shape `(6,3)` and one of `(3,3)`, depending who is placed first, `defualt_x_increment` would be either 2 or 4?

nelimee: The formula above only assumes that the distance between 2 neighbouring plaquette origins is `default_x_increment` in the `x` dimension and `default_y_increment` in the `y` dimension. The actual plaquette shapes are not part of this formula, and 2 neighbouring plaquettes can even overlap if the user implemented them that way. 

Also, `default_{x,y}_increment` should be constant for the whole template. 

I am also having hard times thinking about examples of non-constant plaquettes. I will ask an example to @afowler, it might help me understand what should be done better.

Gistbatch: I wasn't ware they could overlap, I thought they could only have touching borders, but if this is the case, then your solution is definitely a lot simpler.
I was thinking of general tilings where something like [this](https://www.pngfind.com/pngs/m/70-705325_truncated-square-tiling-hexagon-and-square-tile-hd.png) would be possible.

src/tqec/templates/orchestrator.py
Gistbatch: It's https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/nested-min-max.html#nested-min-max-w3301. I don't know if I get it because it can't infer the type?

src/tqec/generation/circuit.py
nelimee: Let's wait for Austin's feedback on this. If I remember correctly, he said that overlapping could be a thing and that it was up to the user to check that the circuits provided are correct.
The tiling you gave as an example could be implemented with `default_x_increment == default_y_increment` no?

inmzhang: IMO `default_x_increment`/`default_y_increment` should be some properties specified by the templates and the plaquettes not aware of. All the mission of the templates should be providing a valid region of qubits and a pattern that the plaquettes can be placed on. Whether the plaquettes can be overlapping(BTW I think this can happen) should do nothing to the coordinating process.  The interop between the plaquettes on the templates only happens when validating the scheduled circuits, where the operation should be unique on every qubits at each moment.

inmzhang (commented):
Thanks for sharing your insights on this topic! And I have some comments that appear to align with Adrien's.

Gistbatch:
I'll wait for Austin to answer and implement your suggestions afterward

afowler:
On reflection, by definition, if we focus initially on architectures with
planar nearest neighbor couplings then it is always possible to keep gates
inside the boundary of a plaquette as anything that extends outside the
boundary can be put in a neighboring plaquette.

As mentioned on another thread, regular rectangular plaquettes of user
specified dimensions should be more that sufficient for the foreseeable
future.

When and only when we start lifting the assumption of planar nearest
neighbor, then you can have medium to long range interactions, but let's
leave that for another day, and probably another year.

Best,
Austin.

Gistbatch:
Given the clarification, let us stick to rectangular plaquettes and accept them overlapping.
This simplifies the circuit construction.
I'll do the following, including the other comments:
- simplify the circuit construction
- fix the origin to `Position(0,0)`
- See if the increment parameters can be easily integrated into the template
- write a test

src/tqec/templates/base.py
```diff
@@ -124,10 +137,18 @@ def shape_instance(self) -> BaseShape:
...
+        :returns: a displacement of the default increments in the x and y directions.
+        """
+        return Displacement(self._default_x_increment, self._default_y_increment)
+
 
 @dataclass
 class TemplateWithIndices:
-    """A wrapper around a Template instance and the indices representing the plaquettes it should be instanciated with."""
+    """A wrapper around a Template instance and the indices
+    representing the plaquettes it should be instanciated with."""
```
nelimee, line 151: The first line of this docstring was longer than expected from the default Python formatting, but as far as I am aware the documentation tools expect the **first line** to be a quick and short explanation of the class/function/method/... 
Separating that on 2 lines might break the documentation when setup.

src/tqec/templates/orchestrator.py
```diff
@@ -313,6 +315,7 @@ def _build_array(self, indices_map: tuple[int, ...]) -> numpy.ndarray:
         shape = self._get_shape_from_bounding_box(bbul, bbbr)
 
         ret = numpy.zeros(shape.to_numpy_shape(), dtype=int)
+        increments = numpy.zeros(shape.to_numpy_shape(), dtype=object)
```
nelimee, line 318: As increments are ultimately 2-dimensional tuples, it might be more "numpy" to have a 
```suggestion
        increments = numpy.zeros((*shape.to_numpy_shape(), 2), dtype=int)
```
and to accordingly change the other use of this array.

src/tqec/generation/circuit.py
```diff
@@ -54,6 +48,7 @@ def generate_circuit(
     # Index 0 is "no plaquette" by convention.
     _indices = list(range(len(plaquettes) + 1))
     template_plaquettes = template.instanciate(*_indices)
+    plaquette_increments = template.get_template_increments(*_indices)
```
nelimee, line 51: This ends up calling the same method twice under-the-hood. Maybe make the `build_array` method public (renaming it with a more descriptive name) and use it here?

src/tqec/templates/base.py
```diff
@@ -57,10 +62,18 @@ def __init__(self, shape: BaseShape) -> None:
         Each template should have a shape, represented by a Shape instance, and encoding
         how the template scales and its plaquettes.
 
+        The default increments define the distance between two plaquettes.
+        For example a default_x_increment of 2 means that two 2x2 plaquettes will share
```
nelimee, line 66: Shouldn't it be `3x3` plaquettes that share a common edge (i.e., common qubits)?

nelimee (commented):
Additionally to the in-code comments, it might be interesting to enforce somewhere that "all the `Template` instances managed by a given `TemplateOrchestrator` instance should have the same increments". 
Or did I miss something?

src/tqec/templates/base.py
Gistbatch: I was wondering that, too. But from the slides, we have plaquettes touching on two qubits, so wouldn't that make each plaquette 2x2?

src/tqec/templates/orchestrator.py
Gistbatch: I wasn't sure what's "pythonic", I can do this though

src/tqec/generation/circuit.py
Gistbatch: I realized this, too. My solution would have been to cache this, because I didn't want to change how the `instanciate` method works. I'll make it public but keep the others.

src/tqec/templates/base.py
nelimee: I though that and just checked. Seems like I was wrong: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html
I'm fine keeping it with a newline then, this comment is based on incorrect assumptions.

src/tqec/templates/orchestrator.py
nelimee: Let's maybe wait for another opinion on that. On my side, I always expect `numpy` arrays to contain native numbers (complex, float and int, mainly), which is the reason why I made this comment.

src/tqec/generation/circuit.py
nelimee: Maybe make the `_build_array` method "public" again by renaming it to `build_array` (or a more descriptive name) and:
```suggestion
    template_plaquettes, plaquette_increments = template.build_array(*_indices)
```
?

src/tqec/templates/base.py
nelimee: There is a qubit in the middle that is used to measure the stabiliser. On the qubit grid, the basic square plaquettes are 3x3. I think saying 2x2 might be more prone to errors by the user than 3x3. Maybe let's also wait for another opinion on that?

src/tqec/templates/orchestrator.py
Gistbatch: I went with yours because I think its more efficient. The intention is a little less clear, but I think that's fine

src/tqec/generation/circuit_test.py
inmzhang, line 1: The tests can be improved in the future to include more test cases.

src/tqec/templates/orchestrator.py
```diff
@@ -332,19 +349,30 @@ def build_array(self, indices_map: tuple[int, ...]) -> numpy.ndarray:
             ret[y : y + tshapey, x : x + tshapex] = template.instanciate(
                 *plaquette_indices
             )
-        return ret
+            increments[y : y + tshapey, x : x + tshapex] = template.get_increments()
+        return ret, increments
```
inmzhang, line 353: I do not get the point of returning the increments array. If we have the constraint that all the template within a orchestrator should have the same increments along each dimension, why do we need to compute and save this increments array?

Gistbatch: @nelimee, what do you think? I also felt like we were moving in circles with this one. We basically made the shape into the default increment.

nelimee: In the case of non-overlapping plaquettes, I agree that this seems like a reformulation of the original code. I am even questioning the lifting of the assumption right now. I do not have an answer to that for the moment, sorry.

inmzhang: I think the underlying assumption we should lift is that we need to know the "shape" of the plaquette. The template can be totally unaware of the "shape" of the plaquette to be well defined. And more, the plaquettes can also be well defined with the local coordinate system and the qubits offsets without the shape property.

nelimee: For @afowler comment on the Google group, it seems like we misunderstood the requirements. We only need to be able to represent rectangles plaquette. For a given QEC, all the plaquettes will be the same (i.e., the same rectangle, the same square, ...).
This was already possible with the `shape` implementation, but as noted above it is beneficial to remove completely the `shape` information in favour of an increment between plaquettes.

So to summarise, we "simply" want to have a `(increment_x, increment_y)`, global to a given QEC, in order to place correctly the plaquette circuits.

I am missing something else here?

I am sorry for misunderstanding the root of issue #34.

afowler:
+1 this summarizes the current plan well.

src/tqec/templates/orchestrator.py
Gistbatch: So if I understood correctly, I can remove again the `increments` information from the templates and move it to the the orchestrator?

afowler:
+1 yes.

src/tqec/templates/orchestrator.py
nelimee: Seems like the `increments` data is only needed in the `generate_circuit` function. 

As a side note, even though this has not been done yet and the class hierarchy does not reflect that, the `TemplateOrchestrator` class has been thought as a `Template` with the idea that we could, in the future, have `TemplateOrchestrator` orchestrating other `TemplateOrchestrator` instances. Meaning that we should not consider a `TemplateOrchestrator` instance to represent the template of a full QEC.

Gistbatch:
I've left the increment information in the templates as a sanity check. Should I remove them?

src/tqec/generation/circuit.py
```diff
@@ -70,20 +62,33 @@ def generate_circuit(
...
-                for qubit in scheduled_circuit.raw_circuit.all_qubits()
-            }
-            scheduled_circuit.map_to_qubits(qubit_map, inplace=True)
+
+            if plaquette_index > 0:
+                offset = (column_index * increments[0], row_index * increments[1])
+                plaquette = plaquettes[plaquette_index - 1]
+                qubit_map = _create_mapping(plaquette, scheduled_circuit, offset)
+                scheduled_circuit.map_to_qubits(qubit_map, inplace=True)
             all_scheduled_circuits.append(scheduled_circuit)
```
nelimee, line 76: In my opinion, the `if` statement should either enclose all the code in the `for`-loop or be removed.

src/tqec/generation/circuit.py
```diff
@@ -70,20 +62,33 @@ def generate_circuit(
...
 
     # Merge everything!
     return merge_scheduled_circuits(all_scheduled_circuits)
+
+
+def _create_mapping(
+    plaquette: Plaquette, scheduled_circuit: ScheduledCircuit, offset: tuple[int, int]
+) -> dict[cirq.Qid, cirq.Qid]:
+    origin = plaquette.origin
+    assert origin == Position(0, 0), "Only origin (0,0) is supported for now"
```
nelimee, line 86: Is this assert still needed?

src/tqec/generation/circuit.py
```diff
@@ -70,20 +62,33 @@ def generate_circuit(
...
+
+    qubit_map = {
+        # GridQubit are indexed as (row, col), so (y, x)
+        # Qubits are given relative to an origin, so we need to add the offset
+        qubit: qubit + (offset[1], offset[0]) + (origin.y, origin.x)  # type: ignore
```
nelimee, line 91: To help with the readability, I think it would be nice to have `offset` be a `Position`-like structure, for which the coordinates should be explicit (i.e., `offset.x` instead of `offset[0]`).

nelimee (commented):
Now that the `increments` are constant for a given QEC experiment, it would probably be simpler to provide them once in `generate_circuit` and remove the increments from the template. This would simplify the API and remove some checks (all increments are equal) that would not be relevant anymore.

Gistbatch:
@nelimee, if this is fine now, I'll resolve the merge conflict and prepare it to be merged

src/tqec/plaquette/plaquette.py
```diff
@@ -26,15 +30,18 @@ def __init__(self, qubits: list[PlaquetteQubit], circuit: ScheduledCircuit) -> N
             coordinate system.
         :param circuit: scheduled quantum circuit implementing the computation that the
             plaquette should represent.
+
+        :raises ValueError: if the number of qubits doesn't match the number of qubits
         """
+        # if len(qubits) != len(circuit.raw_circuit.all_qubits()):
+        #     raise ValueError("Number of qubits doesn't match number of circuit qubits")
+
```
nelimee, line 38: Either the commented code should be uncommented or the docstring part about `:raises:` should be removed.

src/tqec/templates/base.py
```diff
@@ -57,10 +62,18 @@ def __init__(self, shape: BaseShape) -> None:
...
+        a common edge.
+
         :param shape: the underlying template shape.
+        :param default_x_increment: default increment in the x direction between two plaquettes.
+        :param default_y_increment: default increment in the y direction between two plaquettes.
         """
         super().__init__()
         self._shape_instance = shape
+        self._default_x_increment = default_x_increment
+        self._default_y_increment = default_y_increment
```
nelimee, line 76: These might be directly stored as a `Displacement`. It does not bring a lot of value to do so compared to the current state though.

src/tqec/templates/orchestrator.py
```diff
@@ -313,7 +327,6 @@ def build_array(self, indices_map: tuple[int, ...]) -> numpy.ndarray:
         shape = self._get_shape_from_bounding_box(bbul, bbbr)
 
         ret = numpy.zeros(shape.to_numpy_shape(), dtype=int)
-        # tid: template id
```
nelimee, line 316: This comment was here to explain the variable name used in the code and avoid super-long names that would make the code harder to read. Is there a specific reason why you removed it?

nelimee (approved):
I still stand by my previous comment
> Now that the `increments` are constant for a given QEC experiment, it would probably be simpler to provide them once in `generate_circuit` and remove the increments from the template. This would simplify the API and remove some checks (all increments are equal) that would not be relevant anymore.

but I think this can be though about in a follow-up issue.

src/tqec/templates/base.py
nelimee: This is just a comment, I am OK with the current state, and I let you decide if changing it is worthwhile or not.

src/tqec/plaquette/plaquette.py
Gistbatch: I was wondering why this wasn't working, but I don't have the time atm. to look into it.

src/tqec/templates/orchestrator.py
Gistbatch: This was an accident

src/tqec/plaquette/plaquette.py
nelimee: Turns out it is because the `cirq.Circuit` only considers the qubits on which gates are applied and some `ScheduledCircuit` instances have `cirq.Circuit` instances with unused qubits.
I found a way to make that work, the check was a good idea. I will propose patches.

src/tqec/plaquette/plaquette.py
```diff
@@ -3,7 +3,7 @@
 from tqec.enums import PlaquetteOrientation
 from tqec.plaquette.qubit import PlaquetteQubit
 from tqec.plaquette.schedule import ScheduledCircuit
-from tqec.position import Position, Shape2D
+from tqec.position import Position
```
nelimee, line 6: To make the check below possible:
```suggestion
from tqec.position import Position
from tqec.exceptions import TQECException
```

src/tqec/plaquette/plaquette.py
```diff
@@ -26,15 +30,17 @@ def __init__(self, qubits: list[PlaquetteQubit], circuit: ScheduledCircuit) -> N
             coordinate system.
         :param circuit: scheduled quantum circuit implementing the computation that the
             plaquette should represent.
+
         """
+        # if len(qubits) != len(circuit.raw_circuit.all_qubits()):
+        #     raise ValueError("Number of qubits doesn't match number of circuit qubits")
+
```
nelimee, line 37:
```suggestion
        plaquette_qubits = {qubit.to_grid_qubit() for qubit in qubits}
        circuit_qubits = set(circuit.raw_circuit.all_qubits())
        if not circuit_qubits.issubset(plaquette_qubits):
            wrong_qubits = plaquette_qubits.difference(circuit_qubits)
            raise TQECException(
                f"The following qubits ({wrong_qubits}) are in the provided circuit "
                "but not in the list of PlaquetteQubit."
            )
```

nelimee (approved):
This PR works as expected. Let's merge it. Should I do it, or do you do it @Gistbatch ?

src/tqec/generation/circuit.py
nelimee: Just realized that you were using `plaquettes` and not `plaquette_circuits` here. I read too quickly. The `if` statement was indeed needed, and my comment was wrong.

PR #81, opened 2024-01-28
Move logical qubit along the Z observable First Attempt

inmzhang:
This pull request adds an example demonstrating the movement of the logical qubit along its Z-observable, which addresses issue #50.

I've found that moving data under the current architecture isn't a straightforward task (or it might be an issue with my implementation). The circuits evolve as data moves, which poses significant challenges in defining the plaquettes.

As of now, the detector slices from the output stim circuits appear to be correct to me. However, I'm wondering if there might be a more concise or elegant way to implement this task. I'm open to suggestions from the team and am prepared to make changes in the near future.

Additionally, I've realized that this example could serve as a valuable starting point for further discussions about the tool's design.


nelimee:
I just read the notebook, ran it and replicated the results, good job! I agree with you that the library should offer more convenience functions / classes to address more complex codes.

> Additionally, I've realized that this example could serve as a valuable starting point for further discussions about the tool's design.

This was really the principal goal of the issues #50, #51, #52 and #53: find the friction points that makes the library hard to use, as well as potential bugs.

I have 2 main remarks.

## The implementation

I think we can benefit from viewing the generation of a QEC differently: instead of having to think about the whole experiment from the start and define everything accordingly, define each part of the code piece by piece. By that I mean that we could:
1. Generate the initial and last memory experiments using the **exact same** code as in [logical_qubit_memory_experiment.ipynb](./notebooks/logical_qubit_memory_experiment.ipynb). 
2. Generate the extended memory using the **exact same** code as in [logical_qubit_extended_memory_experiment.ipynb](notebooks/logical_qubit_extended_memory_experiment.ipynb).
3. Generate the "glue" circuits between these three QEC steps.
4. Assemble each generated **circuit**.

With the methodology above, the only missing piece that will need to be addressed is how to correctly offset the qubits before merging each circuit in order to have quantum circuits on matching qubits.

Basically, just like the other parts of the code, we define re-usable base situations that will be often encountered in real-codes, and make them easy to generate, only leaving the glue to implement to the user of the library.

In particular, the view above means that we will instanciate several different templates (instances of `TemplateOrchestrator`). But as long as we can offset the qubits correctly, I do not see any problem with that.

## The `Plaquette` classes

I am not satisfied with the different `Plaquette` classes in `tqec.plaquette.library` module. They are quite cumbersome and error-prone to use. What is your opinion on that now that you used them? And if you agree with the above opinion, do you have any idea for improvement?

nelimee (commented):
Previous comment is the review

inmzhang:
Thanks for your remarks!

> In particular, the view above means that we will instanciate several different templates (instances of TemplateOrchestrator).

I agree with that. Actually, this is what I tried to do at first. However, the implementation complexity heavily came from gluing  different parts of the circuits together and we didn't have the template orchecstrator library(assets) yet, then I chose to implement in the simple brute-force way.

> What is your opinion on that now that you used them? And if you agree with the above opinion, do you have any idea for improvement?

The `Plaquette` classes are indeed heavy to use in the code. However, when I think about the use cases after the frontend is ready, we will hopefully define complicated plaquettes on the web with clicking and dragging. The cumbersome definition of plaquettes should only be done for thoes common plaquettes, which will be tested carefully. As for how to test the circuits associated to the plaquettes, I think we can use something like [`stim.Circuit.has_flow`](https://github.com/quantumlib/Stim/blob/main/doc/python_api_reference_vDev.md#stim.Circuit.has_flow) to verify it does have the desired stabilizer flow.

---

And here are some problems I came across during the implementation in more detail, followed by the idea to tackle them:

1. How to define a plaquette with empty circuit when all the qubits of it are idling: add `with_empty_circuit()` to the `Plaquette` class, which returns a new plaquette with the same shape but a empty scheduled circuit.
2. How to easily add more measurement records to a detector gate already defined within the plaquettes, which will happen during the circuit gluing process(Or maybe anywhere else during more complicated logical computation): add something like `modify_records()` to the `DetectorGate`. Though this is still cumbersome to specify the measurements to be appended to the detector, it will hopefully be made handy at the frontend side. BTW, I think making detector/observable a `cirq.Operation` rather than `cirq.Gate` is more appropriate.
3. How to easily define the glue part of different standard circuit layers: not aware of a simple solution for this yet since the complex overlapping structure between different circuit layers. However, the work at the frontend side will definitely make this part easier.
4. How to verify the correctness of the built circuits: the most fundmental test we can/should run is that using `stim` to verify the validity of the definition of the detectors/observables with methods like `stim.Circuit.detector_error_model`. The simulation as integration tests should be run to verify it further.

nelimee:
> I agree with that. Actually, this is what I tried to do at first. However, the implementation complexity heavily came from gluing different parts of the circuits together and we didn't have the template orchecstrator library(assets) yet, then I chose to implement in the simple brute-force way.

Interesting! I though that this would be simpler. 

> The Plaquette classes are indeed heavy to use in the code. However, when I think about the use cases after the frontend is ready, we will hopefully define complicated plaquettes on the web with clicking and dragging. The cumbersome definition of plaquettes should only be done for thoes common plaquettes, which will be tested carefully.

I agree with your comment. 

> As for how to test the circuits associated to the plaquettes, I think we can use something like [stim.Circuit.has_flow](https://github.com/quantumlib/Stim/blob/main/doc/python_api_reference_vDev.md#stim.Circuit.has_flow) to verify it does have the desired stabilizer flow. 

Seems like a good idea.

I agree with most of your listed points, except for
> How to easily add more measurement records to a detector gate already defined within the plaquettes, which will happen during the circuit gluing process(Or maybe anywhere else during more complicated logical computation): add something like modify_records() to the DetectorGate.

for which I do not see why we would like to add measurements after plaquette creation. Do you have an example in mind?

Also, I think we should take a little bit of time to think about the changes that are required in order to simplify this task. They may be quite important, and may also impact already existing code.

inmzhang:
> Do you have an example in mind?

0f7717613a4a2c6b57bc8fbf0e815a9

As an example, at the moment we measure out the data qubits along the path, a square plaquette turned into a rounded plaquette. If we do not change the definition of the plaquette in the glue circuit, we need to accout for the two data qubit measurements in the detector gate of the rounded plaquette. Also note that this will need to be added only for the detector in the glue circuit and the following detectors in the memory block remain unchanged.

This is also what I did with the circuit modifier `modify_detector_then_repeat` in the notebook.

Is there any better idea to address this problem instead of adding additional measurements to the exsiting plaquettes(Or maybe we can change the definition of the plaquettes themself in the glue layer)?

nelimee:
> Or maybe we can change the definition of the plaquettes themself in the glue layer 

I was thinking about that yes. Create a specific plaquette for that specific purpose, and use it once, in the needed timeslice.

I will try to implement what I has in mind today to help me understand the limitations you encountered.

afowler:
As another suggestion, note that I think it is completely okay for
neighboring plaquettes to have gates that are common, meaning you detect
that two identical gates overlap and only put one in the circuit. For
example, in a sea of data qubit measurements, in my mind, it would be
perfectly okay to use plaquettes that have four measurements, one on each
corner, and tile them, meaning neighboring pairs of plaquettes have two
measurements in common, only one of which makes it to the circuit.

nelimee:
> As another suggestion, note that I think it is completely okay for neighboring plaquettes to have gates that are common, meaning you detect that two identical gates overlap and only put one in the circuit. For example, in a sea of data qubit measurements, in my mind, it would be perfectly okay to use plaquettes that have four measurements, one on each corner, and tile them, meaning neighboring pairs of plaquettes have two measurements in common, only one of which makes it to the circuit.

I had to do that for resets and measurements already. For reference, the merging is done here:

https://github.com/QCHackers/tqec/blob/655c86b85260ded3e6268432d00c983d8b210e99/src/tqec/plaquette/schedule.py#L342-L373

and plaquettes are defined with the `mergeable_tag` in the plaquette library (example for `XXPlaquette`, the other plaquettes are similar):

https://github.com/QCHackers/tqec/blob/655c86b85260ded3e6268432d00c983d8b210e99/src/tqec/plaquette/library/xx.py#L46-L62
https://github.com/QCHackers/tqec/blob/655c86b85260ded3e6268432d00c983d8b210e99/src/tqec/plaquette/library/xx.py#L139-L151

nelimee:
I got the time to look at the problem, and though it was a good idea to open another draft PR to discuss about potential ways to simplify the implementation you provided.

I would be fine merging that PR as it is solving #50 and the resulting implementation is working as expected. We can then iteratively improve the `tqec` library to simplify the QEC creation process showcased in the notebook you wrote.

What do you think @inmzhang ? Also, I am eager for other opinions, either on the code here or about the plan laid above (merging this one and then improving in follow-up PRs).

inmzhang:
Great, I think it's fine to merge this PR as a starting point of further iteration. And I'll review #100 in a few hours.

nelimee (approved):

PR #83, opened 2024-01-29
Linting and refactoring

smburdick:
This is a continuation of #41 which needed to close due to local Git file corruption.

* Perform file linting -- address https://github.com/QCHackers/tqec/issues/85
* Replace GET with POST in download button
* Refactored several classes


rryoung98:
Thanks @smburdick  ill review in a bit

PR #88, opened 2024-01-31
update eslint, allow tabs

rryoung98:
Updated eslint, relaxed rules for indentation, tabs, and react. We can enforce them later

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -18,6 +18,11 @@
 	},
 	"plugins": ["react", "react-hooks"],
 	"rules": {
+		"no-tabs": "off",
+		"react/jsx-indent": "off",
+		"react/jsx-indent-props": "off",
+		"react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+		"comma-dangle": ["error", "never"]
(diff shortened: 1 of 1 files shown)


frontend/.eslintrc.json
```diff
@@ -18,6 +18,11 @@
 	},
 	"plugins": ["react", "react-hooks"],
 	"rules": {
+		"no-tabs": "off",
+		"react/jsx-indent": "off",
+		"react/jsx-indent-props": "off",
+		"react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+		"comma-dangle": ["error", "never"]
 		//add customize rules here as per your project's needs
```
smburdick, line 26: This comment is not needed.

smburdick (approved):
LGTM. I am personally ok with only using .jsx for React files, but we didn't start with that for this app (I did rename App.js to App.jsx in #83 , though)

frontend/.eslintrc.json
rryoung98: Changes made

rryoung98:
Yea, we can switch .js / .jsx later if we want, just don't want our eslint to bog us down 😄

PR #97, opened 2024-02-01
Update `tqec.templates.constructions` classes to the new scale definition `k`

nelimee:
Correct the constructors of `tqec.templates.constructions.{ScalableQubitSquare,ScalableCorner}`.


inmzhang (commented):
I'm not sure whether we need to change the `dim` to `k` here. Actually, I left these unchanged on purpose when I changed the definition of `k`.

On the one hand, we define a `ScalableQubitRectangle` with `width` and `height`, which is the raw dimension of the underlying shape. For consistency consideration, it might be beneficial to define the raw dimension/size in construction. On the other hand, it can be more intuitive for the user to use `dim` instead of `k`. The underlying scale `k` of the shape is responsible for imposing the even-size constraint under the hood and only need to be used during scaling.

nelimee:
Currently, trying to execute
```py
from tqec.templates.constructions.corner import ScalableCorner

ScalableCorner(3)
```
raises an error that is not really descriptive:
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 ScalableCorner(3)

/[path]/src/tqec/templates/constructions/corner.py in line 41, in ScalableCorner.__init__(self, dim)
     13 def __init__(self, dim: int) -> None:
     14     """A scalable corner template.
     15 
     16     This corner template can be used to move an error-corrected qubit to another
   (...)
     37     :param dim: dimension (code distance - 1) of the initial error-corrected qubit.
     38     """
     39     _templates = [
     40         # 0
---> 41         TemplateWithIndices(ScalableRectangle(dim, 1), [0, 1]),
     42         TemplateWithIndices(ScalableRectangle(1, dim), [2, 0]),
     43         TemplateWithIndices(ScalableAlternatingSquare(dim), [3, 4]),
     44         TemplateWithIndices(ScalableRectangle(1, dim), [0, 5]),
     45         TemplateWithIndices(FixedRaw([[0]]), [2]),
     46         # 5
     47         TemplateWithIndices(ScalableRectangle(dim, 2), [3, 4]),
     48         TemplateWithIndices(FixedRaw([[0]]), [6]),
     49         TemplateWithIndices(ScalableRectangle(dim, 1), [7, 0]),
     50         TemplateWithIndices(ScalableRectangle(1, dim), [2, 0]),
     51         TemplateWithIndices(
     52             ScalableAlternatingCornerSquare(dim, CornerPositionEnum.LOWER_LEFT),
     53             [
     54                 3,
     55                 4,
     56                 8,
     57                 9,
     58                 11,
     59             ],
     60         ),
     61         # 10
     62         TemplateWithIndices(ScalableRectangle(2, dim, scale_width=False), [9, 8]),
     63         TemplateWithIndices(ScalableAlternatingSquare(dim), [9, 8]),
     64         TemplateWithIndices(ScalableRectangle(1, dim), [10, 0]),
     65         TemplateWithIndices(ScalableRectangle(dim, 1), [0, 12]),
     66         TemplateWithIndices(FixedRaw([[0]]), [12]),
     67         # 15
     68         TemplateWithIndices(ScalableRectangle(dim, 1), [0, 12]),
     69     ]
     70     relative_positions = [
     71         (0, ABOVE_OF, 2),
     72         (1, LEFT_OF, 2),
   (...)
     83         (14, LEFT_OF, 15),
     84     ]
     85     pinned_corners = [
     86         ((6, CornerPositionEnum.LOWER_LEFT), (9, CornerPositionEnum.UPPER_RIGHT)),
     87         ((4, CornerPositionEnum.UPPER_RIGHT), (2, CornerPositionEnum.LOWER_LEFT)),
     88         ((14, CornerPositionEnum.UPPER_RIGHT), (11, CornerPositionEnum.LOWER_LEFT)),
     89     ]

File /[path]/src/tqec/templates/scalable/rectangle.py:23, in ScalableRectangle.__init__(self, width, height, scale_width)
      7 def __init__(
      8     self, 
      9     width: int, 
     10     height: int, 
     11     scale_width: bool | None = None,
     12 ) -> None:
     13     """Rectangle with scalable width **or** height.
     14 
     15     A scalable rectangle can only scale its width **or** height, but not both.
   (...)
     21         are even and equal, the width will be scaled by default.
     22     """
---> 23     assert (width % 2 == 0) or (height % 2 == 0), "At least one dimension must be even to be scalable!"
     24     # Determine which dimension to scale
     25     if scale_width is None:

AssertionError: At least one dimension must be even to be scalable!
```
In my opinion we should at least fix that by checking the input `dim`. But if we know from the beginning that `dim` should be even, why not baking that directly into the API? Also, this would harmonise the API between nearly all the `Template`-like instances.

Thinking about `ScalableQubitRectangle`, I was a little bit too quick on that and I think we should replace `width` and `height` by `k_width` and `k_height`, for the same reasons explained above.

Also, this raises a more fundamental question: the dimension/scale provided at construction is kind of redundant if we impose a `to_scale` call. I can see two ways of "removing" that:
1. remove the `dim`/`scale`/`k` parameter at construction, define the template internally with a valid "standard" dimension, and ask to the user to systematically use the `to_scale` method.
2. keep the `dim`/`scale`/`k` parameter in the construction API, do the same as point 1. and finish by calling `to_scale(k)` in the constructor.

nelimee:
> The problem with that is a ScalableQubitRectangle can have an odd-size width/height.

It can, but does it makes any sense? Having an odd-sized width/height does not improve the number of errors the QEC can detect/correct w.r.t the even size just smaller. 

> I agree with your opnion and prefer the first way. For the ScalableQubitRectangle, we can specify only the fixed dimension during construction while leaving the scalable dimension to be determined by the later call of scale_to.

I agree with that, it seems like the correct way of doing, I'll open an issue.

inmzhang:
> It can, but does it makes any sense? Having an odd-sized width/height does not improve the number of errors the QEC can detect/correct w.r.t the even size just smaller.

You are right. I forgot the rectangle here is used for a qubit. Then I think it's totally fine to make the fixed dimension specified by `k`. 

For another consideration, there are circumstances where we want to scale both of the dimension of a rectangle but with a fixed aspect ratio(for example, when study the surface code for the biased-noise). we can make it more general with an option to scale both of the dimension with a fixed aspect ratio, but this needs a change to the underlying shape implementation.

nelimee:
> For another consideration, there are circumstances where we want to scale both of the dimension of a rectangle but with a fixed aspect ratio(for example, when study the surface code for the biased-noise). we can make it more general with an option to scale both of the dimension with a fixed aspect ratio, but this needs a change to the underlying shape implementation.

Let's open an issue to track that and discuss it there.

nelimee:
For the moment, the `constructions` templates are directly created at the correct scale. 
I can re-write that to create them at the simpler scale and then call `scale_to`. It seems more robust to me, but also I am wondering if it is needed at all.

afowler:
My vote goes for no rectangular fixed aspect ratio templates. This would
get very complex very quickly. In real hardware, you balance out the error
rates and I think in practice we are going to only need square templates.
My vote also goes for using k as originally defined. I think the user can
cope with the idea that the dimensions of the template are twice this and
this enforces even dimensions.

inmzhang (approved):

PR #105, opened 2024-02-02
Change tab rule to use 4 spaces; linting files

smburdick:
I am in favor of the [traditional approach](https://www.linkedin.com/pulse/striking-balance-2-spaces-per-tab-vs-narrow-fonts-coding-jatasra-ly98f#:~:text=Traditionally%2C%20coding%20standards%20have%20favored,organized%20and%20easy%20to%20read.) of using four spaces per tab; I would be open to using two instead.


rryoung98 (approved):
sounds good. lets merge it.

PR #108, opened 2024-02-02
Refactor of the `Template` hierarchy

### LINKED ISSUE #107: Refactor of the `Template` hierarchy
The current `Template` hierarchy has several long-standing issues that were not annoying enough to be worth refactoring. Due to recent developments in #100, these become annoying enough. This issue lists all the problems with the `Template` hierarchy.

## `TemplateOrchestrator` is not a `Template`

The goal of the `Template` hierarchy is to have a very expressive way of constructing templates. The idea behind having a root `Template` class and sub-classes that builds upon it was to be able to decouple the complexity of building more advanced templates.

The fact that the `TemplateOrchestrator` is not a `Template` was not an issue until now because it was:
1. the only class that was supposed to be a `Template` without actually being one and,
2. behaving just like a `Template`.

This is still not an issue per se, due to the fact that Python is duck-typed ("if it walks like a duck, quacks like a duck and fly like a duck, then surely it must be a duck"). But the semantic of the class is unclear.
Also, the interfaces of `TemplateOrchestrator` and `Template` are not exactly the same, which breaks duck-typing in some cases.

## `Template` instances might not need a `Shape`

> [!WARNING] 
> Here `Shape` is **not** in the `numpy` sense, it is **not** a n-ary tuple of integers, it is an instance of the [`BaseShape`](https://github.com/QCHackers/tqec/blob/4a2a03d1bd22a3746cadf92114492754c9e6048e/src/tqec/templates/shapes/base.py#L16) class.

The example of the `TemplateOrchestrator` is quite obvious: we would like it to be a `Template`, but it has no definite `shape` and does not need one.
This raises questions about the usefulness of the `shape` parameter when constructing a `Template`.

## `Template` interface is lacking a few interesting things

The [`expected_plaquettes_number`](https://github.com/QCHackers/tqec/blob/4a2a03d1bd22a3746cadf92114492754c9e6048e/src/tqec/templates/orchestrator.py#L394) property has been implemented in the `TemplateOrchestrator` class but not in the generic `Template` sub-classes. It would be convenient to have it in all `Template` instances.

nelimee:
This PR refactors the `Template` hierarchy to remove a few long-standing pain points.

Fixes #107.


nelimee:
@inmzhang following your comment in #70 you might want to comment on #107 to list the potential refactoring you were mentioning.

inmzhang:
Great, I'll look into this and related issues tomorrow.

src/tqec/templates/orchestrator.py
```diff
@@ -48,7 +49,7 @@ def get_corner_position(
     )
 
 
-class TemplateOrchestrator(JSONEncodable):
+class TemplateOrchestrator(Template):
```
inmzhang, line 52: Now the `TemplateOrchestrator` is a `Template`, maybe a name like `ComposedTemplate` is better.

src/tqec/templates/orchestrator.py
```diff
@@ -48,7 +49,7 @@ def get_corner_position(
...
     def __init__(self, templates: list[TemplateWithIndices]) -> None:
```
inmzhang, line 53: Instead of wrapping a `Template` with an extra `TemplateWithIndices`, I prefer reserving the plaquette indices as a property of the template itself which can be changed with something like `template.with_indices(...)`. This makes the `TemplateOrchestrator` behave more like a composer of `Template`.

src/tqec/templates/shapes/base.py
inmzhang, line 1: I wonder the necessity of having underlying shape instances. It appears that these shape instances serve little purpose beyond their potential reuse when defining both a scalable template and a fixed template with the same underlying shape. However, we can simply make every `Template` inherently scalable and introducing a frozen property to designate it as fixed. What's your opinion on that, did I miss anything else?

src/tqec/generation/circuit.py
```diff
@@ -6,11 +6,11 @@
...
 from tqec.plaquette.schedule import ScheduledCircuit, merge_scheduled_circuits
 from tqec.position import Shape2D
-from tqec.templates.orchestrator import TemplateOrchestrator
+from tqec.templates.base import Template
 
 
 def generate_circuit(
-    template: TemplateOrchestrator,
+    template: Template,
     plaquettes: list[Plaquette],
```
inmzhang, line 14: We can make the type of the plaquettes as `list[Plaqutte] | dict[int, Plaqutte]` to make it more flexible.

inmzhang (approved):

src/tqec/templates/orchestrator.py
nelimee: I need to think a little bit more about that. My initial opinion is that `ComposedTemplate` is too generic and not descriptive enough. 
The example I have in mind is the soon-to-be-included `TemplateStack` (see in #100). In my mind, it is a "composed" template in the sense that it is a composition of templates, but will not be a subclass of `TemplateOrchestrator`.
I am fine with `TemplateOrchestrator` staying, but I understand that it might be more meaningful to have a name that finishes with `Template` to make the inheritance obvious.

src/tqec/templates/orchestrator.py
nelimee: This is a good remark. Again I would need to think deeper into it, but at first glance the current approach does not bother me too much because:
- it is simple,
- it keeps `Template` as a *description* of something, that can then be instantiated at will with concrete indices,
- the `TemplateWithIndices` structure is not visible from the public API.

I would be fine having a `Template.with_indices(...)` returning a `TemplateWithIndices` instance, as a way of writing that part in a more readable manner.

But storing indices within `Template` instances would break the "a `Template` is a description that can be instantiated at will".

Also I realise that something might be unclear with the current code: the indices provided as second argument to `TemplateWithIndices` are not the plaquette indices that are used to instantiate the `Template` instance given as first argument. They are indices that will be used on the list of plaquette indices provided to the `instanciate` method. So `TemplateWithIndices` is not storing any plaquette indices, only how to index in the list of plaquette indices that will eventually be provided to `instanciate`.
With an example,
https://github.com/QCHackers/tqec/blob/4a2a03d1bd22a3746cadf92114492754c9e6048e/src/tqec/templates/constructions/qubit.py#L24-L25
means that:
- if the `instanciate` method is called with `[0, 1, 2, 3, 4, 5, 6]` (which is nearly always the case), the 3rd and 4th indices from that list will be recovered: `3` and `4`,
- if the `instanciate` method is called with `[0, 2, 4, 6, 1, 3, 5]` (which might make sense in some context), the 3rd and 4th indices from that list will be recovered: `6` and `1`.

So having these indices in the `Template` description (here, `TemplateOrchestrator`) does not break point 2. above: these are not the indices the `Template` will be instantiated with, only a way to know which plaquette indices should be used.

src/tqec/templates/shapes/base.py
nelimee: I fully agree. Will try to do that on Monday and see if there are any implications I did not think about.

src/tqec/generation/circuit.py
nelimee: Yes, I also thought about that when starting to use the `TemplateStack` in #100. I agree, and it should not be too hard to add.

nelimee: Note to myself: the `TemplateStack` requires to provide empty plaquettes anyways as it does not simplify out the plaquettes. This idea is still good, but is more about "future proofing" and less about immediate need, except if I miss something.

afowler:
minor comment: instanciate -> instantiate

src/tqec/templates/orchestrator.py
nelimee: A few ideas for names:
1. `PlanarTemplate` (miss the "several templates together" part),
2. `TemplateOrchestratorTemplate` (in my opinion, not clearer than `TemplateOrchestrator` but longer),
3. `ComposedTemplate` (with my opinion above),
4. `TemplateOrchestrator` (not ending in "Template", which could help, if present, making the fact that it is a `Template` sub-class clearer).

Any feelings / other proposition?

src/tqec/generation/circuit.py
nelimee: Should be done!

src/tqec/templates/shapes/base.py
nelimee: Removing the `BaseShape` instance has the following initial effects:
- The `AtomicTemplate` class is not needed anymore (as its only purpose was to store the `BaseShape` instance and delegate most of the `Template` API to this stored instance).
- Refactor of most of the template implementations.

I think this is worth its own PR. But before starting a new PR, here is what I remember from my design decisions at the time.

I initially wanted `FixedTemplate` subclasses (e.g., `FixedAlternatingSquare`) to inherit from their `ScalableTemplate` counterpart (here, `ScalableAlternatingSquare`) and simply re-define the `scale_to` method to return `self` without scaling anything. The main issue I had with that were:
- a `FixedAlternatingSquare` should not have `ScalableTemplate` in its hierarchy (at least it made no sense to me at the time, now I kind of do not care),
- having `ScalableTemplate` in the whole `TemplateHierarchy` calls for a "`NonScalableTemplate`" class somewhere, or `FixedTemplate` as I called it back then, and it makes sense that fixed templates such as `FixedAlternatingSquare` are subclasses of `FixedTemplate`.

The initial implementation has been done by using multiple-inheritance: a fixed template was inheriting its scalable counterpart and was a subclass of `FixedTemplate`, mostly used as a tag. But:
- multiple inheritance is (close to) an anti-pattern in most languages,
- multiple inheritance is annoying to deal with in Python,
- `ScalableTemplate` was still in the hierarchy of fixed templates.

I thought of 2 ways to solve that issue:
1. Make 2 separate branches in the inheritance hierarchy and copy-paste the code that instantiate the integer array for each branch.
2. Make 2 separate branches in the inheritance hierarchy and factor out the code that instantiate the integer array into a separate class.

I ended up doing 2. which is the current state with the `BaseShape` class hierarchy.

src/tqec/templates/orchestrator.py
inmzhang: My thoughts on the naming hierarchy are as follows:

- `Template`: serves as the foundational class.
- `ComposedTemplate`: represents a template that is constructed from multiple non-overlapping templates arranged on a 2D plane.
- `StackedTemplate`: denotes a template that originates from several templates layered in 3D space, though it is designed to be viewed solely from a top-down perspective to project it onto 2D plane.

Once the docstrings for the classes are clearly written, confusion should be minimal. However, the decision rests with you, as I might not fully grasp the intentions behind your original design.

nelimee: > However, the decision rests with you, as I might not fully grasp the intentions behind your original design.

Original design is not as important as current design, and your opinion is, in my opinion, worth more than mine here as you are not biased by the original design decisions (and I am).

Let's go with `ComposedTemplate` then. If my points above did not make you change your mind then it means they are not strong enough to keep the original name.

src/tqec/templates/shapes/base.py
nelimee: I can see a path forward that may also solve the scaling issue in #100 by allowing more expressive scaling.
It may require a lot of changes though, so I will try it in another PR.

nelimee: #114 is open to demonstrate what I had in mind on that matter.

nelimee:
@inmzhang Should I merge that PR and open am issue/discussion to discuss about the comment you made on indices being part of `Template` (https://github.com/QCHackers/tqec/pull/108#discussion_r1477087558)? Or do you prefer to have that solved directly in this PR?

inmzhang:
It's fine to open a new thread for that.

inmzhang (approved):

PR #111, opened 2024-02-05
Replace the gate definition with operations

inmzhang:
This is a large refactor of the codebase. I try to lift the `ShiftCoordsGate`, `RelativeMeasurement`, `DetectorGate` and `ObservableGate` to the `cirq.Operation`. We can have the following benefits from the refactoring:

1. These operations are just annotations in the circuit that do not depend on the qubits they applied to. With the gate implementation, we have to choose some qubit to apply the gates on. This is a cumbersome constraint we have to deal with when we build circuit[^1].
2. Simplify the implementation. Since we do not to deal with the messy `Gate.on()` method anymore, the implementation will be simple and clean.
3. Be consistent with `stimcirq`, which defined these annotations as `cirq.Operation`s.

However, during the implementation, I found some caveats and limitations:

1. We can not easily add the `cirq.VirtualTag` tag to the operation when initializing it. Therefore, I define some factory methods to handle it. However, this will require the users to add `cirq.VirtualTag` by themselves if they use the raw operation classes and might turn out to be a rabbit hole.
2. `cirq.Operation` with tags will become a `cirq.TaggedOperation`. And if we want to get the underlying operation we need to use `op.untagged` to get it.
3. Under current implementation, the `qubits` property of the operations is an empty tuple. When we add the operation directly to a `cirq.Circuit`, the default insert strategy will move the operation to the head of the circuit, which will be a disaster. There are two options to handle that: 1. attach some qubits to the operation's `qubits` property, however, this kind of violates the princeple of "annotation should be independent of the qubits it applied on"; 2. Always containerize the operations with a `cirq.Moment` to pin them at the certain timestep in the circuit. Currently, I use the 2nd approach in the refactoring.

For some other changes should be noted in the refactoring:

1. All the `_stim_conversion_` in the definition were removed and the operations were converted to those correspondences defined in the `stimcirq` during circuit transformation.
2. `compute_global_measurements_lookback_offsets` was moved out of the operation definition and used in a more functional way during circuit transformation.
3. `fill_in_global_record_indices` was renamed to `transform_to_stimcirq_compatible` which covered the two main functionality of the transformer: 1. fill in the global record indices to make it compatible with the `stimcirq`'s `relative_key` definition; 2. convert the annotation operations to those defined in `stimcirq`.

There are definitely many improvements can be made to the implementation and documentation. I want to get some opinions or suggestions before moving on.

[^1]: Take [adding a ShiftCoordsGate](https://github.com/QCHackers/tqec/pull/100#discussion_r1474632738) for example


nelimee:
At first glance, I like the way it looks. 

Definition of detectors is clearer, the complications due to `Detector`s being `cirq.Gate` instances are gone, the `Detector` instances end up transformed into "standard" gates (defined by `stimcirq`, but more standard than what we had).

I initially thought that removing the qubit the `Gate` was supposed to be applied to will raise issues such as not knowing the global coordinates of the local coordinate system origins, but it seems that the `origin` management is even simpler.

> We can not easily add the cirq.VirtualTag tag to the operation when initializing it. Therefore, I define some factory methods to handle it. However, this will require the users to add cirq.VirtualTag by themselves if they use the raw operation classes and might turn out to be a rabbit hole.

We can make that clear in the second line of the docstring (first line that is the body of the docstring). For example:
```python
class ShiftCoords(cirq.Operation):
    def __init__(self, *shift: int) -> None:
        """
        Annotates that the qubit/detector coordinate origin is being moved.

        WARNING: you should use `make_shift_coords` to create instances of this class.
        If you do not, read attentively the documentation below.

        This is a replication of the [stimcirq.ShiftCoordsAnnotation](https://github.com/quantumlib/Stim/blob/main/glue/cirq/stimcirq/_shift_coords_annotation.py)
        class. We can directly use `stimcirq.ShiftCoordsAnnotation` here, however,
        replication brings the class into the `tqec` namespace and is useful for the
        potential future iteration.
        :param shift: How much to shift each coordinate.
        """
        self._shift = shift
```
and add details in the docstring to explain why they should use `make_shift_coords` and what they should do if they want to use the class directly.
In theory, these classes are quite internal and should not leak to the public API, so we can even define a `__all__` property in the file that only lists the factories, and not the classes. An external user would then have to explicitly import something that is not listed in the "normal" import list, which would be another indication that these classes are to be handled with care.

> cirq.Operation with tags will become a cirq.TaggedOperation. And if we want to get the underlying operation we need to use op.untagged to get it.

There are probably some places in the code that use directly `op` instead of `op.untagged`. I only became aware of that property late in the development process, but as nothing broke since then, I did not take the time to have a look at it. Anyway, this is a little inconvenience that should only impact a small portion of the code, I think it is fine.

> Under current implementation, the qubits property of the operations is an empty tuple. When we add the operation directly to a cirq.Circuit, the default insert strategy will move the operation to the head of the circuit, which will be a disaster. There are two options to handle that: 1. attach some qubits to the operation's qubits property, however, this kind of violates the princeple of "annotation should be independent of the qubits it applied on"; 2. Always containerize the operations with a cirq.Moment to pin them at the certain timestep in the circuit. Currently, I use the 2nd approach in the refactoring.

I agree with the second approach. Maybe we could add some preventive checks / warnings / exceptions to try to catch potential mistakes early? For example, print a warning if there is a `Detector` / `Observable` in the first `Moment` of a circuit, as I do not see any case in which this would make sense and so this would be likely caused by human error. This might prevent some "funny" debugging sessions in the future.

inmzhang:
> In theory, these classes are quite internal and should not leak to the public API, so we can even define a __all__ property in the file that only lists the factories, and not the classes.

Agree. This is also what I thought of when I defined the factories.

> For example, print a warning if there is a Detector / Observable in the first Moment of a circuit, as I do not see any case in which this would make sense and so this would be likely caused by human error.

Actually, this will already cause an exception when we fill the global record indices during circuit transformation since no measurements can be found before the first moment of the circuit. But we can still add the warining to clarify the expectation and make the bug easier to find as you suggested.

nelimee:
> Actually, this will already cause an exception when we fill the global record indices during circuit transformation since no measurements can be found before the first moment of the circuit. But we can still add the warining to clarify the expectation and make the bug easier to find as you suggested.

Yes but this is not really a tailored error. Ideally, and because this is a predictable error that users might reasonably fall in, we should have a clear explanation as to why this happened and how to solve it, maybe with a link to this PR and thread of comments for interested users. Having a `MeasurementRecordNotFound`-like exception might not help a beginner finding its mistake.

Also, I will have a deeper look at the code this afternoon.

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+
+STIM_TAG = "STIM_OPERATION"
+
+
+class ShiftCoords(cirq.Operation):
+    def __init__(self, *shift: int) -> None:
+        """
+        Annotates that the qubit/detector coordinate origin is being moved.
+
+        This is a replication of the [stimcirq.ShiftCoordsAnnotation](https://github.com/quantumlib/Stim/blob/main/glue/cirq/stimcirq/_shift_coords_annotation.py)
```
nelimee, line 14: Maybe
```suggestion
        This is a replication of the 
        [stimcirq.ShiftCoordsAnnotation](https://github.com/quantumlib/Stim/blob/main/glue/cirq/stimcirq/_shift_coords_annotation.py)
```
to shorten as much as possible the line?

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
+from typing import Any
+from dataclasses import dataclass
+
+import cirq
+
+STIM_TAG = "STIM_OPERATION"
+
+
+class ShiftCoords(cirq.Operation):
+    def __init__(self, *shift: int) -> None:
```
nelimee, line 10:
```suggestion
    def __init__(self, *shifts: int) -> None:
```
In the code, `shift` is a `tuple[int, ...]` and so might contain several shifts. It might be easier to understand if this is clear from the variable name.

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+        potential future iteration.
+
+        :param shift: How much to shift each coordinate.
+        """
+        self._shift = shift
+
+    @property
+    def shift(self) -> tuple[int, ...]:
+        """The shift the operation represents."""
+        return self._shift
```
nelimee, line 26:
```suggestion
        :param shifts: How much to shift each coordinate.
        """
        self._shifts = shifts

    @property
    def shifts(self) -> tuple[int, ...]:
        """The shifts the operation represents."""
        return self._shifts
```
see previous comment.

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+
+def make_shift_coords(*shift: int) -> cirq.Operation:
+    """This is a helper function to make a :class:~ShiftCoords operation with the
+    `cirq.VirtualTag` tag.
+
+    :param shift: How much to shift each coordinate.
+
+    :return: A :class:`ShiftCoords` operation with the `cirq.VirtualTag` tag.
+    """
+    return ShiftCoords(*shift).with_tags(cirq.VirtualTag(), STIM_TAG)
```
nelimee, line 175:
```suggestion
def make_shift_coords(*shifts: int) -> cirq.Operation:
    """This is a helper function to make a :class:~ShiftCoords operation with the
    `cirq.VirtualTag` tag.

    :param shifts: How much to shift each coordinate.

    :return: A :class:`ShiftCoords` operation with the `cirq.VirtualTag` tag.
    """
    return ShiftCoords(*shifts).with_tags(cirq.VirtualTag(), STIM_TAG)
```
see previous comment.

src/tqec/detectors/transformer.py
```diff
@@ -46,45 +53,66 @@ def _fill_in_detectors_global_record_indices_impl(
...
-                    global_measurement_map, current_moment_index
+                continue
+
+            # All the stim-related annotations should have been tagged with STIM_TAG
+            if STIM_TAG not in operation.tags:
+                operations.append(deepcopy(operation))
+                continue
+            untagged = operation.untagged
+            if isinstance(untagged, ShiftCoords):
+                operations.append(stimcirq.ShiftCoordsAnnotation(untagged.shift))
```
nelimee, line 64:
```suggestion
                operations.append(stimcirq.ShiftCoordsAnnotation(untagged.shifts))
```
see previous comment.

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def _circuit_diagram_info_(self, _: Any) -> str:
+        row, col, t = self.coordinates
+        return f"Detector({row},{col},{t})"
+
+    @property
+    def coordinates(self) -> tuple[int, ...]:
+        return (
+            self._local_coordinate_system_origin.row,
+            self._local_coordinate_system_origin.col,
```
nelimee, line 133:
```suggestion
            self.origin.row,
            self.origin.col,
```

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+        time_coordinate: int = 0,
+) -> cirq.Operation:
+    """This is a helper function to make a :class:`Detector` operation with the
+    `cirq.VirtualTag` tag.
+
+    :param local_coordinate_system_origin: origin of the local coordinate system.
+        The origin along with the local coordinate system will be pinned to the
+        global coordinate system to resolve the actual qubit coordinates the
+        measurements applied to.
+    :param relative_measurements: a list of relative measurements that composed the
```
nelimee, line 190:
```suggestion
    :param relative_measurements: a list of relative measurements that compose the
```

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+        parity check of the detector. Each element of the list is a tuple of
+        (relative_qubit_position, relative_measurement_offset) or a :class:`RelativeMeasurementData`
+        instance. When a tuple is provided, the first element is the position of
+        the qubit relative to a local coordinate system origin and the second element
```
nelimee, line 194:
```suggestion
        the qubit relative to the local coordinate system origin and the second element
```

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+
+    :param local_coordinate_system_origin: origin of the local coordinate system.
+        The origin along with the local coordinate system will be pinned to the
+        global coordinate system to resolve the actual qubit coordinates the
+        measurements applied to.
+    :param relative_measurements: a list of relative measurements that composed the
+        parity check of the observable. Each element of the list is a tuple of
+        (relative_qubit_position, relative_measurement_offset) or a :class:`RelativeMeasurementData`
+        instance. When a tuple is provided, the first element is the position of
+        the qubit relative to a local coordinate system origin and the second element
```
nelimee, line 229:
```suggestion
        the qubit relative to the local coordinate system origin and the second element
```

src/tqec/detectors/operation.py
```diff
@@ -0,0 +1,244 @@
...
+        observable_index: int = 0,
+) -> cirq.Operation:
+    """This is a helper function to make a :class:`Observable` operation with the
+    `cirq.VirtualTag` tag.
+
+    :param local_coordinate_system_origin: origin of the local coordinate system.
+        The origin along with the local coordinate system will be pinned to the
+        global coordinate system to resolve the actual qubit coordinates the
+        measurements applied to.
+    :param relative_measurements: a list of relative measurements that composed the
```
nelimee, line 225:
```suggestion
    :param relative_measurements: a list of relative measurements that compose the
```

nelimee (approved):
I really like how you implemented that. I made a few minor comments, but overall I do not see any issue with the code in this PR.

One side remark that I had when reading part of your code: I stuck with the regular Spinx docstring format, but saw that you used what seems to be Google docstring format. I think it might be beneficial to change the existing docstrings to another format, that will be more readable than the one I initially used, Google format being a solid candidate. Any quick opinion on that? We can open another issue to track that.

inmzhang:
> I think it might be beneficial to change the existing docstrings to another format, that will be more readable than the one I initially used, Google format being a solid candidate.

I will open an issue for this.

nelimee (approved):
I tested again and had a look at your last commits, this PR can be merged in my opinion.

PR #112, opened 2024-02-05
Fix typo and naming

inmzhang:
1. "instanciate" -> "instantiate"
2. `recurse_in_operation_if_CircuitOperation` -> `recurse_in_operation_if_circuit_operation`

### DIFF
--- a/specification/schemas/template.schema.json
+++ b/specification/schemas/template.schema.json
@@ -29,7 +29,7 @@
               "description": "Raw plaquette indices in the 'raw_rectangle'.",
               "items": {
                 "type": "array",
-                "items": {"type": "integer", "description": "Plaquette indices used to instanciate the template."}
+                "items": {"type": "integer", "description": "Plaquette indices used to instantiate the template."}
               }
             },
             {
--- a/src/tqec/display.py
+++ b/src/tqec/display.py
@@ -12,11 +12,11 @@ def display_template(
 
     :param template: the Template instance to display.
     :param plaquette_indices: the plaquette indices that are forwarded to the
-        call to template.instanciate to get the actual template representation.
+        call to template.instantiate to get the actual template representation.
     """
     if isinstance(template, TemplateOrchestrator) and len(plaquette_indices) == 0:
         plaquette_indices = tuple(range(template.expected_plaquettes_number))
-    arr = template.instanciate(*plaquette_indices)
+    arr = template.instantiate(*plaquette_indices)
     for line in arr:
         for element in line:
             element = str(element) if element != 0 else "."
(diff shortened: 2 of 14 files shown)


nelimee:
Even though I understand that the original version of 2. does not agree with the usual naming scheme, I find the original version way clearer in its effects and intent. Reading the original name, it seems easy to understand that the method recurses in instances of `CircuitOperation`. This is not obvious anymore in the updated version in my opinion.

inmzhang:
> This is not obvious anymore in the updated version in my opinion.

I agree on this. However, I suggest we convey the intention in the function's docstring instead of deviating from the style, which could annoy the linter.

nelimee:
I do not know what to think about that.

My opinion is that the linter argument is not enough.
I view linters as guides rather than ground truth. If the linter tells me "this name should rather be XXX", this is for me an indication that I should consider the "XXX" name. If I do not agree with the linter recommendation, then I try to motivate why I should derogate to the linter rule. 
In this case, I think it is fine to not abide to linter rules because the "invalid" version is actually more expressive and clearer in what it does.

Maybe we can have other opinions? From @vtomole, @Gistbatch or @smburdick ?

vtomole:
> Maybe we can have other opinions?

Cirq invented `CircuitOperation` and they name their functions as [circuit_operation](https://github.com/quantumlib/Cirq/blob/f2c6f3ceb50771caa44aa1f6c316f5eb4b3bdd49/benchmarks/circuit_construction.py#L118)

nelimee:
> Cirq invented CircuitOperation and they name their functions as [circuit_operation](https://github.com/quantumlib/Cirq/blob/f2c6f3ceb50771caa44aa1f6c316f5eb4b3bdd49/benchmarks/circuit_construction.py#L118)

I'm fine with `recurse_in_operation_if_CircuitOperation` -> `recurse_in_operation_if_circuit_operation` then :+1:

nelimee (changes requested):
Using
```sh
grep -r instanciate *
```
in the project root, I find a few references to "instanciate" in the notebooks:
```text
notebooks/move_qubit_along_a_line.ipynb:    "    def instanciate(\n",
notebooks/move_qubit_along_a_line.ipynb:    "        arr = super().instanciate(x_plaquette, z_plaquette, *_)\n",
notebooks/doubling_distance.ipynb:    "    def instanciate(\n",
notebooks/doubling_distance.ipynb:    "        arr = super().instanciate(x_plaquette, z_plaquette, *_)\n",
notebooks/doubling_distance.ipynb:    "arr = doubling_template.scale_to(500).instanciate(*plaquette_indices)\n",
notebooks/corner.ipynb:    "array = corner.instanciate(*plaquette_indices)\n",
```
Everything else seems fine.

PR #114, opened 2024-02-05
Tentative at removing BaseShape

nelimee:
This is the first commit for an attempt at removing the BaseShape class, unifying Template implementations and remove some limitation with respect to scaling.

Documentation, tests and some refactoring are still missing and will be added in future commits.

Sits on top of #108.


src/tqec/templates/scale.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def get_fixed_scaling_function(fixed_scale: int) -> ty.Callable[[int], int]:
+    return lambda _: fixed_scale
+
+
+class Dimension:
+    def __init__(
+        self,
+        initial_value: int,
```
inmzhang, line 19: It seems useless providing an initial value that cannot be reached from the `scale_to`. I think it makes more sense to set the `initial_scale` instead of `initial_value`, just like what we did for templates. When the `is_fixed=True`, it can implicitly set the `scaling_function=lambda _: initial_scale` and unify the logic afterwards.

src/tqec/templates/scale.py
```diff
@@ -0,0 +1,96 @@
...
+        scaling_function: ty.Callable[[int], int] | None = None,
+        is_fixed: bool | None = None,
```
inmzhang, line 21: `is_fixed: bool = False` should work？

src/tqec/templates/scale.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def value(self) -> int:
+        """Returns the instance value."""
+        return self._value
+
+    @value.setter
+    def value(self, other: int) -> None:
+        """Sets directly the instance value."""
+        self._value = other
```
inmzhang, line 87: I don't think we should expose the setter. Once the dimension is fixed or provided with a valid `initial_scale`, it should not be changed anymore without calling `scale_to`.

src/tqec/templates/scale.py
```diff
@@ -0,0 +1,96 @@
...
+
+def get_fixed_scaling_function(fixed_scale: int) -> ty.Callable[[int], int]:
+    return lambda _: fixed_scale
+
+
+class Dimension:
+    def __init__(
+        self,
+        initial_value: int,
+        scaling_function: ty.Callable[[int], int] | None = None,
```
inmzhang, line 20: I understand that using a `scaling_function: Callable` instead of a `scaling_factor: int` can make the dimension more flexible. However, it might add some complexity to the serialization (when serialized to bytes with marshal) and communication between the frontend and backend, although I'm not very familiar with frontend development.

According to Austin's suggestion, in most cases, `2*k` should be sufficient. While there might be situations where `2*k+1` could be useful, linear scaling with `m*k + b` should cover most cases. Therefore, it seems reasonable to use a scaling factor like `m` or `(m, b)` instead of introducing the `scaling_function`.

What's your thoughts on this?

src/tqec/templates/constructions/qubit.py
```diff
@@ -1,95 +1,120 @@
...
             (3, RIGHT_OF, 0),
             (4, BELOW_OF, 0),
         ]
-        TemplateOrchestrator.__init__(self, _templates)
+        super().__init__(_templates)
         for source, relpos, target in _relations:
             self.add_relation(source, relpos, target)
+
+    @staticmethod
+    def get_dimensions(
```
inmzhang, line 89: This is kind of messy to me. Maybe we can enforce the user to use `Dimension`?

src/tqec/templates/atomic/rectangle.py
```diff
@@ -0,0 +1,219 @@
...
+            height_dim = Dimension(height, is_fixed=True)
+        else:
+            width_dim = Dimension(width, is_fixed=True)
+            height_dim = Dimension(height, lambda k: 2 * k)
+        super().__init__(
+            width_dim, height_dim, default_x_increment, default_y_increment
+        )
+
+
+class FixedRectangleTemplate(AlternatingRectangleTemplate):
```
inmzhang, line 202: Do we still need to define the "fixed" variant of the template? This is another reason why I think we can enforce the user to use `Dimension` to specify the scaling behavior. All the scaling logic is decoupled from the `Template` itself and injected by the `Dimension` instances.

src/tqec/templates/constructions/qubit.py
nelimee: Yes, I agree. I wanted to keep backward compatibility for a while, but it is way easier to simply enforce `Dimension` everywhere.

src/tqec/templates/atomic/rectangle.py
nelimee: These were helper classes, mainly used to keep the old API and ease the change to `Dimension`, but as noted in another comment we do not need to do that, I will remove.

src/tqec/templates/scale.py
nelimee: I agree. Removing the setter for the value makes that change safe. It removes the possibility to check the correctness of the input value (is it odd or even?) in the constructors of some classes, but this is a small price to pay, and we can find other ways to ensure that.

src/tqec/templates/scale.py
nelimee: Following other changes from your comments, the `is_fixed` has been completely removed. I am thinking about creating a `FixedDimension` child class that will enforce the `is_fixed` property by providing `lambda _: value` to the constructor of `Dimension`.

src/tqec/templates/scale.py
nelimee: I agree, this is fixed in a soon-to-be-pushed commit.

src/tqec/templates/scale.py
nelimee: I agree. Using a function as input is way too much freedom and makes the task of encoding this in JSON very challenging for both the backend and frontend. I agree with `(m, b)` being likely sufficient. I will try to make that change.

nelimee: Should be done. Also the `LinearFunction` type is really simple, and we can extend it without any issue to a polynomial type.

nelimee:
This PR is quite large, but it is expected for such a large refactor. I think it is ready for review, most of the newly introduced code should be tested (with the notable exception of `AlternatingCornerSquareTemplate` that seems broken and that will need a full re-implementation in my opinion) and documented correctly.

inmzhang:
I'll review the code at tomorrow :)

inmzhang (commented):
In overall, it looks good to me. Some comments on `RawRectangleTemplate`.

src/tqec/templates/atomic/rectangle.py
```diff
@@ -0,0 +1,183 @@
...
+                "when calling this method."
+            )
+
+    def scale_to(self, _: int) -> "RawRectangleTemplate":
+        return self
+
+    @property
+    def shape(self) -> Shape2D:
+        if not self._indices:
+            return Shape2D(0, 0)
```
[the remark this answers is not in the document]
nelimee: Not anymore, I will change it.

src/tqec/templates/atomic/rectangle.py
```diff
@@ -0,0 +1,183 @@
...
+        else:
+            return Shape2D(len(self._indices[0]), len(self._indices))
```
nelimee, line 176:
```suggestion
        return Shape2D(len(self._indices[0]), len(self._indices))
```

src/tqec/templates/atomic/rectangle.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def to_dict(self) -> dict[str, ty.Any]:
+        return super().to_dict() | {"indices": self._indices}
+
+    @property
+    def expected_plaquettes_number(self) -> int:
+        return max(max(line) for line in self._indices) + 1
```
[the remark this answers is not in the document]
nelimee: `expected_plaquettes_number` is the number of templates expected in the `instantiate` method.

This class is indeed special, and yes the indices used here are just like `TemplateWithIndices`, as noted in the exception in the `instantiate` method. 
This class is basically the last option for weird non-scaling shapes that are not encodable by other classes in the `Template` hierarchy. I think it is only used for `1x1` atomic templates in the corner at the moment.
What is interesting with this class is that it gives full freedom to the user, it is simply an array of numbers that will index in the plaquette indices provided.

I was initially more on the "let the user have holes in its indices if he/she wants to", but I do not see the potential use cases for that anymore.
Anyway, I think that "transforming" silently `[[0, 10], [38, 21]]` to `[[0, 1], [3, 2]]` is too error-prone. If we go on that road, I am more on the side of checking that the indices provided at construction are contiguous from `0` to the maximum index found and raising an exception if that is not the case.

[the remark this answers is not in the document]
inmzhang: > Anyway, I think that "transforming" silently [[0, 10], [38, 21]] to [[0, 1], [3, 2]] is too error-prone. If we go on that road, I am more on the side of checking that the indices provided at construction are contiguous from 0 to the maximum index found and raising an exception if that is not the case.

Agree. I think this will make **"the number of templates expected in the `instantiate` method"** the same as **"the number of different plaquettes contained in the templates"**, isn't it?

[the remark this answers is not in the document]
nelimee: For this class in particular, the plaquette indexed 0 that, by convention, means "no plaquette here" or "the empty plaquette", need to be provided to the instantiate method, which is different from the `Template` API and all the other classes that "knows about" 0.
To illustrate, if you want the "no plaquette" somewhere in your `RawRectangleTemplate`, you have to provide it:
```py
template = RawRectangleTemplate(
    [
        [0, 0, 1, 0],
        [1, 1, 1, 0],
        [0, 1, 1, 1],
        [0, 1, 0, 0],
    ]
)
>>> print(template.instantiate(1, 2))
#[
#    [1, 1, 2, 1],
#    [2, 2, 2, 1],
#    [1, 2, 2, 2],
#    [1, 2, 1, 1],
#]
>>> print(template.instantiate(1, 0))
#[
#    [1, 1, 0, 1],
#    [0, 0, 0, 1],
#    [1, 0, 0, 0],
#    [1, 0, 1, 1],
#]
```
So "the number of templates expected in the `instantiate` method" may or may not be the same as "the number of different plaquettes contained in the templates". It is definitely the same if you count the plaquette 0 as a plaquette, but else these two quantities are only equal if the plaquette number 0 is not in your template.

Thinking more about it, we might not really need that `Template` instance anymore. The fact that its API differs from all the other `Template` is quite a deal breaker. I'll check where this class is needed before thinking about removing it.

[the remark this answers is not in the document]
nelimee: So as I suspected, this is only used in `ScalableCorner` (that I will rename `ScalableCornerTemplate` soon):
```python
class ScalableCorner(ComposedTemplate):
    def __init__(self, dim: Dimension) -> None:
        # ....
        _templates = [
            # ....
            TemplateWithIndices(RawRectangleTemplate([[0]]), [2]),
            # ....
            TemplateWithIndices(RawRectangleTemplate([[0]]), [6]),
            # ....
            TemplateWithIndices(RawRectangleTemplate([[0]]), [12]),
            # ....
        ]
        # ....
```
The problem of replacing it with another template (for example `AlternatingSquareTemplate`) lies in the number of plaquettes expected by the instantiate method: `RawRectangleTemplate` only requires one, the other `Template` classes we have require at least two.

I do not see a nice way of replacing that class. As a potentially satisfying fix for the API problem raised in my previous comment, we can also try to check that this template is never instantiated with the `0` plaquette, but it does not seem very useful to me.

src/tqec/templates/atomic/base.py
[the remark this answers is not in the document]
nelimee: Removed in [b0acb04](https://github.com/QCHackers/tqec/pull/114/commits/b0acb041f527616ab2f3f70d40d17684f101fb8f)

src/tqec/templates/atomic/rectangle.py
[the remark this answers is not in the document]
nelimee: I just enforced contiguous indices starting from `0` (and fixed a bug that was allowing `RawRectangle([[]])`).

src/tqec/templates/atomic/rectangle_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def test_raw_rectangle_larger_expected_plaquettes_number():
+    template = RawRectangleTemplate([[0], [1], [2]])
+    assert template.expected_plaquettes_number == 3
+
+
+def test_raw_rectangle_expected_plaquettes_number_with_holes():
+    template = RawRectangleTemplate([[0], [1], [24]])
+    assert template.expected_plaquettes_number == 25
```
inmzhang, line 143:
```suggestion
```

src/tqec/templates/atomic/rectangle_test.py
```diff
@@ -0,0 +1,179 @@
...
+def test_raw_rectangle_larger_instantiate_hole():
+    template = RawRectangleTemplate([[0], [34], [9]])
+    arr = template.instantiate(*list(range(1, template.expected_plaquettes_number + 1)))
+    numpy.testing.assert_equal(arr, [[1], [35], [10]])
+
+
+def test_raw_rectangle_larger_instantiate_hole_raising():
+    template = RawRectangleTemplate([[0], [34], [9]])
+    with pytest.raises(TQECException):
+        template.instantiate(1, 2, 3)
```
inmzhang, line 179:
```suggestion
```

src/tqec/templates/atomic/rectangle_test.py
```diff
@@ -0,0 +1,179 @@
...
+    arr = template.instantiate(3, 2, 1)
+    numpy.testing.assert_equal(arr, [[3], [2], [1]])
+
+
+def test_raw_rectangle_larger_instantiate_different_order():
+    template = RawRectangleTemplate([[0], [2], [1]])
+    arr = template.instantiate(3, 2, 1)
+    numpy.testing.assert_equal(arr, [[3], [1], [2]])
+
+
```
inmzhang, line 169:
```suggestion

def test_raw_rectangle_with_noncontiguout_indices():
    with pytest.raises(TQECException, match="CONTIGUOUS indices"):
        template = RawRectangleTemplate([[0], [1], [24]])
```

inmzhang (changes requested):

inmzhang (approved):

nelimee:
Thanks for the approval, I'm merging!

PR #120, opened 2024-02-07
Change the docstring style to Google style

### LINKED ISSUE #118: Docstring style
From [@nelimee's comment](https://github.com/QCHackers/tqec/pull/111#issuecomment-1926810482):
> I think it might be beneficial to change the existing docstrings to another format, that will be more readable than the one I initially used, Google format being a solid candidate. 

Currently, we use Sphinx style for the docstrings. We are considering to turn to some other style like the Google style, which is more readable and simple. Any other opinion on that?

For me, I'm used to the Google style docstring and will be glad to use that in this project.

inmzhang:
Closes #118 

This PR changes the docstring style of all the functions and classes from rest to google style.

Initially, I used [pyment](https://github.com/dadadel/pyment) for this, however, it turns out `pyment` cannot handle the multi-line argument descriptions correctly[^1]. Then I turn to a less well-known package [docconvert](https://github.com/cbillingham/docconvert) and it worked out quite well.

I have checked the auto-converted files with `diff` preview in my IDE, but slight errors might still exist. The conversion process can be reproduced following the procedure below:

1. Install the `docconvert` python package: 
  ```shell
  pip install docconvert
  ```
2. Create the configuration file `docconvert.json` under the directory of your `tqec` project:
  ```json
  {
      "input_style": "rest",
      "output_style": "google",
      "accepted_shebangs": [
          "python"
      ],
      "output": {
          "first_line": true,
          "replace_quotes": "\"\"\"",
          "standard_indent": "    ",
          "tab_length": 4,
          "realign": true,
          "max_line_length": 80,
          "use_optional": false,
          "remove_type_back_ticks": "true",
          "use_types": false,
          "separate_keywords": false
      }
  }
  ```
3. Run `docconvert` from the command line:
  ```shell
  docconvert --threads 4 --in-place --config docconvert.json src/tqec
  ```

[^1]: https://github.com/dadadel/pyment/issues/76

### DIFF
--- a/src/tqec/display.py
+++ b/src/tqec/display.py
@@ -8,9 +8,10 @@ from tqec.templates.composed import ComposedTemplate
 def display_template(template: Template, *plaquette_indices: int) -> None:
     """Display a template instance with ASCII output.
 
-    :param template: the Template instance to display.
-    :param plaquette_indices: the plaquette indices that are forwarded to the
-        call to template.instantiate to get the actual template representation.
+    Args:
+        template: the Template instance to display.
+        *plaquette_indices: the plaquette indices that are forwarded to the call
+            to ``template.instantiate`` to get the actual template representation.
     """
     if len(plaquette_indices) == 0:
         plaquette_indices = tuple(range(1, template.expected_plaquettes_number + 1))
@@ -35,13 +36,15 @@ def display_templates_svg(
     the templates, it can be very slow for large templates and results in huge
     svg files.
 
-    :param templates: the TemplateOrchestrator instance to display.
-    :param write_svg_file: the path to the SVG file to write.
-    :param canvas_height: the height of the canvas in pixels to draw on.
-    :param plaquette_indices: the plaquette indices that are forwarded to the
-        call to template.instantiate to get the actual template representation.
+    Args:
+        templates: the TemplateOrchestrator instance to display.
+        write_svg_file: the path to the SVG file to write.
+        canvas_height: the height of the canvas in pixels to draw on.
+        *plaquette_indices: the plaquette indices that are forwarded to the call
+            to ``template.instantiate`` to get the actual template representation.
 
-    :return: the SVG string.
+    Returns:
+        the SVG string.
     """
     if len(plaquette_indices) == 0:
         plaquette_indices = tuple(range(1, templates.expected_plaquettes_number + 1))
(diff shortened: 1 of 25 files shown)


src/tqec/templates/base.py
```diff
@@ -39,10 +40,15 @@ def to_dict(self) -> dict[str, ty.Any]:
...
-        :raises DefaultKeyInKwargs: if the "default" key is present in kwargs.
+        Args:
+            **kwargs: keyword arguments forwarded to the json.dumps function.
+                The "default" keyword argument should NOT be present.
+
+        Returns:
+            the JSON representation of the instance.
+
+        Raises:
+            DefaultKeyInKwargs: if the "default" key is present in kwargs.
```
nelimee, line 51:
```suggestion
            TQECException: if the "default" key is present in kwargs.
```

src/tqec/display.py
```diff
@@ -35,13 +36,15 @@ def display_templates_svg(
...
-    :param write_svg_file: the path to the SVG file to write.
-    :param canvas_height: the height of the canvas in pixels to draw on.
-    :param plaquette_indices: the plaquette indices that are forwarded to the
-        call to template.instantiate to get the actual template representation.
+    Args:
+        templates: the TemplateOrchestrator instance to display.
+        write_svg_file: the path to the SVG file to write.
+        canvas_height: the height of the canvas in pixels to draw on.
+        *plaquette_indices: the plaquette indices that are forwarded to the call
+            to ``template.instantiate`` to get the actual template representation.
```
nelimee, line 44: Are the double-backticks needed? Isn't a simple backtick enough?

nelimee (approved):
A made 2 very minor comments, this is nice! Thanks for that PR.

src/tqec/display.py
inmzhang: Yes, I think both [single](https://google.github.io/styleguide/pyguide.html#3821-test-modules) and [double](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) backticks should work. I will change this into single backtick for simplicity.

nelimee: Seems like they are not equivalent: https://stackoverflow.com/a/57302493/4810787.

inmzhang: I wasn't aware of that before. So it appears safer to use double backticks for inline code snippets, although single backticks should rarely cause any issues.

inmzhang: And I think we can keep on using single backtick.

nelimee (approved):
Can be merged

Issue #86, opened 2024-01-30
Add automatic checks to pull requests


smburdick:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

PR #121, opened 2024-02-07
CI

Gistbatch:
Work for #86
At the moment, this runs a check on every push. We should change this once it is ready.


Gistbatch:
I created a minimal CI running pytest and jest.
@smburdick I'm not sure how testing in React works. For now, I moved the `tests` directory in the search path, but the test that was there failed. As far as I could tell, we would need to adapt the package.json test script somehow.
@nelimee For Python, I just did a basic pylint -> pytest -> coverage. I had to remove python 3.9 as it doesn't work with the type hints.

setup.cfg
```diff
@@ -21,7 +21,6 @@ classifiers =
     Operating System :: POSIX :: Linux
     Programming Language :: JavaScript
     Programming Language :: Python :: 3 :: Only
-    Programming Language :: Python :: 3.9
```
smburdick, line 24: we are adding support for this in #124

src/tqec/generation/circuit_test.py
smburdick, line 1: Was this added by mistake?

setup.cfg
smburdick: also in `ci.yml`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "pip"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt -r dev-requirements.txt
```
smburdick, line 28: I think there is a more specific command expected

frontend/src/__tests__/App.test.js
smburdick, line 1: We should probably add some frontend tests, which can be spun off into a separate issue.

smburdick (commented):
Looks fine in general, I think we need more frontend support though -- `npm ci` doesn't do anything meaningful. We may want to talk about an informal "release window" fir both the Python package and live service (if we pursue that route), which would motivate tests being completed by a particular time frame.

src/tqec/generation/circuit_test.py
Gistbatch: This fixes the test, another pr broke it

.github/workflows/ci.yml
Gistbatch: This is the command to completely install all dependencies, I'm not aware of any other way to do this in our project setup. This is different from the full build command which is `pip install .`

frontend/src/__tests__/App.test.js
Gistbatch: The one we have also fails at the moment.
I don't know how react usually handles testing.
I just moved the tests directory to `src/__tests__` because this is the default regex for the search path and I don't know how to change the test script.

smburdick:
> > Looks fine in general, I think we need more frontend support though -- `npm ci` doesn't do anything meaningful.
> 
> I looked at other repositories, I understood this to be `npm install` for automated systems.
> 
> Summary of stuff left to do in this pr:
> 
> * [ ]  Fix frontend test / change test script

I think this can be its own issue. We can remove the NodeJS check to make this pass as-is

> * [ ]  Include all supported Python versions (3.9)
> * [ ]  Only trigger on merge into main

Should be doable.

nelimee (dismissed):
One minor additional minor comment: we might want to reformat the code using black/ruff on push. This should in theory be done by the pusher, but checking for code formatting would be nice.

smburdick:
> One minor additional minor comment: we might want to reformat the code using black/ruff on push. This should in theory be done by the pusher, but checking for code formatting would be nice.

This can also be enforced with a `pylintrc` file

Gistbatch:
> > * [ ]   Fix frontend test / change test script
> 
> I think this can be its own issue. We can remove the NodeJS check to make this pass as-is

Unfortunately `npm test` fails when it doesn't find any tests, and I couldn't figure out how to pass the `--passWithNoTests` option to the test script.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+  # pull_request:
+  #   branches:
+  #     - main
+
+jobs:
+  build-and-test-python:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
```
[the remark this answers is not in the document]
nelimee: 3.9 should now be supported. If the tests fail on `3.9`, we have things to fix.
```suggestion
        python-version: ["3.9", "3.10", "3.11", "3.12"]
```

Gistbatch:
I added python 3.9 back and some dummy tests to make the js script pass.
I also changed the policy to check only on pushes to main or prs on main.

nelimee (approved):
For the Python part (and only for that part), I think we reached a good state. 
Future improvements might add checks for formatting errors, but this can be done in a followup PR.

frontend/src/__tests__/setupTests.js
```diff
@@ -0,0 +1,15 @@
...
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom';
+
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+// dummy test to make jest not complain about no tests
+test('renders learn react link', () => {
```
smburdick, line 11: Was this verified to be compatible with the newest Redux changes in #133 ?

src/tqec/generation/circuit_test.py
```diff
@@ -78,22 +78,54 @@ def test_generate_circuit():
     generate_qubits = [(q.row, q.col) for q in generated_circuit.all_qubits()]
     target_qubits = [
         (7, 3),
```
smburdick, line 80: May want to move this to a `test-inputs` file for reusability/readability

smburdick (commented):
Seems fine, just wanted to check a couple things.

src/tqec/generation/circuit_test.py
Gistbatch: I think its fine for now, the test is not very expressive anyway.

frontend/src/__tests__/setupTests.js
Gistbatch: I've merged main into this branch yesterday, so it should work.

src/tqec/generation/circuit_test.py
nelimee: I will try to replace that test by more "atomic"/unitary tests in #130 if it turns out that it makes sense and is doable. For the moment, I agree that this test can be kept like that.

rryoung98:
Tests on the frontend are typically a combination of browserstack for checking browser compatibility and using cypress for E2E or unit/ states tests @Gistbatch  @smburdick. 

I believe that for now it's worth going without it until we have a working solution and just move quickly

smburdick:
Agreed. Go ahead and merge @Gistbatch

PR #122, opened 2024-02-07
Fix some exception types in docstring

inmzhang:
I noticed several exception types in the docstring that weren't updated properly after the merge of #73. This pull request primarily addresses those types. Additionally, I identified one argument docstring that wasn't properly converted in #112, and I fixed that as well.

### DIFF
--- a/src/tqec/detectors/operation.py
+++ b/src/tqec/detectors/operation.py
@@ -31,7 +31,8 @@ class ShiftCoords(cirq.Operation):
         with the correct tags. Otherwise, you might need to manually tag the operation with the
         `cirq.VirtualTag` and the `STIM_TAG`.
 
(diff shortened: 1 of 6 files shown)


src/tqec/detectors/operation.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, *shifts: int) -> None:
         with the correct tags. Otherwise, you might need to manually tag the operation with the
         `cirq.VirtualTag` and the `STIM_TAG`.
 
-        :param shifts: How much to shift each coordinate.
+        Args:
+            shifts: How much to shift each coordinate.
```
nelimee, line 35:
```suggestion
            *shifts: How much to shift each coordinate.
```
I think Google style also includes the `*`.

nelimee (approved):

Issue #106, opened 2024-02-02
Implement more tests


nelimee:
We need to start testing things to avoid regressions like #104.

PR #124, opened 2024-02-07
Add support for Python 3.9

### LINKED ISSUE #123: Which minimum version of Python should we support?
#121 is a good way for us to see if the code contains features that are incompatible with the current minimum supported Python version (`3.9`, because this is the minimum supported Python version of `cirq`).

The feature introduced in Python 3.10 with [PEP 604](https://peps.python.org/pep-0604/) introducing the possibility to use `|` on types (`Optional[int]` -> `int | None`) is one of the blocking point.

I would be in favour of bumping the minimum supported Python version to 3.10 for that reason. Even though I previously said that we should try to maintain as high compatibility as possible, I think that using `|` is way more readable for the code base.

One argument against bumping the version (and so in favour of replacing all the `|` by either `Optional[X]` or `Union[X, Y]`) is that `3.10` was released in Oct. 4, 2021, which is only a little bit more than 2 years ago, and Debian 11 Bullseye, which has not reached EOL, does not seem to support `3.10`.

Any opinion on that?

nelimee:
This PR adds support for Python 3.9, mostly by adding `from __future__ import annotations` at the top of files that are using [PEP 604](https://peps.python.org/pep-0604/).

Fixes #123.

### DIFF
--- a/src/tqec/detectors/measurement_map.py
+++ b/src/tqec/detectors/measurement_map.py
@@ -1,7 +1,8 @@
-import cirq
+from __future__ import annotations
 
-from tqec.exceptions import TQECException
+import cirq
 from tqec.detectors.operation import RelativeMeasurementsRecord
+from tqec.exceptions import TQECException
 
 
 def flatten(obj: cirq.Moment | cirq.AbstractCircuit) -> cirq.Circuit:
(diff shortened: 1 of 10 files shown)


smburdick (approved):
LGTM. Can you add your testing workflow (e.g., what commands did you run)?

nelimee:
> LGTM. Can you add your testing workflow (e.g., what commands did you run)?

I ran the different notebooks using a Python 3.9 conda environment. I am planning to fill up the tests (#106) in the near future, just after #100 is merged.

PR #125, opened 2024-02-08
Add issue templates

smburdick:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
```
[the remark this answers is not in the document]
smburdick: Hash marks are preferred by me as well

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,38 +1,41 @@
...
-**To Reproduce**
-Steps to reproduce the behavior:
-1. Go to '...'
-2. Click on '....'
-3. Scroll down to '....'
-4. See error
+## Steps to reproduce the behavior:
+1.
+1.
+1.
```
nelimee, line 16:
```suggestion
1.
2.
3.
```

nelimee (approved):
I had the idea that we could maybe split the issue template in 2: one targeted at the backend part, and another one targeted at the frontend part.
This would allow us to automatically assign tags to the issue, and maybe ask for more specialised information. 
Also, this is more work and I am not sure we need that separation now. It can come at a later stage.

smburdick:
> I had the idea that we could maybe split the issue template in 2: one targeted at the backend part, and another one targeted at the frontend part.
> 
> This would allow us to automatically assign tags to the issue, and maybe ask for more specialised information. 
> 
> Also, this is more work and I am not sure we need that separation now. It can come at a later stage.

I think we can have a single one for now, since it should be fairly obvious from the context of the issue, assuming the ticket was filled out sufficiently. Will merge.

.github/ISSUE_TEMPLATE/bug_report.md
smburdick: Markdown will render 1,1,1 as 1,2,3.

PR #127, opened 2024-02-08
Prepare for mypy check and fix typing errors

### LINKED ISSUE #126: Code contains some type errors
The backend code contains some typing issues. 

In this particular case, they do not seem to impact the code correctness, but it is always better to have as few as possible type errors.

nelimee:
This PR tries to fix a few typing errors that accumulated in the code base.

As a nice side-effect, it might also be beneficial to #121 and might help also including type checking the full library in the CI.

Command used to check for types:
```sh
python -m mypy --ignore-missing-imports --package tqec
```
with the `tqec` package re-installed to include the `py.typed` file in the package data.

Fixes #126.


nelimee:
Current report from `mypy`:
```
src/tqec/templates/shapes/rectangle.py:14: error: Signature of "instantiate" incompatible with supertype "BaseShape"  [override]
src/tqec/templates/shapes/rectangle.py:14: note:      Superclass:
src/tqec/templates/shapes/rectangle.py:14: note:          def instantiate(self, *_: int) -> ndarray[Any, Any]
src/tqec/templates/shapes/rectangle.py:14: note:      Subclass:
src/tqec/templates/shapes/rectangle.py:14: note:          def instantiate(self, x_plaquette: int, z_plaquette: int, *_: int) -> ndarray[Any, Any]
src/tqec/templates/shapes/square.py:57: error: Signature of "instantiate" incompatible with supertype "Rectangle"  [override]
src/tqec/templates/shapes/square.py:57: note:      Superclass:
src/tqec/templates/shapes/square.py:57: note:          def instantiate(self, x_plaquette: int, z_plaquette: int, *_: int) -> ndarray[Any, Any]
src/tqec/templates/shapes/square.py:57: note:      Subclass:
src/tqec/templates/shapes/square.py:57: note:          def instantiate(self, x_plaquette: int, z_plaquette: int, x_plaquette_flipped: int, z_plaquette_flipped: int, corner_plaquette: int, *_: int) -> ndarray[Any, Any]
src/tqec/templates/shapes/square.py:57: error: Signature of "instantiate" incompatible with supertype "BaseShape"  [override]
src/tqec/templates/shapes/square.py:57: note:      Superclass:
src/tqec/templates/shapes/square.py:57: note:          def instantiate(self, *_: int) -> ndarray[Any, Any]
src/tqec/templates/shapes/square.py:57: note:      Subclass:
src/tqec/templates/shapes/square.py:57: note:          def instantiate(self, x_plaquette: int, z_plaquette: int, x_plaquette_flipped: int, z_plaquette_flipped: int, corner_plaquette: int, *_: int) -> ndarray[Any, Any]
src/tqec/plaquette/schedule.py:278: error: Argument 1 to "transform_qubits" of "Operation" has incompatible type "dict[GridQubit, GridQubit]"; expected "dict[Qid, Qid] | Callable[[Qid], Qid]"  [arg-type]
Found 4 errors in 3 files (checked 56 source files)
```

The first 3 errors are instances of [Incompatible overrides](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides). This can be fixed by changing the signature of all `instantiate` methods to 
```py
def instantiate(self, indices: tuple[int]) -> numpy.ndarray:
    pass
```
with the notable regression that the parameter names will not exist anymore. Parameter names were used to give a little bit of context to an hypothetical user, see for example 
https://github.com/QCHackers/tqec/blob/890dc366e9048613f8cf1fd6bcf90255bbc5817c/src/tqec/templates/shapes/square.py#L57-L65
(note that #114 will remove the `shape` folder).
In practice these raw `Template` instances (qualified as `AtomicTemplate` instances in #114) will nearly never be used directly, and so these parameter names might not matter that much.

The last error is an instance of [Invariance vs covariance](https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance) and is caused by the fact that
- we provide a `dict[cirq.GridQubit, cirq.GridQubit]`
- `cirq`'s `transform_qubit` method expects a `dict[cirq.Qid, cirq.Qid]`
- the `dict` type annotation is invariant (the types given in parameter should **exactly** match the provided types) and we would need it to be covariant.

It is unclear to me what we could do here.
I think the proper way of dealing with that would be for `cirq` to change the type annotation to accept `dict[cirq.GridQubit, cirq.GridQubit]`, for example using `Mapping[cirq.Qid, cirq.Qid]`, because they do not mutate the input dictionary. 

I am waiting for other opinions before "fixing" the issues above.

inmzhang:
> In practice these raw Template instances (qualified as AtomicTemplate instances in https://github.com/QCHackers/tqec/pull/114) will nearly never be used directly, and so these parameter names might not matter that much.

Agree.

> It is unclear to me what we could do here.

I think it's safe to simply cast the type of qubit map to avoid the typing error with `cast_qubit_map = typing.cast(dict[cirq.Qid, cirq.Qid], qubit_map)`.

src/tqec/templates/composed.py
```diff
@@ -252,19 +252,23 @@ def add_corner_relation(
...
-        coordinates.
-
-        Returns:
-            a mapping between templates indices and their upper-left corner
-            absolute position.
+        <<<<<<< HEAD
+                This is the main method of the ComposedTemplate class. It explores templates
+        =======
+                This is the main method of the ``ComposedTemplate`` class. It explores templates
+        >>>>>>> main
```
inmzhang, line 259:
```suggestion
                This is the main method of the ``ComposedTemplate`` class. It explores templates
```

inmzhang (changes requested):
Thanks for adding mypy! There's an issue left from a merge.

inmzhang:
And I note that the [test for circuit generation](https://github.com/QCHackers/tqec/blob/main/src/tqec/generation/circuit_test.py), included in #80, also caused the mypy check to fail. This test has limited value and fails to capture the implications of some earlier modifications. I suggest we remove it for now and consider reimplementing it in the future.

nelimee:
I addressed all the comments except for the circuit test. The test is updated in #121. I will likely improve it in #130 so I have an inclination towards leaving it like this for the moment. I do not have any strong opinion on that though, so if you really think that it should be gone (and eventually added back in an improved version) I am fine with it.

inmzhang (approved):

PR #128, opened 2024-02-08
Fix CircuitMeasurementMap not adhering to the documented API

nelimee:
The `CircuitMeasurementMap.get_measurement_relative_offset` was documented to return `None` when the measurement was not found, but raised in the case of a measurement not found because there was none. This is fixed by returning `None` instead of raising.


inmzhang (approved):
LGTM

Issue #136, opened 2024-02-14
Refactor the `Plaquette` hierarchy


nelimee:
There are several things in the `Plaquette` hierarchy that may be improved:
1. [`SquarePlaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L63) has no way of recovering its qubits positionally. Having something like `plaquette.get_qubits_on(RIGHT)` will be beneficial to finalise #100.
2. [`RoundedPlaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L112) is not a subclass of [`SquarePlaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L63) but should likely implemented as such, as a [`RoundedPlaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L112) is simply a [`SquarePlaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L63) that is not using 2 of its qubits.
3. There is a lot of code repetition in the different plaquettes implemented in the plaquette library.

To be clear: the `tqec` library only uses the [`Plaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L9) class internally. **All** the other classes in the `Plaquette` hierarchy are targeted at the end user and their only reason of existence is to make QEC definition easier for the user. I think we should improve/refactor the other classes. The [`Plaquette`](https://github.com/QCHackers/tqec/blob/126271af21c2c0e0fa9d73b496d7c5ed7fee8466/src/tqec/plaquette/plaquette.py#L9) class is doing its job perfectly and does not have to be changed in my opinion.

PR #130, opened 2024-02-08
Add tests to the library

nelimee:
This branch has been created from #100 because both #114 and #100 added/changed a lot of things and I did not want to write tests for things that will be removed soon.

There is no need to look at the changes here, the immense majority of the files changed are caused by #100 and #114 and will disappear from this PR once they are merged.

Should fix or at least help fixing #106

PR #137, opened 2024-02-14
Refactor Plaquette hierarchy

nelimee:
This PR try to address #136.


nelimee:
Note: removing the `PlaquetteList` class made a few notebooks fail. `PlaquetteList` has always been too specialised to be in the main library in my opinion (only useful for simple memory experiments), so I will change the notebooks accordingly.

nelimee:
The notebook [`move_qubit_along_a_line.ipynb`](https://github.com/QCHackers/tqec/blob/main/notebooks/move_qubit_along_a_line.ipynb) should still be updated to the new plaquette hierarchy, but the other notebooks are working as expected.

smburdick:
@nelimee Would you like my review on this?

nelimee:
I am still missing the rewriting of one notebook, I am having a hard time getting the observable correct, but everything else should be near final so if you have the time, your review is more than welcome 😊
The more opinions we have, the better it is!

nelimee:
With this last commit, I did everything I planned with this PR. I am ready to discuss/make changes according to your reviews :)

smburdick:
> With this last commit, I did everything I planned with this PR. I am ready to discuss/make changes according to your reviews :)

I'll see what I can do. I would like to familiarize myself with the backend more.

nelimee:
> I'll see what I can do. I would like to familiarize myself with the backend more.

Feel free to ask me questions if you have any, we can also schedule a meeting like last time :)

nelimee:
I will also, either this week-end or at the beginning of next week, refactor the different measurement plaquettes. All the `{XX,ZZ,XXXX,ZZZZ}FinalMeasurementPlaquette` will be replaced by two plaquettes `MeasurementRoundedPlaquette` and `MeasurementSquarePlaquette` in a new `tqec.plaquettes.library.memory` module. 

With the exception of that change (that is simply a code factorisation step), this PR should still be considered finalised and ready to be reviewed.

inmzhang:
I've gone through most of the refactoring changes in the associated PR #100 and initially had few opinions. I will revisit this tomorrow for a closer look.

src/tqec/detectors/operation.py
```diff
@@ -59,6 +59,9 @@ def qubits(self) -> tuple[cirq.Qid, ...]:
     def with_qubits(self, *new_qubits: cirq.Qid) -> "ShiftCoords":
         return self
 
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self._shifts})"
```
inmzhang, line 63:
```suggestion
    def __repr__(self) -> str:
        return f"{self.__class__.__name__}{self._shifts}"
```

src/tqec/plaquette/library/initialisation.py
inmzhang, line 1: All the initialization/measurement plaquettes for X/Z basis and square/rounded variants are almost the same. I think we can refactor it further by extracting the circuit construction process.  Currently, the difference between the square/rounded variants is only the way we get the data qubits. If we can make `get_data_qubits_cirq(self)` as well as `get_syndrome_qubits_cirq(self)` as an abstract method in the `Plaquette` level, we can eliminate some code repetition in these cases.

src/tqec/plaquette/plaquette.py
```diff
@@ -148,19 +164,21 @@ def __init__(
...
-        PlaquetteQubit(Position(0, 0)),
-        PlaquetteQubit(Position(2, 0)),
-        PlaquetteQubit(Position(0, 2)),
-        PlaquetteQubit(Position(2, 2)),
-    ]
-    _syndrome_qubits = [PlaquetteQubit(Position(1, 1))]
+    @staticmethod
+    def _get_data_qubit_side_from_plaquette_orientation(
+        orientation: PlaquetteOrientation,
+    ) -> PlaquetteSide:
```
inmzhang, line 173: This relationship can be defined within the `PlaquetteOrientation` or `PlaquetteSide` enum.

src/tqec/plaquette/library/initialisation.py
nelimee: The problem of making the `get_data_qubits` and `get_data_qubits_cirq` methods non-static is that, most of the time, we need them before actually calling `__init__` on the `Plaquette` instance. For example in
https://github.com/QCHackers/tqec/blob/3df4a0588ef9aacd27e553af06b1af6ab38d40d9/src/tqec/plaquette/library/zz.py#L11-L33
we need to access the data qubits to create the circuit, before actually calling the `RoundedPlaquette.__init__` (which is exactly `super().__init__` here) method.

That could be solved by first creating the `Plaquette` instance with an empty circuit and then change that circuit, but this is not really desirable in my opinion.

Any idea how we could proceed?

inmzhang: > The problem of making the get_data_qubits and get_data_qubits_cirq methods non-static is that, most of the time, we need them before actually calling __init__ on the Plaquette instance. 

That makes sense to me. Anyway, we can at least refactor some of the circuit construction code, take syndrome measurement for example:
```python
def make_syndrom_measurement_circuit(
    syndrome_qubit: cirq.GridQubit, 
    ordered_data_qubits: list[cirq.GridQubit],
    detector: cirq.Operation | None,
    x_basis: bool = False
) -> cirq.Circuit:
    ...
```
Then we can use it in all XXXX/ZZZZ/XX/ZZ and Square/Rounded plaquettes.

nelimee:
Following the last commit, I have a few conceptual questions. 

First, currently, two "features" of a `Plaquette` may be passed down by inheritance: 
1. its shape: is it square or rounded
2. its effect: is it a `XX` syndrome measurement, `XXXX`, `ZZ`, `ZZZZ`?

Initially, the `RoundedPlaquette` and `SquarePlaquette` have been introduced to have a common definition **and order** of the data qubits the plaquette should operate on, as well as to have helper methods to recover the correct data qubits in the case of `RoundedPlaquette` instances. 

Now, I really think that it is more meaningful to have `XXMemoryPlaquette` (that should probably be renamed `XXSyndromeMeasurementPlaquette`) inherit from a `PauliMemoryPlaquette` (should be renamed too if introduced) than from a `RoundedPlaquette`.

Any opinion on that?

inmzhang:
> Now, I really think that it is more meaningful to have XXMemoryPlaquette (that should probably be renamed XXSyndromeMeasurementPlaquette) inherit from a PauliMemoryPlaquette (should be renamed too if introduced) than from a RoundedPlaquette.

LGTM. This kind of syndrome measurements should have the same structure at least at the stabilizer level. One comment which is not quite related to this PR, on the native gateset we used in the circuit implementation: While you have chosen the CNOT gate, I can think of another two ways to implement the controlled-pauli gates:

1. Use A control B gates, where A/B are control and target basis for the gate. This has the benefits of completely removing the basis change gates and is natively supported by `stim`. In the simple X/Z plaquette setting, this might be overkilling. But it can be useful in more complicated pauli syndrome measurements. However, these gates seem nontrivial in `cirq` and will need to be transpiled to some experimental native gateset if we want to simulate for experiment setups.
2. Use CZ with basis-change single qubit gates. This has the benefits of being symmetry and is native to superconducting qubits(and some other systems). But this can complicate the simple X/Z syndrome circuits a little bit.

I'm still fine with the CNOT, which can be transpiled to CZ gateset in the post circuit transformation.

inmzhang (approved):

nelimee:
I re-tested and re-read everything, this PR was a little bit too large, thanks @inmzhang for reviewing! I'll adapt #130 and merge as soon as possible.

Issue #134, opened 2024-02-12
Re-organise backend files


nelimee:
## Issue
The current file organisation in the `src/tqec` folder does not make a lot of sense.

<details><summary>Tree-view of the tqec backend files</summary><pre>
src/tqec
├── detectors
│   ├── measurement_map.py
│   ├── operation.py
│   └── transformer.py
├── display.py
├── enums.py
├── exceptions.py
├── generation
│   ├── circuit.py
│   ├── circuit_test.py
│   └── topology.py
├── noise_models
│   ├── after_clifford_depolarization.py
│   ├── after_reset_flip.py
│   ├── base.py
│   ├── before_measure_flip.py
│   ├── before_round_data_depolarization.py
│   ├── idle_qubits.py
│   ├── __init__.py
│   └── multi_qubit_gates.py
├── non_template_scaling
│   ├── animated_shape_display.py
│   ├── display_data.py
│   ├── __init__.py
│   └── non_template_scaling.py
├── plaquette
│   ├── library
│   │   ├── __init__.py
│   │   ├── xx.py
│   │   ├── xxxx.py
│   │   ├── zz.py
│   │   └── zzzz.py
│   ├── plaquette.py
│   ├── qubit.py
│   └── schedule.py
├── position.py
├── position_test.py
├── templates
│   ├── atomic
│   │   ├── rectangle.py
│   │   ├── rectangle_test.py
│   │   ├── square.py
│   │   └── square_test.py
│   ├── base.py
│   ├── composed.py
│   ├── constructions
│   │   ├── corner.py
│   │   └── qubit.py
│   ├── scale.py
│   └── scale_test.py
└── _version.py
</pre>
</details>

## Enhancement
It would be nice to re-organise the file hierarchy above to have a file organisation that makes more sense.

Example of files that, in my opinion, are not in the right place:
- `plaquette/schedule.py`,
- all the `detectors/` folder,
- `display.py` can likely go in the `templates/` folder,

Additionally, the `non_template_scaling` folder might not be relevant anymore. 

Any opinion on that? I would love to have the most feedback possible in order to do these changes only once.

inmzhang:
The structure should definitely be refactored, though I don't have strong opinions for now. A few first impressions without deep thought:

- Those circuit related modules can go to the same folder, i.e. `plaquette/schedule.py`, `generation/circuit.py`. Even the modules in the `detectors` folder can also be placed in the same place as those.
[...]

PR #153, opened 2024-02-22
Add top level imports

### LINKED ISSUE #113: Top level imports
Including imports at the top level of the package can be beneficial from the user's point of view, similar to what cirq has done in its [top-level](https://github.com/quantumlib/Cirq/blob/main/cirq-core/cirq/__init__.py).

What are your thoughts on this?

inmzhang:
Fixes #113

I'm not sure whether we need the `__all__` statements in each `__init__.py`. I understand it served as a convention for public api, but it's kind of cumbersome to add every method name twice within a single `__init__.py`.

### DIFF
--- a/src/tqec/templates/__init__.py
+++ b/src/tqec/templates/__init__.py
@@ -6,7 +6,6 @@ from .atomic import (
     AlternatingRectangleTemplate,
     RawRectangleTemplate,
     AlternatingSquareTemplate,
-    AlternatingCornerSquareTemplate,
 )
 
 from .composed import ComposedTemplate


src/tqec/templates/atomic/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from .square import (
+    AlternatingSquareTemplate,
+    AlternatingCornerSquareTemplate,
+)
+
+__all__ = [
+    "AlternatingRectangleTemplate",
+    "RawRectangleTemplate",
+    "AlternatingSquareTemplate",
+    "AlternatingCornerSquareTemplate",
```
nelimee, line 15: The `AlternatingCornerSquareTemplate` is currently on hold. If I remember correctly, the implementation uses permutations and I found at least one issue with that approach, as noted in the docstring:
https://github.com/QCHackers/tqec/blob/c1bf3d44697b77f04be9ee36436dca606d07b505/src/tqec/templates/atomic/square.py#L60-L63
I should have opened an issue back then, it seems like I forgot, I will do that just after finishing this review.
In the meantime, I am in favour of commenting out this class from the top-level imports, as a weak sign that it is not yet ready for the public API.

src/tqec/plaquette/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    RoundedPlaquette,
+    PlaquetteList,
+)
+
+from .qubit import PlaquetteQubit
+
+from .schedule import (
+    ScheduleException,
+    ScheduledCircuit,
+    ScheduledCircuits,
```
nelimee, line 13: The `ScheduledCircuits` class is only an implementation detail and does not appear in the public API, I am in favour of removing it from the top-level imports.
```suggestion
```

src/tqec/templates/atomic/__init__.py
nelimee:
```suggestion
```

src/tqec/templates/atomic/__init__.py
```diff
@@ -0,0 +1,16 @@
+from .rectangle import (
+    AlternatingRectangleTemplate,
+    RawRectangleTemplate,
+)
+
+from .square import (
+    AlternatingSquareTemplate,
+    AlternatingCornerSquareTemplate,
```
nelimee, line 8:
```suggestion
```

src/tqec/plaquette/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+__all__ = [
+    "Plaquette",
+    "SquarePlaquette",
+    "RoundedPlaquette",
+    "PlaquetteList",
+    "PlaquetteQubit",
+    "ScheduleException",
+    "ScheduledCircuit",
+    "ScheduledCircuits",
```
nelimee, line 24:
```suggestion
```

src/tqec/detectors/__init__.py
```diff
@@ -0,0 +1,26 @@
+from .operation import (
+    ShiftCoords,
+    RelativeMeasurementData,
+    RelativeMeasurementsRecord,
+    Detector,
+    Observable,
```
nelimee, line 6: Do we want to expose `ShiftCoords`, `Detector` and `Observable` here? Not exposing them in the top-level API would be a good sign to tell to the user "if you have to dig into the library to get something, you should know what you are doing, so read the documentation first" in my opinion.

nelimee: Also, I am not sure that the `RelativeMeasurementsRecord` should be exposed. Does it have another usage than avoiding code duplication between `Observable` and `Detector`?

nelimee (commented):
I have made per-file comments. 

I also have a question: do we need to import everything on the top-level `tqec` module? I think a user would be fine with `from tqec.templates import AlternatingSquareTemplate` (adding the `.template`). It would reduce the number of entries in the top level module and would, in my opinion, make it easier to discover for new users.

As a side note, the `__all__` attribute in the `__init__.py` file was initially added by me to silence quickly a linter warning about unused imports. I find that quite annoying now that I see it on several files with more imports, and see no point in doing that. I am in favour of removing that.

inmzhang:
> also have a question: do we need to import everything on the top-level tqec module? 

Yes, we can encapulate some modules within the sub-packages and not expose them directly in the top-level. I think the imports hierachy can be better determined when #134 is addressed.

nelimee (approved):
LGTM

PR #156, opened 2024-02-24
Replace `setup.cfg` with `pyproject.toml`

inmzhang:
This PR replaced `setup.cfg` with the modernized `pyproject.toml` for project setup. Currently, `setup.py` is kept for the potential more dynamic configurations such as compiled extensions in the future.


nelimee (approved):
This looks good! I tested on a Python 3.9 environment and all the notebooks seem to pass.

As a side question (not blocking the merge), what should we do about tool-specific parameters in your opinion? Should we include them in the `pyproject.toml` file, or keep the configuration elsewhere?

inmzhang:
I'd like to include thoes just within `pyproject.toml` since it  should work well for most configurations of the tools we used, i.e. `pylint`, `ruff`.

Issue #151, opened 2024-02-21
Persist unit cell


smburdick:
**Is your feature request related to a problem? Please describe.**
When the page closes or reloads, user loses the unit cell they've created, in addition to any plaquettes they've made

**Describe the solution you'd like**
When the user creates a unit cell, it should be saved to their local storage, and when starting the app, be presented the option to load an existing unit cell.

- [x] When user specifies a unit cell, save it to the Redux store.
- [ ] When the app starts, check the Redux store. If there's a unit cell, offer to populate the grid, otherwise proceed as-is

Once this is implemented, we can move on to plaquettes, templates, circuits, etc.

These can also be saved to the backend, if/when user login info is stored and tied to constructed circuits.

**Describe alternatives you've considered**
Task 2 integration will proceed asynchronously to this. Once more circuit components are added, we can persist those as well.

smburdick:
There can be a dedicated "Save" button, but saving it to the browser cache or the cloud (once logins are established) are fine. This will be especially important for templating.

smburdick:
Need to define this user story a little better

nelimee:
This issue is not relevant anymore. Closing it. See #323 for more details.

PR #157, opened 2024-02-26
Save unit cell to Redux store

smburdick:
* Address #151 part 1.
* Add reducers, store, and actions
* Minor refactoring


smburdick:
Ping @vtomole @rryoung98 @giangiac

rryoung98 (approved):
Let's keep pushing, again merging this without a rigorous check (we can fix any issues in a new PR + issue if necessary)

PR #173, opened 2024-03-04
Decouple Plaquette and qubits

### LINKED ISSUE #172: Improve `Plaquette` hierarchy by removing shape-related plaquettes

Currently, in the backend, [`Plaquette`](https://github.com/QCHackers/tqec/blob/a3ef9db7d0abebd13825030f540bd13b8f6c458b/src/tqec/plaquette/plaquette.py#L10) is defined as:
1) qubits, that basically store the relative physical location of the qubits used by the `Plaquette`,
2) a scheduled circuit that should be applied on the plaquette qubits.

Because specifying the qubits is quite error-prone, two helper classes have been created: [`SquarePlaquette`](https://github.com/QCHackers/tqec/blob/a3ef9db7d0abebd13825030f540bd13b8f6c458b/src/tqec/plaquette/plaquette.py#L64) and [`RoundedPlaquette`](https://github.com/QCHackers/tqec/blob/a3ef9db7d0abebd13825030f540bd13b8f6c458b/src/tqec/plaquette/plaquette.py#L129). These classes are inherited by the different plaquettes in the plaquette library.

This architecture makes less and less sense as:
- it still requires some convention on qubit order and adaption to this convention for inheriting plaquettes, which is easy to forget and will result in a non-descriptive error such as "overlapping gates in a Moment" raised from `cirq` at the very last stage of circuit creation. See https://github.com/QCHackers/tqec/blob/a3ef9db7d0abebd13825030f540bd13b8f6c458b/src/tqec/plaquette/library/zzzz.py#L21-L23 for an example.
- it leads to code duplication, e.g. [`tqec.plaquette.library.initialisation`](https://github.com/QCHackers/tqec/blob/a3ef9db7d0abebd13825030f540bd13b8f6c458b/src/tqec/plaquette/library/initialisation.py#L11).

The shape / qubit placement of the `plaquette` should avoid code duplication as much as possible in the plaquette library.

The main way I see to solve that issue is to create a new class hierarchy that will handle **separately from the `Plaquette` hierarchy** the different qubit placements. Each `Plaquette` instance would then require an instance from a class in that hierarchy to position its qubits.

nelimee:
This PR decouples the `Plaquette` classes from their `qubits`, removing `SquarePlaquette` and `RoundedPlaquette` and allowing the description of higher-level plaquettes with less code duplication.

Fixes #172

### DIFF
--- a/notebooks/corner.ipynb
+++ b/notebooks/corner.ipynb
@@ -30,7 +30,7 @@
          "source": [
             "from tqec.templates.constructions.corner import ScalableCorner\n",
             "from tqec.templates.scale import Dimension, LinearFunction\n",
-            "from tqec.display import display_template\n",
+            "from tqec.templates.display import display_template\n",
             "\n",
             "# Corner made of 2*2=4 plaquettes for the scaled distances.\n",
             "corner = ScalableCorner(Dimension(2, LinearFunction(2)))\n",
@@ -114,7 +114,15 @@
          "name": "python3"
       },
       "language_info": {
+         "codemirror_mode": {
+            "name": "ipython",
+            "version": 3
+         },
+         "file_extension": ".py",
+         "mimetype": "text/x-python",
          "name": "python",
+         "nbconvert_exporter": "python",
+         "pygments_lexer": "ipython3",
          "version": "3.12.0"
       }
    },
--- a/notebooks/doubling_distance.ipynb
+++ b/notebooks/doubling_distance.ipynb
@@ -108,7 +108,7 @@
                 "        self,\n",
                 "        plaquette_indices: ty.Sequence[int],\n",
                 "    ) -> numpy.ndarray:\n",
-                "        self._check_plaquette_number(plaquette_indices)\n",
+                "        self._check_plaquette_number(plaquette_indices, 3)\n",
                 "        x_plaquette, z_plaquette, special_plaquette = plaquette_indices[:3]\n",
                 "        arr = super().instantiate([x_plaquette, z_plaquette])\n",
                 "        arr[0, -1] = special_plaquette\n",
@@ -329,7 +329,7 @@
             },
             "outputs": [],
             "source": [
-                "from tqec.display import display_template\n",
+                "from tqec.templates.display import display_template\n",
                 "\n",
                 "display_template(doubling_template)"
             ]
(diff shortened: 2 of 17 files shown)


inmzhang (approved):
In the current implementation, It appeared to me that a `Plaquette` is just a wrapper around a `ScheduledCircuit`, as the qubits should be available from the specified local circuit. And the `PlaquetteQubits` only helps with the circuit construction within child classes. Then a `Plaquette` is equivalent to a `ScheduledCircuit` in the sense. From the frontend's perspective, the equivalence should be more obvious.

As for the plaquette hierachy, in my mind, a single `Plaquette` base class with some factory functions to create different library plaquettes should be enough and can eliminate the huge amount of child classes. For example:

```py
# only used for constructions of library plaquettes
class PlaquetteShape(enum.Enum):
    Square,
    LeftRounded,
    UpRounded,
    RightRounded,
    DownRounded,

make_initialization_plaquette(
    initialization_basis: Literal['X', 'Z'],
    plaquette_shape: PlaquetteShape, # determine the plaquette shape
) -> Plaquette:
    pass

make_measurement_plaquette(
    plaquette_shape: PlaquetteShape,
) -> Plaquette:
    pass

make_syndrome_extraction_plaquette(
    plaquette_orientation: PlaquetteShape,
    schedule: list[int],
    include_detector: bool = True,
    is_first_round: bool = False,
) -> Plaquette:
    pass
```

The OOP-style class hierachy should work well too, I just prefer the FP way personally.

nelimee:
> In the current implementation, It appeared to me that a `Plaquette` is just a wrapper around a `ScheduledCircuit`, as the qubits should be available from the specified local circuit. 

Right, for some reason it took me several days to realise that and I think I even said otherwise in a previous issue. 

> And the PlaquetteQubits only helps with the circuit construction within child classes.

Yes, and we can likely reflect that on the definition of `Plaquette` by not storing the `qubits` and only using the qubit-related classes for construction.

> As for the plaquette hierachy, in my mind, a single Plaquette base class with some factory functions to create different library plaquettes should be enough and can eliminate the huge amount of child classes.

I am fine either ways. The functional-way may produce larger chunks of code, but it will also produce less boilerplate. We could also provide both, implementing the functionalities in functions that are in the public API and re-using these functions internally to implement the classes. No hard opinion on that, we can ask to the larger community their preference and change the code accordingly.

nelimee:
As you approved this PR, I will merge and we can then iterate on it to change the interface using FP if this is the consensus.

PR #180, opened 2024-03-05
Canonical form circuit builder

smburdick:
* User can specify ancilla, measure, CX|CZ qubits in a plaquette
* Print out ASCII form to console when done

Demo: https://drive.google.com/file/d/1yChGhv0ErEw5g9OsJp4cRlKj_izdvWBQ/view?usp=sharing


Gistbatch:
Suggested changes from the call today:

- Ancilla and Measure are synonyms
- ASCII code is not formatted correctly; I assume the console is not in a monospace font
- "Cycle" through X and Z qubits in the plaquette

smburdick:
> Suggested changes from the call today:
> 
> * Ancilla and Measure are synonyms

Will remove specification of M

> * ASCII code is not formatted correctly; I assume the console is not in a monospace font

It is, but I should use more consistent formatted strings to achieve the results we want.

> * "Cycle" through X and Z qubits in the plaquette

Will replace individual CX and CZ flows with a singe CX/CZ flow.

rryoung98 (approved):
Haven't tested but let's expedite this so we can unblock the frontend migration

smburdick:
> Haven't tested but let's expedite this so we can unblock the frontend migration

OK, I'll open an issue for my/@Gistbatch's discussion.

PR #188, opened 2024-03-11
Rename frontend_v2/ to tutorial/

smburdick:
Since the canonical frontend is already located, rename this to reflect its intent.


rryoung98:
This is a trivial change g2g

Issue #189, opened 2024-03-12
Save plaquette data to plaquette libary


smburdick:
User should be able to select "Save plaquette" from its drop-down; clicking it will save the plaquette's shape, color, rotation, and circuit selection to the 

**Describe the solution you'd like**
- [ ] Implement a full `serialized()` function 
- [ ] Save JSON string to Redux store
- [ ] On start, get the plaquette templates from the Redux store.

**Additional context**
Tasks 4/5.

nelimee:
This issue is not relevant anymore. Closing it. See #323 for more details.

PR #193, opened 2024-03-13
🛩️  🏁  Removes the template specification process and cleans up template class 🐰 

### LINKED ISSUE #186: Change template specification to footprint specification
## Describe the bug
I shouldn't have to specify a template before I specify a plaquette.

1. Specify unit cell
1. I must specify template, when I should be able to specify a plaquette right away.

## Expected behavior
Once the qubits have populated, I should be able to start clicking them and when 3+ have been clicked, specify a plaquette.

## Screenshots and MP4s
image

## Other context:
From discussions with @afowler .

## Coding task

Bypass the `template` functions: hoist plaquette addition into the control flow function.

rryoung98:
Closes #186

### DIFF
--- a/frontend/src/plaquettes/Footprint.js
+++ b/frontend/src/plaquettes/Footprint.js
@@ -43,7 +43,7 @@ export default class Footprint {
    * @returns {void}
    */
   initializeFootprint = () => {
-    // Create the plaquettes and footprint
+    // Create the footprint
     this.selectQubitsButton.on('click', () => {
       this.app.view.addEventListener('click', this.selectQubit);
       this.unselectQubitsButton.visible = true;
--- a/frontend/src/plaquettes/Footprint.js
+++ b/frontend/src/plaquettes/Footprint.js
@@ -33,6 +33,7 @@ export default class Footprint {
     this.rectangle = new Graphics();
     this.workspace = workspace;
     this.container.name = 'footprint';
+    this.footprintQubits = [];
     this.initializeFootprint();
   }
 
@@ -203,7 +204,7 @@ export default class Footprint {
     this.isDragging = true;
 
     // Untint the qubits
-    this.templateQubits.forEach((qubit) => {
+    this.footprintQubits.forEach((qubit) => {
       qubit.changeColor('black');
     });
 
@@ -246,12 +247,12 @@ export default class Footprint {
           ) {
             // Change the color of the qubits
             child.changeColor('red');
-            this.templateQubits.push(child);
+            this.footprintQubits.push(child);
             return child;
-          } if (this.templateQubits.includes(child)) {
-            // If the qubit is no longer in the template area,
-            // remove it from the template
-            this.templateQubits = this.templateQubits.filter(
+          } if (this.footprintQubits.includes(child)) {
+            // If the qubit is no longer in the footprint area,
+            // remove it from the footprint
+            this.footprintQubits = this.footprintQubits.filter(
               (qubit) => qubit !== child
             );
             child.changeColor('black');
--- a/frontend/src/plaquettes/Footprint.js
+++ b/frontend/src/plaquettes/Footprint.js
@@ -4,15 +4,15 @@ import notification from '../components/notification';
 import Button from '../components/Button';
 
 /**
- * Template class to create the template for the plaquettes
- * @class Template
+ * Footprint class to create the footprint for the plaquettes
+ * @class footprint
  * @param {Container} workspace - The workspace container
  * @param {PIXI.Application} app - The PIXI application
  * @param {number} x - The x coordinate
  * @param {number} y - The y coordinate
  * @returns {void}
  */
-export default class Template {
+export default class Footprint {
   constructor(workspace, app, x, y) {
     // UI Properties
     this.app = app;
@@ -32,17 +32,17 @@ export default class Template {
     this.selectedQubits = [];
     this.rectangle = new Graphics();
     this.workspace = workspace;
-    this.container.name = 'template';
-    this.initializeTemplate();
+    this.container.name = 'footprint';
+    this.initializeFootprint();
   }
 
   /**
-   * Initialize the template
-   * @function initializeTemplate
+   * Initialize the footprint
+   * @function initializefootprint
    * @returns {void}
    */
-  initializeTemplate = () => {
-    // Create the plaquettes and template
+  initializeFootprint = () => {
+    // Create the plaquettes and footprint
     this.selectQubitsButton.on('click', () => {
       this.app.view.addEventListener('click', this.selectQubit);
       this.unselectQubitsButton.visible = true;
@@ -77,7 +77,7 @@ export default class Template {
   };
 
   /**
-   * Select a qubit from the template
+   * Select a qubit from the footprint
    * @function selectQubit
    * @param {Event} e - The click event
    * @returns {void}
@@ -116,12 +116,12 @@ export default class Template {
   };
 
   /**
-   * Create the plaquette from the selected qubits and assign it to the template
+   * Create the plaquette from the selected qubits and assign it to the footprint
    * @function createPlaquette
    * @returns {void}
   */
   createPlaquette = () => {
-    // Check that the selected qubits are part of the template area
+    // Check that the selected qubits are part of the footprint area
     if (this.selectedQubits.length < 3) {
       notification(this.app, 'Plaquette requires 3+ qubits');
       return;
@@ -130,7 +130,7 @@ export default class Template {
     const plaquette = new Plaquette(this.selectedQubits, this.workspace, this.app);
     // Add the plaquette to the tile container
     this.container.addChild(plaquette);
-    // Remove seleected qubits from the template qubits, so they can be used again
+    // Remove seleected qubits from the footprint qubits, so they can be used again
     // For each qubit, remove the text
     this.selectedQubits.forEach((qubit) => {
       qubit.removeChildren();
--- a/frontend/src/plaquettes/Footprint.js
+++ b/frontend/src/plaquettes/Footprint.js
@@ -1,4 +1,4 @@
-import { Container, Graphics } from 'pixi.js';
+import { Container, Graphics, Color } from 'pixi.js';
 import Plaquette from './Plaquette';
 import notification from '../components/notification';
 import Button from '../components/Button';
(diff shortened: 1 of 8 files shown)


smburdick:
Thanks @rryoung98 . This is exactly what I had in mind when we last spoke.

Per this morning's discussion, we'll want to be able to produce "plaquette footprints," which aren't the same as plaquette templates per se, hence why we needed to change the frontend to reflect that.

I think it would be nice to have the ability to specify the plaquette footprint using the infrastructure that exists within the `Template` class, which we can then rename.

To summarize, the workflow would look like:

Unit cell -> specify footprint (using click-and-drag) ->  specify circuit on plaquette, based on the footprint.

Additionally, once a footprint has been specified, we could be able to choose it from the library. Task #189 will provide the infrastructure for the library, which can include footprints, completed circuits, and scalable templates.

@afowler Do you have any more questions/commentary here? In particular, how would we define a template, now that we have footprints defined?

frontend/src/plaquettes/Template.js
```diff
@@ -1,230 +1,87 @@
...
-    });
-    this.container.addChild(this.templateButton);
-    this.container.addChild(this.plaquetteButton);
     this.container.name = 'template';
-    // this.workspace.addChild(this.container);
+    this.initializeTemplate();
   }
 
-  // Render the template control buttons
-  renderTemplateControlButtons() {
```
smburdick, line 43: We can still keep this functionality. That goes for most of the deleted code here

frontend/src/plaquettes/Template.js
```diff
@@ -1,230 +1,87 @@
...
+/**
+ * Template class to create the template for the plaquettes
+ * @class Template
+ * @param {Container} workspace - The workspace container
+ * @param {PIXI.Application} app - The PIXI application
+ * @param {number} x - The x coordinate
+ * @param {number} y - The y coordinate
+ * @returns {void}
+ */
 export default class Template {
```
smburdick, line 15: rename to `class Footprint`.

frontend/src/plaquettes/Template.js
```diff
@@ -1,230 +1,87 @@
...
+   * @returns {void}
+   */
+  initializeTemplate = () => {
+    // Create the plaquettes and template
+    this.selectQubitsButton.on('click', () => {
       this.app.view.addEventListener('click', this.selectQubit);
-      this.plaquetteButton.visible = true;
-      this.plaquetteButton.on('click', () => {
-        // Create the plaquettes and tile
-        this.createPlaquette();
```
smburdick, line 109: Footprint can still create a plaquette on top of it. In fact, I'm not sure I fully understand the difference between plaquettes and footprints at this point. @afowler Do you have any ideas?

afowler:
For the moment they should be only one footprint for everything. For the
moment we don't plan to combine plaquettes with different footprints in a
single circuit.

frontend/src/plaquettes/Template.js
```diff
@@ -247,14 +104,22 @@ export default class Template {
...
       // Show the button
-      this.plaquetteButton.visible = true;
+      this.selectQubitsButton.visible = false;
+      this.createPlaquetteButton.visible = true;
     }
   };
 
-  // Create the plaquette from the selected qubits and assign it to the template
+  /**
+   * Create the plaquette from the selected qubits and assign it to the template
```
smburdick, line 119: assign it to the `Footprint`? Again, I still need to understand the difference.

frontend/src/plaquettes/Template.js
```diff
@@ -263,13 +128,9 @@ export default class Template {
     }
     // Render the plaquette
     const plaquette = new Plaquette(this.selectedQubits, this.workspace, this.app);
-    // Remove seleected qubits from the template qubits, so they can be
```
smburdick, line 266: To be clear, I don't think a `Footprint` should remember the specific set of qubits (in the viewport) used to create it, but should remember a cell of qubit positions (just like the unit cell) for re-creating the footprint anywhere on the viewport. Can we add that to this class?

frontend/src/plaquettes/Template.js
```diff
@@ -1,230 +1,87 @@
...
     this.mode = 'static';
     this.isDragging = false;
+    this.x = x;
+    this.y = y;
     this.startX = 0;
     this.startY = 0;
     this.plaquette = null;
-    this.plaquetteButton = plaquetteButton;
-    this.templateQubits = selectedQubits || [];
+    this.selectQubitsButton = new Button('Select 3 qubits to make a Plaquette', x, y + 50);
```
smburdick, line 29: Before this point, qubits should not be clickable (when my mouse hovers over them, the Mickey Mouse hand should not appear). Can we change that?

smburdick: Also, it should say "Select 3+ qubits to make a Plaquette Footprint"

frontend/src/plaquettes/Template.js
```diff
@@ -1,230 +1,87 @@
...
-        this.createPlaquette();
-        this.workspace.addChild(this.container);
-        // Clear the selected qubits
-        this.selectedQubits = [];
-        notification(this.app, 'Step 4: Define the circuit');
+      this.unselectQubitsButton.visible = true;
+      this.selectQubitsButton.visible = false;
+      this.createPlaquetteButton.visible = false;
+    });
+    this.unselectQubitsButton.on('click', () => {
```
smburdick, line 52: Nice. I'm planning on adding undo/redo buttons in general (for the whole app), but this is good to have until that's implemented.

smburdick (changes requested):
Looks good overall, but making some suggestions based on this morning's meeting.

smburdick:
> For the moment they should be only one footprint for everything. For the moment we don't plan to combine plaquettes with different footprints in a single circuit.

@afowler OK, we can leave footprint specification as-is then. I'll add plaquette footprint saving/specification to the next milestone.

afowler:
The best example of the difference between a footprint and a plaquette is
our usual architecture and a memory surface code. The footprint of every
plaquette will be a square including 5 qubits. The plaquette itself, and
it's shape symbol, will fit inside this footprint, and could be a subset,
for example a triangle that attaches to the bottom, top, right, or left of
the footprint. Every plaquette should have the same footprint so that
plaquettes connect neatly when inserted into templates which are just
integer arrays.

Does this clarify the need for a footprint?

smburdick:
>  could be a subset, for example a triangle that attaches to the bottom, top, right, or left of the footprint.

@afowler That makes sense, but would the triangular plaquette still be something we want in the plaquette library? What would the use case for creating one be? I think the purpose of the library would be to have different *circuits*.

frontend/src/plaquettes/Template.js
rryoung98: do we want to leave it uncommented?

frontend/src/plaquettes/Template.js
rryoung98: Not sure what you mean here....I have made the name change though :)

frontend/src/plaquettes/Template.js
rryoung98: yes, would you mind pointing me to how we are instantiating the qubit positions?

frontend/src/plaquettes/Template.js
rryoung98: Done

rryoung98: the clicking functionality resides in the qubit class and we can do it in a separate issue.

frontend/src/plaquettes/Template.js
smburdick: Yes. If it's not being invoked at the moment, it should remain uncommented.

frontend/src/plaquettes/Template.js
smburdick: The notion of a plaquette is different from a plaquette footprint, and the goal of the name change is to make that clear. We should discuss in the meeting, I think.

frontend/src/plaquettes/Template.js
smburdick: When it is placed, that information should be known to the workspace. When it is stored (say, in redux), we only need to store the *relative* positions. There's a 1-1 correspondence, and storing relative positions helps us keep things general.

rryoung98: Okay makes sense, w.r.t. relative position could you point me towards any existing code that lays out the foundation?

frontend/src/plaquettes/Template.js
rryoung98: That's clear to me. I wasn't sure if you were commenting on the edit from single line comment to multiline

frontend/src/plaquettes/Template.js
rryoung98: added it back

frontend/src/plaquettes/Template.js
smburdick: I'm saying it should say `assign it to the footprint`

smburdick:
@rryoung98 Is this ready for re-review?

rryoung98:
@smburdick i haven't done the relative positioning since the frontend version 2, if i recall might already have the relative positioning notion. Otherwise, it's all set for re-review

smburdick:
> @smburdick i haven't done the relative positioning since the frontend version 2, if i recall might already have the relative positioning notion. Otherwise, it's all set for re-review

Thanks, and in the future, be sure to request that from the reviewer when ready.

frontend/src/plaquettes/Footprint.js
```diff
@@ -3,123 +3,194 @@ import Plaquette from './Plaquette';
...
+    const relativeY = e.clientY - canvasRect.top;
+    // Get all the qubits
+    const qubits = this.workspace.children.filter(
+      (child) => child.isQubit === true
+    );
+    const qubit = qubits.find(
+      // Find the qubit that was clicked
+      (q) => q.checkHitArea(relativeX, relativeY) === true
+    );
+    if (!qubit && !(qubit?.isQubit === true)) return; // Check that the qubit exists
```
smburdick, line 101: `qubit.isQubit` should be deprecated, since a qubit is always a qubit; perhaps you meant to check its visibility?

frontend/src/plaquettes/Footprint.js
```diff
@@ -3,123 +3,194 @@ import Plaquette from './Plaquette';
...
+    this.createPlaquetteButton.visible = false;
+    this.unselectQubitsButton.visible = false;
+  };
+
+  /**
+   * Render the footprint
+   * @function renderFootprintControlButtons
+   * @returns {void}
+   */
+  renderFootprintControlButtons() {
```
smburdick, line 153: A better name would be something like `updateVisiblityAndEventListeners()` since rendering is not the job of a `Container` per se

smburdick (commented):
Minor comments.

frontend/src/plaquettes/Footprint.js
rryoung98: This is a built in for the QubitClass since it's a convenient way to check if child components, such as graphics or containers are indeed representing a qubit. Without the .isQubit state, it would be difficult to filter through a set of child containers and determine if its a qubit or not.

smburdick: In that case, you should use `instanceof Qubit`.

rryoung98: sounds good! thanks

rryoung98:
@smburdick  please re-review

smburdick (approved):
image

I think we should remove this label as a button, since I don't have to actually click it to make a new footprint -- I can still make plaquettes by clicking on qubits, and clicking "create plaquette". So, instead of having the button, we can simply add the text to the topbar.

Looks good otherwise, you can make the changes here, or create a new issue to address that.

rryoung98:
> image
> I think we should remove this label as a button, since I don't have to actually click it to make a new footprint -- I can still make plaquettes by clicking on qubits, and clicking "create plaquette". So, instead of having the button, we can simply add the text to the topbar.
> 
> Looks good otherwise, you can make the changes here, or create a new issue to address that.

Done

PR #196, opened 2024-03-21
New arithmetic operators on Dimension

nelimee:
This PR is the only feature that is worth cherry-picking from #100 (and not already in main).


src/tqec/templates/scale.py
```diff
@@ -13,6 +15,33 @@ def __call__(self, x: int) -> int:
...
+        return LinearFunction(self.slope + other.slope, self.offset + other.offset)
+
+    def __sub__(self, other: "LinearFunction") -> "LinearFunction":
+        return LinearFunction(self.slope - other.slope, self.offset - other.offset)
+
+    def __mul__(self, other: int) -> "LinearFunction":
+        return self.__rmul__(other)
+
+    def __rmul__(self, other: int) -> "LinearFunction":
+        return LinearFunction(other * self.slope, other * self.offset)
```
inmzhang, line 28: Is there use cases for these arithmetic operators?

nelimee: I find it easier and less error-prone to use such operators to place templates representing logical qubits on the logical qubit grid. For example:
```py
# Before
logical_qubit_dimension = Dimension(2, LinearFunction(2, 2))
shift_for_logical_qubit = Dimension(2, LinearFunction(6, 6))

# With arithmetic operators
one_fixed_dimension = FixedDimension(1)
inner_qubit_dimension = Dimension(2, LinearFunction(2))
# decomposition not strictly needed here, the "logical_qubit_dimension" fropm before also makes sense
logical_qubit_dimension = 2 * one_fixed_dimension + inner_qubit_dimension
shift = 3 * logical_qubit_dimension
```
In the example above, computing the shift is easier, more readable and less error-prone. Also, I think this change is a low-effort, low-risk, low-reward one: it does bring little (but not none) value, cost nothing to implement, and is pretty unlikely to lead to bugs.

inmzhang: The `LinearFunction` with the given arithmetic rules is not a [dual vector space](https://en.wikipedia.org/wiki/Dual_space) so it's a little weired for me.

nelimee: I may be completely off on that but isn't the test
```py
def test_linear_function_operators():
    a, b = LinearFunction(2, 5), LinearFunction(3, 1)
    assert (a + b)(10) == a(10) + b(10)
    assert (a - b)(3) == a(3) - b(3)
    assert (3 * a)(54) == 3 * a(54)
    assert (a * 3)(54) == 3 * a(54)
```
exactly checking that it is indeed a dual vector space?

inmzhang (approved):

PR #203, opened 2024-04-10
Functional style plaquette

nelimee:
## Goal of the PR

Per [this comment](https://github.com/QCHackers/tqec/pull/173#pullrequestreview-1914500988) from @inmzhang and [the discussion/poll that followed](https://github.com/QCHackers/tqec/discussions/179) it seems like the active developers in this repository prefer a functional-style rather than an inheritance based style. This means that instead of 
```py
class Plaquette:
    pass

class EmptyPlaquette(Plaquette):
    pass

class ZInitialisationPlaquette(Plaquette):
    pass

# many more
```
with a `Plaquette` class whose sole purpose is to provide a common interface between plaquette, everyone voted for a functional-style:
```py
class Plaquette:
    pass

def empty_plaquette(qubits: PlaquetteQubits) -> Plaquette:
    pass

def z_initialisation_plaquette(qubits: PlaquetteQubits) -> Plaquette:
    pass

# many more
```
with `Plaquette` being the only class used, and sub-classes are replaced by functions that initialise `Plaquette` instances.

## How was it done?
This change is mostly (entirely?) cosmetic, as the sub-classes of `Plaquette` were only initialising correctly their `Plaquette` subclasses. This shows quite well in this PR: the changes are quite small and can be summarised in 2 points:
- Change the subclass name by the function name. The new functions have exactly the same parameters as their corresponding `Plaquette` subclass, so this change can be done using a "dumb" tool (i.e., just search & replace, no need for code-aware tools).
- Change the subclasses to functions:
  ```py
  class [CLASSNAME](Plaquette):
      def __init__(self, [PARAMETERS]):
          [CODE]
          super().__init__([SUPER_PARAMETERS])
  ```
  becomes
  ```py
  def [SNAKE_CASE(CLASSNAME)](PARAMETERS) -> Plaquette:
      CODE
      return Plaquette([SUPER_PARAMETERS])
  ```
  with a slight variation for classes that were 2 or more levels under `Plaquette` in the inheritance hierarchy (e.g. `ZRoundedInitialisationPlaquette` -> `ZInitialisationPlaquette` -> `Plaquette`) for which the `super().__init__` call should be replaced with their superclass corresponding function.


nelimee:
Pinging for review, I would like this branch to be merged if you are still convinced that this is the right way to go. Feel free to comment on anything like @smburdick did, this is helpful to see if I missed anything.

afowler:
+1 to getting some help reviewing this!

inmzhang (approved):
LGTM

PR #204, opened 2024-04-10
Improve plaquette number checking

### LINKED ISSUE #197: Improve `_check_plaquette_number`

The `Template` classes are using inheritance. When `instantiating` a `Template`, it is easy to give the wrong number of plaquette indices as input. This error-prone interface is made a little bit less cumbersome to work with by checking systematically that the number of plaquette indices provided agrees with what the `Template` instance is expecting, by calling `_check_plaquette_number`. 

Currently, the `_check_plaquette_number` method is quite hard to use and easy to misuse, a better way of checking would be nice.

Also, in the case of inheriting `Template`s, the `self.__class__.__name__` part in the error message will always resolve to the final type in the inheritance graph, which is not what is expected.

A better way to check that the user provided the right number of plaquettes that is also able to raise meaningful and useful error messages.

None yet.

nelimee:
I do not see a better way of checking the number of plaquettes provided to the `instantiate` method. This PR adds documentation to `Template._check_plaquette_number` to explicitly provide guidance on how to use that method.

Fixes #197

### DIFF
--- a/src/tqec/templates/base.py
+++ b/src/tqec/templates/base.py
@@ -100,7 +100,7 @@ class Template(JSONEncodable):
         """
         if len(plaquette_indices) < expected_plaquettes_number:
             raise TQECException(
-                f"{self.__class__.__name__}.instanciate needs "
+                f"Calling an instanciate method that requires "
                 f"{expected_plaquettes_number} plaquettes, but only "
                 f"{len(plaquette_indices)} were provided."
             )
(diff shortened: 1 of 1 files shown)


src/tqec/templates/base.py
```diff
@@ -100,7 +109,7 @@ def _check_plaquette_number(
         """
         if len(plaquette_indices) < expected_plaquettes_number:
             raise TQECException(
-                f"{self.__class__.__name__}.instanciate needs "
+                f"Calling an instanciate method that requires "
```
vtomole, line 112:
```suggestion
                f"Calling an instantiate method that requires "
```

vtomole (commented):
One nit then LGTM

vtomole (approved):
LGTM

PR #206, opened 2024-04-10
Tag Pauli memory syndrome qubit as mergeable

### LINKED ISSUE #205: Resets in pauli memory plaquette should be annotated as mergeable
## Describe the bug
The bug is described in #202: the reset conditionally applied to the syndrome qubit in 
https://github.com/QCHackers/tqec/blob/47cb5e9712ab02959df9e2ce552cf8c663a2b110/src/tqec/plaquette/library/utils/pauli.py#L56
should be annotated as mergeable. 

Without this annotation, it will not be merged by the call to [`remove_duplicate_operations`](https://github.com/QCHackers/tqec/blob/47cb5e9712ab02959df9e2ce552cf8c663a2b110/src/tqec/circuit/schedule.py#L440) when merging plaquette circuits in
https://github.com/QCHackers/tqec/blob/47cb5e9712ab02959df9e2ce552cf8c663a2b110/src/tqec/circuit/schedule.py#L475-L498

## Expected behavior
The reset applied on the first memory plaquette syndrome qubit should be merged with initialisation plaquette resets.

nelimee:
Fix #205

### DIFF
--- a/src/tqec/plaquette/library/utils/pauli.py
+++ b/src/tqec/plaquette/library/utils/pauli.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import cirq
 from tqec.exceptions import TQECException
+from tqec.plaquette.plaquette import Plaquette
 from tqec.plaquette.qubit import PlaquetteQubit
 
 _SUPPORTED_PAULI: set[str] = set("xz")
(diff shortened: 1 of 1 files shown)


vtomole (approved):
LGTM

PR #209, opened 2024-04-11
Fix list syntax

smburdick:
There can be up to 4 sponsored usernames here. Right now, it's only me, but it needs to be in list format.


vtomole (approved):
LGTM

Issue #141, opened 2024-02-15
Specifying an `Observable` requires too much knowledge about implementation details


nelimee:
## The issue
We have an [`Observable`](https://github.com/QCHackers/tqec/blob/09229fe45ede2827fa1718cf6fac6c3613254733/src/tqec/detectors/operation.py#L200) operation that can be used to add an observable to the quantum circuit generated. 

Creating correctly an observable currently requires to know implementation details. For example, because you can only create an observable by providing global qubit positions, you **need** to understand how the code is working in order to pick the right qubits.

## Expected improvement
It would be nice to be able to create an observable without needing any knowledge about the internal implementation details. 

I do not have any idea to make that easy at the time being, I am open to all suggestions.

afowler:
Hi Adrien,

Some ideas below...

A final observable will be associated with an array of measurements, so
potentially fairly simple plaquettes. Is it possible to write a function
that detects whether a square template with just measurement plaquettes has
a well defined central vertical line or horizontal line of qubits? The idea
is to scalably choose the observable. A function that takes a final
measurement layer in, and H or V, and produces an observable. I believe
Stim is capable of detecting whether an observable is valid. It's ok if the
function fails to cope with some inputs.

I believe that ultimately to automate things we are going to need to be
able to detect such midlines.

When it comes to moving observables horizontally, and assuming we have well
defined midlines, it seems that templates with plaquettes with single
measurements as well as clean midlines would be a good starting point.
Moving observables from midline to midline would then typically mean
choosing to extend and contract along midlines, or using one half of the
checkerboard of plaquettes to move a midline operator sideways.

Happy to chat more 1-1 or in our next meeting.

Best,
Austin.

Gistbatch:
I can have a look at this

Issue #198, opened 2024-03-21
Functional-style Plaquette


nelimee:
**Is your feature request related to a problem? Please describe.**
As discussed in #179, functional-style `Plaquette` is the preferred way. 

**Describe the solution you'd like**
The `Plaquette` class should be made `final` and the actual classes that inherit from `Plaquette` should be replaced by functions.

**Describe alternatives you've considered**
Alternative: keep the the class hierarchy. Not the path picked in #179.

nelimee:
I forgot about #183, closing this as duplicate

PR #213, opened 2024-04-22
Observables

Gistbatch:
closes: #141

I started to develop this, and I have some open questions:

- I've implemented a test for the `RawRectangleTemplate`. @nelimee, can you clarify if this is the intended behavior?
- What should the scope of this Issue/PR be? Only build this for the available templates, including moves and so on.
- How do we define midlines for uneven-sized templates, e.g., where the number of qubit rows is even?

Todos:
- [ ] Add all templates
- [ ] Write tests, including scaling
- [ ] Figure out edge cases: Moving, uneven ...
- [ ] Documentation


afowler:
In my opinion there would be very little benefit and significant additional
complexity allowing even code distances, so I would support not allowing
this at all and consequently always having well defined mid lines.

Best,
Austin.

nelimee:
> I've implemented a test for the `RawRectangleTemplate`. @nelimee, can you clarify if this is the intended behavior?

The output you fixed is dependent on the actual `Plaquette` implementation. What if the origin (i.e., `(0, 0)` point) was on the top-left of the `Plaquette`? Then both the output coordinates would change. 
I cannot remember if we completely ruled out that case though.

In any case, I think this method needs the `Plaquette` instances to be able to function correctly and generically.

> What should the scope of this Issue/PR be? Only build this for the available templates, including moves and so on.

For this PR, only build this for the available `Template`s. Moving will come after.

> How do we define midlines for uneven-sized templates, e.g., where the number of qubit rows is even?

As Austin said, I would rule out that case. To do that, I would change the `default_observable_qubits` return type to `ty.Sequence[tuple[cirq.GridQubit, int]] | None`, returning `None` being a signal that the midline could not be computed.

Also, I think the implementations of `default_observable_qubits` might be factorisable. But let's have something working for the moment.

Gistbatch:
@nelimee, I've updated this; let me know what you think.

If the midline is supposed to depend on the plaquettes, we need this information, so I built `observable_qubits_from_template` similar to the `generate_circuit` function. The template now only returns the row and column indices of the templates that have the observable qubits, by convention, above/left of the midline.

nelimee:
> If the midline is supposed to depend on the plaquettes

I do not get how you can generate correct results without any idea of the plaquette used. For the moment, all the plaquettes we use have the support
```
1   2
  5 
3   4
```
with:
1. `cirq.GridQubit(-1, -1)`
2. `cirq.GridQubit( 1, -1)`
3. `cirq.GridQubit(-1,  1)`
4. `cirq.GridQubit( 1,  1)`
5. `cirq.GridQubit( 0,  0)`

but (as far as I remember) we never explicitly said that was the only possible default. This means that a plaquette with the same support but with
1. `cirq.GridQubit(0, 0)`
2. `cirq.GridQubit(2, 0)`
3. `cirq.GridQubit(0, 2)`
4. `cirq.GridQubit(2, 2)`
5. `cirq.GridQubit(1, 1)`

is possible.
If that is the case, then we need to have access to that information, else the midline qubits will be completely wrong. I am missing something here?

I am making a few comments on the code before commenting again on the overall approach.

src/tqec/templates/atomic/rectangle.py
```diff
@@ -202,3 +205,16 @@ def to_dict(self) -> dict[str, ty.Any]:
...
     def expected_plaquettes_number(self) -> int:
         return max(max(line) for line in self._indices) + 1
+
+    def get_midline_plaquettes(self, horizontal: bool = True) -> list[tuple[int, int]]:
+        if horizontal:
+            if self.shape.y % 2 == 1:
+                raise TQECException("Midline is not defined for odd height.")
+            midline = self.shape.y // 2 - 1
+            indices = self._indices[midline]
+            return [(midline, column ) for column, _ in enumerate(indices)]
```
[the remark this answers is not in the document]
nelimee: Or even 
```suggestion
            return [(midline, column) for column in range(self.shape.x)]
```

nelimee (commented):
I did not check thoroughly the code validity for the moment, but your tests help on that. 
For the `Template.get_midline_plaquettes` method, I completely agree with the fact that we need a convention such as the one you used (plaquettes left of the vertical midline or above the horizontal midline). The two branches (depending on the value of the `horizontal` boolean) are very similar, maybe they can be "merged" in one more generic version.

afowler:
Some reasonable restrictions. We already restrict ourselves to plaquettes
with a rectangular footprint. The midline calculation is associated with a
square 2kx2k template of such plaquettes. It would be reasonable to
restrict ourselves to plaquettes with exactly four qubits in the four
corners of the plaquette, then the midline calculation should depend on the
shape of the plaquette, rather than the precise labeling of the qubits in
the plaquette.

Best,
Austin.

Gistbatch:
Using their coordinates, I take the qubits furthest to the right (bottom). This works only for rectangular shapes but has the advantage that this works independently of the orientation.
As an alternative, we could implement this as the default behavior in the plaquette, which can be overwritten.
I'll clean up the code once we've decided on the mechanism and implement the rest of the templates.

nelimee:
> then the midline calculation should depend on the shape of the plaquette, rather than the precise labeling of the qubits in the plaquette.

I agree, the issue here is that, in the current state of the code, the positions of the plaquette qubits can change all the positions. Basically, the positioning algorithm starts by positioning the origin of the top-left-most plaquette on the qubit `(0, 0)`, all the other qubits position being ultimately determined relative to that position. That mean that in the 2 cases I presented above, all the qubits of the template are shifted by `(1, 1)`. And we need exact coordinates for the midline.

This limitation should be quite easy to remove though, using the top-left corner of the plaquette instead of the origin in the first step would solve it on the current use-case. 
Finally, this is very much an implementation detail for the moment, but it has implications on what the user see and experience, so we might want to change the current default anyway.

> Using their coordinates, I take the qubits furthest to the right (bottom). This works only for rectangular shapes but has the advantage that this works independently of the orientation.
> As an alternative, we could implement this as the default behavior in the plaquette, which can be overwritten.

Both approaches seems fine for me. The second one is likely more generic, but we are only interested in rectangular plaquette for the moment so this is not really a problem to implement a simpler, less generic, version.

Gistbatch:
@nelimee, I'm confused about why the test keeps failing. Do you have any idea why this is the case? I can run it locally without this problem, and I'm not sure how to fix this as it seems it works completely fine with `circuit_test` which has a similar import structure.

nelimee:
> @nelimee, I'm confused about why the test keeps failing. Do you have any idea why this is the case? I can run it locally without this problem, and I'm not sure how to fix this as it seems it works completely fine with `circuit_test` which has a similar import structure.

This might be due to #198 being merged in main. I am not sure what https://github.com/actions/checkout is checking out, but it might be worth updating this branch with the latest changes from `main` and see if that resolves the issue.

src/tqec/circuit/observable_qubits.py
```diff
@@ -0,0 +1,96 @@
+from typing import Any, Mapping, Sequence
```
nelimee, line 1:
```suggestion
from __future__ import annotations

from typing import Any, Mapping, Sequence
```
fixes some test issues.

nelimee (commented):
One suggestion to fix at least one test.

The other errors are expected as not all `Template` subclasses implement the `get_midline_plaquettes` method, which makes them abstract and non-instantiable

Gistbatch:
I've now implemented the midline definition for the existing template instances.
For the `ComposedTemplate` and its children, I'm wondering what the default behavior should be.
It seems impossible to know which qubits are on the midline without prior knowledge.
For example, the `QubitSquareTemplate` has a midline, but none of the associated template midlines coincide.
We could hard code this, but figuring out which templates to look at for qubits seems arbitrary and complex.
@afowler, @nelimee any suggestions?

nelimee:
It seems at least hard, maybe impossible (there are likely instantiations of the `CompositeTemplate` that do not represent a valid code, and so do not have observables), to me too.
I will review the code as it is right now, let's wait for @afowler answer.

src/tqec/templates/atomic/square.py
```diff
@@ -171,3 +172,17 @@ def scale_to(self, k: int) -> "AlternatingCornerSquareTemplate":
...
+        midline_shape, iteration_shape = self.shape.x, self.shape.y
+        if midline_shape % 2 == 1:
+            raise TQECException(
+                "Midline is not defined for odd "
+                + f"{'height' if horizontal == TemplateOrientation.HORIZONTAL else 'width'}."
+            )
+        midline = midline_shape // 2 - 1
+        if horizontal == TemplateOrientation.VERTICAL:
+            return [(midline, column) for column in range(iteration_shape)]
+        return [(row, midline) for row in range(iteration_shape)]
```
nelimee, line 188: Isn't this one already implemented by the `AlternatingRectangleTemplate.get_midline_plaquettes`? Why aren't you swapping `midline_shape` and `iteration_shape` depending on the orientation here?

src/tqec/templates/stack.py
```diff
@@ -105,3 +107,32 @@ def instantiate(self, plaquette_indices: ty.Sequence[int]) -> numpy.ndarray:
...
+            if (
+                horizontal == TemplateOrientation.HORIZONTAL
+                and template.shape.y == midline_shape
+            ):
+                return template.get_midline_plaquettes(horizontal)
+            if (
+                horizontal == TemplateOrientation.VERTICAL
+                and template.shape.x == midline_shape
+            ):
+                template.get_midline_plaquettes(horizontal)
```
nelimee, line 135:
```suggestion
            elif (
                horizontal == TemplateOrientation.VERTICAL
                and template.shape.x == midline_shape
            ):
                return template.get_midline_plaquettes(horizontal)
```

src/tqec/circuit/observable_qubits.py
```diff
@@ -0,0 +1,96 @@
...
+    horizontal: TemplateOrientation = TemplateOrientation.HORIZONTAL,
+) -> Sequence[tuple[cirq.GridQubit, int]]:
+    """Return the default observable qubits for the given template and its plaquettes.
+
+    Args:
+        template (Template): The template to get the default observable qubits from.
+        plaquettes (list[Plaquette] | dict[int, Plaquette]): The plaquettes to use to get the
+        acurate positions of the observable qubits.
+        horizontal (bool, optional): Whether to get the observable qubits from
+            the horizontal or vertical midline. Defaults to True -> Horizontal.
```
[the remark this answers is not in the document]
Gistbatch: I haven't updated docs and tests yet, I'll do this once functionality is fixed

src/tqec/templates/atomic/square.py
Gistbatch: since this is a square `midline_shape == iteration_shape`, and this class inherits from `Template` so we can't inherit the function.

nelimee: Right, I missed that, thanks for the answer :)

Gistbatch:
I wrote tests for all the templates now.
I don't know what the intended behavior for the stacked template is.
Also, I left the `CornerTemplate` and `StackedTemplate` unimplemented.
I could hardcode the corner template similar to the qubit template if needed.
There isn't a good way of knowing what the midline is supposed to be for the stacked template, so I just left it. 
The only option is to give all possible midlines to the user, and they have to piece them together themselves.

Gistbatch:
@nelimee I'll merge this once you approve

src/tqec/circuit/observable_qubits.py
```diff
@@ -0,0 +1,94 @@
...
+from tqec.exceptions import TQECException
+from tqec.plaquette.plaquette import Plaquette
+from tqec.plaquette.qubit import PlaquetteQubit
+from tqec.templates.base import Template
+
+
+def observable_qubits_from_template(
+    template: Template,
+    plaquettes: Sequence[Plaquette] | Mapping[int, Plaquette],
+    horizontal: TemplateOrientation = TemplateOrientation.HORIZONTAL,
```
nelimee, line 16:
```suggestion
    orientation: TemplateOrientation = TemplateOrientation.HORIZONTAL,
```

src/tqec/circuit/observable_qubits.py
```diff
@@ -0,0 +1,94 @@
...
+) -> Sequence[tuple[cirq.GridQubit, int]]:
+    """Return the default observable qubits for the given template and its plaquettes.
+
+    Args:
+        template: The template to get the default observable qubits from.
+        plaquettes: The plaquettes to use to get the acurate positions of the observable qubits.
+        horizontal: Whether to get the observable qubits from
+            the horizontal or vertical midline. Defaults to True -> Horizontal.
```
nelimee, line 24:
```suggestion
        orientation: Whether to get the observable qubits from
            the horizontal or vertical midline. Defaults to horizontal.
```

src/tqec/circuit/observable_qubits.py
```diff
@@ -0,0 +1,94 @@
...
+        observable_qubits += [
+            (qubit.to_grid_qubit() + offset, 0)
+            for qubit in _get_edge_qubits(plaquette, horizontal)
+        ]
+    return sorted(set(observable_qubits))
+
+
+def _get_edge_qubits(
+    plaquette: Plaquette,
+    horizontal: TemplateOrientation = TemplateOrientation.HORIZONTAL,
```
nelimee, line 80:
```suggestion
    orientation: TemplateOrientation = TemplateOrientation.HORIZONTAL,
```

src/tqec/circuit/observable_qubits.py
nelimee: WARNING: should also be changed in the function body if changed here.

src/tqec/circuit/observable_qubits.py
nelimee: WARNING: should also be changed in the function body if changed here.

src/tqec/templates/atomic/rectangle.py
```diff
@@ -90,6 +91,23 @@ def to_dict(self) -> dict[str, ty.Any]:
     def expected_plaquettes_number(self) -> int:
         return 2
 
+    def get_midline_plaquettes(
+        self, horizontal: TemplateOrientation = TemplateOrientation.HORIZONTAL
```
nelimee, line 95:
```suggestion
        self, orientation: TemplateOrientation = TemplateOrientation.HORIZONTAL
```
WARNING: should also be changed in the function body if changed here.

nelimee (approved):
Some very minor comments, mostly about renaming `horizontal` to `orientation`. 
There is also some duplication of code, but this can be addressed later. 

A few other comments for the future of that feature:
- I can see the `_get_edge_qubits` function being implemented in the `PlaquetteQubits` class. With a little bit of geometry, we could even implement something like `PlaquetteQubits.extremum_qubits_in_direction(self, x: float, y: float)`. This can be thought about in a follow-up issue.
- The more I think about the short-term and end goal of the community, the more I think I made a mistake by introducing `StackedTemplate` and `ShiftedTemplate`. These are 1. not really useful and 2. making a few PR more difficult than what they should be (this PR being a good example, another exploration work I have locally being another). 

In the end, except the minor comments, I think this is a good first step to include automatic observable computation.

Issue #212, opened 2024-04-17
Merge frontend and alternative-frontend


smburdick:
The purpose of this issue is to unify all the work we've put in to build a comprehensive service. The `alternative-frontend` has several useful features we'd like to incorporate into the live frontend, `tqec.app`.

## UI components
- [x] Click-cycle `A, X, Z` qubit labels and corresponding colors
- [ ] `App.js`: ~~Migrate `handleTabClick()` to `frontend/`~~ integrate tabbing and menu selection into `frontend/`
- [ ] ~~Migrate `TqecLibrary()` from `index.js` to its own file in `frontend/`.~~ Implement library functionality
- [x] Do not use `console.log`, use a dialog box instead.

## Data exportation
- [x] Migrate `StimCircuitWriter.js` to `frontend/`
- [x] `circuit.js`: migrate `createCircuitStimCode()` to `frontend`, along with other ASCII circuit writers that are not already present.
- [x] Update qubit role enumeration for ASCII conversion.

## Security and performance
- [x] Do not use `document.getElementById`; use React `jsx` syntax to define interactive components.

## CI/CD
- [x] Deploy `alternative-frontend` to [tqec2live.org](http://tqec2live.org/)

## Clean coding
- [x] `gridSize` must be defined as a constant in a file and the Redux store.
- [x] Use `Object.freeze()` to define constant enumerations.
- [x] Code comments must be consistent and adherent to ESLint and clean coding.
- [x] Source code lines should not exceed 120 characters.
- [x] Open separate issue for `.eslintrc.json` changes.
- [x] Remove `alternative-frontend/`.

smburdick:
Now that the task flow has been revised, the next step is to extract the UI updates into their own issues, and resolve this when all the other issues have been resolved.

smburdick:
Going to make the library and tabbing functionality its own issue, since that wasn't in the other frontend to begin with anyway.

PR #216, opened 2024-04-27
Relocate alternative frontend

smburdick:
The alternative frontend now lives at https://github.com/QCHackers/tqec-alternative-frontend.

Related: #212


vtomole:
Why the relocation? Is this on the way to merging the projects?

smburdick:
> Why the relocation? Is this on the way to merging the projects?

It is indeed.

vtomole (approved):
LGTM

vtomole:
@giangiac I think the plan is to cherry pick the important changes from the alternative frontend and wisely merge them to the existing one; which is already under way by @smburdick

smburdick:
> @giangiac I think the plan is to cherry pick the important changes from the alternative frontend and wisely merge them to the existing one; which is already under way by @smburdick

I thought having two frontends would be confusing for newcomers.

The alternative frontend is preserved at  https://alt-tqecjs-421617.wl.r.appspot.com/ and https://github.com/QCHackers/tqec-alternative-frontend

I've already cherry picked the features into https://tqec.app (this repo frontend)

giangiac:
I believe we discussed in the weekly call to provide both frontends since they have different features.

I understand that it may be confusing having two frontends if we agree that:
1) one design is clearly superior and the other just a distraction
2) or if the TQEC community has no bandwidth to pursue more than one approach.

There can be a case for 2) since only a handful of members are contributing to the frontend.
Concerning 1), the situation is less clear for me. The current frontend has a more professional codebase (good coding practices, enhanced security, ...) while the alternative one offers more functionalities.

smburdick:
> I believe we discussed in the weekly call to provide both frontends since they have different features.
> 
> I understand that it may be confusing having two frontends if we agree that:
> 
> 1. one design is clearly superior and the other just a distraction
> 2. or if the TQEC community has no bandwidth to pursue more than one approach.
> 
> There can be a case for 2) since only a handful of members are contributing to the frontend. Concerning 1), the situation is less clear for me. The current frontend has a more professional codebase (good coding practices, enhanced security, ...) while the alternative one offers more functionalities.

@giangiac See my above comment (you may not have received the email):
> The alternative frontend is preserved at https://alt-tqecjs-421617.wl.r.appspot.com/ and https://github.com/QCHackers/tqec-alternative-frontend

smburdick:
Alternatively, http://tqec2live.org.

Issue #219, opened 2024-05-04
[Rev.] Task 2: footprint specification and retention


smburdick:
- [x] Ensure that the bounding box is rectangular.
- [x] Ensure the constellation boundary doesn't lead to overlapping qubits
- [x] Specify bounding quad before the qubits.
- [x] Rename unit cell to footprint
- [x] Provide additional instructions via `Notification()`s
- [x] Don't erase the footprint after it has been made.
- [ ] ~~Move "create canonical circuit" button functionality out of the individual Plaquette class~~ Added to #228

smburdick:
Working on subtask 2 in https://github.com/smburdick/tqec/tree/219-valid-bounding-quad. There's a bit of math involved, so it will take me some time.

PR #220, opened 2024-05-04
Fix bounding box footprint as rectangle bug

smburdick:
* Address first item in #219 
* Fix linting issues in Plaquette class


vtomole (approved):
LGTM

PR #224, opened 2024-05-08
Fix typing issue in CircuitMeasurementMap

### LINKED ISSUE #223: Small typing issue in backend
## Describe the bug

The last line of

https://github.com/QCHackers/tqec/blob/d2a9d2fce649fe1b614dbcb90e0494716109f4b1/src/tqec/circuit/operations/measurement_map.py#L151-L155

is raising the typing issue

```
Index 0 is out of range for type Tuple[()]Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
```

## Expected behavior
No typing issue.

## Additional context
Currently, we check if the number of qubits is strictly superior to 1. If that is the case, an exception is raised. That means that the number of qubits can be either 0 or 1. The typing analyser is showing this error because, if it turns out that there are no qubits (i.e. 0), trying to get the first one will raise an error.

nelimee:
This PR fixes #223

### DIFF
--- a/src/tqec/circuit/operations/measurement_map.py
+++ b/src/tqec/circuit/operations/measurement_map.py
@@ -148,9 +148,9 @@ class CircuitMeasurementMap:
             global_measurement_indices.append(dict())
             for op in moment.operations:
                 if isinstance(op.gate, cirq.MeasurementGate):
-                    if len(op.qubits) > 1:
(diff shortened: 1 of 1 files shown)


vtomole (approved):
LGTM

PR #226, opened 2024-05-11
[Rev. Task 2-5] Plaquette specification

smburdick:
Implement issue #219. Redesign from the revised task flow, task 2.

* Retain the footprint after specifying it, along with the lattice of qubits.
* When I specify the qubit constellation against the footprint, ensure that if the footprint were to be copied over, there would be no overlapping qubits.
* Revised qubit population technique
* Cycle through X and Z, and specify a time stamp.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15712,10 +15712,6 @@
         "node": "^16.14.0 || >=18.0.0"
       }
     },
-    "node_modules/npm/node_modules/ip": {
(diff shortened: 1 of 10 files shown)


smburdick:
Converting this to draft since I have a couple questions about the overall design, and may need more debugging.

afowler:
Footprint first is a good idea.

vtomole (approved):
LGTM

Issue #199, opened 2024-03-22
Implement automatic detector computation


nelimee:
**Is your feature request related to a problem? Please describe.**
Currently, the `tqec` library requires users to explicitly implement the detectors. To ease that process, detectors are automatically included in `Plaquette` instances for the "default" situation, when that makes sense. But this way of adding detector is not optimal:
- the "default" detectors have been chosen arbitrarily, and are not a good fit for all possible codes,
- requiring users to explicitly annotate detectors should remain a possibility, but is a tedious and error-prone process,
- there are ways to find detectors automatically.

**Describe the solution you'd like**
We need to provide a way for users to automatically add detectors. This should be opt-in, as the automatic computation of detectors will likely, at least in the beginning, be driven by heuristics (i.e., not always correct/exhaustive), and because we want the library to provide the most control possible to the users that require it.

**Describe alternatives you've considered**
For the moment, I am not sure how this should be implemented. I have a theoretical reformulation that works generally, but that is not convenient to compute and may be made more efficient. We also have very localised information: propagated Paulis are, in the current setup in which we only consider surface codes, very much local and cannot spread globally (in the spatial dimension). There might be a way to use that feature by considering the QEC at the plaquette level, before generating the actual circuit.

Finally, we will likely require another abstraction: "layer". It seems reasonable, at the moment and (again) in the setup we are considering, that a QEC is composed of a sequence of layers, each following the scheme:
- `0` or `1` Moment (in the `cirq` sense) of reset,
- `k` Moments of computation with Clifford gates,
- `0` or `1` Moment of measurement.

**Additional context**
- [@inmzhang answer in a PR](https://github.com/QCHackers/tqec/pull/100#issuecomment-1952002691)
- [Inplace Access to the Surface Code Y Basis (Appendix C)](https://arxiv.org/abs/2302.07395)

inmzhang:
I'll start working on this when I have time. As a starting point, I will try to implement exactly what Austin proposed, which I think should be sufficient for the near-term goal.

nelimee:
@inmzhang any update on that issue?

inmzhang:
Sorry for taking such a long time. I have done most of the work except for the anticommute grouping and several edge cases. Hopefully I'll complete the remaining stuff and start iterating on the design in this week.

PR #227, opened 2024-05-11
WIP: Add detector automation but without anti-commute grouping

inmzhang:
This is a very basic attempt to adress #199. I tried to implement the [proposal by Austin](https://github.com/QCHackers/tqec/issues/199#issuecomment-2015637349) at the level of `stim` instead of `cirq` because I want this part of work can be applied to other stim-based projects more easily. The main pipeline can be splitted into the following steps:

1. Split the `stim.Circuit` into fragments. By design, a fragment is the sub-circuit of a single error correction round with related resets/measurements collapsing operations.
2. Use a strategy similar to that Craig Gideny used in the [midout paper source code](https://github.com/Strilanc/midout/blob/main/src/midout/gen/_flow_util.py) to compile each fragment. When compiling the fragment, the boundary stabilizers are constructed and matched to form the detectors. Note that I haven't implement anticommute grouping as proposed by Austin yet.

Currently, there are some problems in the codebase as far as I can notice and remember:

1. The structure of the round-based error correction circuit is not well-tested. The boundary of a fragment is identified by the collapsing moments, which tried to deal with the consequtive measurements/resets moments and the circuit repeat block, this part is error-prone.
2. At the early stage, I thought a boundary stabilizer with trivial stabilizer(all pauli operator is Identity) should alway form a detector by itself. However, consider a simplest circuit `R -- MR -- M`, the second detector should be `DETECTOR rec[-1][-2]` instead of `DETECTOR rec[-1]`. Though the two definitions seem equivalent when multiplying the first detector into it, there are critical differences when decoding the circuits. Thus I always try to match the boundary stabilizers between two consequtive fragments even when the weight of the stabilizer is zero.
3. Currently, the detector annotation for the simple r=1 repetition code circuit will fail. Consider the following circuit: 图片, if we allow resets on data qubits to generate the boundary stabilizers, it might generate detectors unwanted. Maybe we can forbid resets on some specific set of qubits to generate end stabilizers to handle it. However, when the circuit gets more complex, the forbiden set can be changed over the circuit(e.g. the wiggling/sliding circuit in http://arxiv.org/abs/2302.02192). I did not have a good solution to this yet.
4. Though the performance of the code and algorithm is not optimized, it already took around 14s to run the tests for 80 small circuits in the test cases. Note that the unimplemented anticommute grouping step might increase the time further significantly. We might need to start considering whether we should implement this in pure Python.
5. I have passed the tests for the surface code circuits generated by `stim`. In my plan, I would like to test it more on the various circuits in this [paper](http://arxiv.org/abs/2302.02192). Currently I have only included a single circuit in the directory `extra_test_circuits` for anticommute grouping test purpose(so it should fail for now).

Ideally, I would have completed the remaining tasks and submitted a finalized PR. However, I'm quite occupied these days and I'm not sure when I will be able to finish this. Therefore, I'm submitting it as a draft and would appreciate any help from the community to improve, refactor or even complete it.


src/tqec/circuit/detector/detector_util.py
```diff
@@ -0,0 +1,406 @@
...
+            qubits: A dictionary mapping qubit indices to Pauli operators. The
+                Pauli operators should be one of "I", "X", "Y", or "Z".
+        """
+        for qubit, pauli in qubits.items():
+            if pauli not in "IXYZ":
+                raise TQECException(
+                    f"Invalid Pauli operator {pauli} for qubit {qubit}, expected I, X, Y, or Z."
+                )
+        self.qubits = {q: qubits[q] for q in sorted(qubits.keys())}
+        self._hash = hash(tuple(self.qubits.items()))
```
nelimee, line 34: Maybe 
```suggestion
    def __init__(self, qubits2pauli: dict[int, str]):
        """A mapping from qubits to Pauli operators that represent a Pauli string.

        Args:
            qubits2pauli: A dictionary mapping qubit indices to Pauli operators. The
                Pauli operators should be one of "I", "X", "Y", or "Z".
        """
        for qubit, pauli in qubits2pauli.items():
            if pauli not in "IXYZ":
                raise TQECException(
                    f"Invalid Pauli operator {pauli} for qubit {qubit}, expected I, X, Y, or Z."
                )
        self.qubits = {q: qubits2pauli[q] for q in sorted(qubits2pauli.keys())}
        self._hash = hash(tuple(self.qubits.items()))
```
or another name that would be more expressive about the fact that the datastructure is not a simple list of qubits but rather a mapping from qubit indices to paulis?

src/tqec/circuit/detector/detector_util.py
```diff
@@ -0,0 +1,406 @@
...
+        ):
+            continue
+        if not stim.gate_data(inst.name).is_reset:
+            raise TQECException(
+                f"The reset moment contains non-reset instruction: {inst}."
+            )
+    return True
+
+
+def collapse_pauli_strings_at_moment(moment: stim.Circuit, is_reset: bool):
```
nelimee, line 265: Missing return type annotation

nelimee:
```suggestion
def collapse_pauli_strings_at_moment(moment: stim.Circuit, is_reset: bool) -> list[PauliString]:
```
right?

nelimee (commented):
I started to read the code, for the moment I need more time to familiarise myself with it and provide real feedback. 
Do not feel obligated to answer to my comments on this PR. I leave them mostly to keep track of what I think about some portion of the code, and if you do not have the time to finish that PR I will likely try to step in and finish it, asking for your opinion on the result (a review) before merging.

For the code I read: it seems very good! Well done!

Also, a more general comment, I completely agree with your approach of directly going with the `stim` representation. Ideally, in the long term, I would like the whole code base to be only dependent on `stim` too and remove the `cirq` dependency entirely if that is possible and not too intrusive.

Another side note: because this code, once finished, may be useful outside of the `tqec` code base, it might be worth to eventually extract it into its own package that the `tqec` package will depend on.

src/tqec/circuit/detector/detector_util.py
```diff
@@ -0,0 +1,406 @@
...
+        self.begin_stabilizer_sources = list(begin_stabilizer_sources)
+        self.sources_for_next_fragment = list(sources_for_next_fragment)
+
+    @property
+    def have_detector_sources(self) -> bool:
+        return not (
+            len(self.end_stabilizer_sources) == 0
+            and len(self.begin_stabilizer_sources) == 0
+            and len(self.sources_for_next_fragment) == 0
+        )
```
nelimee, line 301: From what I have gathered, `end_stabilizer_sources` stores the `PauliString` instances representing the `reset` collapsing operations and `start_stabilizer_sources` stores the same thing, but for `measurement` collapsing operations right? If that is the case, I really do not understand the `start`/`end` naming, to the point where I think they should even be inverted as, in my head, a `reset` is starting a new stabilizer and a measurement is ending one.

Could you explain the rationale behind that naming?

src/tqec/circuit/detector/detector_util.py
```diff
@@ -0,0 +1,406 @@
...
+
+        elif state.seen_measurement_at_tail:
+            fragments.append(state.to_fragment())
+            state.clear()
+            state.cur_fragment += moment
+        else:
+            state.cur_fragment += moment
+    if state.cur_fragment:
+        fragments.append(state.to_fragment())
+    return fragments
```
nelimee, line 399: This function is quite complex to understand. I think it would help to provide in comments a few invariants that are always checked through the for loop. A few examples of invariants that might be useful to help someone understanding that function:

- `state` stores and gather the necessary information to build a `Fragment` and is reset each time the final `Fragment` is constructed with `SplitState.to_fragment`.
- `state` represents a complete `Fragment` when any of the following conditions is met: `[list here]`

nelimee (commented):
A few questions I have been asking myself for a few days without coming up with a satisfactory answer on my side. I am still familiarizing myself with the code.

src/tqec/circuit/detector/extra_test_circuits/r=12,d=3,p=0.001,noise=SI1000,b=X,style=3-CZ,q=17.stim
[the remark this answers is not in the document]
inmzhang: I think it won't work on Windows. The file is copied directly from the ancillary files of the paper mentioned above and I'm currently working on Unix-type systems so I did not bother to change them. It should definitely be renamed in the future or for now.

src/tqec/circuit/detector/detector_util.py
inmzhang: > From what I have gathered, `end_stabilizer_sources` stores the `PauliString` instances representing the `reset` collapsing operations and `start_stabilizer_sources` stores the same thing, but for `measurement` collapsing operations right? 

Yes, it is.

> If that is the case, I really do not understand the start/end naming.

It is indeed a little confusing. I followed the convention/terminology from Austin's proposal. An `end_stabilizer` refers to the stabilizer at the end of a circuit fragment propagated from the start of the fragment. In particular, it is the stabilizer introduced by one/several reset(s) at the beginning of the fragment, passing through the circuit till the end. So the 'end' means the stabilizer is at the end of a fragment, which explains the `end_stabilizer_sources` are resets.

nelimee: Ok, I got it. I am fine with that naming then, but I think a comment to explain would be welcomed. I am working on adding comments and documentation right now, so you do not need to do it. 

I will do a draft PR on your branch shortly to show what I added for the moment.

src/tqec/circuit/detector/construction.py
```diff
@@ -0,0 +1,408 @@
...
+            index = i - len(boundary_collapse)
+            if not collapse.intersects(before_collapse):
+                continue
+            if before_collapse.commutes(collapse):
+                commute_collapse_indices.append(index)
+            else:
+                anticommute_collapse_indices.append(index)
+        if inverse:
+            if source.weight == 0:
+                source_measurement_indices = frozenset()
```
nelimee, line 268: Is that branch even possible? Looking for the usages of this function, the `sources` argument is given:

- `fragment.begin_stabilizer_sources`
- `state.resets_passed_in + fragment.end_stabilizer_sources`

It seems like, in both cases, the `PauliString` should be at least weight 1 as they originate from a reset/measurement. Do I miss something?

src/tqec/circuit/detector/detector_util.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+def pauli_string_mean_coords(
+    pauli_string: PauliString, qubit_coords_map: dict[int, list[float]]
+) -> tuple[float, ...]:
+    all_coords_items = [qubit_coords_map[i] for i in pauli_string.qubits.keys()]
+    return tuple(np.mean(np.asarray(all_coords_items), axis=0)) + (0.0,)
```
nelimee, line 406: What is the meaning of the last `+ (0.0,)`? Is it to add the time dimension?

src/tqec/circuit/detector/construction_test.py
```diff
@@ -0,0 +1,163 @@
...
+                after_clifford_depolarization=0.001,
+                before_round_data_depolarization=0.005,
+                after_reset_flip_probability=0.01,
+                before_measure_flip_probability=0.01,
+            )
+        )
+        if code_task.startswith("repetition_code"):
+            circuit = prepend_qubit_coords_for_repetition_code(circuit)
+        if decompose_MR:
+            circuit = replace_MR_with_M_then_R(circuit)
```
nelimee, line 140: Why do you test both cases (keeping `MR` and replacing them with `M` followed by `R`)?

I am trying to enforce explicitly some invariants on the dataclasses, and ended-up with that for the `Fragment` class:
```py
@dataclass(frozen=True)
class Fragment:
    """A sub-circuit guaranteed to span the locations between two nearest collapsing
    moments or those outside the error detection regions.

    Attributes:
        circuit: the circuit represented by the instance.
        end_stabilizer_sources: sources (reset instructions) that generate the
            stabilizers that might be found at the end of the circuit.
        begin_stabilizer_sources: sources (measurement instructions) that generate
            the stabilizers that might be found at the beginning of the circuit.
        sources_for_next_fragment: sources (reset instructions) that generate
            stabilizers after the measurements of this Fragment instance and so
            should be accounted for in the next Fragment.
    """

    circuit: stim.Circuit
    end_stabilizer_sources: list[PauliString] = field(default_factory=list)
    begin_stabilizer_sources: list[PauliString] = field(default_factory=list)
    sources_for_next_fragment: list[PauliString] = field(default_factory=list)

    def __post_init__(self):
        moments = list(
            circ.copy() if isinstance(circ, stim.Circuit) else circ
            for circ in iter_stim_circuit_by_moments(self.circuit)
        )
        print(f"Moments: {moments}")
        if not moments:
            raise TQECException("Empty circuit found in Fragment.")
        if any(isinstance(m, stim.CircuitRepeatBlock) for m in moments):
            raise TQECException(
                "Found a stim.CircuitRepeatBlock in a Fragment instance. "
                "Please use FragmentLoop."
            )
        # Check that the first moment of the fragment contains reset operations.
        if not has_reset(moments[0]):  # type: ignore we already checked that there is no instance of stim.CircuitRepeatBlock
            print(moments)
            raise TQECException("First moment of Fragment not containing any reset.")
        # Check that the last moment of the fragment contains measurement operations.
        if not has_measurement(moments[-1]):  # type: ignore we already checked that there is no instance of stim.CircuitRepeatBlock
            print(moments)
            raise TQECException(
                "Last moment of Fragment not containing any measurement."
            )
```

Basically, for any `Fragment`, it seems to me that we expect the first moment of the stored circuit to contain reset operations, and the last moment to contain measurement operations. But replacing `MR` with `M` + `R` breaks that assumption.

Is my constraint incorrect, in which case what is the invariant for `Fragment`, or is the constraint correct, in which case the test case seems incorrect?

nelimee: I finally understood. There is no real invariant on the `circuit` attribute of `Fragment` (except that it is not supposed to contain an instance of `stim.CircuitRepeatBlock`). Everything is encoded in the other attributes, and so even if the circuit does not start with resets, or does not strictly end with measurements, this is still fine.

src/tqec/circuit/detector/detector_util.py
inmzhang: Yes it is the time dimension.

src/tqec/circuit/detector/construction.py
inmzhang: At the first matching interface(the beginning boundary of the first fragment), the stabilizers from the measurements at the end of the fragment will match to the "I" stabilizers intentionally made for each qubit as the end stabilizers from the virtual previous fragment:
```python
def compile_fragments_to_circuit(
    fragments: Iterable[Fragment | FragmentLoop],
    qubit_coords_map: dict[int, list[float]],
) -> stim.Circuit:
    all_qubits = sorted(qubit_coords_map.keys())
    state = State(
        resets_passed_in=list(),
        previous_measurements=list(),
        end_stabilizers=[
            BoundaryStabilizer(
                before_collapse=PauliString({q: "I"}),
                after_collapse=PauliString({q: "I"}),
                coords=tuple(qubit_coords_map[q]),
            )
            for q in all_qubits
        ], # Here are the "I" stabilizers
    )
    out_circuit = stim.Circuit()
    _compile_fragments_into_circuit(
        fragments, state, out_circuit, qubit_coords_map, True
    )
    return out_circuit
```

The same situation is for the end of the circuit(the ending boundary of the last fragment):

```python
    if compile_final_state_once_more:
        _compile_fragment_into_circuit(
            Fragment(
                circuit=stim.Circuit(),
                begin_stabilizer_sources=[
                    PauliString({q: "I"}) for q in sorted(qubit_coords_map.keys())
                ],
            ),
            state,
            out_circuit,
            qubit_coords_map,
        )
```

nelimee:
So I have been exploring this code for the past 2 days. I also made [a tentative at implementing a generic method to compute stabilizers a few months ago](https://github.com/QCHackers/tqec/compare/main...nelimee:tqec:introduce-layer).

From that, here are my main conclusions.

First, the problem of finding stabilizers can be split into 3 sub-problems:

1. Splitting correctly the circuit according to the collapsing operations (resets and measurements) contained in this circuit.
2. Pre-computing various quantities on the split sub-circuits (stabilizer propagation for example).
3. Using the data pre-computed in step 2. to match stabilizer flows and find detectors.

Now, in more details.

### 1. Splitting the input circuit

The collapsing operations (measurement and resets) are the main point of interest here. What I did in my attempt, and what I think you also did in this PR, is to constraint the input circuit to avoid a (probably large) number of edge cases.
The constraint I have in mind is: collapsing operations (again, measurements and resets) cannot appear everywhere in the circuit and can only appear in their own moment (sub-circuit between two `TICK` instructions).

This constraint makes sense as Austin already told us, this is in fact something that most hardware will want to ensure.

With that constraint in mind, there are still a few edge cases that need to be handled:

- `REPEAT` blocks need to be treated differently,
- combined measurement/reset operations (such as `MR` that performs a measurement in the `Z` basis directly followed by a reset in the `Z` basis) makes it difficult to split the input circuit cleanly into sub-circuits.

In the end, I dislike how I split the circuit in my previous attempt because I was using `cirq` (which would be best avoided here, I agree) and I was storing a lot of `stim.Tableau`.

I do not fully adhere to your method either, I feel like the code is too complex for what it is doing. In particular, as you noted, the `split_stim_circuit_into_fragments` is quite hard to reason about.

I think my preferred way of solving that sub-problem would be to re-use your code, but ensuring that `Fragment` instances start with a moment of resets and end with a moment of measurements, splitting the combined operations into two separate operations if needed. This simple change in the input circuit would remove the need to store begin/end stabilizer sources and to keep track of the sources for the next `Fragment`. Basically, only performing `stim.Circuit` manipulation and not computing anything (yet).

### 2. Pre-computation

We need to pre-compute a lot of information in order to hope to perform step 3 efficiently. In particular, for each collapsing operation, we need to compute how the stabilizer it propagates evolves in the circuit.
Here, there is a difference between measurements and resets:

- measurements propagate stabilizers backwards where resets propagate stabilizers forward,
- the final result, the detectors, are composed of measurements, not of resets,

These differences mean that we will need to store enough information to recover the exact measurement associated with a given stabilizer propagation, but we do not really care about the reset associated to a propagation.

In our case, I fully agree with the step-by-step description by Austin. So our `Fragment` instances have beginning and end stabilizers, the beginning stabilizers being annotated with the measurement that started the (back-)propagation and the end stabilizers storing the measurement(s) that commuted and anti-commuted with the pauli string that propagated from the reset.

The goal of that step is simply to produce as efficiently as possible a data-structure that we will be able to query afterward to match stabilizers together.

If we only restrict ourselves to detectors only including measurements from the current and previous rounds, it is sufficient in this step to pre-compute the stabilizers for each `Fragment` separately: evolve forward each reset-generated stabilizer, evolve backward each measurement-generated stabilizer, and store the resulting stabilizer with the commuting/anti-commuting measurements in a data-structure that will be accessed at the next step.

### 3. Matching stabilizer flows

The approach described by Austin and implemented in your different `match_*` methods works for me. I only have a doubt on a specific part of the code in `find_cover`:

```py
def find_cover(
    target: BoundaryStabilizer,
    sources: list[BoundaryStabilizer],
) -> list[BoundaryStabilizer] | None:
    subsets = [
        source
        for source in sources
        if target.after_collapse.contains(source.after_collapse)
    ]
```
I am not sure the restriction to sources that are contained by `target` is correct, as some sources might cancel out the Pauli terms that are not contained in `target` and end up forming a stabilizer. 

## Overall opinion

All in all, I really think that we should look at these 3 sub-problems separately for several reasons:

- it helps tackling one problem at a time,
- it will likely help organising the code in well define, self-contained, portions,
- it will allow us to think about the interfaces between the code solving each sub-problem.

If you agree with what I wrote above, my plan is to reformat the code in this PR following these 3 sub-problems.

inmzhang:
The sub-problems are reasonably well to me and it's exactly the mental model I followed when I wrote the code.

> I am not sure the restriction to sources that are contained by target is correct, as some sources might cancel out the Pauli terms that are not contained in target and end up forming a stabilizer.

I guess you are right and I really cannot remember why I made this assumption, maybe because it just work well for the usual surface code circuit.

I have added some comment on the `split_stim_circuit_into_fragments` method, but it's still complex and kind of messy. A well-structured `Fragment` as you proposed should work better than the current one.

nelimee:
I have made significant (in size) but quite trivial (in actual code change) changes on your PR in #248. I am currently starting fresh to explore the idea exposed above, with the plan to copy-paste large portions from here or #248, and adapt a few things.
I will likely open another PR to show these changes, and discuss.

afowler:
I would vote for banning the measure-reset (MR) gate and insisting on reset always being a separate gate that always follows measurement before any other gates are allowed on that qubit. I'd also vote for banning repeat blocks.

nelimee:
> I would vote for banning the measure-reset (MR) gate and insisting on reset always being a separate gate that always follows measurement before any other gates are allowed on that qubit. I'd also vote for banning repeat blocks.

The solution proposed in #249 bans all combined gates (`MR`, `MRZ`, `MRX`, `MRY`) and replaces them with their equivalent (minor note on that in #249) measurement + reset gates.

Repeat blocks are not that hard to handle, and should be already handled correctly by this PR (and #249 that started from here and made significant changes to the code).

inmzhang:
Closes as there is the more robust and correct implementation in #249 .

PR #235, opened 2024-05-16
Improve ASCII visualization of a composed template

giangiac:
Method added in `<repo>/src/tqec/templates/display.py`.

It is tested in notebook `<repo>/notebooks/doubling_distance.ipynb`.

Example of output:
```
     ■-------------------■                                  
  .  | .    1    .    1  | .    .    .    .    .    .    .  
■----■-------------------■----■                             
| 1  | 1    1    1    1  | .  | .    .    .    .    .    .  
|    |                   |    |                             
| .  | 1    1    1    1  | 1  | .    .    .    .    .    .  
|    |                   |    |                             
| 1  | 1    1    1    1  | .  | .    .    .    .    .    .  
|    |                   |    |                             
| .  | 1    1    1    1  | 1  | .    .    .    .    .    .  
■----■-------------------■----■                             
| 1  | 1    1    1    1  | .    .    .    .    .    .    .  
■----■                   ■----■    ■-------------------■    
  .  | 1    1    1    1  | 1  | .  | 1    .    1    .  | .  
■----■-------------------■----■----■-------------------■----■
| 1  | 1    1    1    1  | 1    1  | 1    1    1    1  | .  |
|    |                   |         |                   |    |
| .  | 1    1    1    1  | 1    1  | 1    1    1    1  | 1  |
|    |                   |         |                   |    |
| 1  | 1    1    1    1  | 1    1  | 1    1    1    1  | .  |
|    |                   |         |                   |    |
| .  | 1    1    1    1  | 1    1  | 1    1    1    1  | 1  |
■----■-------------------■---------■-------------------■----■
| 1  | 1    1    1    1  | 1    1  | 1    1    1    1  | .  
■----■                   |         |                   ■----■
  .  | 1    1    1    1  | 1    1  | 1    1    1    1  | 1  |
■----■-------------------■---------■-------------------■----■
| 1  | 1    1    1    1  | 1    1  | 1    1    1    1  | .  |
|    |                   |         |                   |    |
| .  | 1    1    1    1  | 1    1  | 1    1    1    1  | 1  |
|    |                   |         |                   |    |
| 1  | 1    1    1    1  | 1    1  | 1    1    1    1  | .  |
|    |                   |         |                   |    |
| .  | 1    1    1    1  | 1    1  | 1    1    1    1  | 1  |
■----■-------------------■----■----■-------------------■----■
  .  | 1    .    1    .  | 1  | .  | 1    .    1    .  | .  
     ■-------------------■----■    ■-------------------■    
```


nelimee:
I will review that in the afternoon, thanks for the contribution, seems nice!

src/tqec/templates/display.py
```diff
@@ -29,6 +30,74 @@ def display_template(
         print()
 
 
+def display_templates_ascii(
+    templates: ComposedTemplate,
```
nelimee, line 34: I would name that parameter `template` (singular), but this is a very small comment.

src/tqec/templates/display.py
```diff
@@ -29,6 +30,74 @@ def display_template(
...
+        # the represented code to the axes and avoid having negative indices.
+        x = tul.x - bbul.x
+        y = tul.y - bbul.y
+        # Recall that numpy indexing is (y, x) in our coordinate system convention.
+        tshapey, tshapex = templates._templates[tid].shape.to_numpy_shape()
+        ul_pos[tid] = (y, x)
+        br_pos[tid] = (y + tshapey, x + tshapex)
+    # Format of the ASCII art.   
+    h_space = 5
+    empty_line = ' '*numpy.shape(arr)[1]*h_space
```
nelimee, line 65: I would create a named variable for `numpy.shape(arr)[1]`, just to make that more explicit. Also, the convention used in the code base is to enclose strings in `"` instead of `'`:
```suggestion
    x_size = numpy.shape(arr)[1]
    empty_line = " "*x_size*h_space
```

src/tqec/templates/display.py
```diff
@@ -29,6 +30,74 @@ def display_template(
...
+    cross_mark = '■'
+    # Instead of printing to screen, we create a multi-line buffer string.
+    # When completed, we print the buffer.
+    buffer_lines = []
+    buffer_lines.append(empty_line)
+    for line in arr:
+        buffer = ''
+        for element in line:
+            element = str(element) if element != 0 else "."
+            buffer += ' ' + f"{element:^{h_space-1}}"
```
nelimee, line 75:
```suggestion
            buffer += f"  {element:^{h_space-1}}"
```

src/tqec/templates/display.py
```diff
@@ -29,6 +30,74 @@ def display_template(
...
+        col = tul[1] * h_space
+        num_h_sep = br_pos[tid][1]*h_space - col
+        # Add horizontal lines at the top and bottom.
+        for row in [tul[0], br_pos[tid][0]]:
+        #for row in [tul[0]]:
+        #for row in [br_pos[tid][0]]:
+            row *= 2
+            line = buffer_lines[row]
+            line = line[:col] + cross_mark + '-'*(num_h_sep-1) + cross_mark + line[col+num_h_sep+1:]
+            buffer_lines[row] = line
```
nelimee, line 89:
```suggestion
            buffer_lines[row][col:col+num_h_sep+1] = cross_mark + '-'*(num_h_sep-1) + cross_mark
```
should be functionally equivalent but will (I think) save a few string concatenation / creation / destruction operations that might matter for large templates. Also, personal opinion, I prefer the above one line version.

src/tqec/templates/display.py
```diff
@@ -29,6 +30,74 @@ def display_template(
...
+        # Add vertical lines on the left and right.
+        for col in [tul[1], br_pos[tid][1]]:
+            col *= h_space
+            for row in range(tul[0]*2+1, br_pos[tid][0]*2):
+                line = buffer_lines[row]
+                line = line[:col] + '|' + line[col+1:]
+                buffer_lines[row] = line
```
nelimee, line 96:
```suggestion
                buffer_lines[row][col] = "|"
```
same as above, but quite obviously better in that case.

nelimee (commented):
A few general comments.

The current code has not been formatted by an automatic tool. I can do that for you, or you can do it, your call. You can use `black` or `ruff`. (I want to add that to CI in the short term).

A few parameters / hard-coded values could be exposed as parameters. 
This is for example the case of:
- `h_space` that can be a parameter with a default value of `5`,
- `cross_mark` to let users replace that with a non-unicode character if they know that their terminal / text-display-thing does not support unicode.
- The number `2` that seems to be `v_space` (vertical space) and that appear 3 times in the code (I think).

Also, it might be worth implementing the "bounding-box coordinate shifting" algorithm to avoid negative coordinates directly as a (private) method in the `ComposedTemplate` class, and re-use that method both in `ComposedTemplate` and in your implementation here. That would also reduce the length and complexity of the `display_template_ascii` function.

Seeing the operations you are performing (replacing rows/columns), it might be worth using a `numpy.ndarray` with `dtype` of [`numpy.str_`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.str_). This would likely improve readability, performance and memory consumption.

src/tqec/templates/display.py
giangiac: I took the name from `display_templates_svg()`.
I can change both, or just the name in `display_templates_ascii()`.

src/tqec/templates/display.py
giangiac: Thanks!

giangiac:
Revert some of the suggested changes:
`str` object is immutable and do not support item assignment

giangiac:
> A few general comments.
> 
> The current code has not been formatted by an automatic tool. I can do that for you, or you can do it, your call. You can use `black` or `ruff`. (I want to add that to CI in the short term).
> 
> A few parameters / hard-coded values could be exposed as parameters. This is for example the case of:
> 
> * `h_space` that can be a parameter with a default value of `5`,
> * `cross_mark` to let users replace that with a non-unicode character if they know that their terminal / text-display-thing does not support unicode.
> * The number `2` that seems to be `v_space` (vertical space) and that appear 3 times in the code (I think).
> 
> Also, it might be worth implementing the "bounding-box coordinate shifting" algorithm to avoid negative coordinates directly as a (private) method in the `ComposedTemplate` class, and re-use that method both in `ComposedTemplate` and in your implementation here. That would also reduce the length and complexity of the `display_template_ascii` function.
> 
> Seeing the operations you are performing (replacing rows/columns), it might be worth using a `numpy.ndarray` with `dtype` of [`numpy.str_`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.str_). This would likely improve readability, performance and memory consumption.

Thanks for the quick review!

- changed the function to take some of the hardcoded values as parameters
- replaced the str delimiters
- introduced v_space, corresponding to "1 + num additional lines between the plaquette rows"

It would be great if you could apply the automatic formatting tool. Thanks!

src/tqec/templates/display.py
nelimee: This is very minor, let's leave it like that for the moment as it does not impact external API.

src/tqec/templates/display.py
```diff
@@ -29,6 +30,88 @@ def display_template(
...
+    """
+    if plaquette_indices is None:
+        plaquette_indices = tuple(range(1, templates.expected_plaquettes_number + 1))
+    if h_space < 4:
+        print(f"WARNING: horizontal space value {h_space} is not enough. Changed to default value 5.")
+        h_space = 5
+    if v_space < 2:
+        print(f"WARNING: vertical space value {v_space} is not enough. Changed to default value 2.")
+        v_space = 2
+    if len(corner_mark) != 1:
```
nelimee, line 57: This *should* work most of the time because Python `str::len` seems to behave as you expect on unicode characters, but I can pretty much guarantee that it will not work for all inputs (see for example [this thread](https://discuss.python.org/t/peculiar-behaviour-involving-unicode-len-s-0-len-s-0-lower-is-this-expected/24469/5)).

For the moment, and because this will likely never be an issue, I am fine not changing that portion of the code. I just wanted to note that length of unicode characters might not work as we expect in all cases.

nelimee (commented):
I think that using a `numpy.ndarray` of `dtype = numpy.str_` would still be worth. I can try to hack a version with that if you do not want to test yourself, to see what it ends up looking like.
Nonetheless, the current code seems to work as expected. 

Before approving I will:
1. format the code,
2. try out locally with different input values and check that I did not miss any edge-case.

PS: Sorry for the non-working suggestions, I forgot about strings being immutable.

src/tqec/templates/display.py
```diff
@@ -29,6 +30,88 @@ def display_template(
         print()
 
 
+def display_templates_ascii(
+    templates: ComposedTemplate,
+    plaquette_indices: ty.Sequence[int] | None = None,
+    h_space: int = 5,
+    v_space: int = 2,
+    corner_mark: str = "■",
```
nelimee, line 38:
```suggestion
    corner_mark: str = "+",
```

Having a non-unicode default separator seems like the best option as:
- it is in the direction of maximum compatibility,
- it still allow anyone to use unicode,
- it might avoid users thinking that unicode characters are perfectly handled. There might be very_few edge cases with the` len(corner_mark) == 1` check for strange unicode characters. These edge cases will likely never be triggered, but let's be on the safe side.

src/tqec/templates/display.py
```diff
@@ -29,6 +30,88 @@ def display_template(
...
+    buffer_lines = []
+    buffer_lines.append(empty_line)
+    for line in arr:
+        buffer = ""
+        for element in line:
+            element = str(element) if element != 0 else "."
+            buffer += f" {element:^{h_space-1}}"
+        buffer_lines.append(buffer)
+        for _ in range(v_space-1):
+            buffer_lines.append(empty_line)
```
nelimee, line 91: It would be a little bit more readable to have vertically-centered plaquette numbers (they are already correctly horizontally centered).

I *think* the following should work:
```suggestion
    buffer_lines = []
    buffer_lines.append(empty_line)
    empty_lines_number_before = (v_space - 1) // 2
    empty_lines_number_after = v_space - 1 - empty_lines_number_before
    for line in arr:
        for _ in range(empty_lines_number_before):
            buffer_lines.append(empty_line)
        buffer = ""
        for element in line:
            element = str(element) if element != 0 else "."
            buffer += f" {element:^{h_space-1}}"
        buffer_lines.append(buffer)
        for _ in range(empty_lines_number_after):
            buffer_lines.append(empty_line)
```
from my local tests, that works but the labels are one line too low on odd `v_space` values (personal taste).

src/tqec/templates/display.py
```diff
@@ -29,6 +30,88 @@ def display_template(
...
+        plaquette_indices = tuple(range(1, templates.expected_plaquettes_number + 1))
+    if h_space < 4:
+        print(f"WARNING: horizontal space value {h_space} is not enough. Changed to default value 5.")
+        h_space = 5
+    if v_space < 2:
+        print(f"WARNING: vertical space value {v_space} is not enough. Changed to default value 2.")
+        v_space = 2
+    if len(corner_mark) != 1:
+        print(f"WARNING: corner mark '{corner_mark}' is unacceptable. Changed to default '■'")
+        corner_mark = "■"
```
nelimee, line 59:
```suggestion
        corner_mark = "+"
```
if you agree with the change above.

nelimee (changes requested):
A few more notes and ask for changes.

Also see https://github.com/giangiac/tqec/pull/2 for formatting.

nelimee:
> I think that using a `numpy.ndarray` of `dtype = numpy.str_` would still be worth. I can try to hack a version with that if you do not want to test yourself, to see what it ends up looking like.

I tried briefly, and it seems to shift code complexity from string manipulation to index manipulation. I personally prefer index manipulation, but string manipulation is **way** more readable, so please ignore that part of my comment. Trying `numpy`-based implementation might not be worth.

src/tqec/templates/display.py
giangiac: Thanks for pointing out these corner cases.
Below you are suggesting '+' instead of '■', I will commit your suggestions.
The choice of '■' was to provide a very visible corner mark.

giangiac:
> > I think that using a `numpy.ndarray` of `dtype = numpy.str_` would still be worth. I can try to hack a version with that if you do not want to test yourself, to see what it ends up looking like.
> 
> I tried briefly, and it seems to shift code complexity from string manipulation to index manipulation. I personally prefer index manipulation, but string manipulation is **way** more readable, so please ignore that part of my comment. Trying `numpy`-based implementation might not be worth.

I'd also value readability more than efficiency. We are printing to screen, a functionality that is not meaningful for ASCII with more than a few 100s lines.

giangiac:
> A few more notes and ask for changes.
> 
> Also see [giangiac#2](https://github.com/giangiac/tqec/pull/2) for formatting.

Thanks!
Merged.

nelimee (approved):
This seems ready to merge, thanks a lot @giangiac for that!

PR #238, opened 2024-05-24
[Do not Merge] This is an example for PYTHON <-> JS interaction

Gistbatch:
@giangiac 

This is a very brief example of how Python and JS can interact.
Let me know if you have any questions

### DIFF
--- a/frontend/src/components/download/SimpleButton.js
+++ b/frontend/src/components/download/SimpleButton.js
@@ -1,0 +1,68 @@
+import axios from 'axios';
+import Button from '../Button';
+import config from './config';
+
+function postExample(url, data) {
+  console.log(url);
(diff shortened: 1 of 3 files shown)


giangiac:
@Gistbatch Thanks for the PR.

I moved the alternative frontend back to the codebase (see [#PR250](https://github.com/QCHackers/tqec/pull/250))
Then I tried to adapt your example to that context (see [#PR3](https://github.com/giangiac/tqec/pull/3) in my fork).
In this way you can see the difference w.r.t. the current alternative backend.

Not sure what is supposed to happen when "get" and "post" are clicked.
could you give a look if things are set up properly?

Gistbatch:
no longer needed, see #325

Issue #170, opened 2024-02-29
Convert TQEC json format to OpenQASM/Stim


smburdick:
**Is your feature request related to a problem? Please describe.**
User specifies plaquettes, qubits, circuits, etc. in a way conducive to TQEC. This work should be saved, and can be easily represented in JSON. But we'd also like to export to OpenQASM or Stim. That file may not be able to reconstruct the user's work, so we should still save the TQEC JSON file

**Describe the solution you'd like**
Provide to the user the ability to export their circuit to openqasm/stim

**Additional context**
[JSON specifications: ](https://github.com/QCHackers/tqec/tree/main/specification/schemas)

nelimee:
Note that I do not think we should / can reasonably represent **all** the information in OpenQASM/Stim format.

In my opinion, we should just use the formats for what they were made for:
- OpenQASM/Stim to represent a quantum computation (i.e., quantum circuit),
- JSON to represent quite generic data.

The way I see it (that is also reflected in the tentative JSON specification) is that only the quantum circuit (not the qubits not the detectors) will be represented as OpenQASM/Stim, the other information being encoded in JSON.

nelimee:
Then I misunderstood your message, my bad, I am sorry 👍

Issue #169, opened 2024-02-29
Enable import/export of local circuits


smburdick:
**Is your feature request related to a problem? Please describe.**
I can't save my work and load it later.

**Describe the solution you'd like**
Options to save JSON circuit to local files, and load into the browser (including the unit cell)

**Describe alternatives you've considered**
Saving things on the cloud, which wouldn't be good for security-minded users.

Related -- #160

smburdick:
Related: #170

nelimee:
This issue is not relevant anymore. Closing it. See #323 for more details.

Issue #61, opened 2024-01-23
Implement `from_json` in backend


nelimee:
The backend (i.e., the Python library `tqec`) and the frontend (i.e., the Javascript library `tqecjs`) need to interact in some way. The current scheme, retained for its simplicity, is that the `tqec` backend should be able to answer as a REST server to some queries and the `tqecjs` frontend will perform these queries on behalf of the user using the frontend.

The unique query, for the moment, that is needed is the one that will send from the frontend to the backend all the information needed to describe the QEC setting wanted by the user, to which the server should answer with the generated `Stim` file(s).

#43 is discussing the JSON format that both the backend and the frontend need to agree on in order to allow a good communication between the two. Note that this is still a work in progress and the specification discussed in #43 can still change.

On the backend side, we need to be able to create all the necessary Python structures from the JSON sent from the frontend.

This issue can be solved by:
- implementing a `from_json` method for each class that might be communicated and a generic `from_json` function that will handle the instantiation of the correct class. 
- using the [`pydantic`](https://docs.pydantic.dev/latest/) library to handle all the serialisation-related work.

The solutions above are up to discussion with anyone wanting to help on that.

nelimee:
For the moment, the `pydantic` approach is being explored.

Issue #18, opened 2024-01-05
Save tile set as Stim file


smburdick:
https://github.com/quantumlib/Stim/blob/main/doc/file_format_stim_circuit.md

smburdick:
Do we want to be able to save the circuit that the user created on the frontend to a file? If so, we can add a button on the frontend, which could invoke a function on the backend (passing the current circuit as a parameter), and return a stim file to the browser.

nelimee:
Yes I agree, but this scheme does not require the UI part to know about the structure of a Stim file, only to know that it is a textual file. Am I missing something?

smburdick:
Right, the UI can be totally agnostic of the file structure, it would just have to invoke the Python backend to get the file back

smburdick:
I'm pretty sure this is Task 11.

nelimee:
I am still having hard times understanding what this issue is about. The title is about "tile set", but it seems to me that you are rather referring to the final Stim files.

Also, having a Stim file does not equate task 11. For task 11, we still have to simulate the different Stim files and perform the graph. This is made easy by the `sinter` package, but it is more than simply "getting the Stim files".

smburdick:
My terminology may be off here. By "tile set", I mean the set of plaquettes the user has made in the frontend. My impression is that these plaquettes and qubits (as well as templates, which we don't have in the UI yet) will need to be communicated to the backend, which will perform some work on the input to generate a circuit or other object that is meaningful to the user, and return that to them.

My most recent pull requests #40 and #41 establish this communication in an early stage -- crucially, the real processing on the backend is missing, (the input is returned as the output by the backend, as a placeholder) as we appear to need templates as well to construct a circuit.

nelimee:
As a note, I tried to provide a first version of what the backend would need as input to do its job. This can be found [here](https://github.com/QCHackers/tqec/tree/main/specification/schemas) (and I will modify it in the next few days).

smburdick:
@nelimee Awesome! Thank you for making those specifications. Once #43 is finalized, I will update the serialization in #41

smburdick:
Updated #41 to incorporate the current specs

smburdick:
This is being addressed in #192

nelimee:
This issue is not relevant anymore. Closing it. See #323 for more details.

PR #239, opened 2024-05-24
Add a REST server with FastAPI

### LINKED ISSUE #38: Finalize TQEC REST API
Currently, both the frontend (javascript UI / code) and the backend (Python code) are separate pieces of code that do not interact with each other. 

The currently envisioned process to create a QEC experiment is the following:
1. The user starts the frontend.
2. The user builds the experiment he/she wants to perform, still 100% in the frontend.
3. Once the experiment is fully caracterised in the frontend, the user asks for the `Stim` file generation (or multiple `Stim` files).
4. [internally] the frontend represents the QEC experiment in a JSON format.
5. [internally] the frontend sends the generated JSON to the backend (Python code).
6. [internally] the backend generates a `Stim` file from the input JSON.
7. [internally] the backend sends back the generated `Stim` file to the frontend.
8. The frontend provides a way to download / export the generated `Stim` file.

The above process is still in discussion. It has shortcomings, but also has the good property of being really simple to implement on both sides (frontend and backend).

The current communication scheme using JSON, both the frontend and backend need to agree on the JSON format.
A tentative format can be found in the [`specification/schemas`](https://github.com/QCHackers/tqec/tree/main/specification/schemas) folder. We would like to improve this format and actually implement the serialisation / deserialisation procedures in both the frontend and the backend.

nelimee:
This is a draft PR trying to implement a REST server on the backend side.

This is the initial implementation that will allow a link between the frontend and the backend. 

Currently, the idea is the following: all the QEC-related state is stored in the backend and the frontend is making requests through the provided REST API to 1. query the current QEC state or 2. change it. 
Here, "QEC-related state" means that:
- the frontend will **not** store the plaquettes, it will rather query the backend to have a description of the currently implemented plaquettes when needed,
- the frontend will **not** store the templates, it will rather query the backend to have a description of the currently implemented plaquettes when needed, 

Because we want to avoid a communication at each user inputs, the frontend may store partial state.

Note that everything I wrote above and the code from this PR is a proposal, I am very much open to any other idea if it turns out to be easier to use.

To use the code from that PR and experiment with the REST server, follow the instructions:
1. checkout the branch from that PR (https://github.com/nelimee/tqec/tree/feat/fast-api-rest),
2. re-install the `tqec` package to get the new dependencies
    ```sh
    python -m pip install -e ./
    ```
    or install the added dependencies directly
    ```sh
    python -m pip install pydantic fastapi
    ```
3. Start the REST server with
    ```sh
    fastapi dev ./web/server.py
    ```
4. If everything succeeded, you can access the server on http://127.0.0.1:8000/ (try out http://127.0.0.1:8000/templates/list or http://127.0.0.1:8000/plaquettes/list). The REST api documentation is available at http://127.0.0.1:8000/docs.

### Related issues:

- #170: if the decision to pursue in the direction explained above is taken, the backend will be responsible for that, and already implement the feature.
- #169: the backend would become responsible for the generation of the file, the frontend will still need to provide a way to save the file to the user.
- #61: nicely handled by using `pydantic` and implementing methods to convert from/to the defined `pydantic.BaseModel`.
- #38: the main related issue, this PR try to fix #38.
- #18: this PR introduces the possibility to export to Stim files, and so should fix that issue.


Gistbatch:
Just as a reminder:
We already had a Flask server in place, which was initially set up by @smburdick, and the front end has some unused functionality.
The file server root is in `/web/main.py`, also deployed to gcloud.
I don't have any strong preference between `flask` and `fastapi`, and most of the concepts are very similar.

- From a dev perspective, we might eventually think about hosting a proper database, which is straightforward with Pydantic.
- If we plan to host the backend and the frontend on different sites, we also need to enable `CORS`
- We should start to think about what information needs to be transferred at which points in time
- Also we should set up a proper structure for the webserver

nelimee:
I have no personal preference on that. As I already said multiple times, I am not competent for web-related development task (and I try to stay as far away as possible from anything related to web-technologies). 

My choice of `pydantic` + `fastapi` was based on the easy of use of `pydantic` and the ultra-fast integration of `fastapi` once you have `pydantic` models in place. There is really no smart thinking that happened on my side, an I will adopt whatever is judged to be the best solution by others. 

Let's consider this PR as a playground to start exploring the integration between the frontend and backend, and see what is required from both sides. At any time (if possible, early), if the choice of `pydantic` + `fastapi` seems worse than another library, please tell me, or change the framework directly.

nelimee:
I will resolve all the conflicts and push to a repository branch in the afternoon.

nelimee:
Replaced by #240

Issue #242, opened 2024-05-28
Validate `ComposedTemplate`


nelimee:
**Is your feature request related to a problem? Please describe.**
Some instances of `ComposedTemplate` are invalid when scaled up. For example,
![image](https://github.com/QCHackers/tqec/assets/12374487/4d2d4903-90d8-480b-b4b8-12c8b54f0b08)
will eventually be invalid as, when `k` grows, the two scalable templates at the bottom will eventually overlap.

It would be nice to be able to check that kind of problems directly within the `ComposedTemplate` class and return a nice error message when an error is found.

**Describe the solution you'd like**
Following #241, and assuming that template bounding boxes cannot overlap (which might not be valid if we have a `ComposedTemplate` instance containing non-atomic templates, such as other `ComposedTemplate` instances), we might be able to compute if two template might eventually overlap from their scalable bounding box. 
If such a case happens, an error should be raised.

PR #243, opened 2024-05-28
Use scalable variants when possible

### LINKED ISSUE #241: Use as much `Scalable` structures as possible
Currently, `Template` instances have methods such as `shape` (and the core private methods of `ComposedTemplate`) that return `Shape2D` instances. 

In practice, it would be possible and desirable to return a scalable instance (something like `ScalableShape2D`) to let the user have as much information as possible on the template instance. 
Having scalable instances would be a first step to check that a given `ComposedTemplate` instance is valid (in the sense that its templates **never** overlap).

1. Implement scalable versions of `Shape2D` and `Position`. `Displacement` is less an issue here as it used for `Plaquette` instances and is not really expected to be scalable in the near future.
2. Return instances of these scalable classes from calls to `Template.shape` and internal methods of `ComposedTemplate`.

**Additional context**
The final goal being to check the validity of a given `ComposedTemplate`, we will need something more expressive than `LinearFunction`. In practice, it seems like a `PiecewiseLinearFunction` would be sufficient to implement the above-mentioned methods. 
Note that user input will still only be `LinearFunction` instances. `PiecewiseLinearFunction` instances will likely only have to appear internally.

nelimee:
*This is a work in progress*

This PR introduces several scalable variants to `Shape2D` or `Position` that may ultimately be used to check `ComposedTemplate` validity (see #242).

Fixes #241


nelimee:
Replaced by #254

PR #245, opened 2024-06-06
Implement `TemplateGrid`

### LINKED ISSUE #244: Implement a grid template
For the moment, we can limit ourselves to regular grids of templates (that will likely represent logical qubits). There is nothing in the code that is explicitly here to build such a grid.

Constructing a `GridTemplate` should be quite easy with the `ComposedTemplate` class.

Having a specific implementation for a grid, just like `StackedTemplate` and `ShiftedTemplate` are not implemented using `ComposedTemplate` (because it does not provide the necessary features) but rather directly as a subclass of `Template` with a specific behaviour. Turns out that a grid should be describable quite easily with a `ComposedTemplate`, so this alternative has not been explored.

nelimee:
This PR implements a generic `TemplateGrid` that can be used to generate a grid of templates representing logical qubits (for example).

Closes #244

### DIFF
--- a/src/tqec/templates/constructions/grid.py
+++ b/src/tqec/templates/constructions/grid.py
@@ -1,0 +1,37 @@
+from tqec.enums import BELOW_OF, RIGHT_OF
+from tqec.templates.base import Template, TemplateWithIndices
+from tqec.templates.composed import ComposedTemplate
+
+
+class TemplateGrid(ComposedTemplate):
+    def __init__(self, rows: int, cols: int, template: Template) -> None:
+        plaquettes_indices_per_qubit = template.expected_plaquettes_number
+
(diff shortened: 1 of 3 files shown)


nelimee:
This is ready for review.

inmzhang (approved):

afowler:
Just as a comment, while we are definitely moving towards having an even
convention for our surface code and standardizing the global 3D structure,
there will be cases where we need to build specialized structures at a
lower level, so we do need to maintain generality. For example,
implementing the latest state distillation factory, which we will talk
about in due course, will require very custom layers of scalable circuits.
The idea is that the entire factory will become a block available in our 3D
template and this block will be carefully constructed using all techniques
available.

Best,
Austin.

PR #252, opened 2024-06-21
Port the C++ parsing scripts for dae files to Python

inmzhang:


nelimee:
I am reviewing this today.

src/tqec/sketchup/dae_parse.py
```diff
@@ -0,0 +1,152 @@
...
+    Args:
+        filepath: Path to the DAE file.
+
+    Returns:
+        A list of Node objects representing the nodes in the SketchUp model.
+    """
+    nodes = []
+    tree = ET.parse(filepath)
+
+    _remove_namespace(tree)
```
nelimee, line 80: It seems like there is no namespace in the provided test files, is that line (and function) really needed?

src/tqec/sketchup/dae_parse.py
```diff
@@ -0,0 +1,152 @@
...
+    matrix = np.fromstring(matrix_text, dtype=float, sep=" ")
+    if matrix.size != 16:
+        raise TQECException("The transformation matrix must have exactly 16 elements.")
+    x, y, z = matrix[3], matrix[7], matrix[11]
+    for value in (x, y, z):
+        if not _can_cast_to_int_safely(value):
+            raise TQECException(
+                f"The translation component of the transformation matrix must be integers, but got {(x, y, z)}."
+            )
+    return NodePosition(int(round(x)), int(round(y)), int(round(z)))
```
nelimee, line 125: Even though this is a correct translation from the original C++ code, I think it would be nice to add a quick explanation about why the entries `3`, `7` and `11` are encoding the position.

nelimee (approved):
The re-write seems correct. I left a few comments, but nothing is important enough to block this PR.

An important note: in my opinion this is a first step, and I think it would be nice to use a third-party library to parse `.dae` files in the future. I found [`pycollada`](https://pycollada.readthedocs.io/en/latest/) and successfully used it to parse `library.dae`. It will likely be more robust than any hand-written parser we can implement ourselves, and probably (I did not look extensively at the documentation yet) provide nice abstraction around things we are interested in.

src/tqec/sketchup/dae_parse.py
inmzhang: There is the default namespace defined in `<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">`. And according to the [documentation of xml](https://docs.python.org/3/library/xml.etree.elementtree.html):
 
> Also, if there is a [default namespace](https://www.w3.org/TR/xml-names/#defaulting), that full URI gets prepended to all of the non-prefixed tags.

We need to remove it here.

src/tqec/sketchup/dae_parse.py
inmzhang: I'll add comments for that.

inmzhang:
> I found [pycollada](https://pycollada.readthedocs.io/en/latest/) and successfully used it to parse library.dae

I'll leave this PR open and look into `pycollada`. If that's a better solution, I'll use that instead.

src/tqec/sketchup/dae_parse.py
nelimee: Then I do not understand the implementation:
```py
def _remove_namespace(tree: ET.ElementTree):
    for elem in tree.iter():
        if "}" in elem.tag:
            elem.tag = elem.tag.split("}", 1)[1]
```

Why is the function searching for `}`?

inmzhang: Default namespace will also be prepended to each element's tag in the form of `{url}` when parsing with `xml.etree.ElementTree`.

nelimee: Ok, got it, I am fine with it then :) Thanks for the explanation.

inmzhang:
See #253

PR #253, opened 2024-06-21
Implement SketchUp-exported dae file parsing with `pycollada`

inmzhang:
A direct port of C++ parse script like #252  does work but using the existing COLLADA parsing library like `pycollada` seems more robust and performant.

### DIFF
--- a/src/tqec/sketchup/dae_parse_test.py
+++ b/src/tqec/sketchup/dae_parse_test.py
@@ -1,0 +1,69 @@
+import pathlib
+
+from tqec.sketchup.dae_parse import parse_dae_file_to_nodes, Node
+
+TEST_DIR = pathlib.Path(__file__).parent / "dae_parse_test"
+
+
+def _group_nodes_by_type(nodes: list[Node]) -> dict[str, list[tuple[int, int, int]]]:
+    nodes_by_type = {}
+    for node in nodes:
+        nodes_by_type.setdefault(node.node_type, []).append(node.position.to_tuple())
+    for node_type, positions in nodes_by_type.items():
+        nodes_by_type[node_type] = sorted(positions)
+    return nodes_by_type
+
+
+def test_parse_templates_nodes():
+    library_nodes = parse_dae_file_to_nodes(TEST_DIR / "templates.dae")
+    nodes_by_type = _group_nodes_by_type(library_nodes)
+    assert nodes_by_type == {
+        "xoz": [(7, 2, 0)],
+        "zox": [(9, 2, 0)],
+        "ozxh": [(4, 6, 1)],
+        "oxzh": [(4, 4, 1)],
+        "zoxh": [(9, 5, 1)],
+        "ozx": [(4, 2, 0)],
+        "xzx": [(2, 2, 0)],
+        "zxoh": [(2, 6, 0)],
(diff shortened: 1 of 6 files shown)


src/tqec/sketchup/dae_parse_test.py
```diff
@@ -0,0 +1,69 @@
...
+            (0, 0, 3),
+            (0, 0, 6),
+            (0, 0, 9),
+            (0, 3, 6),
+            (3, 3, 0),
+            (3, 3, 3),
+            (3, 3, 6),
+            (3, 3, 9),
+        ],
+    }
```
nelimee, line 69: Minor comment, up to your preference:
It works as-is, but you do not necessarily have to sort. Here we only want to perform a unordered equality comparison, so it is equivalent to return a `set`/`frozenset`. 

The following **only** replace the inner `list` type by `set` and removes the second loop that was sorting the `list` values.
```suggestion

def _group_nodes_by_type(nodes: list[Node]) -> dict[str, set[tuple[int, int, int]]]:
    nodes_by_type: dict[str, set[tuple[int, int, int]]] = {}
    for node in nodes:
        nodes_by_type.setdefault(node.node_type, set()).add(node.position.to_tuple())
    return nodes_by_type


def test_parse_templates_nodes():
    library_nodes = parse_dae_file_to_nodes(TEST_DIR / "templates.dae")
    nodes_by_type = _group_nodes_by_type(library_nodes)
    assert nodes_by_type == {
        "xoz": {(7, 2, 0)},
        "zox": {(9, 2, 0)},
        "ozxh": {(4, 6, 1)},
        "oxzh": {(4, 4, 1)},
        "zoxh": {(9, 5, 1)},
        "ozx": {(4, 2, 0)},
        "xzx": {(2, 2, 0)},
        "zxoh": {(2, 6, 0)},
        "zxz": {(0, 0, 0)},
        "oxz": {(4, 0, 0)},
        "xzoh": {(2, 8, 0)},
        "xozh": {(7, 5, 1)},
        "zzx": {(2, 4, 0)},
        "xxz": {(0, 4, 0)},
        "zxx": {(0, 2, 0)},
        "xzo": {(0, 8, 0)},
        "xzz": {(2, 0, 0)},
        "zxo": {(0, 6, 0)},
    }


def test_parse_cnot_nodes():
    cnot_nodes = parse_dae_file_to_nodes(TEST_DIR / "cnot.dae")
    nodes_by_type = _group_nodes_by_type(cnot_nodes)
    assert nodes_by_type == {
        "oxz": {(1, 3, 6)},
        "zxx": {(0, 3, 3)},
        "zox": {(0, 3, 3)},
        "zxo": {
            (0, 0, 1),
            (0, 0, 4),
            (0, 0, 7),
            (0, 3, 4),
            (3, 3, 1),
            (3, 3, 4),
            (3, 3, 7),
        },
        "zxz": {
            (0, 0, 0),
            (0, 0, 3),
            (0, 0, 6),
            (0, 0, 9),
            (0, 3, 6),
            (3, 3, 0),
            (3, 3, 3),
            (3, 3, 6),
            (3, 3, 9),
        },
    }
```

nelimee (approved):
Looks good, one very minor comment, you can ignore it and merge with the current code without problem if you prefer the current version.

PR #254, opened 2024-06-25
Use scalable variant in templates

### LINKED ISSUE #241: Use as much `Scalable` structures as possible

nelimee:
This PR changes how scaling is performed in `tqec`.

This fix #241.

**Before**

Before that PR, the `Dimension` class was storing a `LinearFunction` describing how the `Dimension` was supposed to scale and a `value` representing the current value of the `Dimension`.

The `Template` instance were taking the `Dimension` instances as input, and were using them to scale internally, but were only outputting integers resulting from the "instantiation" of such `Dimension` instances.

**In this PR**

This PR completely removes the `Dimension` class. `Template` instances are now responsible for storing the scaling parameter `k` along with the `LinearFunction` instances needed to scale.

The `Template` class now requires the implementation of a new `scalable_shape` property that returns the shape of the `Template` instance as a fully scalable structure. It basically returns the full information (the full function) instead of returning its value on one specific point (before). For that, `ComposedTemplate` needed a `PiecewiseLinearFunction` class due to the use of `min`/`max` in the computation of the methods needed for `scalable_shape`.

As side (but important) notes:
- the `ShiftedTemplate` and `StackedTemplate` classes are removed from the code base, because they are not currently used and are more classes that would need maintenance.
- all the JSON-related methods are also removed, due to their replacement in #240.


nelimee:
This is ready for review.

src/tqec/templates/scale.py
```diff
@@ -12,110 +17,347 @@ class LinearFunction:
...
-            )
-        return (value - self.offset) // self.slope
+    def intersection(self, other: LinearFunction) -> float | None:
+        """Compute the intersection between two linear functions.
+        Args:
+            other: the `LinearFunction` instance to intersect with `self`.
+        Returns:
+            an instance of `IntersectionResults` containing the necessary
+            data to know when the two `LinearFunction` instances intersect
+            and which one is above/below before/after the intersection.
```
inmzhang, line 43:
```suggestion
            If they intersect, return x such that `self(x) = other(x)`.
            Otherwise, return None.
```
`IntersectionResults` is not defined.

inmzhang (changes requested):
LGTM, some minor changes are requested.

inmzhang (approved):

PR #256, opened 2024-06-27
Validate `ComposedTemplate`

### LINKED ISSUE #242: Validate `ComposedTemplate`

nelimee:
This PR implements the necessary classes and functions to be able to validate any `ComposedTemplate` instance. Here, validating means checking that, for any possible input `k`, the `ComposedTemplate` instance will not overlap with itself.

Closes #242


nelimee:
I just realised that a lot of documentation was missing, I will add it before asking for another review.

nelimee:
Since last review, I added documentation, fixed some edge cases and allowed `Interval` instances to explicitly know whether or not each boundary should be considered included in the interval or excluded from it.

I think this is ready for review.

inmzhang (approved):
LGTM. Just wanted to mention that, I found a package [`portion`](https://github.com/AlexandreDecan/portion?tab=readme-ov-file#interval-operations) that appear to do something similar to `Interval`.

nelimee:
> LGTM. Just wanted to mention that, I found a package [`portion`](https://github.com/AlexandreDecan/portion?tab=readme-ov-file#interval-operations) that appear to do something similar to `Interval`.

This package seems way better than my hand-written solution! I would like to use it instead of my custom-wrote code that is very likely way less robust. Do you think I should do that in this PR or in another one?

PR #261, opened 2024-07-02
Manipulating dae file

inmzhang:
I'm working on manipulating the dae files. To be more specific, I'm trying to do the following things in this PR:

## Template Blocks

Define all the template blocks(xzz, zox, zoxh, ...) in a **canonical** way in the backend, which will help with reconstructing the dae file after making changes to the block instances like scaling the connectors. To be canonical, the definition of the cubes/rectangles in the template satisfy the following condition:
- All vertex coordinates are non-negative and one vertex is located at the coordinate origin.
- All edges are aligned along the coordinate axes. Cube's edge length equals to 1. Rectangle is of shape (1,1,2) with its long edge aligned along the connecting ("o") axis direction.

The **canonical orientation** of a connector(rectangle) is defined as pointing towards the positive axis along the connecting direction.

A template block is defined as:
```py
@dataclass(frozen=True)
class Block:
    block_type: BlockType
    faces: list[Face]
```
where the `block_type` is an enum of 18 types of blocks and `faces` is a collection of the `Face` that consists of the geometry of the cube/rectangle. Each `Face` is just a collection of the vertices and triangle indices that describe a face geometry.

## SketchUp Model

To reconstruct a SketchUp model, we need to make a dae file that conforms to the structure required by SketchUp. I encapsulate all the library nodes(template blocks), geometries, materials and basic scene setup in an internal `_BaseSketchUpModel` object. And it will only be created when we need to write the varied/constructed model to the file.

All we care about is the block instances, which are the references to the template blocks defined in our library. I made the `SketchUpModel` to handle the creation of the block instances and resolve the connectivity between the instances. Internally, it only maintains a connectivity directional graph without the actual positions of the blocks. The positions of the block instances will be infered from the shape of the blocks and the connectivity graph when we write the model to the file.

## Problems and Limitations

Some problems that I'm aware of:

- The current template file provided by Austin has some problems in the definition: `oxzh`, `ozxh`, `zoxh`, `zox`, `xozh`, `xoz` are not defined by the "RGB" convention. You can see the template in SketchUp's components browser. For example, for the `xoz` block, the component in the SketchUp is as follows: 
    image
    which is actually a `oxz` instead. The instance in the scene looks correct because there is an underlying rotation operation to the instance. I found this problem when I try to check all the rotation associated to the instance are identity. This definition problem leads to the wrong position of some block instances, which then make the import of `exportable_cnot.dae` in the code fail since I simply ignore all the rotation operations.

- When SketchUp imports a dae model, it erases all the library node's name. If we export a model from SketchUp, then manipulate it and then import it into SketchUp, all the node's definition name will be lost. That means we can not export a well-defined dae file after a export-import roundtrip unless we change the definition names by hands, which is a heavy work and not realisitic.

Some limitations of the current implementation:

- The graph representation of the model currently works for the cube/rectangle blocks. I'm not sure whether it will work for the general blocks with some more complex shapes.
- The model should be represented as a single connected components since there are no stored position information.
- Do not support open connectors, which means each connector should be connected to exactly two cubes
at its twp ends. This should not be a problem for a valid model. However, it also means we can not import `exportable_templates.dae` as a model in the code.
- Rotation of the blocks is not supported.
- There are little documentation and no unit tests available.

### Model Construction

Let's make a CNOT SketchUp model:
```py
from tqec.sketchup.model import SketchUpModel
from tqec.sketchup.block import BlockType

model = SketchUpModel()

zxz_cubes = [model.add_cube(BlockType.ZXZ) for _ in range(9)]
zxx_cube = model.add_cube(BlockType.ZXX)
for pair in [(0, 1), (1, 2), (2, 3), (5, 6), (6, 7), (7, 8)]:
    model.add_connector(BlockType.ZXO, zxz_cubes[pair[0]], zxz_cubes[pair[1]])
model.add_connector(BlockType.ZXO, zxx_cube, zxz_cubes[4])
model.add_connector(BlockType.ZOX, zxz_cubes[1], zxx_cube)
model.add_connector(BlockType.OXZ, zxz_cubes[4], zxz_cubes[7])
model.write("cnot_from_scratch.dae")
```

We can see there are no actual positions involved during the construction of the model. All we care about is the types of the blocks and the connectivity between them. Import it into SketchUp, we get:

### Model Import

```py
from tqec.sketchup.model import SketchUpModel

model = SketchUpModel.from_dae_file("cnot_from_scratch.dae")
print(model.num_cubes) # 10
print(model.num_connectors) # 9
```

### Rescale the Connectors

By the canonical definition of template blocks, all the connectors have length 2 along the connecting direction. The length can be rescaled by a factor `scale` to make the connector length as `2*scale`. The rescaling will have effect on the whole model when infering the positions of the blocks.

```py
from tqec.sketchup.model import SketchUpModel

model = SketchUpModel.from_dae_file("cnot_from_scratch.dae")
model.set_scale_for_all_connectors(0.5)
model.write("cnot_from_scratch_half_scale.dae")

model.set_scale_for_all_connectors(2)
model.write("cnot_from_scratch_twice_scale.dae")
```

We manually import all the three scale of the same models into SketchUp and get:


afowler:
This is excellent :-) thank you for the detailed text and images.

Best,
Austin.

src/tqec/sketchup/model.py
```diff
@@ -0,0 +1,511 @@
...
+    def num_cubes(self) -> int:
+        return self._instance_graph.number_of_nodes()
+
+    @property
+    def num_connectors(self) -> int:
+        return self._instance_graph.number_of_edges()
+
+    @property
+    def num_instances(self) -> int:
+        return self.num_cubes + self.num_connectors
```
nelimee, line 75: In practice, this `num_instances` property and the `_instance_id` one can be used interchangeably right?

nelimee (commented):
Thanks for the contribution, I think this is the right path, nice work!

I added a few comments/questions I had on the code, here are more generic comments that did not really fit in a single place in the code.

I would advocate for the use of constants for weights in networkx graphs. Basically, taking "block_type" as an example, having
```py
_BLOCK_TYPE_KEY = "block_type"
```
at the beginning of the file, and using code like
```py
self._instance_graph.add_edge(
    src, dst, **{_BLOCK_TYPE_KEY: block_type}
)
```
to add an edge with a `block_type` weight and 
```py
edge[_BLOCK_TYPE_KEY]
```
to recover the weight, with maybe an additional public function to recover the block type of a specific edge/node.
Same remark for `scale` and `connector_id`.

I think that the restriction `All vertex coordinates are non-negative and one vertex is located at the coordinate origin.` is impossible to check in some cases. Basically, it can only be checked if the logical qubit that has a corner at the spatial origin is one of the qubits that is initialised at "time 0". If not, the above condition will not be checked.

src/tqec/sketchup/model.py
inmzhang: It is for now. But if we want to add functionality to remove instances, the two might not be equal since `_instance_id` is monotonically increasing by design.

inmzhang:
> I think that the restriction All vertex coordinates are non-negative and one vertex is located at the coordinate origin. is impossible to check in some cases. Basically, it can only be checked if the logical qubit that has a corner at the spatial origin is one of the qubits that is initialised at "time 0". If not, the above condition will not be checked.

Note that this restriction is only posed on the templates, which satisfy the restriction natively by our construction. The templates do not exists in the real scene but only provides the definition of geometry/material for instances that point to it. For instances, they are the instantiation of templates and their position have no restrictions and can be anywhere in the 3D space. Since every block on the scene are all instances, so their is no problem with the restriction.

To explain more on template vs. instance. Template is the asset that can be reused, which corresponds to the nodes in the <library_nodes> of a COLLADA file:
```xml
<library_nodes>
        <node id="ID3" name="zxz">
            <instance_geometry url="#ID4">
                <bind_material>
                    <technique_common>
                        <instance_material symbol="Material2" target="#ID6">
                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
                        </instance_material>
                    </technique_common>
                </bind_material>
            </instance_geometry>
            ...
        </node>
       ...
</library_nodes>
```

An instance will reference a template and specify the transformation on the template. The position of the instance is specified by the translation part of the transformation matrix. In COLLADA file, the instance is represented by the <instance_node>:
```xml
<library_visual_scenes>
        <visual_scene id="ID1">
            <node name="SketchUp">
                <node id="ID2" name="instance_0">
                    <matrix>1 0 0 2 0 1 0 0 0 0 1 0 0 0 0 1</matrix>
                    <instance_node url="#ID27" />
                </node>
                ...
        <visual_scene id="ID1">
<library_visual_scenes>
```

So we pin one vertex of the template block to the origin just to make sure the translation is relative to the origin.

notebooks/sketchup_demo.ipynb
```diff
@@ -0,0 +1,171 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Library Blocks\n",
+    "\n",
+    "All the library/template blocks are defined in the backend to be the building blocks for a model. A library block is defined by its geometries(faces) and materials(X/Z/H face type). For example, a `zxx` block is represented by its 6 faces as follows:\n",
+    "```txt\n",
```
KabirDubey, line 29: Put ````python`  instead of ````txt` for readability in VS Code.

KabirDubey:
@nelimee mentioned in 7/24 meeting that we need to merge this to use it for the backend. @inmzhang are you still making changes or can we move this out of drafts?

KabirDubey:
> The model should be represented as a single connected components since there are no stored position information.

Could you clarify this? It seems we are tracking the positions of the library nodes.

inmzhang:
> > The model should be represented as a single connected components since there are no stored position information.
> 
> Could you clarify this? It seems we are tracking the positions of the library nodes.

Currently, we track the positions in the `_instantiate_instances()` method and only call the method when write the model to file. So we do not have the explicit position information before writing it.

notebooks/sketchup_demo.ipynb
inmzhang: I don't have the problem in my IDE but I can make the change.

inmzhang:
> @nelimee mentioned in 7/24 meeting that we need to merge this to use it for the backend. @inmzhang are you still making changes or can we move this out of drafts?

I just watched the recording. There are some changes to make in my mind like allowing to specify the position of cubes explicitly, simplifying the block types we need to specify when constructing a model and observable/correlation surface inferring with the zx graph. But for the MVP purpose, I think this should be ready to be merged after adding some documentation. I will add the docs at today or tomorrow then mark this PR ready to be reviewed.

KabirDubey:
> > @nelimee mentioned in 7/24 meeting that we need to merge this to use it for the backend. @inmzhang are you still making changes or can we move this out of drafts?
> 
> 
> 
> I just watched the recording. There are some changes to make in my mind like allowing to specify the position of cubes explicitly, simplifying the block types we need to specify when constructing a model and observable/correlation surface inferring with the zx graph. But for the MVP purpose, I think this should be ready to be merged after adding some documentation. I will add the docs at today or tomorrow then mark this PR ready to be reviewed.

Sounds good! Feel free to delegate any tasks to me.

nelimee (approved):
I checked the new code since my last review and everything seems alright! 

Thanks a lot for that contribution!

PR #267, opened 2024-07-24
Implement helpers to recover spatial boundaries

### LINKED ISSUE #260: Prepare the `tqec` code base to the concept of spatial and temporal boundaries

Going from "simple" 2-dimensional templates to the more complete picture of 3-dimensional topological structures will require to introduce a few concepts into the `tqec` code base.

One of the main concept that will need to be introduced is "interchangeable spatial and temporal boundaries". Spatial boundaries are the `1` by `2k` and `2k` by `1` rectangular templates at the borders of a logical qubit. Temporal boundaries are basically the initialisation/measurement layers at the beginning/end of a memory experiment.

To implement QEC computations, we need to be able to easily swap the plaquettes used in any of these spatial/temporal boundaries by other, user-provided, plaquettes.

This issue is here to track the fact that we need to think about how to implement that feature, and to track the advancement of the feature once someone will want to look into it.

nelimee:
This PR fixes #260 by introducing a few helper methods to ease the replacement of spatial boundaries.


inmzhang (approved):
LGTM. One question is that do we still need a rectangle version of qubit template?

nelimee:
> LGTM. One question is that do we still need a rectangle version of qubit template?

I agree that we do not.

inmzhang:
I agree that we do not need it anymore and just want to check for that.

PR #278, opened 2024-07-25
Refactor plaquette library for better uniformity

### LINKED ISSUE #276: Fix initialisation plaquettes implementation
## Describe the bug
The initialisation plaquettes are currently implemented with only reset operations (see [src/tqec/plaquette/library/initialisation.py](https://github.com/QCHackers/tqec/blob/0d91fb530934c735cf31392558417ff0c81162b6/src/tqec/plaquette/library/initialisation.py)) but defined in all the external supports as resets + 1 round of memory.

## Expected behavior
Initialisation plaquettes should follow the definition used in other supports.

### LINKED ISSUE #277: Remove detectors from plaquettes
## Describe the bug
`Plaquette` building functions are still including hand-written detectors. These should rather be automatically computed with the result of #249 and #259.

Remove the detectors from the plaquette definitions.

## Expected behavior
Detectors should be removed from the plaquette definition. They will eventually (likely in #259) be included again, in an automated manner.

nelimee:
Fixes #276 and fixes #277 by refactoring the plaquette library.

It basically change two things:

1. Initialisation and measurement plaquettes are now also performing a memory round before/after measurement/initialisation.
2. The interface of functions creating plaquettes is now more uniform.

This breaks most of the notebooks in the repository, but our main goal here is to obtain a working CNOT implementation, which might require performing breaking changes often, so I think it is fine if we update the notebooks only when the topological CNOT is implemented and working.


inmzhang (approved):
LGTM

PR #279, opened 2024-07-25
Introduce temporal `ComputationBlock`

### LINKED ISSUE #275: Introduce a block structure for the temporal dimension
Currently, the `tqec` code base does not include a native way of representing the temporal dimension. It has to be done by hand, for each user, and is quite inconvenient.

It would be nice to introduce a "block" structure like the one drafted in #260 to manage everything related to the temporal dimension **and** manage connecting 2 blocks together with a connector (i.e., replacing spatial/temporal boundaries to glue two blocks together).

**Additional context**
Several complications are expected to rise up during the implementation. A non-exhaustive list is:
- how to devise if a special connector should be used in order to keep the even-parity of logical qubit boundaries?
- extra care should be given to situations such as the edge internal to the L angle 
  image
  that require to carefully merge 2 boundaries together.

### LINKED ISSUE #284: Allow X-basis resets and measurements
There is currently no way to reset or measure a qubit in the `X` Pauli basis, except by manually {ap,pre}pending a `H` gate.

Manually adding a Hadamard gate to use the equivalences `Rx = Rz H` and `Mx = H Mz` will lead to issues with existing scheduling (adding a moment in some plaquette but not everywhere), loose some expressiveness on the developer side, and is quite tedious.

On the other hand, `cirq` has no notion of X-basis reset or X-basis measurement. 

Implement two custom `cirq` gates to represent X-basis reset/measurement.

Remove completely the dependence on `cirq` and only use `stim`, even internally. This is likely desirable in the long term, but right now this is not the priority.

nelimee:
This PR adds the third dimension to our `tqec` library, easing the representation of 3-dimensional topological quantum error-corrected computations.

Fixes #275.
Fixes #284.


inmzhang:
Since this PR is quite large, I need to take some to review it.

src/tqec/block/library/closed.py
```diff
@@ -0,0 +1,172 @@
...
+        defaultdict(empty_square_plaquette)
+        | {
+            6: xx_memory_plaquette(PlaquetteOrientation.UP),
+            7: zz_memory_plaquette(PlaquetteOrientation.LEFT),
+            9: xxxx_memory_plaquette(),
+            10: zzzz_memory_plaquette(),
+            12: zz_memory_plaquette(PlaquetteOrientation.RIGHT),
+            13: xx_memory_plaquette(PlaquetteOrientation.DOWN),
+        },
+        dimension,
```
inmzhang, line 67: For spatial dimension `2k`, the temporal repetition dimension should be `2k-1`.

src/tqec/block/library/open.py
```diff
@@ -0,0 +1,262 @@
...
+def xzo_block(dimension: LinearFunction) -> StandardComputationBlock:
+    xzz = xzz_block(dimension)
+    xzz.repeating_plaquettes = None
+    return xzz
+
+
+def zxo_block(dimension: LinearFunction) -> StandardComputationBlock:
+    zxz = zxz_block(dimension)
+    zxz.repeating_plaquettes = None
+    return zxz
```
inmzhang, line 262: The functionality of time-direction connector blocks is: 
    1. replace the measurement layer of the cube beneath it with a memeory layer
    2. replace the initialization layer of the cube above it with a memory layer

So it seems this implementation is incorrect in the sense.

src/tqec/block/enums.py
```diff
@@ -0,0 +1,35 @@
...
+
+from tqec.exceptions import TQECException
+from tqec.templates.enums import TemplateSide
+
+
+class BlockDimension(Enum):
+    X = auto()
+    Y = auto()
+    Z = auto()
+    T = Z
```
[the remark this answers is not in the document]
nelimee: Yes, we will have to merge a few common structures. I'm adding a TODO.

src/tqec/circuit/schedule.py
```diff
@@ -354,6 +364,39 @@ def _compute_number_of_non_virtual_moments(circuit: cirq.Circuit) -> int:
...
+            self._raw_circuit.insert(insertion_index, moment)
+        else:
+            # Else, the schedule already exists, in which case we just need to add the
+            # operations to an existing moment. Note that this might fail if two
+            # operations overlap.
+            try:
+                self._raw_circuit[schedule_index] += moment
+                self._number_of_non_virtual_moments += (
+                    ScheduledCircuit._is_virtual_moment(moment)
+                )
```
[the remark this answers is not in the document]
nelimee: It is not, good catch, this should be in the above if-branch.

src/tqec/block/library/open.py
inmzhang: I think the current implementation is still problematic. Maybe some changes on the `Computation` class need to be made to account for the action "remove the initialization/measurement layer of the cube it connects to".

nelimee: Yes, the methods are present (`StandardComputationBlock.replace_boundary_with_empty_plaquettes` and `StandardComputationBlock.instantiate_without_boundary`) but I wanted to wait for a graph-like structure in order to avoid computing which boundaries should be removed / replaced. I think this can be done in a follow-up PR, do you think I should rather implement everything here?

inmzhang: I think it's fine to leave this with another PR. Then I have no problems and I'll approve the merge:)

inmzhang (approved):

Issue #268, opened 2024-07-25
Allow specifying topological QEC structure as graphs


nelimee:
**Is your feature request related to a problem? Please describe.**
The current process leading to the creation of a collection of Stim files according to a given noise model, computation and code distance involves, at the very first step, to open SketchUp and create a 3-dimensional structure representing the computation. 

![CNOT in SketchUp](https://github.com/user-attachments/assets/d9098550-10b8-4a59-9ba3-ea4904255868)

This is a perfectly valid way of doing things, and even the main way we should support.

But some users might want to be able to perform that step without SketchUp, generating the topological computation using code. This is currently possible **but** very tedious as the user has to place the different blocks one by one and to ensure (without the visual representation given by SketchUp) that all the blocks are correctly placed.

It turns out that a more concise format might be used: a ZX diagram.

![CNOT in ZX diagram](https://github.com/user-attachments/assets/f3cae833-1416-464b-8f49-f432f2baf5ef)

**Describe the solution you'd like**
Create the API to be able for a user to create a ZX diagram representing a computation.

PR #285, opened 2024-07-30
Refactor 3D spacetime diagram representation

inmzhang:
Relates to #261 and #268.

This PR focus on refactoring the code of representing and constructing the 3D spacetime diagram, enabling simplified ZX graph representation and infer the concrete block structures from the simple ZX graph.


src/tqec/sketchup/zx_graph.py
```diff
@@ -0,0 +1,334 @@
...
+            abs(self.x - other.x) + abs(self.y - other.y) + abs(self.z - other.z) == 1
+        )
+
+    def as_tuple(self) -> tuple[int, int, int]:
+        """Return the position as a tuple."""
+        return astuple(self)
+
+    def __post_init__(self):
+        if any(not isinstance(i, int) for i in astuple(self)):
+            raise TQECException("Position must be an integer.")
```
nelimee, line 50: Small nitpick: I prefer `__init__` and `__post_init__` methods at the top of the method definitions.

src/tqec/sketchup/zx_graph.py
```diff
@@ -0,0 +1,334 @@
...
+        return Direction3D.all()[i]
+
+    @property
+    def axis_index(self) -> int:
+        """Get the axis index."""
+        return Direction3D.all().index(self)
+
+    def __str__(self) -> str:
+        return self.name
+
```
nelimee, line 80: These two classes might be better placed in `tqec.position`, in case we need to re-use them elsewhere.

src/tqec/sketchup/zx_graph.py
```diff
@@ -0,0 +1,334 @@
...
+    def shift_by(self, dx: int = 0, dy: int = 0, dz: int = 0) -> Position3D:
+        """Shift the position by the given offset."""
+        return Position3D(self.x + dx, self.y + dy, self.z + dz)
+
+    def is_nearby(self, other: Position3D) -> bool:
+        """Check if the other position is near to this position, i.e. Manhattan distance is 1."""
+        return (
+            abs(self.x - other.x) + abs(self.y - other.y) + abs(self.z - other.z) == 1
+        )
```
nelimee, line 42: Small nitpick: I think `is_neighbour` would be a better naming, or even `is_direct_neighbour`. I am not a native english speaker, so the probability that I am wrong is high, but to me, "nearby" is looser then "neighbour" and does not really imply that "nearby" nodes are neighbours, they might be 2 edges appart and that would still be a valid "nearby" node.

src/tqec/sketchup/zx_graph.py
```diff
@@ -0,0 +1,334 @@
+"""ZX graph representation of a 3D spacetime defect diagram."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
```
nelimee, line 5: Needed for typing errors below
```suggestion
from typing import TYPE_CHECKING, cast
```

src/tqec/sketchup/zx_graph.py
```diff
@@ -0,0 +1,334 @@
...
+            data[_EDGE_DATA_KEY]
+            for _, _, data in self._graph.edges(position, data=True)
+        ]
+
+    def draw(self, show_title: bool = True) -> None:
+        """Draw the 3D graph using matplotlib."""
+        import matplotlib.pyplot as plt
+
+        fig = plt.figure()
+        ax = fig.add_subplot(111, projection="3d")
```
nelimee, line 242: Fixing a few typing errors returned by ruff (not mypy yet)
```suggestion
    def draw(self, show_title: bool = True) -> None:
        """Draw the 3D graph using matplotlib."""
        import matplotlib.pyplot as plt
        from mpl_toolkits.mplot3d.axes3d import Axes3D

        fig = plt.figure()
        # See https://matplotlib.org/stable/users/explain/toolkits/mplot3d.html
        ax = typing.cast(Axes3D, fig.add_subplot(111, projection="3d"))
```

nelimee (commented):
Started a review instead of just adding comments, sorry.

src/tqec/sketchup/zx_graph.py
nelimee: Disregard that, I will do a proper review when you think the PR is ready :)

src/tqec/sketchup/zx_graph.py
nelimee: Disregard that, I will do a proper review when you think the PR is ready :)

src/tqec/sketchup/zx_graph.py
nelimee: Disregard that, I will do a proper review when you think the PR is ready :)

src/tqec/sketchup/zx_graph.py
nelimee: Disregard that, I will do a proper review when you think the PR is ready :)

src/tqec/sketchup/zx_graph.py
nelimee: Disregard that, I will do a proper review when you think the PR is ready :)

inmzhang:
I believe this PR is ready for review. Although I didn't have enough time to detail the implementation here, the main goal should be clear from the linked issue.

Additionally, I have implemented code to display the Collada (dae) model in the IPython environment and save the generated HTML content to a file, which can then be viewed in a browser.

src/tqec/position.py
```diff
@@ -60,3 +63,59 @@ class Displacement:
...
+    Y = auto()
+    Z = auto()
+
+    @staticmethod
+    def all() -> list["Direction3D"]:
+        """Get all directions."""
+        return [Direction3D.X, Direction3D.Y, Direction3D.Z]
+
+    @staticmethod
+    def from_axis_index(i: int) -> "Direction3D":
```
nelimee, line 111: Due to the very limited valid inputs, I would rather annotate with `typing.Literal`. If you import `typing`, the following should be stricter and provide more information to the user:
```suggestion
    def from_axis_index(i: typing.Literal[0, 1, 2]) -> "Direction3D":
```
If you prefer to keep `int` as typing, I would then add a runtime check and a `TQECException` with a nice explanatory message about why the entry should not be above 2.

src/tqec/position.py
```diff
@@ -60,3 +63,59 @@ class Displacement:
...
+        """Get the direction from the axis index."""
+        return Direction3D.all()[i]
+
+    @property
+    def axis_index(self) -> int:
+        """Get the axis index."""
+        return Direction3D.all().index(self)
```
nelimee, line 118: Instead of using `auto()` for the enumerations values, you can directly use the index of the direction:
```py
class Direction3D(Enum):
    """Axis directions in the 3D spacetime diagram."""

    X = 0
    Y = 1
    Z = 2
    # ...
```
in which case this method becomes 
```suggestion
    @property
    def axis_index(self) -> int:
        """Get the axis index."""
        return self.value
```
Note that, if I understood correctly, `auto` is already doing that but starts by `1` instead of `0`, so you should be able to simply do
```suggestion
    @property
    def axis_index(self) -> int:
        """Get the axis index."""
        return self.value - 1
```
here. I find that quite confusing, so I would advocate against it.

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+from tqec.exceptions import TQECException
+
+
+@dataclass(frozen=True)
+class Color3D:
+    """Get face colors along the x, y, and z axes."""
+
+    x: str | None
+    y: str | None
+    z: str | None
```
nelimee, line 29: As a first improvement, I would change the typing to either
```suggestion
    x: typing.Literal["x", "z"] | None
    y: typing.Literal["x", "z"] | None
    z: typing.Literal["x", "z"] | None
```
or 
```suggestion
    x: typing.Literal["x", "z", "o"]
    y: typing.Literal["x", "z", "o"]
    z: typing.Literal["x", "z", "o"]
```
It might be worth representing one colour as a small enumeration such as
```py
class Colour(Enum):
    X = "x"
    Z = "z"
    O = "o"
```
and define some helper methods on that enumeration. I leave the choice to you :)

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+    def push_color_at_direction(
+        self, direction: Direction3D, color: str | None
+    ) -> Color3D:
+        """Set the color at the given direction."""
+        if direction == Direction3D.X:
+            return Color3D(color, self.y, self.z)
+        elif direction == Direction3D.Y:
+            return Color3D(self.x, color, self.z)
+        else:
+            return Color3D(self.x, self.y, color)
```
nelimee, line 55: Might be replaced with 
```suggestion
        colors = list(astuple(self))
        colors[direction.axis_index] = color
        return Color3D(*colors)
```

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+    """Valid cube types in the library."""
+
+    ZXX = "zxx"
+    XZX = "xzx"
+    XXZ = "xxz"
+    XZZ = "xzz"
+    ZXZ = "zxz"
+    ZZX = "zzx"
+    # Virtual cube for open port
+    VIRTUAL = "virtual"
```
nelimee, line 87: Equivalent to
```suggestion
class CubeType(StrEnum):
    """Valid cube types in the library."""

    ZXX = auto()
    XZX = auto()
    XXZ = auto()
    XZZ = auto()
    ZXZ = auto()
    ZZX = auto()
    # Virtual cube for open port
    VIRTUAL = auto()
```
per the [documentation of `auto()`](https://docs.python.org/3/library/enum.html#enum.auto). I prefer your version, but the proposed version above removes any possibility of typo in the values.

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+        if color == Color3D(None, None, None):
+            return CubeType.VIRTUAL
+        if any(c is None for c in astuple(color)):
+            raise TQECException("All the color must be defined for a non-virtual cube.")
+        return ty.cast(CubeType, parse_block_type_from_str("".join(astuple(color))))
+
+    def normal_direction_to_corner_plane(self) -> Direction3D:
+        """If the cube is at a corner, return the normal direction to the corner plane.
+
+        Due to the color match rule at the corner turn, the cornel plane can be inferred
```
nelimee, line 113:
```suggestion
        Due to the color match rule at the corner turn, the corner plane can be inferred
```

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+    ZOX = "zox"
+    XZO = "xzo"
+    ZXO = "zxo"
+    # pipes with H
+    OZXH = "ozxh"
+    OXZH = "oxzh"
+    ZOXH = "zoxh"
+    XOZH = "xozh"
+    ZXOH = "zxoh"
+    XZOH = "xzoh"
```
nelimee, line 151: Same remark as above with `StrEnum` and `auto()`.

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+
+    @property
+    def has_hadamard(self) -> bool:
+        """Check if the block type has a hadamard transition."""
+        return "h" in self.value
+
+    @property
+    def direction(self) -> Direction3D:
+        """Return the direction of the pipe."""
+        return [Direction3D.X, Direction3D.Y, Direction3D.Z][self.value.index("o")]
```
nelimee, line 161:
```suggestion
        return Direction3D.all()[self.value.index("o")]
```
or even 
```suggestion
        return Direction3D.from_axis_index(self.value.index("o"))
```

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,551 @@
...
+        for c in astuple(color):
+            if c is None:
+                pipe_color.append("o")
+            elif has_hadamard and not src_side:
+                pipe_color.append("x" if c == "z" else "z")
+            else:
+                pipe_color.append(c)
+        if has_hadamard:
+            pipe_color.append("h")
+        return ty.cast(PipeType, parse_block_type_from_str("".join(pipe_color)))
```
nelimee, line 188:
```suggestion
        return PipeType("".join(pipe_color).lower())
```
is enough right? The `parse_block_type_from_str` seems to be adding more code than it removes.

nelimee (commented):
The `.dae` file viewer seems to be working correctly! Very nice! A few minor remarks:
- Is it possible to have to same colours in the javascript viewer as in SketchUp? The sketchup colours are more gray than dark/white, whereas the javascript loader is seems really dark & white:
image
versus 
image
- Is is possible to have a small visual delimiter between blocks, like in sketchup?

I added several comments in the code. I need to do another reading pass eventually, I spent less time reviewing the collada- and ZX-related portions on the code.

inmzhang:
> Is it possible to have to same colours in the javascript viewer as in SketchUp? The sketchup colours are more gray than dark/white, whereas the javascript loader is seems really dark & white.

Actually I used exactly the same color definition as that viewed in SketchUp(the same dae file). It should be the [issue of the light sources](https://forums.sketchup.com/t/i-need-an-explanation-about-coloring-in-the-model/110493/7) in SketchUp are different from thoes I used in `three.js`. I tried some lights setup but did not manage to match the view from SketchUp.

> Is it possible to have a small visual delimiter between blocks, like in sketchup?

In principle, this can be done by showing the edges of each face after merging coplanar triangles in the same face. I spent some time trying to do that but did not get a satisfying result yet...

I'm quite new to 3D modeling and computer graphics, maybe someone more experienced can help with this :)

src/tqec/sketchup/block_graph.py
inmzhang: I have thought about using `StrEnum` for that, but it's available only until Python 3.11.

src/tqec/position.py
inmzhang: I often come across some type issues with `Literal` when I want to pass a `int`/`string` as the argument of the function. For example, the following code should raise a type error when using int literal for the argument type:
```py
a = [1, 3, 2]
i = a.index(3)
Direction3D.from_axis_index(i)
```
What's the general solution to this without tedious `cast` everywhere?

nelimee: The solution I would use is to keep the annotation generic (`int` as initially), but introduce a runtime check in the method:
```suggestion
    def from_axis_index(i: int) -> "Direction3D":
        if not 0 <= i <= 2:
            raise TQECException(
                f"Could not get the axis related to index {i}. Axes "
                f"are indexed 0, 1 or 2 and 0 <= {i} <= 2 should be verified."
            )
```
I do not have a suitable solution to make the typing information right, this is up to the type checkers, and they do not implement that fully.

src/tqec/sketchup/block_graph.py
nelimee: Ho, then I am fine with your implementation, sorry I did not realised that it was `>= 3.11`.

nelimee (approved):
I listed a few comments, I think the code is ready to be merged once all the comments have been addressed (either included or explicitly not included).

src/tqec/sketchup/block_graph.py
```diff
@@ -0,0 +1,571 @@
...
+                block_graph.add_cube(node.position, CubeType.VIRTUAL)
+                nodes_to_handle.remove(node)
+        # Check 3D corner
+        for node in nodes_to_handle:
+            if len({e.direction for e in zx_graph.edges_at(node.position)}) == 3:
+                raise TQECException(f"ZX graph has a 3D corner at {node.position}.")
+        # The color of corner cubes can be inferred locally from the ZX graph
+        # by querying which plane the corner is in and the color of the ZX node.
+        corner_cubes: dict[Position3D, Cube] = {}
+        for node in set(nodes_to_handle):
```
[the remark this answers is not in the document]
nelimee: Right, I missed that. I'm OK with the current code then.

PR #286, opened 2024-07-31
Implement optimised detector computation from the block structure

nelimee:
This PR aims at proposing a tentative implementation to automatically compute detectors from the block structure.

More details to come later.

Note: based on #279, the diff will be cleaner once #279 is merged.


nelimee:
Closing this PR because:

- it was exploratory work,
- I am not sure the path used here is the right one.

Feel free to pick code from it if you think it might help you.

Issue #272, opened 2024-07-25
Plot an observable


nelimee:
**Is your feature request related to a problem? Please describe.**
There is currently no way of plotting an observable to have a visual understanding of what we are measuring. 

**Describe the solution you'd like**
Ideally, a way to plot an observable such as 

![Observable plotted on a ZX diagram](https://github.com/user-attachments/assets/2e63230d-3c08-4fa9-8ab0-1f7d6c34a2b7)

**Additional context**

- This issue might be interesting to implement alongside #270 as a way to visualise the output of the function created in #270 and debug.
- We should be careful in the data obtained from #271. The transformation performed by #271 (transforming the output data of #270 to a "scalable observable") should ideally be reversible, so that we can, from the "scalable observable" recover the graph-data output by #270 and plot it using the result of this issue.

Issue #270, opened 2024-07-25
List all valid observables from a ZX diagram


nelimee:
**Is your feature request related to a problem? Please describe.**
We currently have no way to automatically discover observable. 

**Describe the solution you'd like**
This will be a two-stage process:

1. Create a function that takes a ZX diagram such as 
    ![Image](https://github.com/user-attachments/assets/28b03fa1-4869-4366-adc0-f5b6e28ee6b6)
    as input and returns a list of all the observable that can be considered valid. This amounts to a graph exploration following some exploration rules that have been detailed in our previous meetings and in some presentation.
2. _(separate issue: #271)_ Create a function that takes the data obtained in 1 and the topological computation as input and returns a **scalable** observable.

This issue is only about 1. 

The ZX diagram could be represented by a `networkx.Graph` with some annotations (e.g., is the node black or white?), but if another solution is found to be better then let's use it!

Note that the return format is not defined yet, you will have to think about it and discuss with people on #271 and #272 to be sure that the return format of point 1 is usable in 2.

inmzhang:
Question: how to choose the generating set of the valid observable?

As the deterministic observables are just a way to verifying the underlying (part of)stabilizer flows represented by the block structure, there are some freedom to choose the generating set of the observable. For example, in the CNOT gate shown above, we can choose between "1", "4" and mirrored "h" shape observables. And two of them should be enough. Is there any general idea to choose the independent generating set to track? Austin, could you elaborate on this?@afowler

afowler:
For the moment I'd like to keep all unique (not just generating)
observables. I think they are all (at least initially) worthy of simulation
and study.

Best,
Austin.

PR #290, opened 2024-08-07
Find correlation subgraphs with `ZXGraph` and visualize them

inmzhang:
Relates to #270 and #272.

I try to generalize the problem of finding all the valid logical observables to finding all the "correlation subgraphs". Here I define a correlation subgraph as the correlation surface guided by the `ZXGraph` structure. It's more general than the logical observables because it also works for `ZXGraph` with open ports, and in that senario, the correlation subgraph can be used to verify all the possible stabilizer flow of the computation represented by the `ZXGraph`.

We will eventually use the correlation subgraph to construct scalable observables at the level of `BlockGraph`. And directly showing the correlation surface in the 3D model will be implemented in the future.

Some concrete examples have been added to https://github.com/QCHackers/tqec/blob/feat/observable/notebooks/sketchup_demo.ipynb. You can also view the notebook with the [ReviewNB outputs](https://app.reviewnb.com/QCHackers/tqec/blob/feat/observable/notebooks/sketchup_demo.ipynb/) below.


src/tqec/sketchup/zx_graph.py
```diff
@@ -279,3 +353,109 @@ def __eq__(self, other: object) -> bool:
...
+            )
+            cur_corr_node = ZXNode(cur_zx_node.position, cur_corr_type)
+            edge_between_cur_parent = ZXEdge(
+                parent_corr_node, cur_corr_node, edge.has_hadamard
+            )
+            branched_subgraphs.append(
+                [
+                    subgraph | {edge_between_cur_parent}
+                    for subgraph in self._find_correlation_subgraphs_dfs(
+                        cur_corr_node, cur_zx_node.node_type, set(visited_positions)
```
nelimee, line 384: I think using `copy` would be more explicit here
```suggestion
                        cur_corr_node, cur_zx_node.node_type, copy(visited_positions)
```

src/tqec/sketchup/zx_graph.py
```diff
@@ -279,3 +353,109 @@ def __eq__(self, other: object) -> bool:
...
+            root_corr_node = ZXNode(node.position, correlation_type)
+            for edges in self._find_correlation_subgraphs_dfs(
+                root_corr_node, node.node_type, set()
+            ):
+                if frozenset(edges) in multi_edges_correlation_subgraphs:
+                    continue
+                subgraph = ZXGraph(
+                    f"Correlation subgraph {num_subgraphs} of {self.name}"
+                )
+                subgraph.add_node(node.position, correlation_type)
```
nelimee, line 439: I think this line is not really needed as the parent node will appear in at least one of the edges. Am I missing something?
```suggestion
```

src/tqec/sketchup/zx_graph.py
```diff
@@ -112,18 +127,31 @@ def edges(self) -> list[ZXEdge]:
...
         Args:
             position: The 3D position of the node.
             node_type: The type of the node.
+            raise_if_exist: Whether to raise an exception if the node already exists.
+
         """
-        if position in self._graph:
+        if raise_if_exist and position in self._graph:
             raise TQECException(f"Node {position} already exists in the graph.")
         self._graph.add_node(position, **{_NODE_DATA_KEY: ZXNode(position, node_type)})
```
nelimee, line 156: It was unclear to me what happened if `raise_if_exist == False and position in self._graph` and I had to look it up in the code of `networkx.Graph.add_node`. 
It turns out that, in this case, only the metadata is updated. I think this should be at least written in the documentation, and I would even prefer renaming the `raise_if_exist` flag to something self-explanatory (but I do not have good ideas for a new name, hence my first proposition).

nelimee (approved):
Overall I really like how you solved the problem. 

I think it would help a lot to add a more complete docstring for both `ZXGraph.find_correlation_subgraphs` and `ZXGraph._find_correlation_subgraphs_dfs`. The later in particular would benefit from a high-level description of the recursive algorithm used. 

The notebook changes illustrate very nicely what you have done.

I approve the PR, it would be nice if you could include more documentation on the main methods introduced in this PR before merging, but not a blocker.

Issue #271, opened 2024-07-25
Construct scalable observable


nelimee:
**Is your feature request related to a problem? Please describe.**
#270 is concerned about finding all valid observable in a graph, this issue is about transforming the obtained graph information to a data-structure that can represent an observable (i.e., a list of measurements) that can scale with the templates.

**Describe the solution you'd like**
As explained in #270, the input for the function created to solve this issue is not fixed and will have to be discussed and agreed upon with people working on #270. 
The output format is not fixed either and will need to be easy to use for quantum circuit generation. This might require some changes in the current circuit-generation code, which is not a problem per se, but an output format that require little to no change in the current code will be preferred.

inmzhang:
Closed by mistake. I'll work on this.

PR #296, opened 2024-08-09
Introduce `AbstractObservable` that represents the location of measurements that should be included in a observable.

inmzhang:
Relates to #271.

I'm no sure whether the `AbstractObservable` representation is convenient for people working on circuits generation. But in principle, this simple representation should include all the information to derive a scalable observable that consists of the measurements.

`AbstractObservable` has two attributes:
1. `top_lines: frozenset[Cube | Pipe]`: the measurements along a line at the top of these blocks should be included in the observable.  
2. `bottom_regions: frozenset[Pipe]`: the stabilizer measurements within a region at the bottom of these block should be included in the observable.

I did not explicitly include the measurement basis or line direction here, because they could be inferred easily from the block type.

Any opinions?


src/tqec/sketchup/block_graph.py
```diff
@@ -562,3 +584,45 @@ def display(
...
+            if g.num_nodes == 1:
+                cube = ty.cast(Cube, self.get_cube(g.nodes[0].position))
+                abstract_observables.append(
+                    AbstractObservable(frozenset({cube}), frozenset())
+                )
+                continue
+            top_lines: set[Cube | Pipe] = set()
+            bottom_regions: set[Pipe] = set()
+            for edge in g.edges:
+                pipe = ty.cast(Pipe, self.get_pipe(edge.u.position, edge.v.position))
```
nelimee, line 607: Because having a `None` here should not be possible, I would add an assert instead of a typing cast that has no runtime implication. 
```suggestion
                pipe = self.get_pipe(edge.u.position, edge.v.position)
                assert pipe is not None, f"{edge} is in the graph and should be associated with a Pipe instance."
```

src/tqec/sketchup/block_graph.py
```diff
@@ -562,3 +584,45 @@ def display(
...
+                u, v = pipe.u, pipe.v
+                if pipe.direction == Direction3D.Z:
+                    if is_measured(v):
+                        top_lines.add(v)
+                    continue
```
nelimee, line 612: Is there a reason why this can only be `v` and not `u`? The underlying type is `nx.Graph` and not `nx.DiGraph`, so at first sight I expect edges to also be undirected (i.e., `u` and `v` are in an arbitrary direction).

src/tqec/sketchup/block_graph.py
```diff
@@ -562,3 +584,45 @@ def display(
...
+        self.check_validity(allow_virtual_node=False)
+        correlation_subgraphs = self.to_zx_graph().find_correlation_subgraphs()
+        abstract_observables: list[AbstractObservable] = []
+
+        def is_measured(cube: Cube) -> bool:
+            return self.get_pipe(cube.position, cube.position.shift_by(0, 0, 1)) is None
+
+        for g in correlation_subgraphs:
+            if g.num_nodes == 1:
+                cube = ty.cast(Cube, self.get_cube(g.nodes[0].position))
```
nelimee, line 599: Same as below:
Because having a None here should not be possible, I would add an assert instead of a typing cast that has no runtime implication.
```suggestion
                cube = self.get_cube(g.nodes[0].position)
                assert cube is not None, f"{g.nodes[0]} is in the graph and should be associated with a Cube instance."
```

src/tqec/sketchup/block_graph.py
```diff
@@ -562,3 +584,45 @@ def display(
...
+                    pipe.pipe_type.value.index(correlation_type_at_src)
+                )
+                if surface_attach_direction == Direction3D.Z:
+                    top_lines.add(pipe)
+                    if is_measured(u):
+                        top_lines.add(u)
+                    if is_measured(v):
+                        top_lines.add(v)
+                else:
+                    bottom_regions.add(pipe)
```
nelimee, line 624: I do not understand this part. If `surface_attach_direction` is `Direction.Z` that means that we have CUBE1 -> PIPE -> CUBE2 that are all on the same (x, y) coordinates but with increasing `z` right? In which case, why should the pipe be added?

nelimee (approved):
I left a few questions / comments. I might not be able to re-review before 2 weeks, and even though I did not test the code, I think that it is fine to merge if you tested it yourself and it appears to work. I'll approve to not block that PR during my holidays, but would love to understand the code better before merging.

src/tqec/sketchup/block_graph.py
inmzhang: The pipe's `u`/`v` is always sorted by the position so that `u.position < v.position` in `Pipe.__post_init__` method.

src/tqec/sketchup/block_graph.py
inmzhang: No, here `surface_attach_direction` is `Direction.Z` means that the normal direction of the correlation surface is in the XY plane. Sorry for  the confusing name and I will try to improve that.

inmzhang:
Thank you for the review! I'll keep this PR open for a while to see if anyone who wants to work on the circuit generation part has questions about this representation. Enjoy your holidays!

inmzhang:
I will be taking a two-week vacation to go traveling. During this period, I will keep this PR open.

nelimee:
> Hi Adrien @nelimee, do you need to review this PR again before I merge it?

I will do a quick check again and tell you in less than 30 minutes :)

nelimee (approved):
LGTM

Issue #294, opened 2024-08-07
`BlockGraph` -> `Computation`


nelimee:
**Is your feature request related to a problem? Please describe.**
There is currently a missing piece in the chain of data-structures we use to go from a `.dae` file to a `stim` circuit: we cannot build a `Computation` instance from a `BlockGraph` instance. 

**Describe the solution you'd like**
It is necessary for the tool to be able to build a `Computation` from a `BlockGraph` instance. 

**Additional context**
It may be interesting for #287 to change the data stored in the `Computation` data-structure and also include a graph-like representation of the blocks, to avoid having to re-compute block neighbouring information when calling `Computation.to_circuit`. As such, in accordance with the person on #287, it is entirely possible to re-write the `Computation` class from scratch.

Gistbatch:
I'll have a look at this!

PR #298, opened 2024-08-23
Graph to computation

### LINKED ISSUE #287: Fix `Block` merging
## Describe the bug
The current `Computation.to_circuit` method only has a placeholder implementation that does not work as expected. A few issues:
- the block positions are incorrectly used (currently, the Z component of the position is a number of moments after which the circuit should be scheduled),
- the block boundaries are not replaced when they are connected using connectors.

```py
from copy import deepcopy

from tqec.block.block import Computation, Position3D
from tqec.block.library.closed import zxz_block
from tqec.block.library.open import zxo_block
from tqec.templates.scale import LinearFunction

dim = LinearFunction(2, 0)
zxz = zxz_block(dim)
computation = Computation(
    {
        Position3D(0, 0, 0): deepcopy(zxz),
        Position3D(0, 0, 1): zxo_block(dim),
        Position3D(0, 0, 2): deepcopy(zxz),
    }
)

circuit = computation.to_circuit()
```
raises 
```
Traceback (most recent call last):
  File "/workspaces/tqec/main.py", line 18, in <module>
    circuit = computation.to_circuit()
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/tqec/src/tqec/block/block.py", line 318, in to_circuit
    return merge_scheduled_circuits(instantiated_scheduled_blocks)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/tqec/src/tqec/circuit/schedule.py", line 548, in merge_scheduled_circuits
    all_moments.append(cirq.Moment(*non_duplicated_operations))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vscode/.local/lib/python3.12/site-packages/cirq/circuits/moment.py", line 112, in __init__
    raise ValueError(f'Overlapping operations: {self.operations}')
ValueError: Overlapping operations: (cirq.H(cirq.GridQubit(0, 4)), cirq.H(cirq.GridQubit(0, 8)), cirq.H(cirq.GridQubit(2, 2)), cirq.H(cirq.GridQubit(2, 6)), cirq.H(cirq.GridQubit(4, 4)), cirq.H(cirq.GridQubit(4, 8)), cirq.H(cirq.GridQubit(6, 2)), cirq.H(cirq.GridQubit(6, 6)), cirq.H(cirq.GridQubit(8, 4)), cirq.H(cirq.GridQubit(8, 8)), cirq.H(cirq.GridQubit(10, 2)), cirq.H(cirq.GridQubit(10, 6)), cirq.ResetChannel()(cirq.GridQubit(4, 6)), cirq.ResetChannel()(cirq.GridQubit(6, 6)), cirq.ResetChannel()(cirq.GridQubit(2, 8)), cirq.ResetChannel()(cirq.GridQubit(4, 2)), cirq.ResetChannel()(cirq.GridQubit(4, 10)), cirq.ResetChannel()(cirq.GridQubit(6, 2)), cirq.ResetChannel()(cirq.GridQubit(2, 0)), cirq.ResetChannel()(cirq.GridQubit(6, 0)), cirq.ResetChannel()(cirq.GridQubit(0, 4)), cirq.ResetChannel()(cirq.GridQubit(2, 2)), cirq.ResetChannel()(cirq.GridQubit(6, 8)), cirq.ResetChannel()(cirq.GridQubit(8, 2)), cirq.ResetChannel()(cirq.GridQubit(4, 8)), cirq.ResetChannel()(cirq.GridQubit(10, 2)), cirq.ResetChannel()(cirq.GridQubit(8, 10)), cirq.ResetChannel()(cirq.GridQubit(2, 4)), cirq.ResetChannel()(cirq.GridQubit(8, 4)), cirq.ResetChannel()(cirq.GridQubit(4, 4)), cirq.ResetChannel()(cirq.GridQubit(0, 8)), cirq.ResetChannel()(cirq.GridQubit(10, 6)), cirq.ResetChannel()(cirq.GridQubit(2, 6)), cirq.ResetChannel()(cirq.GridQubit(8, 6)), cirq.ResetChannel()(cirq.GridQubit(8, 8)), cirq.ResetChannel()(cirq.GridQubit(6, 4)))
```
because the gates in the first moment of the first round of the connector are scheduled with the gates of the second moment of the first round of the block in `(0, 0, 0)`.

Devising the number of moments and hard-coding that in the `Position3D` instances is not a solution because it would remove any possibility to automatically replace boundaries of blocks with connectors.

## Expected behavior
The interface should stay relatively stable with respect to the current interface, but the internals of `to_circuit` should take care of correctly generating the quantum circuit. 

## Additional context
It might be easier to also include in the `Computation` class a graph-like data-structure to have an easy access to the information "block A is connected to connector B".

We might also require some pre-conditions on the input blocks.

Gistbatch:
I started to work on #294.
To understand what we need, I tried to fix #287 first.
This also contains a fix for #293.

For me, it is not immediately clear how to best proceed.
I implemented the happy case for merging circuits.
So far, this works based on the assumption that each block has unit size and touches the next one.

**First Question:** What happens if we have a gap in the blocks? e.g.
```python
    computation = Computation(
        {
            Position3D(0, 0, 0): deepcopy(zxz),
            Position3D(0, 0, 2): zxo_block(dim),
            Position3D(0, 0, 3): deepcopy(zxz),
        }
    )
```
How are we supposed to figure out which block to extend? I suggest extending the lower block unless it is an initialization layer.

To tackle the problem of conversion, I decided to implement a graph. This helps determine which blocks are supposed to be connected but still requires the positional information.
For each edge, there are then two (three) cases:
1. only temporal movement which is solved by the above
2. spatial movement
3. (combined movement), which is out of scope for now.

When solving the second part, there needs to be a block filling the gap between two other simultaneous blocks.

**Second Question:** What is supposed to happen when the blocks don't align temporally due to shifts in *1*? What if they have a different number of moments?


Gistbatch:
From a follow-up conversation with Austin:
**First Question:** What happens if we have a gap in the blocks?
We assume the user intends this. Gaps are empty circuits of the *correct* duration.

**Second Question:** What is supposed to happen when the blocks don't align temporally due to shifts in 1? What if they have a different number of moments?
We again assume the user intends this.
For now, we assume blocks have well-defined sizes; mismatches shouldn't happen.
If there is an unexpected mismatch, we *panic*.

src/tqec/block/block.py
```diff
@@ -266,23 +280,30 @@ def instantiate(self) -> cirq.Circuit:
...
     The computation is represented by a mapping from a position to the
     computational block whose origin is located at that position.
     """
 
-    blocks: dict[Position3D, ComputationBlock]
+    def __init__(self, blocks: dict[Position3D, ComputationBlock]) -> None:
+        self.blocks = blocks
+        self._graph = nx.Graph()
+        for position, block in blocks.items():
+            self._graph.add_node(position, **{_COMPUTATION_DATA_KEY: block})
```
nelimee, line 306: I can see why we will end up needing a graph-like structure, but for the moment you are not linking any node (i.e., the graph is only a collection of nodes without any edge) and not using the graph either, is that right?

src/tqec/block/enums.py
```diff
@@ -20,7 +20,7 @@ def to_template_sides(self) -> list[TemplateSide]:
                 TemplateSide.RIGHT,
                 TemplateSide.BOTTOM_RIGHT,
             ]
-        elif self == BlockDimension.Y:
+        if self == BlockDimension.Y:
```
nelimee, line 23: Not sure about the necessity of changing that to `if`. I think `elif` better shows the programmer intent, but `if` also works.

src/tqec/block/block.py
```diff
@@ -295,20 +316,22 @@ def to_circuit(self) -> cirq.Circuit:
...
-
             spatially_shifted_circuit = block.instantiate().transform_qubits(
-                shift_qubits
+                partial(_shift_qubits, position)
             )
             instantiated_scheduled_blocks.append(
-                ScheduledCircuit(spatially_shifted_circuit, position.z)
+                ScheduledCircuit(spatially_shifted_circuit, depth)
             )
+            depth += block.depth
```
nelimee, line 327: This will need to be improved as it will break for any computation containing more than 1 qubit. I think the best way of getting the depth each block should be located would be to perform an exploration of the graph-like representation of the computation, starting from the `t=0` side, compute the depth by exploring the graph, and raise in the process if any incompatibility is found. Something very similar to the `ComposedTemplate` internals, using a DFS exploration, would work in my opinion.

nelimee (commented):
The current direction only works for a reduced subset of computations and will need to be improved.

inmzhang (commented):
Pre-commit needs to be installed to check against the hooks.

I suggest to make the `Computation` rely on a `BlockGraph` explicitly. It will serve as the graph structure you need immediately and make the `Computation` satisfy the pre-conditions that a `BlockGraph` meets, i.e. the manhattan distance between two connected cubes is always one.

src/tqec/block/block.py
Gistbatch: Yes, this is a relict from my testing a bit.

src/tqec/block/enums.py
Gistbatch: It's a suggested change by pylint.

Gistbatch:
> Pre-commit needs to be installed to check against the hooks.
> 
> I suggest to make the `Computation` rely on a `BlockGraph` explicitly. It will serve as the graph structure you need immediately and make the `Computation` satisfy the pre-conditions that a `BlockGraph` meets, i.e. the manhattan distance between two connected cubes is always one.

You mean in the sense of `computation = Computation(block_graph)`?

src/tqec/block/block.py
Gistbatch: This is similar to what I had in mind; I just started with the happy case.
Could two qubits have cubes of different sizes in a case we should be able to treat? I assumed that if we had the same code, each repetition would have the same number of moments. Unless the user specifies a case where we have a different number of repetitions, the blocks should be all equal. If they are not equal I would then raise a warning to the user.

nelimee:
> > Pre-commit needs to be installed to check against the hooks.
> > I suggest to make the `Computation` rely on a `BlockGraph` explicitly. It will serve as the graph structure you need immediately and make the `Computation` satisfy the pre-conditions that a `BlockGraph` meets, i.e. the manhattan distance between two connected cubes is always one.
> 
> You mean in the sense of `computation = Computation(block_graph)`?

I would rather guess something like `computation = Computation.from_block_graph(block_graph)` if we also want to be able to construct a `Computation` from other inputs, but this is really close.

src/tqec/block/block.py
inmzhang: Is there use cases where two cubes can have any inconsistent dimension? In my mind, all the computation should admit an atomic cube definition, at least for our current workflow.

inmzhang:
For the purpose of MVP, I think it’s fine to make things simple. Then it’s easier to assume strong coupling between the two data structures and assume no other input formats for now.

afowler:
Even beyond the MVP, the goal is to manage the complexity of building the
circuits, so unless we are absolutely forced to, the goal will be to have
consistent dimensions and blocks that fit together neatly and simply.

Gistbatch:
From the last discussion, we decided that the `Computation` itself is not necessary as the concept is already expressed by the `BlockGraph`.
@nelimee what's your suggestion on progressing? Should I repurpose this PR to write a `BlockGraph` -> `Circuit` functionality? Or should I clean up and we just cherry-pick the stuff we need from this?

Gistbatch:
I had the same idea in mind. I'm not sure how easy a spatial edge will be but I'll see :+1:

Gistbatch:
@nelimee I've added a replacement logic for two cases:
- **temporal** edges by replacing the templates of the `initial` and `final` plaquettes with the ones from the `repeating` section
- **spatial** edges: I added a *hardcoded* replacement logic for the `DenseQubitSquareTemplate`; I don't see a clear way how to do this without knowing the layout.

At the moment the corners aren't replaced at all. As far as I understand this is only important if two spatial pipes are connected to the same cube.
I don't see an alternative to hardcoding this behavior where every corner has one (if we don't allow rotation) predefined block that can be turned on or off.

Also, I'm not happy with how my code looks, so any feedback is appreciated.

Gistbatch:
I also realized that the Hadamard pipes are missing. Is this as simple as adding a different substitution rule or do we need to insert another layer? @afowler (I was using slides 12/13 as a reference from the SketchUp to Sims slide deck)

afowler:
Hi Philipp,

Hadamard pipes are indeed just different substitution rules. For a temporal
Hadamard, the data qubits need Hadamard gates in addition to the
substitution of ordinary rounds of stabilizer measurement. These Hadamard
gates can go in many places, but close the the middle of the two
substituted layers would be nice. Concurrent with either the first layer's
measurement or the second layer's initialization would be fine.

For spatial Hadamard, you still substitute two cube walls, but one of these
walls will have XZ and ZX stabilizers. Stabilizers that are X on one side
and Z on the other. This changes the color of the stabilizer checkerboard
and implements spatial Hadamard. Please let me know if the above makes
sense, happy to expand.

Best,
Austin.

Gistbatch:
@nelimee 
 >     2. Perform a DFS-like exploration of the `BlockGraph` instance to explore all the edges once. For each `start_node, edge, end_node`:
Is this necessary? We can just iterate over all pipes, the order is not important I think
   
>        * If the `edge` is in the temporal direction:
I did most of this directly without going through the process of instantiating a block for the pipe, do you think I should switch to your recommendations?

> In practice, because the templates have to match (else it is an error), "merging" an edge should simply amount to replace some plaquettes in the `StandardComputationBlock` definition.

I think we should have validation as a separate issue.

> Now about spatial replacement, you can use the interface provided by `ComposedTemplateWithSides` (and add methods to it if needed). The only non-trivial part in my mind is the following: we have plaquette indices from the cube (obtained with `cube.template.get_plaquette_indices_on_sides(sides)`) and we have plaquette indices from the pipe (obtained with `pipe.template.instantiate()`, how to properly know which index from the pipe should replace each index from the cube. We have to find a mapping from plaquette indices from the cube to plaquette indices from the pipe. I think this can be done by:

I just *hardcoded* this which, if I'm not mistaken, should work. This approach is only limited by the hadamards.

nelimee:
I'll have a deeper look today.

src/tqec/block/block.py
```diff
@@ -254,6 +266,48 @@ def replace_boundary_with_empty_plaquettes(
             self.plaquettes.with_updated_plaquettes(plaquettes_to_replace),
         )
 
+    def replace_boundary_plaquettes(
+        self, boundary: Direction3D, outgoing: bool = True
+    ) -> None:
+        """Mutating."""
+        if boundary == Direction3D.Z:
+            if self.repeating_plaquettes is None:
```
nelimee, line 274:
```suggestion
            if self.is_connector:
```

src/tqec/block/block.py
```diff
@@ -266,49 +320,24 @@ def instantiate(self) -> cirq.Circuit:
...
             )
-            instantiated_scheduled_blocks.append(
-                ScheduledCircuit(spatially_shifted_circuit, position.z)
-            )
-
-        return merge_scheduled_circuits(instantiated_scheduled_blocks)
+            self.plaquettes.final_plaquettes[index] = self.plaquettes.final_plaquettes[
+                replacement_index
+            ]
+            if self.repeating_plaquettes is not None:
```
nelimee, line 340:
```suggestion
            if self.plaquettes.repeating_plaquettes is not None:
```
to silence typing error about `self.plaquettes.repeating_plaquettes` that may be `None`.

src/tqec/templates/constructions/qubit.py
```diff
@@ -26,6 +26,10 @@ class ComposedTemplateWithSides(ComposedTemplate):
     def get_plaquette_indices_on_sides(self, sides: list[TemplateSide]) -> list[int]:
         pass
 
+    @abstractmethod
+    def get_corresponding_mid_for_side(self, side_index: int) -> int:
+        pass
+
```
nelimee, line 32: I needed a long time to actually understand what this did. Documentation would be welcomed.

src/tqec/sketchup/block_graph.py
```diff
@@ -551,14 +580,117 @@ def display(
...
+        """
+        blocks: dict[Position3D, StandardComputationBlock] = {}
+        for cube in self.cubes:
+            blocks[cube.position] = cube_to_block(cube, dimension)
+        for pipe in self.pipes:
+            u_computation, v_computation = (
+                blocks[pipe.u.position],
+                blocks[pipe.v.position],
+            )
+            u_computation.replace_boundary_plaquettes(pipe.direction)
```
nelimee, line 617:
```suggestion
            u_computation.replace_boundary_plaquettes(pipe.direction, outgoing=True)
```
for symmetry with the line just below.

src/tqec/sketchup/block_graph.py
```diff
@@ -551,14 +580,117 @@ def display(
...
+            v_computation.replace_boundary_plaquettes(pipe.direction, outgoing=False)
+
+        instantiated_scheduled_blocks: list[ScheduledCircuit] = []
+        depth = 0
+        for position, block in blocks.items():
+            spatially_shifted_circuit = block.instantiate().transform_qubits(
+                partial(_shift_qubits, position)
+            )
```
nelimee, line 625: I think this is broken, there are two different coordinate systems here:

1. the coordinate system used to place cubes and pipes,
2. the physical qubits coordinate system.

Here `position` is in the first coordinate system, but `_shift_qubits` is expecting the second coordinate system. I think, each entry in position `position` should be multiplied by the number of qubits needed by each block spatial dimension (`2k + 2`? I might be wrong on the exact number.)

src/tqec/sketchup/block_graph.py
```diff
@@ -551,14 +580,117 @@ def display(
...
+            instantiated_scheduled_blocks.append(
+                ScheduledCircuit(spatially_shifted_circuit, depth)
+            )
+            depth += block.depth
```
nelimee, line 629: Again, this is broken for any circuit that has more than 1 static logical qubit.

nelimee (commented):
Right now you are not using the `pipes` at all and, if I understood correctly, you "extend" the block sides by using plaquettes defined within the block itself. 
I think you should use the pipes instead for 2 reasons:

1. I am not sure that the substitution rule you use (by using existing plaquettes in the block being connected) is always valid (I would need to check that in more details later).
2. I am sure that this does not work for pipes containing a Hadamard layer.

Did I miss something in my statements above?

src/tqec/templates/constructions/qubit.py
Gistbatch: I'm not sure if this is the correct way to do it, I think getting the plaquettes from the pipe is the better idea.

src/tqec/sketchup/block_graph.py
Gistbatch: I didn't touch this anymore. First, I wanted to get the computation graph right. Once I have the graph, I can go back to fixing this

afowler:
Indeed, the substitution rule is not always valid. A Hadamard pipe is a
good example. Another example is a pipe connecting to the top of a temporal
3 or 4 way junction. In that situation you wouldn't have the necessary
boundaries.

Best,
Austin.

src/tqec/block/block.py
```diff
@@ -327,17 +332,21 @@ def instantiate_without_boundary(self, boundary: Direction3D) -> cirq.Circuit:
...
-            self.plaquettes.initial_plaquettes[index] = (
-                self.plaquettes.initial_plaquettes[replacement_index]
-            )
-            self.plaquettes.final_plaquettes[index] = self.plaquettes.final_plaquettes[
-                replacement_index
+            self.plaquettes.initial_plaquettes[index] = replacement.initial_plaquettes[
+                index
+            ]
+            self.plaquettes.final_plaquettes[index] = replacement.initial_plaquettes[
+                index
```
Gistbatch, line 344: @nelimee this is how I would use the computation blocks from the pipes.
This works based on the assumption that the replacement block has the same indices on the border as the original block, which is not true at the moment.
Do you think it is possible to have `QubitHorizontalBorders` look like
```
1  5  6  5  6 2
3 13 14 13 14 4
```
instead?

nelimee:
Sorry for the 2 commits above, I pushed by mistake a tentative commit (that was meant to be exposed as another PR) and so removed it by force-pushing. See https://github.com/Gistbatch/tqec/pull/1 for the tentative.

nelimee:
Hi @Gistbatch, any news on that PR?

Gistbatch:
@nelimee  I reused your solution, although I had to swap X and Y, I'm not sure if I made a mistake in the test creation.
Now the circuit generation is left to fix. Should I do this in this PR or do you plan to move this somewhere else?

nelimee:
> @nelimee  I reused your solution, although I had to swap X and Y, I'm not sure if I made a mistake in the test creation.
> Now the circuit generation is left to fix. Should I do this in this PR or do you plan to move this somewhere else?

I will probably be able to get e ie that in more details tomorrow, but I think you can do everything in this PR.

Gistbatch:
closed in favor of #321

Gistbatch:
closed in favor of #321, see #325

PR #295, opened 2024-08-07
Improve CI and include coverage data

nelimee:
This PR aims at improving the CI build times and including an automatic report on code coverage.

PR #303, opened 2024-08-27
Improve CI and include coverage data

nelimee:
Replaces #295. The only differences are:
- this PR is from a branch within the `QCHacker` group (and not on the `nelimee/tqec` fork),
- the last 3 commits are removed.

Adding a message to PRs require `write` permission to that PR, which is only given when the branch to be merged is from the same repository as the base branch (here, `QCHackers/tqec`). Trying to circumvent that with `workflow_run` seems too complex for such a low-priority issue, so let's live with that restriction for the moment.

TLDR: PRs from branches in any repository other than `QCHackers/tqec` will fail the CI job about coverage, this is "expected" and is not a code issue but rather a GitHub permission issue.


inmzhang (approved):
This looks great!

KabirDubey:
> TLDR: PRs from branches in any repository other than `QCHackers/tqec` will fail the CI job about coverage, this is "expected" and is not a code issue but rather a GitHub permission issue.

Does this mean that from now on, contributors on should work on branches from the `QCHackers` group instead of their own forks? If so, may implicate step 4 of the [contributions file](https://github.com/QCHackers/tqec/blob/main/CONTRIBUTING.md)

PR #307, opened 2024-08-27
Bump minimum version to 3.10

### LINKED ISSUE #306: Bump minimum supported version of Python to 3.10
#305 is failing because the last version of `cirq` supporting Python 3.9 is `<1.4` (i.e., `1.3.0`, see [`cirq 1.4.0` release notes](https://github.com/quantumlib/Cirq/releases/tag/v1.4.0)) and that version does not include the type hints needed to resolve correctly the result of `cirq.GridQubit < cirq.GridQubit`.

Make the `tqec` library only compatible with Python `>=3.10` as `cirq` is still a crucial dependency that we won't be able to remove in the short term.

nelimee:
Fixes #306

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     needs: pre-commit-ci
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
     steps:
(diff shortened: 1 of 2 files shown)


inmzhang (approved):
I remember that the reason why we support Python3.9 is to match `cirq`, so it makes sense to sync with its requirement. Then finally we can leave behind lots of `from __future__ import annotation` and `match` expression is on the way too!

PR #313, opened 2024-09-02
Fix off-by-1 error in Template display

nelimee:
This commit fixes an off-by-1 error in an indexing. 

The `plaquette_indices` tuple does **NOT** include the `0` plaquette, so the index should be decreased by 1.


KabirDubey (commented):
LGTM

KabirDubey (approved):

PR #297, opened 2024-08-21
Usage

Gistbatch:
We need instructions on how to use our library.
This is the first draft of a `Usage` file.

PR #316, opened 2024-09-05
Fix the translation to stim issue

nelimee:
This should fix the issue raised in #297. I could not test the fix, so not merging for the moment. 
Maybe @Gistbatch you can try on the specific file you mentioned?


nelimee:
Note that coverage failing is expected, but will prevent anyone except administrators merging this. This is unfortunate and will have to be fixed soon.

Gistbatch (approved):

Gistbatch:
I cherry-picked the commit into the computation branch and it was working!

Issue #71, opened 2024-01-24
Improve the function to display `TemplateOrchestrator` as svg


inmzhang:
As discussed in #70, there can be some improvement ideas on the `tqec.display.display_templates_svg` function, which includes the following or more:
1. Display the relationships between templates (corner pinning).
2. Find a way to display the "scalable" property of the outer outer edges. For example, outer edges that will scale when scale_to is called are depicted as double-ended arrows whereas those that does not scale are depicted with inward-arrows.

This may also relate to issue #67.

nelimee:
Better idea about point number 2.

The problem of what I proposed first in 2. is that the arrows of different edges might intersect, which will lead to an hard-to-read representation, which is the opposite of the main goal.
Basically, any visualisation should be obvious at first glance if possible.
In this regard, here are two replacement proposals for 2. :
1. Make "scalable" and "non-scalable" outer edges a different colour.
2. Make "scalable" outer edges dotted and "non-scalable" outer edges plain.

Other ideas are of course welcome.

inmzhang:
And we should also note that the current implementation relies on the assumption that all the plaquette is of the same square shape. According to #34, this function should also change quite a lot once the assumption is lifted.

inmzhang:
> And we should also note that the current implementation relies on the assumption that all the plaquette is of the same square shape. According to https://github.com/QCHackers/tqec/issues/34, this function should also change quite a lot once the assumption is lifted.

As discussed in #34 and #80, the plaquettes should all be rectangular. This constraint implies that the `display_templates_svg` function currently works as expected without any modifications, or with only slight adjustments, once issue #80 is merged.

In cases where `increment_x` is not equal to `increment_y`, and if there is a need to display the plaquettes[^1] as rectangles, some minor code modifications will be required to ensure that displayed as intended.

[^1]: Strictly speaking, the inner grid is not the boundary of the plaquettes in the sense that the shape of a plaquette is useless information. It just shows the tiling which is determined by the `increment` property of the templates.

inmzhang:
This issue is related to #107 and #108, we can rename `display_templates_svg` as `display_template_svg` and accept a `template: Template` as the argument. The implementation needs to be changed to accout for both `AtomicTemplate` and composed `TemplateOrchestrator`.

nelimee:
@smburdick might be nice for UF hack? (I know, this is a little bit late, I am sorry)

giangiac:
[...]

Issue #274, opened 2024-07-25
Double-check AlternatingCornerSquareTemplate implementation


nelimee:
## Describe the bug
The [`AlternatingCornerSquareTemplate`](https://github.com/QCHackers/tqec/blob/0d91fb530934c735cf31392558417ff0c81162b6/src/tqec/templates/atomic/square.py#L69C7-L69C38) implementation does not adhere to the standard that the first plaquette provided to the `instantiate` method is at the top-right corner of an atomic template.

This "standard" of having the first plaquette provided to `instantiate` as the top-left plaquette is a good way to avoid issues from plaquettes not being placed as expected by the user. I think it has never been explicitly written anywhere in the code, but at the moment this "standard" is implicitly followed by all the atomic templates (with the exception of `AlternatingCornerSquareTemplate`).

Now, the question is the following: should we try to enforce that standard even with the `AlternatingCornerSquareTemplate` or not?

The issue with trying to enforce the standard is that, for convenience, `AlternatingCornerSquareTemplate` is implementing a family of 4 different templates (the 4 different possible location for the corner), and each member of this family of template has a different plaquette in the top-right corner.

```
UPPER_LEFT:
5  4  3  4
2  3  4  3
1  2  3  4
2  1  2  3

UPPER_RIGHT:
4  3  4  5
3  4  3  2
4  3  2  1
3  2  1  2

LOWER_LEFT:
2  1  2  3
1  2  3  4
2  3  4  3
5  4  3  4

LOWER_RIGHT:
3  2  1  2
4  3  2  1
3  4  3  2
4  3  4  5
```

So we have two solutions:
1. _(currently implemented)_ do not enforce the "standard" on this atomic template and add documentation to explicitly document that.
2. enforce the "standard", renumbering the plaquettes of each member of the family, which might be even more confusing than 1.

Any thoughts / opinions?

Gistbatch:
Maybe it can have a function `setCorner(value,position)` which explicitly sets the corner and enforce the standard?

PR #317, opened 2024-09-06
Template structure simplification

nelimee:
It has been a few weeks I am thinking about that, and a discussion with Kabir a few days ago finally convinced me that refactoring the Template hierarchy was a good idea.

The main reason for that refactor is the unneeded complexity of the old `Template` code: we only need to represent "simple" templates such as logical qubits and their sides. Being able to represent arbitrary templates using the `ComposedTemplate` class was nice, but not a requirement. With the need to query templates about their middle lines, sides, and likely other properties in the future that were not easily implementable (or even correctly defined) for all the possible `ComposedTemplate` instance, removing that intermediary seemed like a necessity.

This (rather large, sorry) commit aims at greatly simplifying the `Template`-related module. The interfaces are left unchanged, which reduces by a lot the possibility of introducing errors with this refactor and reduces the number of changes.

Note that this is only a draft pull request, I **really** want other maintainer opinions on that before even considering it mergeable. 

Note also that both #71 and #274 will be impacted by this refactor.

Finally, this is also an experiment to test another idea for automatic detector computation. I will use that PR as a base for another PR on that, and see if it turns out to be a nice approach.


inmzhang (commented):
The removal of `ComposedTemplate` indeed simplifies things a lot. And currently I haven't seen any bottlenecks of this simplification considering the current workflow. A few questions in my mind:
1. How do you plan to construct more complex templates like the corners/junctions? I assume that for each "composed template" now we need to implement a rather large `instantiate` method to specify all the plaquette indices.
2. I remembered that the complexity of `PiecewiseLinearFunction` mainly comes from the usage in `ComposedTemplate`. If it does not exist anymore, is it possible to simplify the `scale.py`/`interval.py` as well?

Gistbatch:
I'm not sure what kind of complex structures we would need, but I think both corners and junctions can be represented as a block graph?

inmzhang:
Yes they can be represented by block graph. But when it comes to instantiate them into circuits, there is the possibility that we need to change the underlying template for more complex circuits to fit in. For example, corner/junctions in space need special care about the plaquettes arrangement for the circuit to be distance-preserving.

nelimee:
> 1. How do you plan to construct more complex templates like the corners/junctions? I assume that for each "composed template" now we need to implement a rather large `instantiate` method to specify all the plaquette indices.

The problem is the same with the current architecture. The main complexity is due to the center `2k x 2k` template that has a "less regular" pattern in the case of corner or junctions. Implementing a corner / junction with the `ComposedTemplate` API still needs to specify that `2k x 2k` template by hand, which is the main part of complexity. 
In other words, because the border implementation is not really complex (and can still be factored out in a separate function), I do not see any added complexity by this approach, the same things have to be done anyway in both approaches.



> I remembered that the complexity of `PiecewiseLinearFunction` mainly comes from the usage in `ComposedTemplate`. If it does not exist anymore, is it possible to simplify the `scale.py`/`interval.py` as well?

This is very much true, I will try to remove that now.

nelimee:
In order to demonstrate that it can be done, I implemented the 4-way junction as a new template. Note that 2/3-way junctions are not implemented yet, and there are enough questions we need to collectively answer on them to justify another PR just for that.
I think this PR is ready, but before merging I would like @KabirDubey to react here as I know he was working on #71, which is greatly impacted by this PR.

nelimee:
Now that I included more tests I think this PR is ready to be reviewed. As soon as it is accepted, I will merge this one and just after merge #318. 
Then, I will need to fix any potential merge conflict with #305 and merge #305.
The next steps will be:
1. Finish #319 and merge it,
2. Start and finish a translation method from `AbstractObservable` instances to actual `Observable`, or even better to a new `ScalingObservable`.

inmzhang (approved):
I'll approve this without carefully checking the 4-way junction implementation. But I regard it is more of a showcase of the `Template` and the correctness does not matter so much for now. And it just looks correct...

nelimee:
This is also how I see it: a showcase that will need to be checked and improved. Thanks for the quick approval !

src/tqec/templates/display.py
KabirDubey, line 1: Do we need new display functions for the refactored `Template`? I can work on this (make a new issue and close issue #71). I'm still in the process of understanding the refactoring.

nelimee: > Do we need new display functions for the refactored `Template`?

I am not sure. I think it depends nearly entirely on whether or not we plan to eventually have a frontend that will need to visually represent templates. Another point is that it would be nice to have nice looking images for the documentation. 

So let's say that we might need it, but it is clearly not sure and not a priority. If you still want to work on it, please feel free to do so :)

PR #319, opened 2024-09-11
[WIP] Automatic and spatially-local detector computation

### LINKED ISSUE #259: Integrate automatic detector computation into `tqec`
#249 aims at including in the code base a generic way to automatically compute detectors from a given quantum circuit checking some pre-conditions. The implementation in #249 is voluntarily disconnected from the main `tqec` library, to be able to extract it as an independent library if we think this might be useful to the QEC community. But this disconnection implies that, for the moment, automatic detector computation does not use the additional knowledge provided by the `tqec` library.

In particular, if #249 is merged without significant changes, the workflow that will need to be followed is the following:
1. Create the necessary plaquettes and templates.
2. Assemble the created plaquettes and templates into a structure representing the QEC computation.
3. For each code distance `d` and noise model `e`:
    1. Create the quantum circuit representing the QEC computation.
    2. Compute detectors over the **whole quantum circuit** and add the computed detectors to the circuit.

Even though detector computation is quite efficient most of the time, the computation might get very expensive on large code distance. 

In practice, it might not be needed to perform detector computation at step 3.2. over the whole circuit. Using knowledge on the code locality (e.g., detectors can only include measurements that are within a ball of radius `X` in space around a central point, and maximum `Y` temporal steps before the current step), we might be able to extract from the representation obtained after step 2. a limited number of sub-circuits (subsets of templates with plaquettes) that:
1. are centered on a specific plaquette (i.e., one index particular location in a template that will be replaced with a plaquette),
2. for each sub-circuit, only include plaquettes that are at a distance `X` from the center plaquette,
3. cover each plaquette in the template.

Basically, finding local (both in time and in space) regions that should contain at least one detector involving a defined central measurement, computing the detector for each of these regions, and re-using the computation when the same local region is found multiple times in the QEC computation.

nelimee:
This PR is based on #318.

This is still a work in progress as some changes from #305 are needed to complete this PR.

In the end, this PR should provide all the necessary tools to efficiently compute detectors when making the translation `BlockGraph` -> `cirq.Circuit` (see #298).

Fixes #259


inmzhang:
> I won't be available next week, so this PR might be stale until then. If anyone wants to have a look at it and finish it, please go ahead.

If I have time then, I will review the current stage of this PR and see whether I can move forward based on what you have done.

nelimee:
A few notes before my holidays:
- For the surface code, I am pretty sure we can set the Manhattan radius to `1` instead of the `2` I was using in my tests (and that might have slipped somewhere in the code).
- I think it would be beneficial to have a cache for detector computation, as we will likely perform the same task multiple times (splitting into subtemplates was the first step, having a cache might be the second step).
- If you have any question on the code, I might be able to answer. I won't be able to write code, but textual answers might be OK, so feel free to ask questions here, I will answer them when possible.

nelimee:
Found some time to work on it. Got it *nearly* working.

You can try it with 
```py
from pprint import pprint
import stimcirq

block = xzz_block()
block.scale_to(1) # Keep the list of detectors human-sized

print(stimcirq.cirq_circuit_to_stim_circuit(block.instantiate()))
detectors = block.compute_detectors()
for ds in detectors:
    print("=" * 80)
    pprint(ds)
```
This should output
```
QUBIT_COORDS(0, 4) 0
QUBIT_COORDS(1, 1) 1
QUBIT_COORDS(1, 3) 2
QUBIT_COORDS(1, 5) 3
QUBIT_COORDS(2, 0) 4
QUBIT_COORDS(2, 2) 5
QUBIT_COORDS(2, 4) 6
QUBIT_COORDS(3, 1) 7
QUBIT_COORDS(3, 3) 8
QUBIT_COORDS(3, 5) 9
QUBIT_COORDS(4, 2) 10
QUBIT_COORDS(4, 4) 11
QUBIT_COORDS(4, 6) 12
QUBIT_COORDS(5, 1) 13
QUBIT_COORDS(5, 3) 14
QUBIT_COORDS(5, 5) 15
QUBIT_COORDS(6, 2) 16
R 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
TICK
H 4 6 10 12
TICK
CX 1 5 6 2 8 11 10 7 12 9 13 16
TICK
CX 2 0 4 1 6 3 7 5 10 8 14 11
TICK
CX 2 5 6 8 9 11 10 13 12 15 14 16
TICK
CX 3 0 4 7 6 9 8 5 10 14 15 11
TICK
H 4 6 10 12
TICK
M 0 4 5 6 10 11 12 16
TICK
REPEAT 3 {
    R 0 4 5 6 10 11 12 16
    TICK
    H 4 6 10 12
    TICK
    CX 1 5 6 2 8 11 10 7 12 9 13 16
    TICK
    CX 2 0 4 1 6 3 7 5 10 8 14 11
    TICK
    CX 2 5 6 8 9 11 10 13 12 15 14 16
    TICK
    CX 3 0 4 7 6 9 8 5 10 14 15 11
    TICK
    H 4 6 10 12
    TICK
    M 0 4 5 6 10 11 12 16
    TICK
}
R 0 4 5 6 10 11 12 16
TICK
H 4 6 10 12
TICK
CX 1 5 6 2 8 11 10 7 12 9 13 16
TICK
CX 2 0 4 1 6 3 7 5 10 8 14 11
TICK
CX 2 5 6 8 9 11 10 13 12 15 14 16
TICK
CX 3 0 4 7 6 9 8 5 10 14 15 11
TICK
H 4 6 10 12
TICK
M 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
TICK
================================================================================
[Detector([Measurement(q(0, 4), -1)]),
 Detector([Measurement(q(2, 2), -1)]),
 Detector([Measurement(q(4, 4), -1)]),
 Detector([Measurement(q(6, 2), -1)])]
================================================================================
[Detector([Measurement(q(0, 4), -2), Measurement(q(0, 4), -1)]),
 Detector([Measurement(q(2, 0), -2), Measurement(q(2, 0), -1)]),
 Detector([Measurement(q(2, 2), -2), Measurement(q(2, 2), -1)]),
 Detector([Measurement(q(2, 4), -1), Measurement(q(2, 4), -2)]),
 Detector([Measurement(q(4, 2), -2), Measurement(q(4, 2), -1)]),
 Detector([Measurement(q(4, 4), -2), Measurement(q(4, 4), -1)]),
 Detector([Measurement(q(4, 6), -2), Measurement(q(4, 6), -1)]),
 Detector([Measurement(q(6, 2), -2), Measurement(q(6, 2), -1)])]
================================================================================
[Detector([Measurement(q(1, 1), -1), Measurement(q(2, 2), -2), Measurement(q(3, 1), -1), Measurement(q(3, 3), -1), Measurement(q(1, 3), -1)]),
 Detector([Measurement(q(1, 3), -1), Measurement(q(1, 5), -1), Measurement(q(0, 4), -2)]),
 Detector([Measurement(q(0, 4), -1), Measurement(q(0, 4), -2)]),
 Detector([Measurement(q(1, 1), -1), Measurement(q(2, 2), -2), Measurement(q(3, 1), -1), Measurement(q(3, 3), -1), Measurement(q(1, 3), -1)]),
 Detector([Measurement(q(2, 0), -2), Measurement(q(2, 0), -1)]),
 Detector([Measurement(q(5, 3), -1), Measurement(q(4, 4), -2), Measurement(q(3, 5), -1), Measurement(q(5, 5), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(3, 1), -1), Measurement(q(1, 1), -1), Measurement(q(2, 2), -2), Measurement(q(1, 3), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(1, 3), -1), Measurement(q(1, 5), -1), Measurement(q(0, 4), -2)]),
 Detector([Measurement(q(2, 2), -1), Measurement(q(2, 2), -2)]),
 Detector([Measurement(q(5, 3), -1), Measurement(q(4, 4), -2), Measurement(q(3, 5), -1), Measurement(q(5, 5), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(3, 1), -1), Measurement(q(1, 1), -1), Measurement(q(2, 2), -2), Measurement(q(1, 3), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(1, 3), -1), Measurement(q(1, 5), -1), Measurement(q(0, 4), -2)]),
 Detector([Measurement(q(2, 4), -2), Measurement(q(2, 4), -1)]),
 Detector([Measurement(q(5, 3), -1), Measurement(q(5, 1), -1), Measurement(q(6, 2), -2)]),
 Detector([Measurement(q(5, 3), -1), Measurement(q(4, 4), -2), Measurement(q(3, 5), -1), Measurement(q(5, 5), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(3, 1), -1), Measurement(q(1, 1), -1), Measurement(q(2, 2), -2), Measurement(q(1, 3), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(4, 2), -2), Measurement(q(4, 2), -1)]),
 Detector([Measurement(q(5, 3), -1), Measurement(q(5, 1), -1), Measurement(q(6, 2), -2)]),
 Detector([Measurement(q(5, 3), -1), Measurement(q(4, 4), -2), Measurement(q(3, 5), -1), Measurement(q(5, 5), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(3, 1), -1), Measurement(q(1, 1), -1), Measurement(q(2, 2), -2), Measurement(q(1, 3), -1), Measurement(q(3, 3), -1)]),
 Detector([Measurement(q(4, 4), -1), Measurement(q(4, 4), -2)]),
 Detector([Measurement(q(5, 5), -1), Measurement(q(3, 3), -1), Measurement(q(5, 3), -1), Measurement(q(4, 4), -2), Measurement(q(3, 5), -1)]),
 Detector([Measurement(q(4, 6), -1), Measurement(q(4, 6), -2)]),
 Detector([Measurement(q(5, 1), -1), Measurement(q(6, 2), -2), Measurement(q(5, 3), -1)]),
 Detector([Measurement(q(5, 5), -1), Measurement(q(3, 3), -1), Measurement(q(5, 3), -1), Measurement(q(4, 4), -2), Measurement(q(3, 5), -1)]),
 Detector([Measurement(q(6, 2), -1), Measurement(q(6, 2), -2)])]
```

1. Detectors in the initial layer are correct.
2. Detectors in the repeated layer are correct.
3. Detectors in the final layer are... not what I expected.

I do not know what's causing the behaviour on 3.

inmzhang:
Great, I'll try to figure out what's going on with the unexpected behavior and see whether I can fix it.

inmzhang:
> I do not know what's causing the behaviour on 3.

There were two problems with that:

1. When you filter out detectors in subtemplates, the plaquettes with data qubit measurements escape the predicate and duplicates in the different plaquettes.
2. The offset computation in `get_measurements_from_circuit` is wrong, so that the final round detectors presented itself as `Detector([Measurement(q(1, 3), -1), Measurement(q(1, 5), -1), Measurement(q(0, 4), -2)]),` instead of `Detector([Measurement(q(1, 3), -1), Measurement(q(1, 5), -1), Measurement(q(0, 4), -1)])`.

I have fixed the above two problems and currently the ouput is as expected.

Run:
```py
from pprint import pprint
import stimcirq
from tqec.block.library.closed import xzz_block

block = xzz_block()
block.scale_to(1) # Keep the list of detectors human-sized

detectors = block.compute_detectors(1)
for ds in detectors:
    print("=" * 80)
    pprint(ds)
```

Get:
```txt
================================================================================
[Detector([Measurement(q(0, 4), -1)]),
 Detector([Measurement(q(2, 2), -1)]),
 Detector([Measurement(q(4, 4), -1)]),
 Detector([Measurement(q(6, 2), -1)])]
================================================================================
[Detector([Measurement(q(0, 4), -2), Measurement(q(0, 4), -1)]),
 Detector([Measurement(q(2, 0), -1), Measurement(q(2, 0), -2)]),
 Detector([Measurement(q(2, 2), -2), Measurement(q(2, 2), -1)]),
 Detector([Measurement(q(2, 4), -2), Measurement(q(2, 4), -1)]),
 Detector([Measurement(q(4, 2), -2), Measurement(q(4, 2), -1)]),
 Detector([Measurement(q(4, 4), -2), Measurement(q(4, 4), -1)]),
 Detector([Measurement(q(4, 6), -1), Measurement(q(4, 6), -2)]),
 Detector([Measurement(q(6, 2), -1), Measurement(q(6, 2), -2)])]
================================================================================
[Detector([Measurement(q(1, 5), -1), Measurement(q(1, 3), -1), Measurement(q(0, 4), -1)]),
 Detector([Measurement(q(3, 3), -1), Measurement(q(1, 1), -1), Measurement(q(2, 2), -1), Measurement(q(3, 1), -1), Measurement(q(1, 3), -1)]),
 Detector([Measurement(q(5, 5), -1), Measurement(q(3, 3), -1), Measurement(q(5, 3), -1), Measurement(q(4, 4), -1), Measurement(q(3, 5), -1)]),
 Detector([Measurement(q(5, 1), -1), Measurement(q(6, 2), -1), Measurement(q(5, 3), -1)])]
```

I'll see in the next few days whether I can further make some improvements/refactors.

inmzhang:
> I think it would be beneficial to have a cache for detector computation, as we will likely perform the same task multiple times (splitting into subtemplates was the first step, having a cache might be the second step).

The matching happens at the level of multiple flows, which requires the knowledge of the entire sub-circuit. And the sub-circuit is unique for each sub-template. Currently there is no proper way to do the caching in my mind. Do you have any idea about this?

nelimee:
> > I think it would be beneficial to have a cache for detector computation, as we will likely perform the same task multiple times (splitting into subtemplates was the first step, having a cache might be the second step).
> 
> The matching happens at the level of multiple flows, which requires the knowledge of the entire sub-circuit. And the sub-circuit is unique for each sub-template. Currently there is no proper way to do the caching in my mind. Do you have any idea about this?

So for me there are two opportunities for caches here:

1. in the computation of subtemplates,
2. in the computation of detectors for each subtemplate.

For 1., I think this should be achieved by overriding the method for each Template subclass that we want to "optimise". We might have to "cache" in the sense that we will reuse

nelimee:
> > I think it would be beneficial to have a cache for detector computation, as we will likely perform the same task multiple times (splitting into subtemplates was the first step, having a cache might be the second step).
> 
> The matching happens at the level of multiple flows, which requires the knowledge of the entire sub-circuit. And the sub-circuit is unique for each sub-template. Currently there is no proper way to do the caching in my mind. Do you have any idea about this?

So I think we will eventually need 2 kind of caches:

1. A cache for subtemplate computation,
2. A cache for detector computation.

For subtemplates, the best would be to override the get_spatially_distinct_subtemplates in each of the subclasses of Template, but this will likely lead to very convoluted code, which is not the goal.
According to how we will use that method (a lot of QubitTemplate instances with the same value of k), it might be interesting to have a class-attached store to cache computations and re-use them.
Basically, this is just a class attribute dict[int, UniqueSubTemplates] that will be used as a cache.

Now for detectors, we can cache according to the subtemplate (i.e., a constant size 2-d array that, for the moment, is assumed to be small because the Pauli flows stay spatially localised) and the Plaquettes. Basically saying "if, locally, the situation is the same, then the detector should be the same". In other words, this aims at caching calls to compute_detectors_in_last_timestep.
Now the issue is to find a correct key for the cache. I made a try in one of my previous failed attempt at implementing automatic detector computation, and we can hash the following:

- the subtemplate in its entirety, which should not be a big deal because it is of constant (and small) size. For example, with the surface code, we should be fine just hashing 9 integers (i.e., Manhattan radius of 1).
- the plaquettes, and this is where it becomes a little bit messy. I ended up giving a unique descriptive name to each Plaquette and use that name to compute the hash, I do not know if that is the best solution.

Anyway, this is not a priority, and it will only be if, after benchmarking the code, we find out that the two functions written earlier in this message are main bottlenecks.

inmzhang:
When I'm working on #321, I try to use the code in this PR to construct the detectors. However, the subtemplate cannot work properly across the boundary of different templates(blocks):
1. Currently, subtemplates can only be constructed from a single template, independently from other neighboring templates.
2. Identifying by indices can hardly work across multiple templates since even the same index of different templates can be instantiated with different `Plaquette`s.

This blocks me from adding detectors with scalable approach in #321, though we can still choose to compute the detectors after the construction of the whole circuit. Do you have any idea on that?

nelimee:
> 1. Currently, subtemplates can only be constructed from a single template, independently from other neighboring templates.

I planned to work around this by creating a temporary instance of `tqec.templates.grid.GridTemplate` for each locally space-connected group of logical qubit. 

Something looking like that (there is a typing error on the `GridTemplate` creation, but this will be easy to sort out):
```py
from pprint import pprint

from tqec.block.block import StandardComputationBlock
from tqec.block.library.closed import xzz_block
from tqec.templates.grid import GridTemplate

block = xzz_block()

grid_block = StandardComputationBlock(
    GridTemplate([[1, 1], [0, 0]], block.template, k=1), block.plaquettes
)

detectors = grid_block.compute_detectors(subtemplate_radius_trial_range=range(1, 2))
pprint(detectors)
```
outputs
```
[[Detector([Measurement(q(2, 2), -1)]),
  Detector([Measurement(q(0, 4), -1)]),
  Detector([Measurement(q(2, 10), -1)]),
  Detector([Measurement(q(0, 12), -1)]),
  Detector([Measurement(q(4, 4), -1)]),
  Detector([Measurement(q(4, 12), -1)]),
  Detector([Measurement(q(6, 2), -1)]),
  Detector([Measurement(q(6, 10), -1)])],
 [Detector([Measurement(q(2, 0), -2), Measurement(q(2, 0), -1)]),
  Detector([Measurement(q(2, 2), -2), Measurement(q(2, 2), -1)]),
  Detector([Measurement(q(0, 4), -1), Measurement(q(0, 4), -2)]),
  Detector([Measurement(q(2, 4), -2), Measurement(q(2, 4), -1)]),
  Detector([Measurement(q(2, 8), -2), Measurement(q(2, 8), -1)]),
  Detector([Measurement(q(2, 10), -2), Measurement(q(2, 10), -1)]),
  Detector([Measurement(q(0, 12), -2), Measurement(q(0, 12), -1)]),
  Detector([Measurement(q(2, 12), -1), Measurement(q(2, 12), -2)]),
  Detector([Measurement(q(4, 2), -2), Measurement(q(4, 2), -1)]),
  Detector([Measurement(q(4, 4), -1), Measurement(q(4, 4), -2)]),
  Detector([Measurement(q(4, 6), -2), Measurement(q(4, 6), -1)]),
  Detector([Measurement(q(4, 10), -2), Measurement(q(4, 10), -1)]),
  Detector([Measurement(q(4, 12), -1), Measurement(q(4, 12), -2)]),
  Detector([Measurement(q(4, 14), -2), Measurement(q(4, 14), -1)]),
  Detector([Measurement(q(6, 2), -1), Measurement(q(6, 2), -2)]),
  Detector([Measurement(q(6, 10), -1), Measurement(q(6, 10), -2)])],
 [Detector([Measurement(q(1, 3), -1), Measurement(q(2, 2), -1), Measurement(q(1, 1), -1), Measurement(q(3, 3), -1), Measurement(q(3, 1), -1)]),
  Detector([Measurement(q(2, 0), -1), Measurement(q(2, 0), -2)]),
  Detector([Measurement(q(2, 2), -1), Measurement(q(2, 2), -2)]),
  Detector([Measurement(q(0, 4), -1), Measurement(q(1, 5), -1), Measurement(q(1, 3), -1)]),
  Detector([Measurement(q(0, 4), -1), Measurement(q(0, 4), -2)]),
  Detector([Measurement(q(2, 4), -1), Measurement(q(2, 4), -2)]),
  Detector([Measurement(q(2, 8), -1), Measurement(q(2, 8), -2)]),
  Detector([Measurement(q(1, 11), -1), Measurement(q(2, 10), -1), Measurement(q(1, 9), -1), Measurement(q(3, 11), -1), Measurement(q(3, 9), -1)]),
  Detector([Measurement(q(2, 10), -1), Measurement(q(2, 10), -2)]),
  Detector([Measurement(q(0, 12), -1), Measurement(q(1, 13), -1), Measurement(q(1, 11), -1)]),
  Detector([Measurement(q(0, 12), -1), Measurement(q(0, 12), -2)]),
  Detector([Measurement(q(2, 12), -1), Measurement(q(2, 12), -2)]),
  Detector([Measurement(q(4, 2), -2), Measurement(q(4, 2), -1)]),
  Detector([Measurement(q(3, 5), -1), Measurement(q(5, 3), -1), Measurement(q(3, 3), -1), Measurement(q(4, 4), -1), Measurement(q(5, 5), -1)]),
  Detector([Measurement(q(4, 4), -2), Measurement(q(4, 4), -1)]),
  Detector([Measurement(q(4, 6), -2), Measurement(q(4, 6), -1)]),
  Detector([Measurement(q(4, 10), -2), Measurement(q(4, 10), -1)]),
  Detector([Measurement(q(3, 13), -1), Measurement(q(5, 11), -1), Measurement(q(3, 11), -1), Measurement(q(4, 12), -1), Measurement(q(5, 13), -1)]),
  Detector([Measurement(q(4, 12), -2), Measurement(q(4, 12), -1)]),
  Detector([Measurement(q(4, 14), -1), Measurement(q(4, 14), -2)]),
  Detector([Measurement(q(5, 1), -1), Measurement(q(6, 2), -1), Measurement(q(5, 3), -1)]),
  Detector([Measurement(q(6, 2), -2), Measurement(q(6, 2), -1)]),
  Detector([Measurement(q(5, 9), -1), Measurement(q(6, 10), -1), Measurement(q(5, 11), -1)]),
  Detector([Measurement(q(6, 10), -2), Measurement(q(6, 10), -1)])]]
```

The `GridTemplate` would benefit from a clearer documentation, I will fix that.

> 2\. Identifying by indices can hardly work across multiple templates since even the same index of different templates can be instantiated with different `Plaquette`s.

This should not be a huge issue now. I am not sure the above solution fixes everything, but at least it seems to me that it fixes the ones you mentioned.

nelimee:
I think this is ready for a last review and merge. If there is still an issue regarding the problem @inmzhang mentioned above, I will fix it here. If there are other issues, I can also fix them here. If not, we can merge.

inmzhang:
The current `GridTemplate` can not solve my problem and It might be the problem of the way I implement the substitution rules...

When doing merge-split operation between two qubits, I substitute/change the plaquettes on the border just in place. And no changes on the underlying template will be made. Then with `GridTemplate`, I will have inconsistent definition for the same plaquette indices for different qubits.

inmzhang:
I think this PR can be merged, and I'll try to change the code in #321 instead, like adding a `GridTemplate`-like template per time step to make it work.

inmzhang (approved):

PR #324, opened 2024-10-01
Fix mypy error on windows

inmzhang:
CI on windows failed after the [last commit](https://github.com/QCHackers/tqec/actions/runs/11117093969/job/30888466220#step:8:18) due to incompatible numpy types. This PR includes the fix to the mypy error.

### DIFF
--- a/src/tqec/templates/subtemplates.py
+++ b/src/tqec/templates/subtemplates.py
@@ -185,7 +185,7 @@ def get_spatially_distinct_subtemplates(
         i + 1: situation for i, situation in enumerate(unique_situations)
     }
     if avoid_zero_plaquettes:
-        final_indices = numpy.zeros((y * x,), dtype=numpy.int_)
+        final_indices = numpy.zeros((y * x,), dtype=numpy.intp)
         final_indices[ignored_flattened_indices] = 0
         final_indices[considered_flattened_indices] = inverse_indices
     else:
--- a/src/tqec/templates/subtemplates.py
+++ b/src/tqec/templates/subtemplates.py
@@ -190,4 +190,6 @@ def get_spatially_distinct_subtemplates(
         final_indices[considered_flattened_indices] = inverse_indices
     else:
         final_indices = inverse_indices
-    return UniqueSubTemplates(final_indices.reshape((y, x)), subtemplates_by_indices)
+    return UniqueSubTemplates(
+        final_indices.reshape((y, x)).astype(numpy.int_), subtemplates_by_indices
+    )
--- a/src/tqec/templates/subtemplates.py
+++ b/src/tqec/templates/subtemplates.py
@@ -185,11 +185,9 @@ def get_spatially_distinct_subtemplates(
         i + 1: situation for i, situation in enumerate(unique_situations)
     }
     if avoid_zero_plaquettes:
-        final_indices = numpy.zeros((y * x,), dtype=numpy.intp)
+        final_indices = numpy.zeros((y * x,), dtype=numpy.int_)
         final_indices[ignored_flattened_indices] = 0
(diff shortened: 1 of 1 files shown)


nelimee:
Several things are strange here:

1. numpy.int_ is used everywhere, but only cause an issue here?
2. numpy.intp and numpy.int_ have a different semantic, the former being "an integer large enough to represent a pointer on the platform" while the later is "the native integer type of the platform". In practice they are (nearly?) always the same width, but the semantic is different.

I am not sure this fix should be merged. I would rather investigate the issue at the source (very likely due to numpy.unique) or add a # type: ignore.

inmzhang:
For the following code:
```py
if avoid_zero_plaquettes:
    final_indices = numpy.zeros((y * x,), dtype=numpy.int_)
    final_indices[ignored_flattened_indices] = 0
    final_indices[considered_flattened_indices] = inverse_indices
else:
    final_indices = inverse_indices
```

and the error message:
```txt
 src\tqec\templates\subtemplates.py:192: error: Incompatible types in assignment(expression has type "ndarray[Any, dtype[signedinteger[_64Bit]]]", variable has type "ndarray[Any, dtype[signedinteger[_32Bit]]]")  [assignment]
```
My understanding is that:

- In the `if` branch, the `final_indices` is of data type `np.int_`. `np.int_` is [compatible with C `long`](https://numpy.org/doc/1.26/reference/arrays.scalars.html#numpy.int_) so is [*32-bit* width on Windows](https://stackoverflow.com/questions/384502/what-is-the-bit-size-of-long-on-64-bit-windows). This is consistent with the type of the variable `ndarray[Any, dtype[signedinteger[_32Bit]]]` in the error message.
- In the `else` branch, the `inverse_indices` is of data type `np.intp`. It should be [the size of the pointer](https://numpy.org/doc/1.26/reference/arrays.scalars.html#numpy.intp) so is *64-bit* on the github action runner machine(I suppose). So the expression type is of `ndarray[Any, dtype[signedinteger[_64Bit]]]` in the error message.

nelimee:
> For the following code:
> ```py
> if avoid_zero_plaquettes:
>     final_indices = numpy.zeros((y * x,), dtype=numpy.int_)
>     final_indices[ignored_flattened_indices] = 0
>     final_indices[considered_flattened_indices] = inverse_indices
> else:
>     final_indices = inverse_indices
> ```
> 
> and the error message:
> ```txt
>  src\tqec\templates\subtemplates.py:192: error: Incompatible types in assignment(expression has type "ndarray[Any, dtype[signedinteger[_64Bit]]]", variable has type "ndarray[Any, dtype[signedinteger[_32Bit]]]")  [assignment]
> ```
> My understanding is that:
> 
> - In the `if` branch, the `final_indices` is of data type `np.int_`. `np.int_` is [compatible with C `long`](https://numpy.org/doc/1.26/reference/arrays.scalars.html#numpy.int_) so is [*32-bit* width on Windows](https://stackoverflow.com/questions/384502/what-is-the-bit-size-of-long-on-64-bit-windows). This is consistent with the type of the variable `ndarray[Any, dtype[signedinteger[_32Bit]]]` in the error message.
> - In the `else` branch, the `inverse_indices` is of data type `np.intp`. It should be [the size of the pointer](https://numpy.org/doc/1.26/reference/arrays.scalars.html#numpy.intp) so is *64-bit* on the github action runner machine(I suppose). So the expression type is of `ndarray[Any, dtype[signedinteger[_64Bit]]]` in the error message.
> 

Yes, this is my understanding too. I would just cast the right hand side in the assignment within the else branch with as_type(numpy.int_). It should be sufficient to solve everything no?

nelimee (approved):

PR #329, opened 2024-10-03
Adding benchmarks using topological CNOT generation

### LINKED ISSUE #328: Add benchmarks to the repository

We are currently in a consolidation phase during which we do not aim at implementing new features but rather refactor the current code to make it better. One (but absolutely not the only) way of evaluating if a new code is "better" than the previous one is to look at the time it takes to accomplish its goal, for us the time the `tqec` library takes to generate a valid `stim.Circuit` instance with observables and detectors.

That means that, during this consolidation phase, we may try to optimize the `tqec` code for runtime. [Initial benchmarks](https://github.com/QCHackers/tqec/pull/321#issuecomment-2374198718) seem to show that there are currently two main bottlenecks (again, only speaking about runtime here):

- automatic detector computation,
- transformation from `cirq.Circuit` to `stim.Circuit`,

but these benchmarks should be replicated and, if possible, improved to be sure that the above points are indeed bottlenecks.

A collection of **meaningful** benchmarks to illustrate the current runtime performance of the `tqec` library and guide the development towards a more efficient circuit generation with high confidence that our efforts will be rewarded by a non-insignificant gain in runtime performance.

**Additional context**

Runtime performance is important, but is absolutely not the only way of quantifying whether or not a code is "better" than another one. Among the other important factors I can list:

- documentation,
- testing,
- simplicity of the code,

and I am pretty sure I am forgetting a few. All that to say that runtime performance is not the ultimate goal (yet) and other factors should also be taken into account when making large code base changes.

nelimee:
This PR fixes #328 by adding a few files to benchmark the runtime of the `tqec` library to generate a fully annotated (detectors and observables) `stim` file from a `.dae` file and a scaling factor `k`.

To start a benchmark:

```bash
cd examples/benchmarks/
# Usage: ./bench.sh [python file to benchmark] [scaling factor k]
./bench.sh cnot_all_observables.py 2
```

This should output something along the lines of

```
Saving all benchmark data to data/cnot_all_observables_k=2_2024-10-03_13-02-42
Statistical profiling of cnot_all_observables.py using pyinstrument...
Profiling saved as an HTML webpage in data/cnot_all_observables_k=2_2024-10-03_13-02-42/cnot_all_observables_k=2.html
```

and you will be able to open the benchmark data by opening the static and self-contained HTML file generated.

### DIFF
--- a/examples/benchmarks/bench.sh
+++ b/examples/benchmarks/bench.sh
@@ -5,28 +5,39 @@ set -e
 
 usage() {
     echo "Usage:"
-    echo "\tbench.sh [python file] [value of k]"
+    echo "\tbench.sh [value of k]"
     echo ""
     echo "Examples:"
-    echo "\tbench.sh cnot_all_observables.py 2"
+    echo "\tbench.sh 2"
 }
 
-if [ $# -ne 2 ]
+if [ $# -ne 1 ]
   then
     usage
     exit 1
 fi
 
-PYTHON_EXEC_NAME=$(basename "$1" .py)
-SCALING_FACTOR="$2"
-BENCH_DIRECTORY="data/${PYTHON_EXEC_NAME}_k=${SCALING_FACTOR}_$(date '+%F_%H-%M-%S')"
+PARENT_PATH=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+
+PYTHON_EXEC_NAME="cnot_all_observables"
+SCALING_FACTOR="$1"
+BENCH_DIRECTORY="${PARENT_PATH}/data/${PYTHON_EXEC_NAME}_k=${SCALING_FACTOR}_$(date '+%F_%H-%M-%S')"
 
 echo "Saving all benchmark data to ${BENCH_DIRECTORY}"
 mkdir -p $BENCH_DIRECTORY
 
 echo "Statistical profiling of ${PYTHON_EXEC_NAME}.py using pyinstrument..."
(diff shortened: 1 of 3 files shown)


nelimee:
Because I cannot upload HTML files directly here, below is a screen capture of the important parts of the profiling information for `k=4`:
image

We can see that:

- detector annotation is costly (and scales badly with `k` as it becomes clearly the main bottleneck for `k=6`),
  - `try_merge_anticommuting_flows` ends up costing a significant amount of time,
  - building the flows is not that costly for the moment,
  - building the fragments from a `stim.Circuit` does not even appear in the profiling information, so its cost should be insignificant compared to the other steps.
- `transform_to_stimcirq_compatible` is costly

inmzhang (approved):
This looks great! Just a few minor comments:
- Perhaps we could move the `benchmark` directory to the top level of the project.
- Benchmarking through CLI commands seems simpler and clearer. However, we need to improve the CLI to incorporate the `NoiseModel`. I understand that using Python scripts directly can offer more flexibility, so this should ultimately be your decision.
- The `assets` directory could also be moved to the top level. We could merge `examples/assets` and `notebooks/assets`, or even combine the `examples` and `notebooks` directories altogether. I think some of the notebooks need to be cleaned up.

nelimee:
> * Benchmarking through CLI commands seems simpler and clearer. However, we need to improve the CLI to incorporate the `NoiseModel`. I understand that using Python scripts directly can offer more flexibility, so this should ultimately be your decision.

In my opinion, the goal of the benchmarks in this PR is to guide development, not to evaluate the end-user performance, so I am fine with having a fine-grained Python file instead of a CLI.

A benchmark for the different CLIs should likely also be included, I will try to include one now.

PR #331, opened 2024-10-04
Remove notebooks and update folder structure

### LINKED ISSUE #330: Clean up `notebook` folder and merge with `examples`
## Describe the bug
Several notebooks in the `notebooks` folder are out-of-date and should be removed.

Also, from [this comment](https://github.com/QCHackers/tqec/pull/329#pullrequestreview-2347006994): 

> The assets directory could also be moved to the top level. We could merge examples/assets and notebooks/assets, or even combine the examples and notebooks directories altogether. I think some of the notebooks need to be cleaned up.

nelimee:
Should fix #330. I *think* I changed all the necessary paths, but might have forgot a few.


inmzhang (approved):
In my opinion, the name of the files under `assets` directory can be changed. "clean_exportable" has no real meaning at this point. Names like "logical_cnot"/"template" should work well. 

And `clean_exportable_template.dae` file is not needed anymore. By the end, we should always start with a `template.skp` to build a model. Trying to import a `template.dae` makes no sense and will even raise an exception in the backend because non-integer positioning.

nelimee:
> And `clean_exportable_template.dae` file is not needed anymore. By the end, we should always start with a `template.skp` to build a model. Trying to import a `template.dae` makes no sense and will even raise an exception in the backend because non-integer positioning.

It is needed in `sketchup_demo.ipynb`, but your comment should definitely be written somewhere in the documentation.

PR #338, opened 2024-10-07
Introduce a few helpers in `tqec.simulation`

nelimee:
This PR introduces a new module `tqec.simulation` that defines helpers to run `stim`/`sinter` simulations and plot the result. The goal is to progressively enrich that module with, for example, the ability to transparently save all the computation results or new ways of plotting the data that interest us.


src/tqec/simulation/simulation.py
```diff
@@ -0,0 +1,97 @@
...
+    block_graph: BlockGraph,
+    ks: Sequence[int],
+    ps: Sequence[float],
+    noise_model_factory: Callable[[float], NoiseModel],
+    observables: list[AbstractObservable] | None = None,
+    num_workers: int = multiprocessing.cpu_count(),
+    progress_callback: Callable[[sinter.Progress], None] | None = None,
+    max_shots: int | None = None,
+    max_errors: int | None = None,
+    decoders: Iterable[str] | None = None,
```
inmzhang, line 22: For our usage, it's safe to default this to `("pymatching",)`.

inmzhang (approved):

PR #340, opened 2024-10-07
Refactor `tqec.sketchup`

nelimee:
I simply re-organized a little bit the code, mostly creating new modules when possible. No actual code change happened here.


nelimee:
I am unsure about the necessity to split `block_graph.py` into smaller chunks: it helps me a lot navigating in the code and understanding it, but it might not be the case for other people. 

*Side note: the same splitting can likely be done for `schedule.py`*

My goal here is to pave the way for a deeper refactor that I still think about (and that I will share in an issue before attempting). In a few words, I think the current state of `ZXGraph` and `BlockGraph` is nice, they are interchangeable data-structures that losslessly represent the computation at hand and that can be 1) exported back to a re-usable computed-readable format (`.dae`), 2) plotted with `matplotlib` for `ZXGraph` and within an HTML document for `BlockGraph` and more importantly 3) exported to a data-structure that will then be used to generate the circuits. I think point 3) should be refactored, and in particular I am still not convinced by the current `CompiledGraph` implementation (even though I like the name that is quite effective at saying that you perform a lossy conversion when compiling and loose points 1 and 2 before).

inmzhang (approved):
And do you have a better name for `sketchup` directory? Plenty of things in the directory are not directly related to SketchUp and I really think it's not a good name since I named it...

nelimee:
> And do you have a better name for `sketchup` directory? Plenty of things in the directory are not directly related to SketchUp and I really think it's not a good name since I named it...

Maybe `computation` as the data-structures in this module (`ZXGraph` and `BlockGraph`) are representing a whole QEC computation?

inmzhang:
> Maybe `computation` as the data-structures in this module (`ZXGraph` and `BlockGraph`) are representing a whole QEC computation?

Great, then we can go with that.

Issue #336, opened 2024-10-07
Decreased code distance for logical CNOT circuits with X basis initialization/measurements


inmzhang:
As reported in #334, the logical error rates of the logical CNOT circuits with  X basis initialization/measurements are suspiciously high compared to Z basis ones. And [the exact circuit-level code distance](https://github.com/QCHackers/tqec/pull/334#issuecomment-2396061423) is 2 instead of the expected 3.

The Z basis block graph is imported from the `.dae` file and looks like:
![图片](https://github.com/user-attachments/assets/62cafdfd-1908-4c4d-9118-7c6c1dbb5941)

The X basis block graph is constructed from a `ZXGraph` with the following code:
```py
cnot_zx = ZXGraph("Logical CNOT ZX Graph")

cnot_zx.add_x_node(Position3D(0, 0, 0))
cnot_zx.add_z_node(Position3D(0, 0, 1))
cnot_zx.add_x_node(Position3D(0, 0, 2))
cnot_zx.add_x_node(Position3D(0, 0, 3))
cnot_zx.add_z_node(Position3D(0, 1, 1))
cnot_zx.add_x_node(Position3D(0, 1, 2))
cnot_zx.add_x_node(Position3D(1, 1, 0))
cnot_zx.add_x_node(Position3D(1, 1, 1))
cnot_zx.add_x_node(Position3D(1, 1, 2))
cnot_zx.add_x_node(Position3D(1, 1, 3))

cnot_zx.add_edge(Position3D(0, 0, 0), Position3D(0, 0, 1))
cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 0, 2))
cnot_zx.add_edge(Position3D(0, 0, 2), Position3D(0, 0, 3))
cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 1, 1))
cnot_zx.add_edge(Position3D(0, 1, 1), Position3D(0, 1, 2))
cnot_zx.add_edge(Position3D(0, 1, 2), Position3D(1, 1, 2))
cnot_zx.add_edge(Position3D(1, 1, 0), Position3D(1, 1, 1))
cnot_zx.add_edge(Position3D(1, 1, 1), Position3D(1, 1, 2))
cnot_zx.add_edge(Position3D(1, 1, 2), Position3D(1, 1, 3))
cnot_zx.to_block_graph("Logical CNOT Block Graph")
```
and looks like:

[...]

PR #342, opened 2024-10-08
Refactor plaquette library

inmzhang:
The current `plaquette/library` is complicated and defines individual plaquettes for initialization/memory/measurements. The `pauli_memory_plaquette` aims to provide resuable utilities for general pauli operator syndrome measurements but there are many tweaks when optimizing circuits in practice.

This PR tries to simplify the `plaquette/library` and provides functionality to generate plaquettes for standard CSS-type surface code and ZXXZ-type surface code. Also, #336 will be addressed in the PR.

ZXXZ-type plaquettes will be added later.


src/tqec/plaquette/library/css.py
```diff
@@ -0,0 +1,68 @@
...
+
+from tqec.circuit.schedule import ScheduledCircuit
+from tqec.exceptions import TQECException
+from tqec.plaquette.enums import MeasurementBasis, ResetBasis
+from tqec.plaquette.plaquette import Plaquette
+from tqec.plaquette.qubit import SquarePlaquetteQubits
+
+
+def make_css_surface_code_plaquette(
+    basis: Literal["X", "Z"] | str,
```
nelimee, line 17:
```suggestion
    basis: Literal["X", "Z"],
```

nelimee (approved):
I really like how this simplifies a lot the plaquette creation process. Well done! 
One extra remark: with that change, it might be worth looking again at the usefulness of `SquarePlaquetteQubits` and the likes. It still seem useful, but I am not sure we really need it anymore.

inmzhang:
> One extra remark: with that change, it might be worth looking again at the usefulness of SquarePlaquetteQubits and the likes. It still seem useful, but I am not sure we really need it anymore.

I think it's ok to remove it.

inmzhang:
The plaquettes for ZXXZ type surface code should work after adding `H` gate to the mergeable set, which I haven't done yet.

Adding `H` gate to `_MERGEABLE_INSTRUCTION_NAMES` would cause some tests to fail. And I think it's not safe to add a commonly used single qubit gate to the mergeable instructions to allow qubit collision passing through the check easily. Do you have any better idea on this issue?

nelimee:
> I think it's ok to remove it.

Do you want to do that here, or should we open another PR?

inmzhang:
> Do you want to do that here, or should we open another PR?

I can do it here but maybe serveral hours later.

nelimee:
> Adding `H` gate to `_MERGEABLE_INSTRUCTION_NAMES` would cause some tests to fail. And I think it's not safe to add a commonly used single qubit gate to the mergeable instructions to allow qubit collision passing through the check easily. Do you have any better idea on this issue?

I fully agree, and started a small review before your last commit removing that gate from `_MERGEABLE_INSTRUCTION_NAMES` saying that I think it should not be done.


> The plaquettes for ZXXZ type surface code should work after adding `H` gate to the mergeable set, which I haven't done yet.

I do not see where the potential duplicates might come from. I will have a deeper look later.

inmzhang:
> I do not see where the potential duplicates might come from.

With the current implementation of ZXXZ plaquettes, I think all the moments of H gates should have duplication.

> One extra remark: with that change, it might be worth looking again at the usefulness of `SquarePlaquetteQubits` and the likes. It still seem useful, but I am not sure we really need it anymore.

I have removed `RoundedPlaquetteQubits` from the codebase but keep the `SquarePlaquetteQubits` untouched because I found it still useful in plaquettes factories and in some tests.

nelimee:
> With the current implementation of ZXXZ plaquettes, I think all the moments of H gates should have duplication.

Right, I see it now, I do not know why I just forgot that plaquettes were sharing data qubits... This is a general issue that we will likely have for a lot of types of plaquettes. Below, a dump of my thoughts on that.

We might want to store inside the `Plaquette` instances a list of instructions that are applied on data qubits and that may be deduplicated. This list can even be generated on-the-fly by returning names of 1-qubit instructions that are applied on all the data-qubits at specific moment.
This would give something like
```py
class Plaquette:
    # ...
    @property
    def deduplicable_instructions(self) -> frozenset[str]:
        data_qubit_indices = frozenset(
            self.circuit.q2i[dq] for dq in self.qubits.data_qubits
        )
        data_qubit_instructions_to_merge: list[str] = []
        for moment in self.circuit.moments:
            # Assumes that instructions are grouped in each moment, i.e. that we
            # do not have
            #     TICK
            #     H 0 1
            #     H 2
            #     TICK
            for instruction in moment.instructions:
                target_groups = instruction.target_groups()
                if any(len(g) > 1 for g in target_groups):
                    # Ignore, this is not a single-qubit instruction
                    continue
                qubit_targets = frozenset(
                    target.qubit_value
                    for target in instruction.targets_copy()
                    if target.is_qubit_target
                )
                if data_qubit_indices.issubset(qubit_targets):
                    data_qubit_instructions_to_merge.append(instruction.name)
        return frozenset(data_qubit_instructions_to_merge)
```

That list can then be used to specifically merge 1-qubit instructions, specifically on data qubits, with a generic version of `remove_duplicate_instructions` that is able to only consider instructions on a subset of the qubits (the data qubits).

nelimee:
Obtained with
```py
"""This is an example of compiling a logical CNOT `.dae` model to
`stim.Circuit`."""

from pathlib import Path

import matplotlib.pyplot as plt
import numpy
import sinter

from tqec import BlockGraph, Position3D, ZXGraph
from tqec.noise_models import NoiseModel
from tqec.simulation.plotting.inset import plot_observable_as_inset
from tqec.simulation.simulation import start_simulation_using_sinter

EXAMPLE_FOLDER = Path(__file__).parent
TQEC_FOLDER = EXAMPLE_FOLDER.parent
ASSETS_FOLDER = TQEC_FOLDER / "assets"
CNOT_DAE_FILE = ASSETS_FOLDER / "logical_cnot.dae"


def create_block_graph(z_basis: bool = True) -> BlockGraph:
    if z_basis:
        return BlockGraph.from_dae_file(CNOT_DAE_FILE)

    cnot_zx = ZXGraph("Logical CNOT ZX Graph")

    cnot_zx.add_x_node(Position3D(0, 0, 0))
    cnot_zx.add_z_node(Position3D(0, 0, 1))
    cnot_zx.add_x_node(Position3D(0, 0, 2))
    cnot_zx.add_x_node(Position3D(0, 0, 3))
    cnot_zx.add_z_node(Position3D(0, 1, 1))
    cnot_zx.add_x_node(Position3D(0, 1, 2))
    cnot_zx.add_x_node(Position3D(1, 1, 0))
    cnot_zx.add_x_node(Position3D(1, 1, 1))
    cnot_zx.add_x_node(Position3D(1, 1, 2))
    cnot_zx.add_x_node(Position3D(1, 1, 3))

    cnot_zx.add_edge(Position3D(0, 0, 0), Position3D(0, 0, 1))
    cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 0, 2))
    cnot_zx.add_edge(Position3D(0, 0, 2), Position3D(0, 0, 3))
    cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 1, 1))
    cnot_zx.add_edge(Position3D(0, 1, 1), Position3D(0, 1, 2))
    cnot_zx.add_edge(Position3D(0, 1, 2), Position3D(1, 1, 2))
    cnot_zx.add_edge(Position3D(1, 1, 0), Position3D(1, 1, 1))
    cnot_zx.add_edge(Position3D(1, 1, 1), Position3D(1, 1, 2))
    cnot_zx.add_edge(Position3D(1, 1, 2), Position3D(1, 1, 3))
    return cnot_zx.to_block_graph("Logical CNOT Block Graph")


def main() -> None:
    z_basis: bool = True
    # 1 Create `BlockGraph` representing the computation
    block_graph = create_block_graph(z_basis)
    zx_graph = block_graph.to_zx_graph("Logical CNOT")
    correlation_surfaces = zx_graph.find_correlation_subgraphs()

    # 2. Find and choose the logical observables
    observables = block_graph.get_abstract_observables()

    stats = start_simulation_using_sinter(
        block_graph,
        range(1, 4),
        list(numpy.logspace(-4, -1, 10)),
        NoiseModel.uniform_depolarizing,
        observables=observables,
        num_workers=30,
        max_shots=10_000_000,
        max_errors=5_000,
        decoders=["pymatching"],
        print_progress=True,
    )

    for i, stat in enumerate(stats):
        fig, ax = plt.subplots()
        sinter.plot_error_rate(
            ax=ax,
            stats=stat,
            x_func=lambda stat: stat.json_metadata["p"],
            group_func=lambda stat: stat.json_metadata["d"],
        )
        plot_observable_as_inset(ax, zx_graph, correlation_surfaces[i])
        ax.grid(axis="both")
        ax.legend()
        ax.loglog()
        ax.set_title("Logical CNOT Error Rate")
        fig.savefig(
            ASSETS_FOLDER
            / f"logical_cnot_{'z' if z_basis else 'x'}_result_observable_{i}.png"
        )


if __name__ == "__main__":
    main()
```
I executed the file twice, once with `z_basis=True` and another time with `z_basis=False`:
logical_cnot_x_result_observable_0
logical_cnot_x_result_observable_1
logical_cnot_x_result_observable_2
logical_cnot_z_result_observable_0
logical_cnot_z_result_observable_1
logical_cnot_z_result_observable_2

inmzhang:
Great! Now the results seem reasonable to me. Several points to note here:
1. Compared to the results in #334, the Z-basis observables has smaller logical error rates because the depth of the syndrom extraction circuits decrease from 8 to 6.
2. The `ZXGraph` built from scratch for X-basis observables is not consistent with the Z-basis observables. We should change the code for that as noted in #336:
```py
def create_block_graph(z_basis: bool = True) -> BlockGraph:
    if z_basis:
        return BlockGraph.from_dae_file(CNOT_DAE_FILE)

    cnot_zx = ZXGraph("Logical CNOT ZX Graph")

    cnot_zx.add_x_node(Position3D(0, 0, 0))
    cnot_zx.add_x_node(Position3D(0, 0, 1))
    cnot_zx.add_z_node(Position3D(0, 0, 2))
    cnot_zx.add_x_node(Position3D(0, 0, 3))
    cnot_zx.add_x_node(Position3D(0, 1, 1))
    cnot_zx.add_z_node(Position3D(0, 1, 2))
    cnot_zx.add_x_node(Position3D(1, 1, 0))
    cnot_zx.add_z_node(Position3D(1, 1, 1))
    cnot_zx.add_z_node(Position3D(1, 1, 2))
    cnot_zx.add_x_node(Position3D(1, 1, 3))

    cnot_zx.add_edge(Position3D(0, 0, 0), Position3D(0, 0, 1))
    cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 0, 2))
    cnot_zx.add_edge(Position3D(0, 0, 2), Position3D(0, 0, 3))
    cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 1, 1))
    cnot_zx.add_edge(Position3D(0, 1, 1), Position3D(0, 1, 2))
    cnot_zx.add_edge(Position3D(0, 1, 2), Position3D(1, 1, 2))
    cnot_zx.add_edge(Position3D(1, 1, 0), Position3D(1, 1, 1))
    cnot_zx.add_edge(Position3D(1, 1, 1), Position3D(1, 1, 2))
    cnot_zx.add_edge(Position3D(1, 1, 2), Position3D(1, 1, 3))
    return cnot_zx.to_block_graph("Logical CNOT Block Graph")
``` 
I haven't checked the code carefully, but the principle is that we should only change the leaf nodes from Z type to X type but keep the structure of the computation consistent.

nelimee:
> Great! Now the results seem reasonable to me. Several points to note here:
> 
>     1. Compared to the results in [Large refactor to remove `cirq` from dependencies #334](https://github.com/QCHackers/tqec/pull/334), the Z-basis observables has smaller logical error rates because the depth of the syndrom extraction circuits decrease from 8 to 6.
> 
>     2. The `ZXGraph` built from scratch for X-basis observables is not consistent with the Z-basis observables. We should change the code for that as noted in [Decreased code distance for logical CNOT circuits with X basis initialization/measurements #336](https://github.com/QCHackers/tqec/issues/336):
> 
> 
> ```python
> def create_block_graph(z_basis: bool = True) -> BlockGraph:
>     if z_basis:
>         return BlockGraph.from_dae_file(CNOT_DAE_FILE)
> 
>     cnot_zx = ZXGraph("Logical CNOT ZX Graph")
> 
>     cnot_zx.add_x_node(Position3D(0, 0, 0))
>     cnot_zx.add_x_node(Position3D(0, 0, 1))
>     cnot_zx.add_z_node(Position3D(0, 0, 2))
>     cnot_zx.add_x_node(Position3D(0, 0, 3))
>     cnot_zx.add_x_node(Position3D(0, 1, 1))
>     cnot_zx.add_z_node(Position3D(0, 1, 2))
>     cnot_zx.add_x_node(Position3D(1, 1, 0))
>     cnot_zx.add_z_node(Position3D(1, 1, 1))
>     cnot_zx.add_z_node(Position3D(1, 1, 2))
>     cnot_zx.add_x_node(Position3D(1, 1, 3))
> 
>     cnot_zx.add_edge(Position3D(0, 0, 0), Position3D(0, 0, 1))
>     cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 0, 2))
>     cnot_zx.add_edge(Position3D(0, 0, 2), Position3D(0, 0, 3))
>     cnot_zx.add_edge(Position3D(0, 0, 1), Position3D(0, 1, 1))
>     cnot_zx.add_edge(Position3D(0, 1, 1), Position3D(0, 1, 2))
>     cnot_zx.add_edge(Position3D(0, 1, 2), Position3D(1, 1, 2))
>     cnot_zx.add_edge(Position3D(1, 1, 0), Position3D(1, 1, 1))
>     cnot_zx.add_edge(Position3D(1, 1, 1), Position3D(1, 1, 2))
>     cnot_zx.add_edge(Position3D(1, 1, 2), Position3D(1, 1, 3))
>     return cnot_zx.to_block_graph("Logical CNOT Block Graph")
> ```
> 
> I haven't checked the code carefully, but the principle is that we should only change the leaf nodes from Z type to X type but keep the structure of the computation consistent.

Note that the code provided in your message produces a faulty circuit with a non-deterministic observable. We will have to check that eventually.

inmzhang:
> Note that the code provided in your message produces a faulty circuit with a non-deterministic observable. We will have to check that eventually.

The problem is with the computed abstract observables. I will fix this at tomorrow.

inmzhang:
Got some time to make a hot fix of the abstract observable computation. The code should be checked again and refactored in the future. But it can run the simulation for now.

inmzhang:
The deduplication process is not working yet, I will let you know when I think it's ready to be reviewed again.

inmzhang:
> Also, what is the added value of storing the instructions names that might be deduplicated in the `ScheduledCircuit` instance instead of the `Plaquette` instance?

I can't really remember why I made this decision and now I do think it will be much cleaner to attach that to `Plaquette`. And I agree that the `ScheduledCircuit` is large enough and deserve to be splitted.

nelimee:
I prefer that version. One final remark: I think that we can remove completely `_DEFAULT_MERGEABLE_INSTRUCTIONS` and require the user (or developer for pre-built `Plaquette` instances) provide a complete list of instruction names that should be considered mergeable.

inmzhang:
> I prefer that version. One final remark: I think that we can remove completely `_DEFAULT_MERGEABLE_INSTRUCTIONS` and require the user (or developer for pre-built `Plaquette` instances) provide a complete list of instruction names that should be considered mergeable.

Agree, I'll make the change.

nelimee (approved):
LGTM, can be merged in my opinion.

PR #344, opened 2024-10-09
Introduce a `QubitMap` structure

nelimee:
During the circuit construction stage, we are performing a lot of operations on qubits such as remapping, changing indices, offsetting, etc. I think it would be beneficial for the long term to have a specific data-structure to represent a map* from indices to qubits.

* it is more than a map, it should be a bijection. In that regard, I think using https://github.com/jab/bidict might be a good idea, but I am a little bit reluctant to add a whole dependency just for that. Any opinion on whether or not we should use that package or keep the implementation as it is for the moment?


src/tqec/circuit/qubit_map.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Returns:
+            a copy of `self` for which the assertion
+            `set(return_value.qubits).issubset(qubits_to_keep)` is `True`.
+        """
+        kept_qubits = frozenset(qubits_to_keep)
+        return QubitMap({i: q for i, q in self.i2q.items() if q in kept_qubits})
+
+
+def get_final_qubits(circuit: stim.Circuit) -> QubitMap:
```
inmzhang, line 88: A name like `get_qubit_map` might be better?

inmzhang (approved):
If we use `bidict`, we still need to wrap the it with some more convenient data structure like `QubitMap`. And I think the performance of the map will never be a bottleneck for our usage. So I prefer keeping the current implementation.

PR #345, opened 2024-10-09
Various fixes for logical CNOT using `ZXXZ` code

nelimee:
This is a PR that includes several fixes and a refactor in order to be able to run correct simulations for the logical CNOT using a different set of plaquettes. 

Modifications in this PR in chronological order:

### 1. Refactor of `tqec.compile.specs`

The `tqec.compile.specs` was hosting the current default specification on how to translate a `CubeSpec` instance into a `CompiledBlock` instance containing the actual quantum circuit. In order to be able to have different pre-built "specs", I refactored into several sub-modules:
- `tqec.compile.specs.base` that contains the core class definitions (`CubeSpec` and the `SpecRule` protocol).
- `tqec.compile.specs.enums` that contains the `JunctionArms` enumeration.
- `tqec.compile.specs.library` that should contain one submodule per pre-built "specs". Right now, it contains `tqec.compile.specs.library.css` that implements the previous default specs and `tqec.compile.specs.library.zxxz` that implements the specification for `ZXXZ` code.

### 2. Update the code in `logical_cnot.py`

The file was generating block_graphs in a poorly manner (reading from a file for `Z` basis initialisation, building a ZXGraph for `X` basis initialisation). This is a detail, and I am fine reverting it if anyone think that the previous code was better.

### 3. Fixing a bug in detector matching

It turns out that there was a typo in the detector matching code that resulted in an out-of-bound access. I am surprised we did not catch this before, but it should be fixed.

------

There are still issues, that I will try to describe here for the moment.

The main issue there is right now is with pipes, replacement of border plaquettes, and insertion of gates when inserting.

Right now, when merging spatially two logical qubits (let's say in the X dimension), we are:

1. Looking at the plaquettes that are used on the merged logical qubits and replacing border plaquettes with them following the image below
image
2. Inserting resets on the line of data qubits that are between the two logical qubit patches to initialise them.

Step 2 is the current issue. This is currently done with
```py

def _inplace_add_operation_on_data_qubits(
    plaquette: Plaquette,
    side: PlaquetteSide,
    operation_basis: ResetBasis | MeasurementBasis,
    moment_schedule: int,
) -> None:
    # TODO: Remove that and make Plaquette frozen.
    qubits = plaquette.qubits.get_qubits_on_side(side)
    # Do not add resets on qubits that have already be reset
    moment = plaquette.circuit.moment_at_schedule(moment_schedule)
    q2i = plaquette.circuit.q2i

    operations_to_add = stim.Circuit()
    for q in qubits:
        if moment.operates_on([q2i[q]]):
            continue
        operations_to_add.append(operation_basis.instruction_name, [q2i[q]], [])
    plaquette.circuit.add_to_schedule_index(moment_schedule, Moment(operations_to_add))
```
but that fails to respect the `specs` provided by the user. For example, if instead of `RX`/`MX` we are supposed to insert `RZ`/`MZ` with the appropriate basis change (`H` gate after/before), this is not possible. 
This also fails for the `ZXXZ` code in which some data qubits have a `H` gate and other do not.

### DIFF
--- a/benchmarks/cnot_all_observables.py
+++ b/benchmarks/cnot_all_observables.py
@@ -5,10 +5,12 @@ import stim
 
 from tqec.circuit.detectors.construction import annotate_detectors_automatically
 from tqec.compile.compile import CompiledGraph, compile_block_graph
-from tqec.noise_models import NoiseModel
-from tqec.position import Position3D
+from tqec.compile.specs.library.css import CSS_SPEC_RULES
+from tqec.compile.substitute import DEFAULT_SUBSTITUTION_RULES
 from tqec.computation.block_graph import BlockGraph
 from tqec.computation.zx_graph import ZXGraph
+from tqec.noise_models import NoiseModel
+from tqec.position import Position3D
 
 BENCHMARK_FOLDER = Path(__file__).resolve().parent
 TQEC_FOLDER = BENCHMARK_FOLDER.parent
@@ -65,6 +67,8 @@ def generate_cnot_circuits(*ks: int):
     # 3. Compile the `BlockGraph`
     compiled_graph = compile_block_graph(
         block_graph,
+        spec_rules=CSS_SPEC_RULES,
+        substitute_rules=DEFAULT_SUBSTITUTION_RULES,
         observables=[observables[1]],
     )
 
--- a/examples/logical_cnot.py
+++ b/examples/logical_cnot.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 import numpy
 import sinter
 
-from tqec import BlockGraph, Position3D, ZXGraph
+from tqec import Position3D, ZXGraph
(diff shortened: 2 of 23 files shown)


inmzhang:
> 2. Inserting resets on the line of data qubits that are between the two logical qubit patches to initialise them.

Yes I also found this issue and noted [here](https://github.com/QCHackers/tqec/pull/342#issuecomment-2401377309).

Maybe it's a good time to bring back the idea of "`Pipe` as plaquettes" for substitution to work properly?

afowler:
+1 for pipes having their own plaquettes.

inmzhang (approved):
> I will open an issue for that. I think this PR should be merged and the specs / plaquette generation should be improved in another PR, what's your opinion on that?

Agree.

nelimee:
See #346

PR #348, opened 2024-10-10
Improve `subtemplates`

nelimee:
This PR:

- ~~bumps `numpy` version explicitly in the requirements to only support `numpy ~= 2.0`,~~ not anymore, see comments,
- improves the documentation of `tqec.templates.subtemplates`,
- performs some minor `import` linting (removing unused, ordering, only relative imports in the main `__init__.py`).


inmzhang:
`Pymatching` still does not support `numpy2`? https://github.com/oscarhiggott/PyMatching/issues/102

nelimee:
> `Pymatching` still does not support `numpy2`? [oscarhiggott/PyMatching#102](https://github.com/oscarhiggott/PyMatching/issues/102)

That seems to be true, good catch, I am rolling back the version bump on numpy.

inmzhang (approved):
Based on the timing, I think it's efficient enough? After all, the realistic `k`s we simulate for should be upper bounded by 10~14.

nelimee:
> Based on the timing, I think it's efficient enough? After all, the realistic `k`s we simulate for should be upper bounded by 10~14.

Yes, I agree.

PR #351, opened 2024-10-11
Adapt to new RGB color convention for sketchup model

### LINKED ISSUE #349: Adapt to new RGB color convention for sketchup model
In the last meeting, a new RGB color convention was introduced to incorporate the Y-basis measurement:
图片

It would be beneficial to adapt to the new convention as soon as possible to avoid distraction. However, the `RGB` value provided in the above picture needs to be checked. In particular, the `RGB` value for the blue face seems incorrect.

Austin @afowler, could you check and decide the color convention here? Also, could you provide the new `template.skp` and `template.dae` files?

inmzhang:
Fix #349.

Additionally, add default value to arguments of `compile_block_graph` for the code in readme/docs to work.


nelimee (commented):
The large size bump of the `assets/logical_cnot.dae` seems suspicious to me, what's justifying such a size bump?

inmzhang:
> The large size bump of the `assets/logical_cnot.dae` seems suspicious to me, what's justifying such a size bump?

Good catch! The old `assets/logical_cnot.dae` file was exported from SketchUp while the new one was generated from the code and saved. SketchUp only keeps used `<library_nodes>` and `<geometry>`s. In the case of logical cnot, that means no geometry for hadamard pipes, spatial junctions and Y cubes. However, currently the backend code will add all the possible geometries to the `.dae` file, which explains the size bump.

I refactored and optimized the code to only append necessary geometry nodes and library nodes on the fly. Now the size of the `logical_cnot.dae` decreases from 202KB to 44KB.

nelimee (approved):
As long as the last tests are successful, this looks ready to merge for me :)

Issue #355, opened 2024-10-11
Circuit of `XZZX` plaquette


inmzhang:
Currently, the plaquettes for `ZXXZ` surface code all measure `ZXXZ` stabilizer follow the "Z" shape or "N" shape CZ order for different plaquette basis:

![图片](https://github.com/user-attachments/assets/6816e48d-b81b-4894-95e1-66cbea79faf8)

However, when I think about the Hadamard transition in the time direction for ZXXZ code, I realize that the measured stabilizer should be transformed from `ZXXZ` to `XZZX` when the boundary types exchange. So it means we need to use `ZXXZ` stabilizers for `XZ?` cubes while using `XZZX` stabilizers for `ZX?` cubes, which is not the case in the current code. The circuit I constructed for `XZZX` stabilizer measurement is as follow, some hadamard gates are inserted deliberately to balance the detection region for errors of different basis:

![图片](https://github.com/user-attachments/assets/dcf69794-43b3-4ca5-b561-eb1d2df3af2c)

[...]

Issue #361, opened 2024-10-13
`DetectorDatabase` does not work


nelimee:
## Describe the bug
Due to the fact that Python `hash` is not deterministic (it can change from Python version to Python version, and more importantly it can, and does, also change the values returned between two executions, even with the same interpreter), the `DetectorDatabase` is not reliable.

## Steps to reproduce the behavior:
To check that Python hash is not deterministic, you can try to run multiple times `python -c 'print(hash("thisisatest"))'`. The same string `"thisisatest"` ends up being hashed differently.

For `DetectorDatabase`, you "only" have to create and populate an instance, save it (e.g., using `pickle`), and try to load it from a different Python execution. Even if you do the exact same tasks (which results in the exact same queries to the database, that are now supposed to returned the saved entries because the database is now populated), the database will never return a saved value. 

## Expected behavior
The `DetectorDatabase` should be as portable as possible. Across different executions of the same environment is the bare minimum, but it would be nice to have a truly portable format.

## Additional context
[This blog post](https://death.andgravity.com/stable-hashing) might be interesting to look at.

PR #362, opened 2024-10-15
Pipe as plaquettes

### LINKED ISSUE #346: Pipes should have their own plaquettes
Right now, pipes are implemented by `SubstititionRule` instances that describe how to substitute the plaquettes on the boundary of blocks depending on the pipe that link them.

The current default (and only) substitution rule perform the following replacement
image
on block plaquettes (pipe in the X direction shown above, Y and time dimensions follow the same algorithm).

This requires a few fixes at runtime because, for spatial replacements, the data qubits that are on the line between the two logical qubit instances should be initialized. We initially did that with
```py
def _inplace_add_operation_on_data_qubits(
    plaquette: Plaquette,
    side: PlaquetteSide,
    operation_basis: ResetBasis | MeasurementBasis,
    moment_schedule: int,
) -> None:
    # TODO: Remove that and make Plaquette frozen.
    qubits = plaquette.qubits.get_qubits_on_side(side)
    # Do not add resets on qubits that have already be reset
    moment = plaquette.circuit.moment_at_schedule(moment_schedule)
    q2i = plaquette.circuit.q2i

    operations_to_add = stim.Circuit()
    for q in qubits:
        if moment.operates_on([q2i[q]]):
            continue
        operations_to_add.append(operation_basis.instruction_name, [q2i[q]], [])
    plaquette.circuit.add_to_schedule_index(moment_schedule, Moment(operations_to_add))
```

but that heuristic fails to adhere to the specs provided by the user. What if the user asked for only Z-basis resets and measurements and wants a Hadamard gate inserted instead of `RX`? What if only some data qubits should have a Hadamard gate applied on them? (#345 requires that for `ZXXZ` code).

Instead of taking plaquettes from the connected blocks, pipes should have their own "instantiation" (i.e., a `Template` instance and some `Plaquettes`).

I think that several changes will be needed:

1. Removing the `_inplace_add_operation_on_data_qubits` should allow us to make `Plaquette` read-only (`frozen=True`).
2. The `PlaquetteBuilder` protocol introduced in #345 will likely have to be changed. I think that we should, for each "specs", define an instance (something like "PlaquetteLibrary"?) that lists all the plaquettes that might be needed. I do not know what is the best way of doing that, right now a map from identifiers to `Plaquette` instances seems a good option to me, something like
    ```py
    plibrary = CSSPlaquetteLibrary(**options)
    xxxx_plaquette = plibrary.get("xxxx")
    xxxx_junction_left_plaquette = plibrary.get("xxxx_junction_left")
    ```
    with `"xxxx"`, `"xxxx_junction_left"`, ... identifiers that describe the associated plaquette and that are documented / enforced in the code.
3. Build the pipes into "PlaquetteLibrary" (naming?) instances.
4. Change the substitution rules.

inmzhang:
This PR tries to fix #346 and #355.

The main change is the introduced `PipeSpec`, which has the similar functionality with `CubeSpec` but specify the plaquettes used to do the substitution.

I still found it not easy to implement all the details correctly, e.g. the space pipe. It took me some time to debug the circuit and some parts of the code may not look clean. It will definitely need some refactor or redesign based on the review. However, the final simulation results seem correct to me:

### css logical cnot with uniform depolarizing
css_logical_cnot_result_X_observable_0
css_logical_cnot_result_X_observable_1
css_logical_cnot_result_X_observable_2
css_logical_cnot_result_Z_observable_0
css_logical_cnot_result_Z_observable_1
css_logical_cnot_result_Z_observable_2

### zxxz logical cnot with uniform depolarizing
zxxz_logical_cnot_result_X_observable_0
zxxz_logical_cnot_result_X_observable_1
zxxz_logical_cnot_result_X_observable_2
zxxz_logical_cnot_result_Z_observable_0
zxxz_logical_cnot_result_Z_observable_1
zxxz_logical_cnot_result_Z_observable_2


src/tqec/compile/block.py
```diff
@@ -34,26 +34,28 @@ class CompiledBlock:
...
-    def with_updated_layer(
+    def update_layers(
         self,
-        plaquettes_to_update: Mapping[int, Plaquette],
-        layer_to_update: int,
-    ) -> CompiledBlock:
-        """Returns a new `CompiledBlock` with the specified layer updated.
+        substitution: Mapping[int, Plaquettes],
+    ) -> None:
+        """Update the plaquettes in a specific layer of the block.
```
nelimee, line 41: Minor comment: it might be worth explicitly writing down that the modification is performed in-place.

src/tqec/compile/compile.py
```diff
@@ -43,8 +43,10 @@ def __post_init__(self) -> None:
                 "The compiled graph should have at least one time slice "
                 "but got an empty layout_slices."
             )
+        # Use warning instead of exception because a qec circuit without observable
+        # may still be useful, e.g. do statistics on the detection events.
         if len(self.observables) == 0:
-            raise TQECException("The compiled graph includes no observable.")
+            raise TQECWarning("The compiled graph includes no observable.")
```
nelimee, line 49: Warnings are supposed to be used with `warnings.warn`. I am not sure this line has the meaning you intended, as I think it will behave just like a regular exception.

src/tqec/compile/specs/base.py
```diff
@@ -70,3 +67,63 @@ def __call__(self, spec: CubeSpec) -> CompiledBlock:
...
+
+    Attributes:
+        src: a mapping from the index of the layer in the source `CompiledBlock` to
+            the plaquettes that should be used to update the layer.
+        dst: a mapping from the index of the layer in the destination `CompiledBlock`
+            to the plaquettes that should be used to update
+    """
+
+    src: Mapping[int, Plaquettes]
+    dst: Mapping[int, Plaquettes]
```
nelimee, line 114: I think that the convention in Python for typing is the following:
- "abstract" or "generic" types (such as `Mapping`) are used in places that are visible to users,
- concrete types are used everywhere else.

Is there any advantage in having `Mapping` instead of a regular `FrozenDefaultDict`/`dict`/`defaultdict` here?

src/tqec/compile/specs/library/_utils.py
```diff
@@ -1,107 +1,262 @@
...
+            4: p1.project_on_boundary(PlaquetteOrientation.DOWN),
+            11: p1,
+            12: p2,
+        }
+    else:  # substitution_side == PlaquetteSide.DOWN
+        mapping = {
+            3: p2.project_on_boundary(PlaquetteOrientation.LEFT),
+            13: p1,
+            14: p2,
+        }
```
nelimee, line 193: These two branches will break if any `Qubit4WayJunctionTemplate` is used. In general, there is a long-standing problem here which is "how to get the correct plaquette indices when performing substitution". We can leave the code like that, just adding a note or a TODO to be sure that we address that issue eventually.

src/tqec/compile/specs/library/_utils.py
nelimee, line 1: Looking at all the "private" functions in this file, I think it might be worth documenting them even though they are private.

src/tqec/gallery/logical_cnot.py
```diff
@@ -0,0 +1,42 @@
...
+from tqec.computation.zx_graph import ZXGraph
+from tqec.position import Position3D
+
+
+def logical_cnot_block_graph(support_z_basis_observables: bool = True) -> BlockGraph:
+    logical_cnot_zx = _create_logical_cnot_zx_graph(support_z_basis_observables)
+    return logical_cnot_zx.to_block_graph(logical_cnot_zx.name)
+
+
+def _create_logical_cnot_zx_graph(support_z_basis_observables: bool = True) -> ZXGraph:
```
nelimee, line 11: It might be worth leaving the `ZXGraph` creation function public, I do not see the downside of having that function public.

src/tqec/plaquette/library/css.py
```diff
@@ -8,64 +8,97 @@
...
-            If None, no measurement is performed.
-        x_boundary_orientation: The orientation of the X boundary of the surface
-            code block. Either "HORIZONTAL" or "VERTICAL". This determines the
-            CNOT order in the plaquette together with the basis to prevent hook
-            error from decreasing the code distance. Default is "HORIZONTAL".
+        basis: the basis of the plaquette.
+        x_boundary_orientation: the orientation of the X boundary.
+        data_initialization: the logical basis for data initialization.
+        data_measurement: the logical basis for data measurement.
+        init_meas_only_on_side: the side for data initialization and measurement.
```
nelimee, line 31:
```suggestion
        basis: the basis of the plaquette.
        data_initialization: the logical basis for data initialization.
        data_measurement: the logical basis for data measurement.
        x_boundary_orientation: the orientation of the X boundary.
        init_meas_only_on_side: the side for data initialization and measurement.
```
ordering of arguments in docstring

src/tqec/plaquette/library/css.py
```diff
@@ -8,64 +8,97 @@
...
+            {0: self._qubits.syndrome_qubits[0]}
+            | {i + 1: q for i, q in enumerate(self._qubits.data_qubits)}
+        )
+
+    def build(self) -> Plaquette:
+        return Plaquette(
+            self._qubits,
+            ScheduledCircuit(
+                self._moments,
+                schedule=list(range(len(self._moments))),
```
nelimee, line 66: Providing `schedule=0` should have the same effect.

src/tqec/plaquette/library/css_test.py
```diff
@@ -11,9 +11,58 @@
...
 def test_css_surface_code_memory_plaquette() -> None:
-    x_plaquette = make_css_surface_code_plaquette("X")
-    assert x_plaquette.qubits == SquarePlaquetteQubits()
-    assert x_plaquette.circuit.get_circuit() == stim.Circuit("""
+    plaquette = make_css_surface_code_plaquette("X")
+    assert plaquette.qubits == SquarePlaquetteQubits()
+    circuit = plaquette.circuit.get_circuit()
+    assert circuit.has_flow(stim.Flow("1 -> _XXXX xor rec[-1]"))
+    assert circuit.has_flow(stim.Flow("_XXXX -> 1 xor rec[-1]"))
+    assert circuit == stim.Circuit("""
```
nelimee, line 19: That's a nice way of testing, good addition to the tests in my opinion!

src/tqec/plaquette/library/zxxz.py
```diff
@@ -6,98 +6,144 @@
...
-            If None, no measurement is performed.
-        x_boundary_orientation: The orientation of the X boundary of the surface
-            code block. Either "HORIZONTAL" or "VERTICAL". This determines the
-            CNOT order in the plaquette together with the basis to prevent hook
-            error from decreasing the code distance. Default is "HORIZONTAL".
+        basis: the basis of the plaquette.
+        x_boundary_orientation: the orientation of the X boundary.
+        data_initialization: the logical basis for data initialization.
+        data_measurement: the logical basis for data measurement.
+        init_meas_only_on_side: the side for data initialization and measurement.
```
nelimee, line 34:
```suggestion
        basis: the basis of the plaquette.
        data_initialization: the logical basis for data initialization.
        data_measurement: the logical basis for data measurement.
        x_boundary_orientation: the orientation of the X boundary.
        init_meas_only_on_side: the side for data initialization and measurement.
```

src/tqec/plaquette/plaquette.py
```diff
@@ -58,8 +58,18 @@ def __post_init__(self) -> None:
...
 
+    def __eq__(self, rhs: object) -> bool:
+        return (
+            isinstance(rhs, Plaquette)
+            and self.qubits == rhs.qubits
+            and self.circuit == rhs.circuit
+        )
+
+    def __hash__(self) -> int:
+        return hash((self.qubits, str(self.circuit.get_circuit())))
```
nelimee, line 69: Note that due to #361 this will eventually need to change to have a reproducible hash.

src/tqec/circuit/generation_test.py
```diff
@@ -32,9 +32,9 @@ def _expected_circuit() -> stim.Circuit:
     R 0
     TICK
     TICK
-    TICK
     CX 1 0
     TICK
+    TICK
```
nelimee, line 37: Why did that change? Was there a bug in the previous implementation?

src/tqec/compile/observables.py
```diff
@@ -144,8 +144,10 @@ def inplace_add_observables(
         for pipe in observable.bottom_regions:
             z = pipe.u.position.z
             template = template_slices[z]
+            origin = template.origin_shift
+            shifted_pipe = pipe.shift_position_by(-origin.x, -origin.y)
```
nelimee, line 148: I do not get why you need to shift the pipe now. What changed to require this shift?

src/tqec/compile/compile_test.py
```diff
@@ -0,0 +1,206 @@
...
+
+@pytest.mark.parametrize(
+    ("spec", "cube_type", "k"),
+    itertools.product(
+        SPECS.keys(),
+        (CubeType.ZXZ, CubeType.ZXX, CubeType.XZX, CubeType.XZZ),
+        (1,),
+    ),
+)
+def test_compile_two_same_blocks_connected_int_time(
```
nelimee, line 57:
```suggestion
def test_compile_two_same_blocks_connected_in_time(
```

nelimee (commented):
I like how this is implemented. It is true that plaquette creation is a complex piece, but the current approach allows to decouple that task from all the others nicely, which helps a lot in robust testing and code review.

From what I remember, the main issue I see right now is with how substitution is implemented: the hard-coded indices are not always correct. This is fine as long as we only have "regular" cubes, but a 4-way (or 3-way) spatial junction would break that part of the code. They are not implemented yet though, so it is reasonable to not support them yet.

I will approve when we finished discussing the few comments I have made.

src/tqec/circuit/generation_test.py
inmzhang: Previously, I misuse the `x_boundary_orientation`. In all codes related to that, I considered it as the "logical X operator orientation" in my mental model. So the cnot ordering was actually wrong. But as I also wrongly parsed the orientation from `CubeType`, the effect canceled out and gave the correct results. Now I fixed the CNOT ordering as well as the orientation parsing part.

src/tqec/compile/specs/base.py
inmzhang: I think no. I will change it to `dict`.

src/tqec/compile/specs/library/_utils.py
inmzhang: Yes, in my plan, more substitution rules can be added as branched to `_build_substitution_in_space` and this one(`_build_substitution_in_space_with_regular_cubes`) should only work for regular cube connections.

src/tqec/circuit/generation_test.py
nelimee: Ok, I just re-checked the `CX` ordering in CSS `ZZ` stabilizer measurement on the left boundary and the updated code seems correct.

src/tqec/compile/specs/library/_utils.py
nelimee: I would rather advocate for an automated way of performing the substitution, but I am not sure it is doable in general (and I am sure it will require more code for this specific case), so I am fine with the current implementation.

src/tqec/compile/observables.py
inmzhang: The `LayoutTemplate` has some shifts in the space and we need to account for it when instantiating the abstract observables. Actually this is a bug left behind. I did not notice that before because the logical cnot and some other tests cases I used have all the `LayoutTemplate`s with origin at `Position2D(0, 0)`. I found this problem when I have added more test cases in `compile_test.py`.

src/tqec/compile/compile.py
```diff
@@ -46,7 +47,7 @@ def __post_init__(self) -> None:
         # Use warning instead of exception because a qec circuit without observable
         # may still be useful, e.g. do statistics on the detection events.
         if len(self.observables) == 0:
-            raise TQECWarning("The compiled graph includes no observable.")
+            warnings.warn("The compiled graph includes no observable.")
```
nelimee, line 50:
```suggestion
            warnings.warn("The compiled graph includes no observable.", TQECWarning)
```
this helps the user filter warnings if needed.

nelimee (approved):

src/tqec/compile/compile.py
nelimee: Note that the import also need to be included back.

src/tqec/compile/specs/library/_utils.py
inmzhang: Yes, we may find out some algorithm to automate the substitution one day. But now I'm not aware of a simple way to do that.

PR #365, opened 2024-10-16
Require `python<3.13` and add `typing-extensions` to requirements

### LINKED ISSUE #363: `typing_extensions` is not in `tqec` dependencies
## Describe the bug
The `typing_extensions` module is not importable from a fresh install, which makes most of `tqec` broken due to import errors.

1. `conda create -n tqec_test python=3.12`
2. `conda activate tqec_test`
3. `python -m pip install git+https://github.com/QCHackers/tqec.git`
4. `python -c "import tqec"`

## Expected behavior
The `tqec` module should be easily installable.

inmzhang:
Fix #363 

From @nelimee:

> - When trying to install tqec, I had an issue due to Python 3.13 being the newest Python version but most of our dependencies (and in particular stim) do not have yet pre-built binaries, leading to some issues related to CPython imports in stim.
> 
> - The `typing-extensions` package seems to be missing from our requirements. I though it was a standard python package, it seems like this is not the case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ description = "A design automation framework for Topological Quantum Error Corre
 readme = "README.md"
 license = { file = "LICENSE" }
 keywords = ["topological quantum error correction", "qec"]
-requires-python = ">= 3.10, <3.13"
+requires-python = ">= 3.10"
 classifiers = [
     "Development Status :: 4 - Beta",
     "Intended Audience :: Developers",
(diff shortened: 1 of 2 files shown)


nelimee:
So Python `3.13` is likely not an issue here. I do not think we have anything in the code base that would prevent us using Python 3.13. The issue was really local to my environment, likely due to the delay between the first Python release and the availability of pre-compiled binaries for our dependencies. I think we should still allow Python 3.13, people using it probably know that there will be some edge-cases due to the fact that it went out recently.

PR #371, opened 2024-10-17
Add name to `Plaquette`

### LINKED ISSUE #360: Add a `name` field to `Plaquette`
#350 requires to be able to hash `Plaquettes` instances, which in turn requires `Plaquette` instances to be hashable. Right now, the hash is not really robust: it is hashing the string representation of the underlying `stim.Circuit` instance, that **should** never be mutated, but that is not strictly enforced (we just cannot replace the `Plaquette.circuit` field, which does not mean that we cannot mutate the `circuit` field with mutating methods from `ScheduledCircuit`).

Moreover, another issue I had recently is my inability to display nicely `Plaquettes` instances when trying to debug the code.

Adding a unique and descriptive name to `Plaquette` instances (that should be made immutable in a more robust way, #358 goes in this direction) would solve the 2 issues above:

1. hashing a `Plaquette` instance would just be hashing its name, which is nice because this is a constant (and quick) operation.
2. debugging will be nicer as I will be able to print `Plaquettes` instances and have a readable output (e.g., `{1: "XX_up", 3: "ZZZZ", default: "XXXX"}`).

nelimee:
This PR adds a descriptive `name` field to each `Plaquette` instance. Because `Plaquette` instances are frozen, it also uses this unique name to implement `__eq__` and `__hash__`. Finally, it uses the name to represent a plaquette with `__str__`.

Fixes #360.

### DIFF
--- a/src/tqec/plaquette/plaquette.py
+++ b/src/tqec/plaquette/plaquette.py
@@ -65,6 +65,9 @@ class Plaquette:
     def __hash__(self) -> int:
         return hash(self.name)
 
+    def __str__(self) -> str:
+        return self.name
+
     def project_on_boundary(
         self, projected_orientation: PlaquetteOrientation
     ) -> Plaquette:
--- a/src/tqec/plaquette/plaquette.py
+++ b/src/tqec/plaquette/plaquette.py
@@ -60,14 +60,10 @@ class Plaquette:
         return Position2D(0, 0)
 
     def __eq__(self, rhs: object) -> bool:
-        return (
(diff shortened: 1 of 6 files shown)


src/tqec/plaquette/plaquette.py
```diff
@@ -40,6 +40,7 @@ class Plaquette:
             `qubits`.
     """
 
+    name: str
```
inmzhang, line 43: The documentation needs to explain the special effect of the `name` field.

inmzhang (approved):

PR #372, opened 2024-10-17
Adding `StimCoordinates` 

### LINKED ISSUE #359: Define a clear standard for coordinate order in `stim` exports
The current code make extensive use of several data-classes with explicit `x` and `y` attributes to avoid having to ask ourselves the question "should `x` coordinate be in the first position of this tuple or the second?". 

The only place where we still have that question is when we manipulate `stim` data-structures, that explicitly have no standard on coordinate order (see for examples the [`DETECTOR` instruction documentation](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#DETECTOR): `There is no required convention for which coordinate is which.`)

We should have a `tqec` standard on coordinate order for `stim` instructions, that we enforce on all `tqec`-generated `stim` data-structures.

Maybe a `StimInstructionCoordinates` data-structure with explicit `x` and `y` (and `z`, or `t`, ...?) attributes and a `to_stim_coords` method fixing the ordering. That would allow us to easily change the ordering if we find that our initial choice was bad.

nelimee:
Fixes #359 by defining coordinate order through the `StimCoordinates` class. Code in `tqec.circuit.detectors` does not use this because we try to keep the module independent of the outside `tqec` module.


src/tqec/compile/coordinates.py
```diff
@@ -0,0 +1,51 @@
...
+    def x(self) -> float:
+        return self.coordinates[0]
+
+    @property
+    def y(self) -> float:
+        return self.coordinates[1]
+
+    @property
+    def spatial(self) -> Position2D:
+        return Position2D(round_or_fail(self.x), round_or_fail(self.y))
```
inmzhang, line 34: Enforce the coordinates to be `int` in a property is weired. Do we really need this property, or simply return a `tuple[float, float]`?

inmzhang (approved):

inmzhang (approved):
LGTM

Issue #369, opened 2024-10-17
Add a CLI command to generate example circuits


inmzhang:
Similar to `stim.Circuit.generated`, it would be beneficial to add a CLI command like `tqec generated` that outputs example circuits along with the models for testing or other purposes.

Gistbatch:
I'll take this over together with #366

Issue #366, opened 2024-10-16
Add documentation to generate graphs


nelimee:
We need to add documentation that covers the steps to go from "how to generate a circuit" to "how to get plots of logical error VS physical error". 

The file [./examples/logical_cnot.py](./examples/logical_cnot.py) is a good example of what is needed to plot the graphs. We just need to include that into the quick start guide (both README and https://qchackers.github.io/tqec/quick_start.html).

nelimee:
There is now a guide in https://tqec.github.io/tqec/pull/381/user_guide/build_computations.html (link will be dead as soon as #381 is merged).

Issue #364, opened 2024-10-16
Improve simulation for multiple observables


inmzhang:
When simulating a circuit with multiple logical observables, currently we need to call `sinter.collect` one time for each observable. This can waste plenty of time because the computation circuit stays the same and in principle we only need to run the simulation once to get all the statistics for the `OBSERVABLE_INCLUDE`s in the circuit.

However, when I looked into `sinter` I found [no easy way](https://quantumcomputing.stackexchange.com/questions/29273/using-sinter-to-collect-error-rates-of-multiple-observables) to do that. 

One thing might be helpful is the `count_observable_error_combos` argument in `sinter.collect`:
```txt
count_observable_error_combos: Defaults to False. When set to to True,
            the returned stats will have a custom counts field with keys
            like `obs_mistake_mask=E_E__` counting how many times specific
            combinations of observables were mispredicted by the decoder.
```

I thought about setting the `custom_error_count_key` to one of the observables with the minimum correlation surface(i.e. we can assume it has relative low logical error rate) like `obs_mistake_mask=_E_`. Then `sinter` will apply `max_errors` to that observable to make the number of samples kind of reasonable for each observables. Some similar strategies might also be viable.

afowler:
+1 for solving this. We definitely only want to be running two simulations,
one for all output Z correlation surface, and one for all output X
correlation surfaces.

inmzhang:
In my mind, this issue is a great starting point for contributing to the community. It is not tightly coupled to the codebase, yet its resolution is very important and beneficial.

purva-thakre:
Hi @inmzhang, is anyone working on this issue? If not, I am interested in working on this!

KabirDubey:
> Hi @inmzhang, is anyone working on this issue? If not, I am interested! 

Hi @purva-thakre, I've actually started thinking about this issue recently, but I do not have a good solution in mind yet. I don't think any one else is working on it--please feel free to assign yourself. 

I'll assign myself as well to avoid confusion. Do you have a solution in mind or would it be helpful to discuss? Let me know and I'll email you. 

Thanks,
Kabir

purva-thakre:
Hi @KabirDubey , I am new to this project. 

I definitely don't have a well sketched-out idea for how to resolve the issue. I was hoping to familiarize myself with the project by working on the issue. 

I wouldn't mind working on this together. I don't have permission to assign myself.

inmzhang:
[...]

PR #377, opened 2024-10-20
Improve getting started 

Gistbatch:
Closes: #369, #366

I've extended the cli and the *getting started* docs.
I wasn't sure how much I should include, let me know what you think

### DIFF
--- a/docs/quick_start.rst
+++ b/docs/quick_start.rst
@@ -43,7 +43,7 @@ use the following code
 
 .. code-block:: python
 
-    observables, _ = block_graph.get_abstract_observables()
+    observables, correlation_surfaces = block_graph.get_abstract_observables()
 
 Any observable can be plotted using the ``tqec dae2observables`` command line. For our
 specific example, the command line
@@ -98,3 +98,72 @@ error-corrected implementation of a CNOT gate shown at the beginning of this pag
 
 You can download the circuit in a ``stim`` format here:
 :download:`media/quick_start/logical_cnot.stim <./media/quick_start/logical_cnot.stim>`.
+
+6. Simulate multiple experiments
+--------------------------------
+The circuit can be simulated using the ``stim`` and ``sinter`` libraries.
+Usually you want to simulate combinations of error rates and code distances, potentially
+for multiple observables.
+Multiple runs can be done in parallel using the ``sinter`` library using the ``start_simulation_using_sinter``.
+The compilation of the block graph is done automatically based on the inputs.
+
+.. code-block:: python
+
+    import numpy as np
+    
+    from tqec.noise_models import NoiseModel
+    from tqec.simulation.simulation import start_simulation_using_sinter
+
+    stats = start_simulation_using_sinter(
+        block_graph,
+        ks=range(1, 4), # k values for the code distance
+        ps=list(np.logspace(-4, -1, 10)), # error rates
+        noise_model_factory=NoiseModel.uniform_depolarizing # noise model
+        observables=[observables[1]], # observable of interest
+        decoders=["pymatching"],
+        num_workers=20,
+        max_shots=10_000_000,
+        max_errors=5_000,
+        print_progress=True,
+    )
+
+``Sinter`` can be additionally supplied with simulation parameters see
+`sinter_api https://github.com/quantumlib/Stim/blob/main/doc/sinter_api.md`_
+
+
+7. Plot the results
+-------------------
+Simulation Results can be plotted with ``matplolib`` using the ``plot_simulation_results``.
+
+.. code-block:: python
+
+    from tqec.simulation.plotting import plot_simulation_results
+
+    for i, stat in enumerate(stats):
+        fig, ax = plt.subplots()
+        sinter.plot_error_rate(
+            ax=ax,
+            stats=stat,
+            x_func=lambda stat: stat.json_metadata["p"],
+            group_func=lambda stat: stat.json_metadata["d"],
+        )
+        plot_observable_as_inset(ax, zx_graph, correlation_surfaces[i])
+        ax.grid(axis="both")
+        ax.legend()
+        ax.loglog()
+        ax.set_title(f"Logical CNOT Error Rate")
+        fig.savefig(f"logical_cnot_result_x_observable_{i}.png")
+
+8. Conclusion
+-------------
+This quick start guide has shown how to use the ``tqec`` library to define a computation,
+import it into the library, compile it to stim circuits.
+Simulations are run and visualized for multiple error rates and code distances.
+For an extensive example, see also the 
+`tqec_example <https://github.com/QCHackers/tqec/blob/main/examples/logical_cnot.py`_.
+
+The process can be repeated through the cli using
+
+..code-block:: bash
+
+    tqec run-example --out-dir ./results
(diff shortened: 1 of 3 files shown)


docs/quick_start.rst
```diff
@@ -98,3 +98,74 @@ error-corrected implementation of a CNOT gate shown at the beginning of this pag
...
+
+    from tqec.noise_models import NoiseModel
+    from tqec.simulation.simulation import start_simulation_using_sinter
+
+    stats = start_simulation_using_sinter(
+        block_graph,
+        ks=range(1, 4), # k values for the code distance
+        ps=list(np.logspace(-4, -1, 10)), # error rates
+        noise_model_factory=NoiseModel.uniform_depolarizing # noise model
+        observables=[observables[1]], # observable of interest
```
nelimee, line 122: Note that, using `start_simulation_using_sinter`, multiple observables can be provided. For the moment they are computed one after the other, but #364 aims at making that more efficient.

docs/quick_start.rst
```diff
@@ -98,3 +98,74 @@ error-corrected implementation of a CNOT gate shown at the beginning of this pag
...
+        decoders=["pymatching"],
+        num_workers=20,
+        max_shots=10_000_000,
+        max_errors=5_000,
+        print_progress=True,
+    )
+
+``Sinter`` can be additionally supplied with simulation parameters see
+`sinter_api https://github.com/quantumlib/Stim/blob/main/doc/sinter_api.md`_
```
nelimee, line 131: For the moment, `sinter` parameters that do not appear in `start_simulation_using_sinter` cannot be forwarded to `sinter`. This is something that may be interesting to fix.

docs/quick_start.rst
```diff
@@ -98,3 +98,74 @@ error-corrected implementation of a CNOT gate shown at the beginning of this pag
...
+
+    for i, stat in enumerate(stats):
+        fig, ax = plt.subplots()
+        sinter.plot_error_rate(
+            ax=ax,
+            stats=stat,
+            x_func=lambda stat: stat.json_metadata["p"],
+            group_func=lambda stat: stat.json_metadata["d"],
+        )
+        plot_observable_as_inset(ax, zx_graph, correlation_surfaces[i])
```
nelimee, line 152: Here, `correlation_surfaces[i]` is not really correct. `start_simulation_using_sinter` will return a list with `len(observables)` entries, `observables` being the list of observables provided to the function. Because in code cell before you only included `observables=[observables[1]]`, `stats` will only include one entry, corresponding to `observables[1]`, and so corresponding to `correlation_surfaces[1]` and not `correlation_surfaces[0]`.

src/tqec/_cli/subcommands/run_example.py
```diff
@@ -0,0 +1,178 @@
...
+    @staticmethod
+    @override
+    def add_subcommand(
+        main_parser: argparse._SubParsersAction[argparse.ArgumentParser],
+    ) -> None:
+        parser: argparse.ArgumentParser = main_parser.add_parser(
+            "run-examples",
+            description=(
+                "Runs the full pipeline on the CNot example."
+                "block graph -> stim circuits and observables -> simulation -> plots"
```
nelimee, line 35:
```suggestion
                "Runs the full pipeline on the CNOT example. "
                "block graph -> stim circuits and observables -> simulation -> plots"
```

src/tqec/_cli/subcommands/run_example.py
```diff
@@ -0,0 +1,178 @@
...
+            block_graph,
+            ks,
+            ps,
+            NoiseModel.uniform_depolarizing,
+            block_builder=block_builder,
+            substitution_builder=substitution_builder,
+            observables=[observables[i] for i in obs_indices],
+            num_workers=20,
+            max_shots=10_000_000,
+            max_errors=5_000,
```
nelimee, line 144: Maybe add parameters in the CLI for these, or use a better default value for `num_workers` (e.g., `multiprocessing.cpu_count()`?)

src/tqec/_cli/subcommands/run_example.py
```diff
@@ -0,0 +1,178 @@
...
+        parser.add_argument(
+            "--add-detectors",
+            help="Whether to add detectors to the circuits.",
+            action="store_true",
+        )
+        parser.add_argument(
+            "--generate-only",
+            help="Whether to run generation only.",
+            action="store_true",
+        )
```
nelimee, line 78: Are these arguments still used?

src/tqec/_cli/subcommands/run_example.py
```diff
@@ -0,0 +1,178 @@
...
+        parser.add_argument(
+            "--code-style",
+            help="Use the CSS or ZXXZ code style.",
+            choices=["css", "zxxz"],
+        )
+        parser.add_argument(
+            "--include-z-basis",
+            help="Whether to include the Z basis in the simulation.",
+            action="store_true",
+        )
```
nelimee, line 88: Maybe change to `--basis {x,z}`? That would be easier to understand in my opinion.

src/tqec/_cli/subcommands/run_example.py
```diff
@@ -0,0 +1,178 @@
...
+        parser.add_argument(
+            "--generate-only",
+            help="Whether to run generation only.",
+            action="store_true",
+        )
+        parser.add_argument(
+            "--code-style",
+            help="Use the CSS or ZXXZ code style.",
+            choices=["css", "zxxz"],
+        )
```
nelimee, line 83: Should be either required or have a default value.

nelimee (approved):
I tested the CLI command, you should also add it to `tqec/_cli/tqec.py`:


```py
import argparse
import sys

from tqec._cli.subcommands.check_dae import CheckDaeTQECSubCommand
from tqec._cli.subcommands.dae2circuits import Dae2CircuitsTQECSubCommand
from tqec._cli.subcommands.dae2observables import Dae2ObservablesTQECSubCommand
from tqec._cli.subcommands.run_example import RunExamplesTQECSubCommand


def main() -> None:
    parser = argparse.ArgumentParser(
        prog="tqec",
        description="The main tqec command-line tool.",
    )
    subparser = parser.add_subparsers(title="subcommands")

    Dae2ObservablesTQECSubCommand.add_subcommand(subparser)
    CheckDaeTQECSubCommand.add_subcommand(subparser)
    Dae2CircuitsTQECSubCommand.add_subcommand(subparser)
    # HERE BELOW
    RunExamplesTQECSubCommand.add_subcommand(subparser)

    args = parser.parse_args(args=None if sys.argv[1:] else ["--help"])
    args.func(args)
```

docs/quick_start.rst
Gistbatch: I just followed the example further up in the document

src/tqec/_cli/subcommands/run_example.py
Gistbatch: I wasn't sure how much to add for the cli. If this was a general `run-simulation` command I'd agree but since this is just generating the example I don't think this is of much use in the cli. I'll change the default though.

nelimee: I think that at least the number of CPU should be changed to something that depends on the host machine. Running 20 processes on a dual-core CPU makes no sense.

Gistbatch:
@nelimee, are any more changes necessary?

nelimee (commented):
It seems fine, thanks for the PR! :)

PR #400, opened 2024-11-13
Error handling during detector assignment

Brendan-Reid1991:
Minor change to the code to raise an informative error message if an exception occurs during automatic detector assignment, due to a missing `QUBIT_COORDS` annotation.

Single test added also.

### DIFF
--- a/src/tqec/circuit/detectors/boundary.py
+++ b/src/tqec/circuit/detectors/boundary.py
@@ -201,6 +201,10 @@ class BoundaryStabilizer:
         Args:
             qubit_coordinates: mapping from qubit indices to coordinates
 
+        Raises:
+            TQECException: If a qubit in self.source_qubits is not contained
+            in the qubit_coordinates mapping.
+
         Returns:
             the boundary stabilizer coordinates.
         """


Gistbatch (approved):

Gistbatch (approved):

Brendan-Reid1991:
Thank you for the approval guys, and for the quick responses! I'm unable to merge the PR, most likely to being outside the TQEC org, which is maybe for the best as I would rather you all know nothing is going to break prior to a merge. Thanks again.

nelimee:
> Thank you for the approval guys, and for the quick responses! I'm unable to merge the PR, most likely to being outside the TQEC org, which is maybe for the best as I would rather you all know nothing is going to break prior to a merge. Thanks again.

I will merge this PR. I cannot remember if only maintainers can merge or if CI is required to pass in order to be able to merge and only maintainers can bypass this restriction. Anyway, this is also something that we will need to check on our end. 

Thanks a lot for the contribution. As a side note, we would be very interested in hearing what you are using the `tqec` tool for :)

PR #405, opened 2024-11-22
Stricter typing

nelimee:
This small PR increases the strictness of the type checkers (both `mypy` that is used in CI and `Pyright` that is used locally in VS Code).


src/tqec/__init__.py
inmzhang, line 1: `__all__ = [...]` is an alternative for `from A import B as B`. I'm not sure which one is better.

inmzhang (approved):

src/tqec/__init__.py
nelimee: I tend to prefer `from A import B as B` to keep the information local. The problem with `__all__` is that when a new re-export is added, or when one is removed, the person has to also think about modifying the `__all__`, which is not an issue with `from A import B as B` as this is on the same line.

Issue #407, opened 2024-12-02
Move `tqec.circuit.detectors` out of the `tqec` repository


nelimee:

Gistbatch:
are you working on this @nelimee? If not, I can take this one

nelimee:
> are you working on this @nelimee? If not, I can take this one

I'll take the last point on moving issues.

The two middle points require to change things in the `tqec` repository, and I do not want to handle the merge with #381. If you want, you can take the 2 middle points.

PR #409, opened 2024-12-03
Feature/tqecd

Gistbatch:
closes: #407 

I removed the whole subdirectory after doing a reverse search for
- `annotate_detectors_automatically`
- `from tqec.circuit.detectors`

Since `tqecd` is not a pip package yet, I added it as a git dependency to the requirements.
I updated the docs pointing to the new read the docs main page.


Gistbatch:
@nelimee, any good idea how to exclude all the `ticed` submodules? Or should I wait until there's a proper export in that package?

docs/user_guide/detectors.rst
nelimee, line 1: We can likely remove that entry. Having a full page just for that might be too much. I would rather add a note somewhere.

pyproject.toml
```diff
@@ -118,6 +118,7 @@ module = [
     "collada",
     "collada.source",
     "mpl_toolkits.mplot3d.axes3d",
+    "tqecd",
```
nelimee, line 121: In theory we should not have to include `tqecd` here, as it is fully typed.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pycollada
 typing-extensions
 pycryptosat!=5.11.23
 python-sat[cryptosat]
+tqecd @ git+https://github.com/tqec/tqecd.git
```
nelimee, line 9: I'll create a tag on the `tqecd` repository, we should use that tag. Not using the tag might be an issue because the code will likely change, and we would like to avoid breaking the `tqec` package due to changes in `tqecd`.

nelimee (changes requested):
We should wait for `tqecd` to have a proper API in my opinion. Right now the PR looks good to me, modulo my remarks. On `tqecd` side we should:

- add a proper `__init__.py` file to re-export the public API,
- add a tag to use it in the `tqecd` dependency for `tqec`.

pyproject.toml
nelimee: Tests are now passing thanks to changes in `tqecd` (the `py.typed` file was not correctly placed). This should definitely be removed.

nelimee:
Copy pasting what I have posted in one of the issues:

------

Here is the chain of things that seem to be blocking this ~~issue~~ PR:

1. We want a bare-bone documentation of `tqecd` (see https://github.com/tqec/tqecd/pull/8).
2. Once the documentation is OK, we want to tag `tqecd` in order to avoid having a git link pointing on the last commit of the `main` branch as a dependency (see https://github.com/tqec/tqecd/pull/10).
3. The `tqecd` dependency should be made simply `tqecd`, or even better `tqecd~=0.1.0` (in beta stage, i.e. when the major version is still `0`, minor changes are considered as API breaking, hence this version specifier). 

Once the above 3 points are done, this PR will be merged and this issue will be closed.

nelimee:
@Gistbatch the external modifications required by this PR are in place. We only need to:

- [ ] merge `main` here and resolve conflicts.
- [ ] add `tqecd~=0.1` to the requirements.

Do you want to do it, or should I?

requirements.txt
nelimee, line 9:
```suggestion
tqecd~=0.1
```

nelimee: You also need to change `pyproject.toml` with the appropriate requirement.

nelimee (changes requested):

Gistbatch:
@nelimee I rebased onto main could you quickly check if I did this correctly?
Also, I'm not sure where I would need to add this in the `pyproject.toml` as the dependencies are generated dynamically?

nelimee:
> @nelimee I rebased onto main could you quickly check if I did this correctly? Also, I'm not sure where I would need to add this in the `pyproject.toml` as the dependencies are generated dynamically?

You are right, `requirements.txt` is enough. I'll accept the PR, tell me if you can merge it (you should be able to) or if I should do it.

Gistbatch:
Can you quickly check the docs if they still reflect the changes correctly? I can merge after

nelimee (approved):
Both `python -m pytest .` and `python -m mypy src/tqec` succeed after a fresh install. I checked that `tqecd` was pulled from PyPI. Thanks for the PR!

nelimee:
> Can you quickly check the docs if they still reflect the changes correctly? I can merge after

Will do, I will re-submit a review

nelimee (changes requested):
In https://tqec.github.io/tqec/pull/409/user_guide/installation.html, the `pycryptosat` is now an indirect dependency of `tqec` as it is a dependency of `tqecd`. 
I guess we could remove that part from the user guide, because `tqecd` also has it.

Except that, I think the documentation is fine.

nelimee (approved):
Seems fine, thanks for the PR!

Issue #410, opened 2024-12-04
Create plaquette from RPNG string


giangiac:
As discussed on 12/04, we decided to describe the plaquettes with a 16 character string (and 3 spaces).
See below for a description of the format.

We need to have a function to validate the RPNG string and a function that generate the plaquette given the RPNG string.

----

**Example of a RPNG string and description of the format:**

```
-z1- -z2- -z3- -z4-
rpng rpng rpng rpng
```

(r) data qubit reset basis or h or -  
(p) data basis for the controlled operation (x means CNOT controlled on the ancilla and targeting the data qubit, y means CY, z means CZ)  
(n) time step (positive integers, all distinct, typically in 1-5)  
(g) data qubit measure basis or h or -

Assumptions on the circuit:
- if not otherwise stated, a basis can be {x,y,z}
- the ancilla is always initialized in $\ket{+}$ and measured in the X basis
- the ancilla is always the control qubit for the CNOT, CY, or CZ gates

giangiac:
I started working on this new feature in my fork (branch `feature/generate-plaquette-from-rpng`):
https://github.com/giangiac/tqec/tree/feature/generate-plaquette-from-rpng

nelimee:
> I started working on this new feature in my fork (branch `feature/generate-plaquette-from-rpng`): https://github.com/giangiac/tqec/tree/feature/generate-plaquette-from-rpng

Could you open a PR in order for everyone to see what's happening?

nelimee:
Also note that it would be nice to have a generic and documented interface, so that users will be able to implement their own translation if they want.

PR #411, opened 2024-12-04
Create plaquette from RPNG string

giangiac:
See issue https://github.com/tqec/tqec/issues/410 in the tqec repo.

This PR includes:
- [x] a function to validate the RPNG string
- [x] a function that generates the plaquette given the RPNG string
- [ ] **[for a future PR]** a generic and documented interface, so that users will be able to implement their own translation if they want

----

Example of a RPNG string and description of the format:
```
-z1- -z2- -z3- -z4-
rpng rpng rpng rpng
```
(r) data qubit reset basis or h or -
(p) data basis for the controlled operation (x means CNOT controlled on the ancilla and targeting the data qubit, y means CY, z means CZ)
(n) time step (positive integers, all distinct, typically in 1-5)
(g) data qubit measure basis or h or -

Assumptions on the circuit:
- if not otherwise stated, a basis can be {x,y,z}
- the ancilla is always initialized in $\ket{+}$ and measured in the X basis
- the ancilla is always the control qubit for the CNOT, CY, or CZ gates


afowler:
Just note typo in the below: z means CZ.

giangiac:
@nelimee I am not sure why the CI/CD is failing.

Also, I am thinking if it is worth merging with the canonical form of the circuits expressed as RPNG strings, leaving the general interface for custom compilation for another PR.

inmzhang:
Hi Gian, 

> I am not sure why the CI/CD is failing.

You need to install and run the [pre-commit hooks](https://pre-commit.com/) locally. And the deploy failure is because you did not create the branch under the tqec repository but use your fork instead, it will cause permission issue.

> Also, I am thinking if it is worth merging with the canonical form of the circuits expressed as RPNG strings, leaving the general interface for custom compilation for another PR.

In my opinion, they should be in separate PR.

giangiac:
> > I am not sure why the CI/CD is failing.
> 
> You need to install and run the [pre-commit hooks](https://pre-commit.com/) locally. And the deploy failure is because you did not create the branch under the tqec repository but use your fork instead, it will cause permission issue.

Would you suggest creating a new branch in the official tqec repo, merge my fork there, and then create a PR internal to tqec?
In general I would find it more convenient to work directly in the tqec repo, but didn't know if that's a good practice.

giangiac:
> A few preliminary notes on the current state of the implementation:
> 
> * The functions are too complex. There are too much if/else branches, too much "magic" (e.g., `v[0]` could be renamed `r` in most cases, to make it easier to read).
> * I think that a dataclass with some methods would be perfect to wrap one RPNG representation. The dataclass would be responsible to validate that the provided RPNG string is locally valid (i.e. that we have correct values for R, P and G) and would provide methods to ease the global validation (e.g., a property `n` returning the timestep at which the 2-qubit gate is scheduled).
> * The dataclass could also help for circuit generation, for example by returning the correct reset/measurement/computation gates using methods.

Thanks for the early feedback, Adrien.

I can wrap the rpng strings into a dataclass to facilitate querying the values of interest.
Is the idea to facilitate writing custom circuits from the rpng?
Or do you think it would affect how the rest of the backend uses it?
(in my mind we just need the plaquette given a valid rpng string)

I'll include the validation inside that class, but would not mind keeping its "complex" implementation for the moment (there is no urgency here and no risk of computational slowdowns).

nelimee:
> I can wrap the rpng strings into a dataclass to facilitate querying the values of interest.
> Is the idea to facilitate writing custom circuits from the rpng?
> Or do you think it would affect how the rest of the backend uses it?
> (in my mind we just need the plaquette given a valid rpng string)

See for example #412. I think we need a simple and robust representation of a RPNG string, and a RPNGPlaquette, as we might have to use them in several places. 

> I'll include the validation inside that class, but would not mind keeping its "complex" implementation for the moment (there is no urgency here and no risk of computational slowdowns).

The problem is mostly about code reuse. I think it is better to have most of the validation code in one place, and #412 would need a little bit more flexibility, so I think we shoule have reusable structures if we can.

giangiac:
> See for example #412. I think we need a simple and robust representation of a RPNG string, and a RPNGPlaquette, as we might have to use them in several places.

Thanks for the concrete example!
I adopted most of it (I am unfamiliar with the template part so I left it out) and the PR should be ready for review.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+        prep_time = 0
+        meas_time = self.ancilla.n
+        circuit_as_list = [''] * (meas_time - prep_time + 1)
+        for q, rpng in enumerate(self.corners):
+            # 2Q gates.
+            if rpng.n and rpng.p:
+                if rpng.a == BasisEnum.Z:
+                    circuit_as_list[rpng.n] += f'C{rpng.p.value.upper()} 4 {q}\n'
+                else:
+                    circuit_as_list[rpng.n] += f'C{rpng.a.value.upper()} {q} 4\n'
```
inmzhang, line 231:
```suggestion
            if rpng.n and rpng.p:
                circuit_as_list[rpng.n] += f'{rpng.a.value.upper()}C{rpng.p.value.upper()} 4 {q}\n'
```

`stim` supports `[X|Y|Z]C[X|Y|Z]` gates like `ZCZ` or `ZCX`. We can simplify the logic here.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+
+class ExtendedBasisEnum(Enum):
+    X = BasisEnum.X.value
+    Y = BasisEnum.Y.value
+    Z = BasisEnum.Z.value
+    H = 'h'
+
+
+@dataclass
+class RPNG:
```
inmzhang, line 28: We should group the docstring for attributes of a `dataclass` into the `Attribute` section.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
+from tqec.circuit.schedule import ScheduledCircuit
+from tqec.circuit.qubit_map import QubitMap
+from tqec.plaquette.plaquette import Plaquette
+from tqec.plaquette.qubit import PlaquetteQubits
+from tqec.plaquette.qubit import SquarePlaquetteQubits
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Optional
```
inmzhang, line 9: We can use `A | None` instead of `Optional[A]` for consistency in the codebase.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+        a_str = 'z'
+        if len(rpng_string) == 4:
+            r_str, p_str, n_str, g_str = tuple(rpng_string)
+        elif len(rpng_string) == 5:
+            r_str, a_str, p_str, n_str, g_str = tuple(rpng_string)
+        else:
+            raise ValueError('The rpng string must be exactly 4- or 5-character long.')
+        # Convert the characters into the enum attributes (or raise error).
+        r = ExtendedBasisEnum(r_str) if r_str in ExtendedBasisEnum._value2member_map_ else None
+        a = BasisEnum(a_str) if a_str in BasisEnum._value2member_map_ else None
```
inmzhang, line 85:
```suggestion
        a = BasisEnum(a_str) if a_str in BasisEnum else None
```
This should work.

inmzhang (commented):
Some format and lint issue need to be fixed with running pre-commit hooks.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+
+    ----
+    Extended RPNG format:
+
+    The extra field (a) is used together with (p) to determine the 2Q gate involving the
+    ancilla and data qubits. For example:
+    - a=z, p=x --> CNOT controlled by the ancilla and targeting the data
+    - a=z, p=z --> CZ   controlled by the ancilla and targeting the data
+    - a=y, p=z --> CY   controlled by the data    and targeting the ancilla
+    By default a=z.
```
nelimee, line 56: This is not needed anymore I think.
```suggestion
```

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    r: Optional[ExtendedBasisEnum]
```
nelimee, line 58: Can't remember what we settled on, but I think the `h` entry was initially only on the measurement part. I just wanted to leave that as a note, I personally prefer to have it on both measurement and reset.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+    p: Optional[BasisEnum]
+    n: Optional[int]
+    g: Optional[ExtendedBasisEnum]
```
nelimee, line 61:
```suggestion
    r: ExtendedBasisEnum | None
    p: BasisEnum | None
    n: int | None
    g: ExtendedBasisEnum | None
```

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+        """Initialize the RPNG object from a 4-character or 5-character string
+        
+        4-character -> simplified format -> RPNG
+        5-character -> extended format -> RAPNG
+
+        Note that the ancilla basis(a) is the second character, just before (p).
+
+        Note that any character different from a BasisEnum / ExtendedBasisEnum
+        value would result in the corresponding field being None.
+        """
```
nelimee, line 75: Do we still want to support the length-5 format?

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+            raise ValueError('Unacceptable character for the G field.')
+        return cls(r, p, n, g, a)
+
+    
+    def get_r_op(self) -> Optional[str]:
+        """Get the reset operation or Hadamard"""
+        op = self.r
+        if op == None:
+            return None
+        elif op.value in BasisEnum._value2member_map_:
```
nelimee, line 108:
```suggestion
        elif op.value in BasisEnum:
```
should be enough.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+            return f'R{op.value.upper()}'
+        else:
+            return f'{op.value.upper()}'
+
+    def get_g_op(self) -> Optional[str]:
+        """Get the measurement operation or Hadamard"""
+        op = self.g
+        if op == None:
+            return None
+        elif op.value in BasisEnum._value2member_map_:
```
nelimee, line 118:
```suggestion
        elif op.value in BasisEnum:
```
should be enough.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+            return f'M{op.value.upper()}'
+        else:
+            return f'{op.value.upper()}'
```
nelimee, line 121: These two methods can very much be factored together. For example:
```suggestion
    def get_op(self, function: Literal["meas", "reset"]) -> Optional[str]:
        """Get the operation or Hadamard"""
        op = self.r if function == "reset" else self.g
        if op == None:
            return None
        elif op.value in BasisEnum:
            return ("R" if function == "reset" else "M") + op.value.upper()
        else:
            return f'{op.value.upper()}'
```

nelimee: Note that I did not test the above code, `Literal` is also not imported, so better apply it in your code editor if you think this is a good suggestion.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@dataclass
+class RGN:
+    """Organize the prep and meas bases for the ancilla, together with the meas time
+    
+    The init time is assumed to be 0.
```
nelimee, line 128: It may be nice to have an attribute for the reset time, that defaults to `0`, but that is changeable by the user? Not sure if that is a meaningful thing to add though, it might be giving too much freedom to the user.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+            raise ValueError('The rgn string must be exactly 3-character long.')
+        r_str, g_str, n_str = tuple(rgn_string)
+
+        try:
+            r = BasisEnum(r_str)
+            g = BasisEnum(g_str)
+            n = int(n_str)
+            return cls(r, g, n)
+        except ValueError:
+            raise ValueError('Invalid rgn string.')
```
nelimee, line 147: Maybe add stack-trace information:
```suggestion
        except ValueError as err:
            raise ValueError('Invalid rgn string.') from err
```

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@dataclass
+class RPNGDescription:
+    """Organize the description of a plaquette in RPNG format
+    
+    The corners of the square plaquette are listed following the order:
+    tl, tr, bl, br
```
nelimee, line 155: This should be explicitly written, and it might even be nice to have a visual representation. Something like
```suggestion
    top-left, top-right, bottom-left, bottom-right. That is a Z-shaped path on the plaquette data-qubits:
    
        +------------+
        | 1 -----> 2 |
        |        /   |
        |      /     |
        |    ∟       |
        | 3 -----> 4 |
        +------------+
```

nelimee (commented):
I have made a few comments. That is only a first-pass, I will have comments on the implementation of a few methods too, but I think I should wait for a few questions to be answered, in particular:

- Should be support the 5-char notation or not? I am not sure we really need that notation, and it adds 1) a lot of complexity and 2) unusual gates to the resulting circuits.

nelimee:
Note to fix CI errors: follow the [contributor guide](https://tqec.github.io/tqec/contributor_guide.html), in particular the parts related to `pre-commit`. If you do not want to re-install from scratch:

```python
python -m pip install pre-commit
pre-commit install
pre-commit run -a
# Re-run to check that everything now pass the tests
pre-commit run -a
git commit -a -m "Fixing pre-commit"
git push
```

afowler:
I would vote against the 5-character notation. In my mind if you need to go
to a circuit that cannot be expressed with the four character notation, you
probably wouldn't express that circuit using plaquettes in the first place.

H can be on both reset and measurement as a typical expansion of cnot into
CZ can easily require H on both ends of the circuit.

I think reset time should always be zero, and the measurement time should
be the latest time of a gate in an entire circuit layer plus one.

Best,
Austin.

inmzhang:
> I would vote against the 5-character notation. In my mind if you need to go
to a circuit that cannot be expressed with the four character notation, you
probably wouldn't express that circuit using plaquettes in the first place.

For circuits that cannot be represented in a canonical plaquette form, we return to the issue of whether we should allow specifying a scalable circuit directly, as an alternative to the template + plaquette method. This might be necessary if we adhere strictly to the canonical plaquette definition. However, this approach seems to have a couple of challenges:

- The replacement rules currently only apply to plaquettes. We would need a new approach to implement pipes.
- The detector computation based on subtemplates won't work with this approach and would need to be modified.
- The current compilation pipeline heavily relies on the template and plaquettes, it will need to be changed.

afowler:
The only structures I believe will need to be implemented as scalable
circuits are state cultivation and Y basis measurement and initialization.
These are all terminal structures, with I believe a standard surface code
layer that could be specified using plaquettes to end or begin the
structures that could be used to connect in the usual way.

Best,
Austin.

src/tqec/plaquette/rpng_description.py
giangiac: Added in a commit.

src/tqec/plaquette/rpng_description.py
giangiac: Applied.

src/tqec/plaquette/rpng_description.py
giangiac: The suggestion does not work but alternatives may be:
`a_str in [item.value for item in BasisEnum]`
`BasisEnum(a_str) in BasisEnum`

If there is no specific issue, I will keep the original expression.

src/tqec/plaquette/rpng_description.py
giangiac: Actually I adopted the old formulation adding the general one as a fail-safe case.
I think this facilitates reading the stim code and eliminates "unfamiliar" gate notations from the default circuits.

src/tqec/plaquette/rpng_description.py
giangiac: The extra field (a) does not affect the RPNG description when left to its default value z.
We can move all generalizations of the basic RPNG description somewhere else, but I thought that they are harmless and may be helpful to users in need to customize the plaquette circuits.

src/tqec/plaquette/rpng_description.py
giangiac: I don't mind leaving the user more freedom to define plaquettes.
The only reason to remove the option seems to be to avoid mistyping the RPNG string (without being flagged by the RPNG validation).

src/tqec/plaquette/rpng_description.py
giangiac: Already committed

src/tqec/plaquette/rpng_description.py
giangiac: The suggestion does not work but alternatives may be:
`op.value in [item.value for item in BasisEnum]`
`BasisEnum(op.value) in BasisEnum`

If there is no specific issue, I will keep the original expression.

src/tqec/plaquette/rpng_description.py
giangiac: See above.

src/tqec/plaquette/rpng_description.py
nelimee: The extra field:

- is 4 more characters to type each time a user wants a Plaquette,
- makes the code more complex.

We can keep it, but if we do I would advocate for it to be completely separate from the "normal" `RPNG` notation. Basically, have `RPNG` only represent a 4-character string, and have another class (`ERPNG`? `RPNGA`? `RAPNG`?) represent the 5-character notation. The 4-character version will be used extensively in the code, and so it should be easy to read, the simplest possible, and as robust as possible, which is in my opinion incompatible with having both the 4- and 5-character approaches in the same implementation.

afowler:
+1 for careful separation, but I'd be even the happier if we stuck with
just the four character representation exclusively and focused on good
conversion tools for other gate sets.

src/tqec/plaquette/rpng_description.py
nelimee: This is unlikely to be useful in the future, so you can just ignore that comment.

nelimee:
@giangiac the goal is to merge that PR as soon as possible to unlock other issues. I'll make a real full review, and we will be able to merge once the review will be addressed.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+    The init time is assumed to be 0.
+
+    Attributes:
+        r   ancillaqubit reset basis  
+        g   ancilla qubit measure basis
+        n   time step of ancilla measurement
+    """
+    r: BasisEnum = BasisEnum.X
+    g: BasisEnum = BasisEnum.X
+    n: int = 6
```
nelimee, line 143: I am pretty sure `n` should not be included in this representation.
```suggestion
```
The problem of including `n` is that it forces the user to have a value (either user-provided, or the default `6`), which is highly misleading because we will eventually impose that resets and measurements all happen at the same timestep at compilation, so the specific value provided here will likely not be respected.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+        r_str, g_str, n_str = tuple(rgn_string)
+
+        try:
+            r = BasisEnum(r_str)
+            g = BasisEnum(g_str)
+            n = int(n_str)
+            return cls(r, g, n)
+        except ValueError:
+            raise ValueError('Invalid rgn string.')
+
```
nelimee, line 159: I am not even sure we need to enclose this in a try-catch block. Why not just letting the potential exception from `BasisEnum` or `int` flow upwards in the stack? 
On the other side, having a clear description like "Invalid RGN string" is nice.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+    corners: tuple[RPNG, RPNG, RPNG, RPNG]
+    ancilla: RGN = RGN()
+
+    def __post_init__(self):
+        """Validation of the initialization arguments
+        
+        Constraints:
+        - the n values for the corners must be unique
+        - the n values for the corners must be in the interval ]0, ancilla.n[
+        - at least one betwween fields a and p in every corner is z
```
nelimee, line 182: These conditions should be removed for RPNG. Maybe for Extended RPNG, but I am unsure of its usefulness.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+            if rpng.p and rpng.a != BasisEnum.Z and rpng.p != BasisEnum.Z:
+                raise ValueError('At least one between a and p must be "z" in every corner.')
+        if len(times) != len(set(times)):
+            raise ValueError('The n values for the corners must be unique.')
+        elif len(times) not in [0,2,4]:
+            raise ValueError('Each plaquette must have 0, 2, or 4 2Q gates.')
+
+
+    @classmethod
+    def from_string(cls, corners_rpng_string: str) -> 'RPNGDescription':
```
nelimee, line 200: Add 
```py
```
at the beginning of the imports and then 

```suggestion
    def from_string(cls, corners_rpng_string: str) -> RPNGDescription:
```

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+        """Initialize the RPNGDescription object from a (16+3)-character string"""
+        rpng_objs = tuple([RPNG.from_string(s) for s in corners_rpng_string.split(' ')])
+        if len(rpng_objs) != 4:
+            raise ValueError('There must be 4 corners in the RPNG description.')
+        return cls(rpng_objs)
+    
+
+    @classmethod
+    def from_extended_string(cls, ancilla_and_corners_rpng_string: str) -> 'RPNGDescription':
```
nelimee, line 209: Add 
```py
```
at the beginning of the imports and then 

```suggestion
    def from_extended_string(cls, ancilla_and_corners_rpng_string: str) -> RPNGDescription:
```

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def get_g_op(self, data_idx: int) -> str | None:
+        """Get the measurement operation or Hadamard for the specific data qubit"""
+        return self.corners[data_idx].get_g_op()
+    
+    def get_plaquette(self, qubits: PlaquetteQubits = SquarePlaquetteQubits()) -> Plaquette:
+        """Get the plaquette corresponding to the RPNG description
+        
+        Note that the ancilla qubit is the last among the PlaquetteQubits and thus
+        has index 4, while the data qubits have indices 0-3.
```
nelimee, line 235: I think that currently the standard was to have the ancilla as the first qubit (index `0`). Is there any reason to change that?

nelimee: Also note that plaquettes on the boundary only have 3 qubits (counting the ancilla). Having a hole in qubit indexing should not be an issue, but I would advocate for not having a hole at all.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+                    circuit_as_list[rpng.n] += f'{rpng.a.value.upper()}C{rpng.p.value.upper()} 4 {q}\n'
+            # Data reset or Hadamard.
+            if rpng.r:
+                circuit_as_list[0] += f'{rpng.get_r_op()} {q}\n'
+            # Data measurement or Hadamard.
+            if rpng.g:
+                circuit_as_list[-1] += f'{rpng.get_g_op()} {q}\n'
+        # Ancilla reset and measurement.
+        circuit_as_list[0] += f'R{self.ancilla.r.value.upper()} 4\n'
+        circuit_as_list[-1] += f'M{self.ancilla.g.value.upper()} 4\n'
```
nelimee, line 257: See comment on docstring, the hard-coded `4` will introduce holes in qubit indexing for plaquettes on the boundary. I think this is not desirable.

nelimee (commented):
I think the main point that needs to be changed is the extraction of the Extended RPNG representation outside of the RPNG code. There are also formatting issue, you will have to run pre-commit.

src/tqec/plaquette/rpng_description.py
giangiac: You mentioned above if the Hadamard should also be part of the preparation. Perhaps we can wait until we decide on that point before merging the implementations (which may perhaps slightly differ). Also, I find it mnemonically easier to keep the descriptive names.

src/tqec/plaquette/rpng_description.py
giangiac: We need to know it when we create the circuit associated with the plaquette, so it should be available then.
My idea is that it can be determined before creating the plaquette.
Alternatively we can add it as argument to the method generating the plaquette (and its circuit), adding checks of compatibility with the n in the corners.

src/tqec/plaquette/rpng_description.py
giangiac: I think it is important to validate the RPNG of the corners:
while each one can be correct in its own merit, further checks are possible when all corners are available.

src/tqec/plaquette/rpng_description.py
giangiac: That's a good point.
I just noticed that if I iterate through the `SquarePlaquetteQubits` the iterator first goes through the data qubits and then the ancilla qubit.
I do not know the consequences of changing that behavior, so I simply adopted that convention.
In the future one may have multiple ancillas to measure a stabilizer (e.g. when connectivity is constrained) so we may have other type of issees then).

src/tqec/plaquette/rpng_description.py
giangiac: Done

src/tqec/plaquette/rpng_description.py
giangiac: done

src/tqec/plaquette/rpng_description.py
nelimee: > You mentioned above if the Hadamard should also be part of the preparation. Perhaps we can wait until we decide on that point before merging the implementations (which may perhaps slightly differ). 

On that, I am fine with both. Having a matching implementation here would be an argument for having the possibility to set `h` on both `R` and `G` in my opinion.

> Also, I find it mnemonically easier to keep the descriptive names.

Even if we restrict `h` to be only on `G`, the code should be factored. You can then re-use the factored code to implement `get_g_op` and `get_r_op`, just avoid as much as possible code duplication.

src/tqec/plaquette/rpng_description.py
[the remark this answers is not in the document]
giangiac: Done

src/tqec/plaquette/rpng_description.py
nelimee: From memory, the iteration order is very much just a detail and is not used anywhere. A good hint that the previous sentence is likely true is that the iteration order is not documented anywhere. 

PS: jsut checked, `all_qubits` is only used to reduce a circuit to the plaquette's qubits by calling `ScheduledCircuit.filter_by_qubits`, so the specific ordering does not matter.

src/tqec/plaquette/rpng_description.py
nelimee: I am fine having that information somewhere. I agree that we will eventually need it. My main issue is that this information should not be accessible and "settable" by the user, because `tqec` internals will set the correct value for it, and so giving the user the impression he/she can fix the measurement time-step would be counter-intuitive.

giangiac:
> I would vote against the 5-character notation. In my mind if you need to go to a circuit that cannot be expressed with the four character notation, you probably wouldn't express that circuit using plaquettes in the first place. H can be on both reset and measurement as a typical expansion of cnot into CZ can easily require H on both ends of the circuit. I think reset time should always be zero, and the measurement time should be the latest time of a gate in an entire circuit layer plus one. Best, Austin.

I am catching up with the comments and it seems that the consensus (at least among the active users @afowler @inmzhang @nelimee ) is to focus exclusively on the 4-character notation.
I will move the 5-character one to separate classes (not to lose its implementation in case of future reevaluation).

Question on N: Is it ok to assume a single digit value or shall we provide the flexibility to have a larger integer?
This may happen when gates take more than a "time step", if we describe time in terms of a certain clock.

Question on the measurement time: We need to know that at the time when we generate the plaquette.
Are we going to populate the templates with Plaquette or RPNGDescription objects?

afowler:
I'm ok with restricting n to be a single digit 1-9. A single rnpg string
doesn't tell you the measurement time, this is a property of an entire
layer of such strings.

nelimee:
> Question on N: Is it ok to assume a single digit value or shall we provide the flexibility to have a larger integer?

Let's have the simple case correct first. If we need 10+ values for `N` we will improve that afterwards.

> Question on the measurement time: We need to know that at the time when we generate the plaquette.
> Are we going to populate the templates with Plaquette or RPNGDescription objects?

Templates will return RPNG strings. These strings will be translated to `Plaquette` instances at circuit creation, when we will have all the global information for each round of error correction, which is needed to synchronise measurements.

src/tqec/plaquette/rpng_description.py
```diff
@@ -0,0 +1,262 @@
...
+        p   data basis for the controlled operation (assuming a=z, x means CNOT controlled on the ancilla and targeting the data qubit, y means CY, z means CZ)  
+        n   time step (positive integers, all distinct, typically in 1-5)  
+        g   data qubit measure basis or h or -
+        a   ancilla basis for the controlled operation (x means that the controlled operation is applied if ancilla is in |+>, y if it is in |-y>, z if it is in |1>)
+    """
+    r: ExtendedBasisEnum | None
+    p: BasisEnum | None
+    n: int | None
+    g: ExtendedBasisEnum | None
+    a: BasisEnum = BasisEnum.Z
```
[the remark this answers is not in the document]
giangiac: Done

src/tqec/plaquette/rpng_description.py
giangiac: Shall I change the order then (first ancilla, then data)?

src/tqec/plaquette/rpng_description.py
giangiac: The compact implementation has been adopted for RAPNG and removed from RAPG.

src/tqec/plaquette/rpng_description.py
giangiac: 5-character description moved to rapng.py

src/tqec/plaquette/rpng_description.py
giangiac: Removed

src/tqec/plaquette/rpng_description.py
giangiac: Ok, I just updated the `PlaquetteQubits` class changing the order of the fields (first `syndrome_qubits`, then `data_qubits`) and the `__iter__` method.

src/tqec/plaquette/rpng_description.py
giangiac: Done

src/tqec/plaquette/rpng_description.py
giangiac: @nelimee  This actually broke several CI/CD tests.
Shall we update those tests or change the way we iterate though the qubits in the `RPNGDescription.get_plaquette()` method?

I think it would be nice to have consistency across the codebase: either always ancilla first or data first

nelimee: > Shall we update those tests or change the way we iterate though the qubits in the `RPNGDescription.get_plaquette()` method?

For the moment, I would take the solution that minimise the amount of code changes for this PR. If that solution is not the prefered one (ancilla qubit first or last, I have a soft preference for ancilla qubit first) then we can open a different PR to change that.

So for the moment, stick to the convention that does not break tests and minimise code changes.

nelimee:
Here is a proposal for the `RPNG.from_string` method that makes it more readable in my opinion. I am also fine with the existing code (modulus my comments), so it is up to preference.

```py
    @staticmethod
    def _parse_or_raise(
        typ: type[RPNGCharType], char: str, pos: Literal["R", "P", "N", "G"]
    ) -> RPNGCharType:
        try:
            return typ(char)
        except ValueError as err:
            raise ValueError(
                f"Unacceptable character for the {pos} field: '{char}'."
            ) from err

    @classmethod
    def from_string(cls, rpng_string: str) -> RPNG:
        """Initialize the RPNG object from a 4-character string

        Note that any character different from a BasisEnum / ExtendedBasisEnum
        value would result in the corresponding field being None.
        """
        if len(rpng_string) != 4:
            raise ValueError("The rpng string must be exactly 4-character long.")
        r_str, p_str, n_str, g_str = tuple(rpng_string)
        # Convert the characters into the enum attributes (or raise error).
        r = RPNG._parse_or_raise(ExtendedBasisEnum, r_str, "R")
        p = RPNG._parse_or_raise(BasisEnum, p_str, "P")
        n = RPNG._parse_or_raise(int, n_str, "N")
        g = RPNG._parse_or_raise(ExtendedBasisEnum, g_str, "G")

        return cls(r, p, n, g)
```

The code 
```py
from tqec import RPNG

rpng = RPNG.from_string2("e---")
```
raises and prints to the screen
```
Traceback (most recent call last):
  File "/workspaces/tqec/src/tqec/plaquette/rpng.py", line 100, in _parse_or_raise
    return typ(char)
           ^^^^^^^^^
  File "/usr/local/lib/python3.12/enum.py", line 757, in __call__
    return cls.__new__(cls, value)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/enum.py", line 1171, in __new__
    raise ve_exc
ValueError: 'e' is not a valid ExtendedBasisEnum

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/workspaces/tqec/main.py", line 3, in <module>
    rpng = RPNG.from_string2("e---")
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/tqec/src/tqec/plaquette/rpng.py", line 117, in from_string2
    r = RPNG._parse_or_raise(ExtendedBasisEnum, r_str, "R")
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/tqec/src/tqec/plaquette/rpng.py", line 102, in _parse_or_raise
    raise ValueError(
ValueError: Unacceptable character for the R field: 'e'.
```

giangiac:
> Here is a proposal for the `RPNG.from_string` method that makes it more readable in my opinion. I am also fine with the existing code (modulus my comments), so it is up to preference.

I am not sure how you defined `RPNGCharType`, but one has to include the desired behavior for `-` or it would not work.
What if we merge this code and leave code readability for later?

src/tqec/plaquette/rpng_description.py
giangiac: Ok, I reverted the commit.
First data and then ancilla qubits.

giangiac:
From my point of view, this is ready to be merged.

A few things left for future PRs:
1. a generic and documented interface, so that users will be able to implement their own translation if they want
2. expand the tqec documentation, for example by adding to the `user_guide` a new section about creating the plaquettes
3. improve code readability (see [Adrien's proposal](https://github.com/tqec/tqec/pull/411#issuecomment-2540821000))

nelimee (approved):
Ok for me to merge that PR quickly and iterate on it. Thanks for the work :)

nelimee:
@giangiac do you want me to merge that PR, or are you waiting for something else to merge?

giangiac:
> @giangiac do you want me to merge that PR, or are you waiting for something else to merge?

Usually, I let someone else merge my PRs, but it seems that the consensus is to merge.
So, I will go for it. :-)

PR #419, opened 2024-12-11
clarified sketch-up os dependency

KabirDubey:
The documentation should explicitly say that SketchUp usage is OS dependent, to avoid confusing first time users. I clarify this portion of the docs in this short PR.


inmzhang (approved):
Thanks Kabir, looks good to me.

PR #421, opened 2024-12-18
Introduce `RPNGTemplate`

### LINKED ISSUE #412: Introduce `RPNGTemplate`

Following the last discussion during our weekly meeting, we reached the conclusion that `Template` should not return indices anymore, mostly due to #386. Instead, integer indices should be replaced by RPNG⁴ textual representation that is able to accurately and concisely describe the plaquette that is expected at each entry of the `Template.instantiate` output.

There are two potential solution I can see:

1. change the output of `Template.instantiate` to a 2-dimensional array of RPNG⁴ strings (or any data-structure representing such string, that might be introduced in #411),
2. do not touch anything on the `Template` implementation, but make `tqec.templates` an implementation detail (and consequently a private or semi-private module) and use it to implement a ""`RPNGTemplate`"" (name to be refined). This would basically follow the idea proposed in #386.

I personally prefer solution number 2 as it allows to re-use the solid, well tested and well understood `tqec.templates` implementation as a foundation, and it seems "simple". Basically, an ""`RPNGTemplate`"" could be implemented like:

```py
from dataclasses import dataclass
from enum import Enum
from typing import Literal, Mapping

from tqec.templates.base import Template


class BasisEnum(Enum):
    X = "X"
    Y = "Y"
    Z = "Z"


@dataclass
class RPNG:
    r: BasisEnum | None
    p: BasisEnum | None
    n: int | None
    g: BasisEnum | Literal["H"] | None


@dataclass
class RPNGPlaquette:
    corners: tuple[RPNG, RPNG, RPNG, RPNG]


@dataclass
class RPNGTemplate:
    template: Template
    mapping: Mapping[int, RPNGPlaquette]

    def instantiate(self, k: int) -> list[list[RPNGPlaquette]]:
        indices = self.template.instantiate(k)
        return [[self.mapping[i] for i in row] for row in indices]
```

The advantages of such an approach would be the following:

- Scaling would still be handled by the `Template` approach that has proven to be quite robust and well tested.
- We could re-use the `QubitTemplate` for most of the `RPNGTemplate` instances of interest, and keep the number of templates implemented as low as possible. This is nice because implementing correctly a `Template` is error-prone and time-consuming.
- It is a relatively small change.

Any remark/question/alternative ways?

nelimee:
This PR introduces the `RPNGTemplate` class that basically implement the `Template` interface with one difference: `instantiate` returns a `list[list[RPNGDescription]]` instead of a `numpy.ndarray`.

Main modifications:

- `tqec.templates` -> `tqec.templates.indices`,
- Creation of `tqec.templates.rpng` implementing the `RPNGTemplate` class,
- `tqec.templates.enums.TemplateOrientation` -> `tqec.enums.Orientation` to fix a circular import,
- a few import reformat.

Closes #412


src/tqec/templates/rpng.py
```diff
@@ -0,0 +1,81 @@
...
+        return self.template.shape(k)
+
+    @property
+    def scalable_shape(self) -> Scalable2D:
+        """Returns a scalable version of the template shape."""
+        return self.template.scalable_shape
+
+    def get_midline_plaquettes(
+        self, k: int, orientation: Orientation = Orientation.HORIZONTAL
+    ) -> list[tuple[int, int]]:
```
inmzhang, line 31: In the current implementation of observable construction( `compile/observables`), the `get_midline_plaquettes` method in the `Template` class, as well as the concept of observable qubits, are not used at all. I am wondering whether these components are still needed.

inmzhang (approved):
LGTM. Maybe a simple test on the `instantiate` method can be added.

nelimee:
> Maybe a simple test on the `instantiate` method can be added.

I would argue that the `instantiate` method will be tested in #422. Is that enough, or do you still think we need a specific test here?

src/tqec/templates/rpng.py
nelimee: I'll open an issue for that. The `Interval` class is also unused at the moment, so that calls for a removal of dead code.

nelimee: See #424

nelimee:
Merging to avoid blocking. If the test was really needed, I'll open a new PR, just ping me.

inmzhang:
> I would argue that the instantiate method will be tested in https://github.com/tqec/tqec/pull/422. Is that enough, or do you still think we need a specific test here?

It's fine for me. Sorry for the delay, I did not have a computer around in the last few days.

PR #426, opened 2024-12-24
Remove unused code

### LINKED ISSUE #266: Introduce the portion package to handle intervals
#256 introduced the `Interval` and `Intervals` classes to represent portions of R, all the real numbers. It turns out that a Python package seem to implement exactly that.

> LGTM. Just wanted to mention that, I found a package [`portion`](https://github.com/AlexandreDecan/portion?tab=readme-ov-file#interval-operations) that appear to do something similar to `Interval`.
> 
> _Originally posted by @inmzhang in https://github.com/QCHackers/tqec/pull/256#pullrequestreview-2195298205_

            
The question is then: "is it worth using that package?".

At first glance I am on the "yes" side, but here are a few points (on both sides) that might help making the best decision:

- the `portion` package seems well maintained,
- using an external package allow us to support less code in `tqec`,
- using an external package would increase the time needed to fix a bug and the friction to implement new features,

Any thoughts?

### LINKED ISSUE #424: Remove dead code
There are multiple portions of the code that are not used anywhere anymore. These should be removed from the code.

A few examples:

- `tqec.plaquette.frozendefaultdict.FrozenException`,
- `tqec.interval.Interval`
- `tqec.templates.indices.Template.get_midline_plaquettes`,

and potentially others we did not see.

Dead code should be removed from the repository.

If we think that a currently-dead code might be useful in the future, we have two options:

1. remove it anyway, and get it back from git history if we eventually need that code again (exactly what has been done for `tqec.templates.indices.qubit.QubitVerticalBorders` for example),
2. keep it in the code base.

I prefer option 1.

nelimee:
This PR removes dead code from the code base.

It closes #266 by removing the `Interval` data-structure entirely.
Closes #424


inmzhang (approved):
LGTM.

Issue #420, opened 2024-12-11
Replace `Template` by `RPNGTemplate`


nelimee:
This issue depends on #416. Once #416 will be merged, we should be able to replace all the uses of `Template` (and its sub-classes) by uses of `RPNGTemplate` (and its sub-classes/instances). Note that the dependence chain for this issue is the following:

- [x] #410 
- [x] #412 
- [x] #416 
- [ ] This issue

This issue will be fixed when `Template` and its sub-classes will **only** be used in `tqec.templates` and nowhere else. They will become an implementation detail for the new API that is `RPNGTemplate` and the accompanying library.

nelimee:
Closing as `RPNGTemplate` will not be used extensively as we need some circuit that are not representable using the RPNG notation.

PR #428, opened 2024-12-26
Introduce `RectangularTemplate.get_border_indices` to correctly manage substitutions

### LINKED ISSUE #387: Spatial boundary replacement has hard-coded plaquette indices
The code

https://github.com/QCHackers/tqec/blob/6014998ed204054803a9dbe5d1f9cef70f8c5348/src/tqec/compile/specs/library/_utils.py#L186-L209

is using hard-coded plaquette indices that are only valid for the `QubitTemplate` template. This will be a blocking issue as soon as we will implement spatial junctions with `Qubit4WayJunctionTemplate`.

The best would be to be able to query a template for one of its spatial boundaries and perform a match between plaquette indices.

nelimee:
This PR introduces a mechanism that should allow us to perform substitution of plaquettes on the spatial border of a template without needing any hard-coded indices in the logic.

It basically requires all templates inheriting from `RectangularTemplate` to implement a function that reports the plaquette indices on each of its borders.

Eventually this should fix #387 but we will likely need to wait for #420 that will refactor the substitution process and will likely use the code introduced here to perform the substitution correctly.


src/tqec/templates/indices/base.py
```diff
@@ -137,5 +138,51 @@ def instantiation_origin(self, k: int) -> Position2D:
...
+            self.second_repeating,
+            self.bottom_right_corner,
+        )
+
+
 class RectangularTemplate(Template):
     """Base class for all the templates that have a rectangular shape."""
+
+    @abstractmethod
+    def get_border_indices(self, border: TemplateBorder) -> BorderIndices:
```
inmzhang, line 187: doc missing

inmzhang (approved):

Issue #418, opened 2024-12-10
`CompiledGraph` does not appear in API documentation


nelimee:
## Describe the bug
The `CompiledGraph` class, that is quite central and user-facing, does not appear in the API documentation (see https://tqec.github.io/tqec/_autosummary/tqec.compile.html).

PR #430, opened 2024-12-28
Ensuring `simulation` and `CompiledGraph` appear in API docs

KabirDubey:
Small PR to address issue #418. I was exploring #364 and noticed some functions in the `simulation` module were not in the API documentation either, so I update the init file similarly.

### DIFF
--- a/src/tqec/__init__.py
+++ b/src/tqec/__init__.py
@@ -4,6 +4,7 @@ from .circuit import ScheduledCircuit as ScheduledCircuit
 from .circuit import ScheduleException as ScheduleException
 from .circuit import generate_circuit as generate_circuit
 from .circuit import merge_scheduled_circuits as merge_scheduled_circuits
+from .compile import CompiledGraph as CompiledGraph
 from .compile import compile_block_graph as compile_block_graph
 from .computation import BlockGraph as BlockGraph
 from .computation import BlockKind as BlockKind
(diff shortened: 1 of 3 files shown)


KabirDubey:
Ready for review--looks like the docs updated correctly:

https://tqec.github.io/tqec/pull/430/_autosummary/tqec.simulation.html#module-tqec.simulation

https://tqec.github.io/tqec/pull/430/_autosummary/tqec.compile.CompiledGraph.html#tqec.compile.CompiledGraph

nelimee:
Thanks for the PR! The two added functions from `tqec.simulation.generation` are semi-private. I do not see why a user would need to use them in isolation. Not having them in the `__init__.py` was mostly to reduce clutter in the public interface. Users that really need these functions in isolation can import them, but I do think that making them part of the public and documented API does not add a lot of value to regular users.

KabirDubey:
> Thanks for the PR! The two added functions from `tqec.simulation.generation` are semi-private. I do not see why a user would need to use them in isolation. Not having them in the `__init__.py` was mostly to reduce clutter in the public interface. Users that really need these functions in isolation can import them, but I do think that making them part of the public and documented API does not add a lot of value to regular users.

This makes sense, thanks for the clarification. This PR just adds `CompiledGraph`, then.

nelimee (approved):
LGTM, thanks!

PR #431, opened 2024-12-29
Plaquette creation and compilation

nelimee:
This PR introduce an `RPNGDecription` -> `Plaquette` translator interface and a default implementation of that interface.

It also introduce the beginning of a plaquette compilation pipeline.


inmzhang:
Happy new year! I'll have time to review this PR at tomorrow :-)

src/tqec/plaquette/compilation/passes/transformer/transformer.py
```diff
@@ -0,0 +1,132 @@
...
+    instruction_simplifier: InstructionSimplifier = NoInstructionSimplification()
+
+    def apply(self, circuit: ScheduledCircuit) -> ScheduledCircuit:
+        # moment_instructions: schedule_index -> instruction list.
+        moment_instructions: dict[int, list[stim.CircuitInstruction]] = {}
+        for schedule, moment in circuit.scheduled_moments:
+            for instruction in moment.instructions:
+                # if the transformation represented by self does not apply to
+                # the current instruction, just add it unmodified.
+                if instruction.name != self.source_name:
```
inmzhang, line 64:
```suggestion
                if instruction.name not in stim.gate_data(self.source_name).aliases:
```

This eliminates the need to specify transformation rules both for `R` and `RZ`.

src/tqec/plaquette/compilation/passes/reset_basis.py
```diff
@@ -0,0 +1,34 @@
...
+        if basis == Basis.X:
+            transformations.append(
+                ScheduledCircuitTransformation(
+                    "R",
+                    {
+                        ScheduleOffset(0): [InstructionCreator("RX")],
+                        ScheduleOffset(1): [InstructionCreator("H")],
+                    },
+                )
+            )
```
inmzhang, line 33: Once we use the aliases as below, this is not needed anymore.

src/tqec/plaquette/compilation/passes/measurement_basis.py
```diff
@@ -0,0 +1,34 @@
...
+        if basis == Basis.X:
+            transformations.append(
+                ScheduledCircuitTransformation(
+                    "M",
+                    {
+                        ScheduleOffset(-1): [InstructionCreator("H")],
+                        ScheduleOffset(0): [InstructionCreator("MX")],
+                    },
+                )
+            )
```
inmzhang, line 33: The same as the reset case.

inmzhang (commented):
This looks great!

Not relevant to this issue but just to note that we might also need some global circuit optimizer for optimization like gate merging across different qec rounds, e.g. cancelling the hadamard gates at the start/end of the round circuits. I think this cannot be done by looking at the plaquette circuit alone. Criag has implemented some global optimizers(transpilers) that keep the layer structure of the circuit [here](https://github.com/Strilanc/magic-state-cultivation/blob/main/src/gen/_layers), which we may learn from.

nelimee:
Thanks for the detailed review! I will address the points you raised tomorrow morning :)

Note to myself: the compilation pipeline does not set the `duplicated_plaquettes` attribute. Add a compilation pass to set that attribute.

src/tqec/plaquette/enums.py
```diff
@@ -37,6 +37,11 @@ def opposite(self) -> PlaquetteSide:
             return PlaquetteSide.DOWN
 
 
+class Basis(Enum):
```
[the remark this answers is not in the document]
nelimee: I changed all the trivial-to-change places. I will try to finish that change in this PR, leaving that discussion open to track that.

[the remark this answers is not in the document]
nelimee: The other uses for `ResetBasis` and `MeasurementBasis` are in the plaquette library, that will have large changes in the near future, so I am keeping the code like that for the moment.

inmzhang (approved):
LGTM.

PR #434, opened 2024-12-31
Use `RPNGTemplate` everywhere

nelimee:
This is a tentative PR. I am pretty sure the approach taken here will not be the right one, but some code might be useful to test other approaches, hence the PR to allow anyone to take some code if needed.


nelimee:
As said in the description, this was not the best path. Closing in favour of #437

Issue #423, opened 2024-12-19
Support constructing observables for the spatial junctions


inmzhang:
**Is your feature request related to a problem? Please describe.**

We can find abstract observables(correlation surface) for a computation containing spatial junctions but cannot construct a stim logical observable correctly with the current implementation.

**Describe the solution you'd like**

Improve the implementation in `compile/observables`. It should be aware of the specification/context of a block to construct the observables correctly.

PR #439, opened 2025-01-12
Construct logical observables for spatial junctions

inmzhang:
This PR addresses #423 and includes some refactoring.

## Observable Construction for Spatial Junctions

For the spatial junctions shown below:

The junction supports two types of correlation surfaces:
1. The **Z-type** correlation surface, which is parallel to the junction plane.
2. The **X-type** correlation surface, which is perpendicular to the junction plane.

For the **Z-type** correlation surface, all `Z-basis` stabilizer measurements in the first layer of the central junction are included in the observable. For the **X-type** correlation surface, the data qubit readouts are included, depending on the correlation surface's shape within the junction:

Image

The current implementation is untested and may contain bugs. I will think about how to properly test it and add tests in the next few days.

## Other Changes

1. Made the `AbstractObservable` data structure private and adopted `CorrelationSurface` throughout the public API. `AbstractObservable` now serves as an implementation detail, allowing modifications without breaking changes.
2. Corrected a typo: `ZXGraph.find_correration_surfaces()` to `ZXGraph.find_correlation_surfaces()`.
3. Added imports for `solo_node_zx_graph` and `solo_node_block_graph` in `tqec/gallery/__init__.py`.


inmzhang:
I've added unit tests for observable calculations, but I left out the key component, `inplace_add_observable`, possibly to include it in future integration tests based on #437.

nelimee (commented):
I still need to review the two main files of the PR (`abstract_observable.py` and `builder.py`), will do that later.

nelimee (commented):
I find the code in `abstract_observables.py` and `builder.py` to be quite hard to read, for different reasons:

- `abstract_observables.py` is difficult to read because I am still not 100% on point with how we handle correlation surfaces (e.g., why some qubits on top and others on the first layer), that is my fault.
- `builder.py` lacks, in my opinion, a little bit of documentation to explain the different conventions used. For example, the convention on qubit positioning, the convention that qubits from a pipe are indexed w.r.t the source block of the pipe, ...

The tests are helping a little for `abstract_observables.py`, but I still have hard times understanding the logic behind `builder.py`, even though it seems to work correctly.

src/tqec/compile/observables/builder.py
```diff
@@ -0,0 +1,268 @@
...
+    local_coords: tuple[float, float] | tuple[int, int],
+    block_position: Position3D,
+    k: int,
+) -> GridQubit:
+    """Transform local coordinates at a block to the global coordinates in the circuit.
+
+    When calculating the coordinates of the measurement qubits, we use a local coordinate system
+    in the individual blocks. The top-left corner of the block starts at (0, 0) and each column
+    is separated by 0.5. That is, all the data qubits are placed at the integer coordinates while
+    all the measurement qubits are placed at the half-integer coordinates.
```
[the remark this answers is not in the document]
inmzhang: This is equivalent to using `template_increments=Displacement(1, 1)` when calculating the local coordinates of the observable qubits. When calling `_transform_coords_into_grid`, it calculates the real coordinates of the qubits in the `LayoutTemplate`, taking into account the actual `template_increments` and block positions. This convention simplifies the code slightly, as calculating the real coordinates of the qubits upfront would require passing `template_increments` and `block position` around and being careful with the top-left data qubit starting at (-1, -1), among other details.

src/tqec/compile/observables/builder.py
```diff
@@ -0,0 +1,268 @@
...
+def _get_top_readout_spatial_junction_qubits(
+    junction_shape: Shape2D,
+    arms: JunctionArms,
+) -> list[tuple[int, int]]:
+    assert len(arms) == 2
+    half_x, half_y = junction_shape.x // 2, junction_shape.y // 2
+    # could use match expressions here
+    # but not sure how is `|` operator evaluated in
+    # `case JunctionArms.LEFT | JunctionArms.RIGHT:`
+    # Is that used as case separator or bitwise OR?
```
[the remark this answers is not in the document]
inmzhang:
```py
a = JunctionArms.UP

match a:
    case JunctionArms.UP | JunctionArms.DOWN:
        print("case or")
    case _:
        print("flag or")
```
This evaluates to `case or`, so the bitflag OR operation will not take effect. Adding parentheses does not help too.

inmzhang:
> I find the code in `abstract_observables.py` and `builder.py` to be quite hard to read, for different reasons:
> 
>     * `abstract_observables.py` is difficult to read because I am still not 100% on point with how we handle correlation surfaces (e.g., why some qubits on top and others on the first layer), that is my fault.
> 
>     * `builder.py` lacks, in my opinion, a little bit of documentation to explain the different conventions used. For example, the convention on qubit positioning, the convention that qubits from a pipe are indexed w.r.t the source block of the pipe, ...
> 
> 
> The tests are helping a little for `abstract_observables.py`, but I still have hard times understanding the logic behind `builder.py`, even though it seems to work correctly.

Thanks for pointing it out! I will add more documentation to the builder and include module docs for `abstract_observable` to explain how it works from `CorrelationSurface` to `AbstractObservable`.

src/tqec/compile/observables/builder.py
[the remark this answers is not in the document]
inmzhang: One solution is to use guards but it degrades back to `if-else`-like syntax.

```py
a = JunctionArms.UP

match a:
    case _ if a == JunctionArms.UP | JunctionArms.DOWN:
        print("case or")
    case _:
        print("flag or")
```

inmzhang:
I ran `docformatter` locally on `abstract_observable.py` but it takes no effect, then I format it a bit manually and the result is awful... Could you try to auto-format it?

inmzhang:
I've added docstrings in `abstract_observables.py` and `builder.py`. Could you take a look and let me know if you think anything needs further clarification?

nelimee:
> I ran `docformatter` locally on `abstract_observable.py` but it takes no effect, then I format it a bit manually and the result is awful... Could you try to auto-format it?

Applying `docformatter` or `ruff` does not change the file on my side either.

> I've added docstrings in `abstract_observables.py` and `builder.py`. Could you take a look and let me know if you think anything needs further clarification?

It is much clearer, I got nearly everything, will do a few comments on the places where it was still slightly unclear for me.

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -0,0 +1,215 @@
...
+    `OBSERVABLE_INCLUDE <https://github.com/quantumlib/Stim/blob/main/doc/gates.md#OBSERVABLE_INCLUDE>`_
+    instructions with the same observable index in the ``stim`` circuit
+    and is composed of a set of measurements. Abstract observable
+    specifies where are the measurements located in the block graph.
+
+    Attributes:
+        top_readout_cubes: A set of cubes of which a straight line of data
+            qubit readouts on the top face should be included in the observable.
+        top_readout_pipes: A set of pipes of which a single data qubit readout
+            on the top face should be included in the observable.
```
nelimee, line 29: Maybe add explicitly that only one data-qubit from the pipe need to be included (and it is the central one) because data-qubits on the borders of the pipe are shared with neighbouring cubes, and the data-qubits are already included by the cube-processing functions?

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -0,0 +1,215 @@
...
+        bottom_stabilizer_pipes: A set of pipes of which a region of stabilizer
+            measurements on the bottom face(in the cubes it connects) should be
```
nelimee, line 31:
```suggestion
            measurements on the bottom face (in the cubes it connects) should be
```
and I do not understand the information added in parenthesis.

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -0,0 +1,215 @@
...
+    We choose the stabilizer measurements at the first layer (i.e., the earliest
+    in time or the bottom face of the block) because these are typically better
+    error-corrected by the decoder.
+
+    The compilation process is as follows:
+
+    1. Find all the spatial junctions involved in the correlation surface. For
+    each junction:
+
+    - If a surface is perpendicular to the junction's normal direction, include
```
nelimee, line 77: > If a surface is perpendicular to the junction's normal direction

i.e., the surface is in XZ or YZ plane, am I right? Just checking I understood correctly.

src/tqec/compile/observables/builder.py
```diff
@@ -0,0 +1,314 @@
...
+    in the local coordinate system of the cube.
+    """
+    stabilizers: list[tuple[float, float]] = []
+    # We calculate the qubits for the connect_to=SignedDirection3D(Direction3D.X, True) case
+    # and post-process(reflect, rotate) to get the correct orientation.
+    for i in range(cube_shape.x // 2):
+        x = cube_shape.x - i - 0.5
+        for j in range(cube_shape.y // 2):
+            y = (1 - i % 2) + 2 * j + 0.5
+            stabilizers.append((x, y))
```
nelimee, line 233: One thing that was unclear for a few minutes for me here is why we are only measuring a quarter of the stabilizer. First, I was wrong because `j` is multiplied by `2` in the expression to compute `y`, so this is a half. Then, I remembered that, by convention, the correlation surface "starts" (when "entering" the cube from below) at the middle of the cube, hence the need to only take into account half of the stabilizers. 

Not sure others would make these mistakes, but I did. Your call to determine if mode details are needed.

src/tqec/compile/observables/builder.py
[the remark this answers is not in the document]
nelimee: Right. With the above information, seems like the current if-else implementation is the best one.

nelimee (commented):
As said before, the comments really help! I made a few comments on the places that were not obvious to me. I would like to do a last review on the code by itself now that I think I understood the conventions used, will likely do that this evening or tomorrow morning.

afowler:
Since we use Stim to simulate and it uses "observable", if we had to choose
just one term I'd actually prefer to just use "observable" everywhere.
Don't recommend dropping "observable" in favor of "correlation surface".

afowler:
>
> ***@***.**** commented on this pull request.
>
> I find the code in abstract_observables.py and builder.py to be quite
> hard to read, for different reasons:
>
>    - abstract_observables.py is difficult to read because I am still not
>    100% on point with how we handle correlation surfaces (e.g., why some
>    qubits on top and others on the first layer), that is my fault.
>
> If you refer to figure 9 of the review, you will see two types of
measurements that are used to move logical operators, final data qubit
measurements, and stabilizer measurements. The final data qubit
measurements clearly must be in the final layer, but we have a choice as
to which layer of stabilizer measurements we use. There is no good reason
not to use the first layer of stabilizer measurements, since with
sufficiently advanced software we will have greater confidence in these
measurements earlier and in principle be able to make decisions based on
these measurements earlier.

Best,
Austin.

afowler:
>
> +    2. The surface spans the XY plane within some blocks. This represents a
> +    region of stabilizer measurements in the basis of the surface, whose products
> +    give the parity of the logical operators on the surface edges. Stabilizer
> +    measurements need to be included in the tracked logical observable to account
> +    for the correlation between logical operators at different spatial locations.
> +    We choose the stabilizer measurements at the first layer (i.e., the earliest
> +    in time or the bottom face of the block) because these are typically better
> +    error-corrected by the decoder.
>
> To be clear, the earlier stabilizer measurements can be known with higher
probability earlier in time. When all of the decoding is done, they will
not have a lower probability of error. Earlier knowledge of these
measurements could potentially be used to drive faster computation
dependent on these measurements. This is why we use the first round of
measurements.

Best,
Austin.

inmzhang:
> When all of the decoding is done, they will
> not have a lower probability of error. Earlier knowledge of these
> measurements could potentially be used to drive faster computation
> dependent on these measurements. This is why we use the first round of
> measurements.

Indeed, it's about real-time instead of accuracy. I will correct the docs.

src/tqec/compile/observables/abstract_observable.py
inmzhang: Pipes don't have spacetime extension, so stabilizer measurements take place in the cubes instead. I'll rephrase it slightly to make it clearer.

src/tqec/compile/observables/abstract_observable.py
inmzhang: The surface is in the XY plane. Since the junction is always in the XY (spatial) plane, the normal direction refers to the Z direction. I will rephrase it more.

src/tqec/compile/observables/builder.py
inmzhang: By convention, we select the middle line of the cube as the logical operator. For two (regular) cubes connected by a horizontal pipe, the product of stabilizer measurements determines the parity of the two logical operators. Both cubes are responsible for half of the measurements:

图片

inmzhang:
> Since we use Stim to simulate and it uses "observable", if we had to choose
just one term I'd actually prefer to just use "observable" everywhere.
Don't recommend dropping "observable" in favor of "correlation surface".

I'll use `observable` as the name for the argument in the public API, although it may accept a `CorrelationSurface` object (type hints will help). I'll also note in the docs that "correlation surface" and "observable" are used interchangeably throughout the library. Does that make sense?

afowler:
+1 for the below. But as mentioned, if we are pushing for consistency I
would go with observable everywhere. In a large computation, we ultimately
only care about the results, so observable is not a bad name.

src/tqec/compile/observables/abstract_observable.py
nelimee: Got it thanks!

src/tqec/compile/observables/builder.py
nelimee: Yes I got that, but only after a few minutes of thinking, hence my comment.

nelimee:
> If you refer to figure 9 of the review, you will see two types of measurements that are used to move logical operators, final data qubit measurements, and stabilizer measurements. The final data qubit measurements clearly must be in the final layer, but we have a choice as to which layer of stabilizer measurements we use. There is no good reason not to use the first layer of stabilizer measurements, since with sufficiently advanced software we will have greater confidence in these measurements earlier and in principle be able to make decisions based on these measurements earlier. Best, Austin.

Thanks, that clears up my confusion :) 

I am starting a last round of review on this PR now.

src/tqec/compile/observables/builder.py
```diff
@@ -0,0 +1,319 @@
...
+    This function calculates the coordinates of the data qubits in the
+    local coordinate system of the junction based on which arms the
+    correlation surface touches.
+    """
+    assert len(arms) == 2
+    half_x, half_y = junction_shape.x // 2, junction_shape.y // 2
+    # could use match expressions here
+    # but not sure how is `|` operator evaluated in
+    # `case JunctionArms.LEFT | JunctionArms.RIGHT:`
+    # Is that used as case separator or bitwise OR?
```
nelimee, line 280: Maybe
```suggestion
```
?

src/tqec/compile/observables/builder.py
```diff
@@ -0,0 +1,319 @@
...
+    elif arms == JunctionArms.UP | JunctionArms.DOWN:
+        return [(half_x, y) for y in range(junction_shape.y + 1)]
+    elif arms == JunctionArms.LEFT | JunctionArms.UP:
+        return [(x, half_y) for x in range(half_x)] + [
+            (half_x, y) for y in range(half_y)
+        ]
+    elif arms == JunctionArms.DOWN | JunctionArms.RIGHT:
+        return [(x, half_y) for x in range(junction_shape.x, half_x, -1)] + [
+            (half_x, y) for y in range(junction_shape.y, half_y, -1)
+        ]
```
nelimee, line 292:
```suggestion
        return [(half_x + x, half_y) for x in range(half_x)] + [
            (half_x, half_y + y) for y in range(half_y)
        ]
```
should be the same because the shapes are always even numbers. The above version is easier to read for me.

nelimee: The above suggestion seems to contain off-by-1 errors, not sure it will be easier to read than the original code after fixing these.

nelimee (approved):
With the comments, I was able to follow the logic and I do not see any major issue. I made a few comments, but the PR looks good to me :)

src/tqec/compile/observables/builder.py
inmzhang: Both the two don't seem that complex, so I'll keep the current implementation.

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -0,0 +1,219 @@
...
+        if zx_flipped.kind != node.kind:
+            # check correlation edges in the junction arms
+            arms = JunctionArms.NONE
+            for arm, shift in JunctionArms.get_map_from_arm_to_shift().items():
+                edges = endpoints_to_edge.get(
+                    frozenset({cube.position, cube.position.shift_by(*shift)})
+                )
+                if edges is not None and any(n == zx for edge in edges for n in edge):
+                    arms |= arm
+            assert len(arms) in {
```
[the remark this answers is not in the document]
inmzhang: Are you suggesting including the assert directly in the `compile/spec/enum.py` module like appending a line of assertion at the bottom of the file?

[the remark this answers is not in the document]
nelimee: Yes, I think that would be the safest way. Something like
```py
# JunctionArms.NONE is expected to be 0 such that len(JunctionArms.NONE | JunctionArms.LEFT) == 1
# in other parts of the code
assert JunctionArms.NONE.value == 0
```
at the end of `compile/spec/enum.py`.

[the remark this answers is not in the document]
inmzhang: I checked the code again, and it’s not immediately clear that `JunctionArms.NONE.value == 0` ensures the condition `len(JunctionArms.NONE | JunctionArms.LEFT) == 1` holds:

```py
    def __len__(self) -> int:
        return sum(arm in self for arm in JunctionArms.single_arms())

    def __iter__(self) -> Iterator[JunctionArms]:
        yield from (arm for arm in JunctionArms.single_arms() if arm in self)
```

It seems that it’s the definition of `JunctionArms.single_arms()` that ensures this behavior.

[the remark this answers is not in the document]
nelimee: You are right, and I though that this behaviour was documented in the `enum.Flag.__len__` method introduced in Python 3.11, but it is not. Not sure what to do then, I guess that leaving it like that should be fine as long as we do not re-define `__len__` incorrectly.

PR #440, opened 2025-01-13
docs: add disclaimer about rapid code evolution to README

mdimado:
This PR adds a disclaimer at the top of the README.md to address the issue raised in the Google group discussion about potential confusion when users can't find files they see in weekly meeting recordings.


nelimee:
Nice! I do not know how to write that clearly and concisely, but I think it would be nice to add a distinction:

- people can open an issue (or maybe a discussion should be preferred?) if they are lost (e.g., cannot find file `X` that appears on a recording but not on the repository),
- people should not open an issue about issues related to things in the recordings (e.g. dead links when the link comes from a recording) but should always be sure that the issue is still present on the current version of the documentation.

Thinking about it, issue templates might be the right tools to make that distinction. What do you think?

mdimado:
Thank you for the feedback! I've revised the disclaimer to be more explicit in differentiating between:
1. Discussions as a way to assist in locating materials mentioned in recordings
2. Issues only for known problems with the current version

I definitely agree that issue templates would be super helpful here. We could make:

1. A discussion template for "Finding Materials from Recordings"
2. An issue template with a checklist including:
- [ ] I have checked for this issue on the latest version
- [ ] It is not content from older recordings
- [ ] I checked the current documentation

Should I create these templates as part of this PR or should we make that a separate issue?

nelimee:
> Should I create these templates as part of this PR or should we make that a separate issue?

You can edit them here. I cannot remember if special GitHub files in https://github.com/tqec/tqec/tree/main/.github/ISSUE_TEMPLATE have a restricted access or not. If you are having permission issues, please let me know.

nelimee:
@mdimado do you still want to finish this PR? Else, someone else might be able to update the issue templates and merge this PR with your changes and the updates.

nelimee:
@mdimado I have updated the issue templates in #485 (that I will improve again soon). This PR can be merged.

nelimee (approved):
LGTM

PR #442, opened 2025-01-14
Add logical CZ gate to gallery

inmzhang:
This PR adds logical CZ gates to the gallery and implements methods for rotating the `ZXGraph` and `BlockGraph` by multiples of 90 degrees.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -11,3 +11,4 @@ types-networkx
 pytest-cov
 pandoc
 pre-commit
+scipy-stubs
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -299,4 +299,5 @@ class BlockGraph(ComputationGraph[Cube, Pipe]):
 
         zx = self.to_zx_graph()
         rotated_zx = zx.rotate(rotation_axis, n, counterclockwise)
-        return rotated_zx.to_block_graph(self.name + " rotated")
+        name_suffix = f" rotated by {n * 90} degrees {'counter' if counterclockwise else ''}clockwise around the {rotation_axis.name} axis"
+        return rotated_zx.to_block_graph(self.name + name_suffix)
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -273,3 +273,30 @@ class BlockGraph(ComputationGraph[Cube, Pipe]):
                 pipe.kind,
             )
         return new_graph
+
+    def rotate(
+        self,
+        rotation_axis: Direction3D = Direction3D.Y,
+        num_90_degree_rotation: int = 1,
+        counterclockwise: bool = True,
+    ) -> BlockGraph:
+        """Rotate the graph around an axis by ``num_90_degree_rotation * 90`` degrees and
+        return a new rotated graph.
+
+        Args:
+            rotation_axis: The axis around which to rotate the graph.
+            num_90_degree_rotation: The number of 90-degree rotations to apply to the graph.
+            counterclockwise: Whether to rotate the graph counterclockwise. If set to False,
+                the graph will be rotated clockwise. Defaults to True.
+
+        Returns:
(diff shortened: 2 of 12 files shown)


src/tqec/computation/block_graph.py
```diff
@@ -273,3 +273,30 @@ def shift_min_z_to_zero(self) -> BlockGraph:
...
+            A data-independent copy of the graph rotated by the given number of 90-degree rotations.
+        """
+        n = num_90_degree_rotation % 4
+
+        if n == 0:
+            return self.copy()
+
+        zx = self.to_zx_graph()
+        rotated_zx = zx.rotate(rotation_axis, n, counterclockwise)
+        return rotated_zx.to_block_graph(self.name + " rotated")
```
nelimee, line 302: I like to be able to follow what transformations have been performed on something from its name. In this spirit, I would rather name the new graph using
```suggestion
        fname = "R" + str(n * 90 if not counterclockwise else (4 - n) * 90)
        return rotated_zx.to_block_graph(f"{fname}({self.name})")
```
This is very much a detail.

requirements-dev.txt
nelimee, line 1: And also add `scipy` to the `requirements.txt` right?

src/tqec/computation/zx_graph.py
```diff
@@ -114,10 +115,12 @@ class ZXEdge:
     has_hadamard: bool = False
 
     def __post_init__(self) -> None:
-        if not self.u.position.is_neighbour(self.v.position):
-            raise TQECException("An edge must connect two nearby nodes.")
-        # Ensure position of u is less than v
         u, v = self.u, self.v
+        if not u.position.is_neighbour(v.position):
+            raise TQECException(
+                f"An edge must connect two nearby nodes, but get {u.position} and {v.position}."
```
nelimee, line 121:
```suggestion
                f"An edge must connect two nearby nodes, but got {u.position} and {v.position}."
```

src/tqec/computation/zx_graph.py
```diff
@@ -370,3 +373,56 @@ def check_invariants(self) -> None:
...
+            elif rotation_axis == Direction3D.Y:
+                rot_vec[1] = -1
+            elif rotation_axis == Direction3D.Z:
+                rot_vec[2] = 1
+            if not counterclockwise:
+                rot_vec *= -1
+            rotated = R.from_rotvec(rot_vec * n * np.pi / 2).apply(p.as_tuple())
+            return Position3D(*[round_or_fail(i) for i in rotated])
+
+        g = self.__class__(self.name + " rotated")
```
nelimee, line 417: Same comment on the naming.

src/tqec/computation/zx_graph.py
```diff
@@ -370,3 +373,56 @@ def check_invariants(self) -> None:
...
+
+        def _rotate(p: Position3D) -> Position3D:
+            rot_vec = np.array([0, 0, 0])
+            if rotation_axis == Direction3D.X:
+                rot_vec[0] = 1
+            # account for the flipped y-axis by our convention
+            elif rotation_axis == Direction3D.Y:
+                rot_vec[1] = -1
+            elif rotation_axis == Direction3D.Z:
+                rot_vec[2] = 1
```
nelimee, line 411: Maybe use a `match`?

nelimee (approved):
Left a few comment, but LGTM.

requirements-dev.txt
inmzhang: It seems that `scipy` has already been on the dependency tree. At least I know `pymatching` has it as dependency.

nelimee: Yes, but if we explicitly import it, we should explicitly include it in our dependencies, to avoid any surprise if `pymatching` removes `scipy` from its dependencies.

inmzhang:
If you are ok with the last few commits, I will merge this PR first and then the observable one.

nelimee:
Yes, I personally prefer concise names, but this is likely to never really matter and the name you gave also include the rotation axis, which I forgot, so I am OK :)

PR #444, opened 2025-01-16
Split template library functions

nelimee:
This PR is part of a split of #437. It splits the different `RPNGTemplate` generation functions into 2:

- one function that returns the `RectangularTemplate` needed to implement the `RPNGTemplate` and,
- one function that returns the `RPNGDescription`s.

This split was motivated by the fact that, for cubes, we have to generate 3 different sets of `RPNGDescription`s without actually changing the template.


src/tqec/templates/library/hadamard.py
```diff
@@ -4,18 +4,39 @@
...
 from tqec.templates.indices.qubit import (
     QubitHorizontalBorders,
     QubitTemplate,
     QubitVerticalBorders,
 )
-from tqec.templates.rpng import RPNGTemplate
 
 
-def get_temporal_hadamard_template(
+def get_temporal_hadamard_raw_template() -> QubitTemplate:
```
inmzhang, line 14: The docstring needs to be fixed. The same for the others.

inmzhang (commented):
Need to fix the docstrings.

inmzhang (approved):

PR #445, opened 2025-01-16
Update compilation specs to implement spatial and hadamard junctions

nelimee:
This PR is part of a split of #437. It implements the different specs needed to build blocks and substitutions that support spatial junctions and Hadamard junctions.


nelimee:
Closed due to wrong base branch (`main` instead of `feat/split-template-library-functions`).

PR #446, opened 2025-01-16
 Update compilation specs to implement spatial and hadamard junctions

nelimee:
This PR is part of a split of https://github.com/tqec/tqec/pull/437. It implements the different specs needed to build blocks and substitutions that support spatial junctions and Hadamard junctions.


nelimee:
I have cherry-picked most of the commits from #437, but not all of them. In particular:

- the new entries in `tqec.gallery` belong to another PR in my opinion (PR that might be best merged between #444 and this one),
- the commit renaming "spatial junction" to "spatial cube" is tied with code related to a new entry in `tqec.gallery`, so I was not able to cherry-pick it without a lot of conflicts, that seemed error-prone to fix. So it is not included here for the moment.
- I included the commit removing the `len(arms)>2` in the spatial cube template, but not the other commits changing schedules. It still requires fixing, and I think it would be better to have one commit that fixes the scheduling issues once and for all (hopefully) once we have a clear view of the issue(s).

At the current commit:
- all previously existing tests are OK,
- none (except one, next point) of the new tests from #437 have been included,
- the test trying to implement a spatial L junction is failing, due to a scheduling issue.

inmzhang:
I can make a PR for the new gallery functions and a sparate one for renaming "spatial junctions", but may need to wait till tomorrow.

nelimee:
I'll wait for that PR to include all the necessary tests in order to be reasonably convinced that the implemented features are working as expected. I do not think that merging that PR with the issues we know about and potentially other issues we do not know about is a good idea.

nelimee:
This PR is out of date with `main` and all the still-useful code should be in #459. Closing for that reason.

PR #448, opened 2025-01-17
Replace most "spatial junction" terms with "spatial cube"

inmzhang:


nelimee (approved):
I quickly read the diff, it seems fine. I made a `grep -ri "spatial junction" .` in the main directory, and made sure that all the occurrences of "spatial junction" left are correct, which seems to be the case.

Issue #273, opened 2024-07-25
Implement a file-based database for simulation results


nelimee:
**Is your feature request related to a problem? Please describe.**
The goal of our initiative is to generate graphs such as

![Image](https://github.com/user-attachments/assets/8f6f66c8-a5ad-450e-acdb-4e4b1c884fdf)

In the above graph, each point is:
- one quantum circuit, generated by `tqec` for a given value of `k`, and that can be represented as a `stim` file,
- a large number of simulations performed by `stim`.

One problem is that `Stim` simulations are not free, and computing one point from the above graph can take minutes to hours of computational time. 

Currently, we have no clever way of storing such data, meaning that the `stim` simulations have to be re-done each time we want to generate a new graph.

**Describe the solution you'd like**

We should have a database-like way of storing simulation data. There are multiple requirements:

- we should be able to retrieve easily already existing results,
- data should be written on disk,
- we should be able to add new results to existing ones (typically, start a simulation with 1000 shots to see the overall look of the plot and check that there is not mistake, and once obvious mistakes have been corrected be able to launch 999000 more shots to reduce the error bars),
- we should be able to remove existing results, but this should be hard to do (i.e., be wary of accidental data loss)

Note that simulation results might be quite heavy in terms of memory, so an optimised storage would be a plus.

inmzhang:
We can think about utilizing the existing sampling tool like `sinter`. But as far as I know, currently there is no API provided by `sinter` to store the intermediate sampled detectors/observables to files.

nelimee:
> We can think about utilizing the existing sampling tool like `sinter`. But as far as I know, currently there is no API provided by `sinter` to store the intermediate sampled detectors/observables to files.

Yep, the goal of this issue is not the generation (which will very likely be handled by `sinter` as you note) but rather the storage of generated results.

afowler:
Craig: can you comment on how Stim/sinter simulation results can be
systematically stored so that one could later gather additional data for a
plot to improve its statistics or explore a wider range of code distances
and error rates?

nelimee:
> Craig: can you comment on how Stim/sinter simulation results can be systematically stored so that one could later gather additional data for a plot to improve its statistics or explore a wider range of code distances and error rates?

Whenever I have a task like that, I really follow the database point of view:
1. I try to find a set of small data points that uniquely identify an "experiment" (in database terms, the primary key),
2. I try to store in the "experiment" (i.e., the value associated to the primary key) whatever I may need in the future.

[...]

PR #451, opened 2025-01-19
Write sinter results to file

rainij:
Resolves https://github.com/tqec/tqec/issues/273.


rainij:
@nelimee could you review this?

I did not implement the strong id for the observable. It is better to be done in a separate PR.

This PR focuses on writing sinter stats and tries to use the sinter utilities as much as possible. Basically it is just a workaround for the fact that sinter does not provide writing to file with optional merging of existing data out of the box.

src/tqec/simulation/io_utils/csv_files.py
```diff
@@ -0,0 +1,55 @@
...
+
+import sinter
+
+
+def write_sinter_stats_to_csv(
+    filepath: str | Path,
+    stats: Iterable[sinter.TaskStats],
+    if_file_exists: Literal["merge"]
+    | Literal["overwrite"]
+    | Literal["raise"] = "raise",
```
nelimee, line 16: Equivalent to (but longer than):

```suggestion
    if_file_exists: Literal["merge", "overwrite", "raise"] = "raise",
```

src/tqec/simulation/io_utils/csv_files.py
```diff
@@ -0,0 +1,55 @@
...
+            - raise: Raise an `AssertionError` if the file exists.
+            - overwrite: Effectively delete the file and pretend none exists.
+            - merge: Read the file and try to match entries with the simulation data from `stats`
+                by strong id. In case of a match merge the entries.
+    """
+    filepath = Path(filepath)
+
+    if filepath.exists():
+        if if_file_exists == "raise":
+            raise AssertionError(f"File `{filepath}` already exists.")
```
nelimee, line 36: Small nitpick: we mostly use single quotes `'` and double quotes `"` in the repository. I do not think we use back-quotes in user-visible messages/documentation (except for formatting purposes).
```suggestion
            raise AssertionError(f"File '{filepath}' already exists.")
```

src/tqec/simulation/io_utils/csv_files.py
```diff
@@ -0,0 +1,55 @@
...
+        elif if_file_exists == "merge":
+            # Writing stats to a file is necessary because sinter does not expose merging via API.
+            # We can only indirectly merge by reading in two files.
+            in_memory_file = io.StringIO()
+            _write_sinter_stats(in_memory_file, stats)
+            _ = in_memory_file.seek(0)
+            stats = sinter.read_stats_from_csv_files(filepath, in_memory_file)
+
+    with open(filepath, "w") as f:
+        _write_sinter_stats(f, stats)
```
nelimee, line 49: I do not understand how that merges. For me here is what is done when `if_file_exists == "merge"` is `True`:

1. write the user-provided `stats` in a file and read them back (never touching the existing file at `filepath`), so basically the identity operation (modulus merging some of the provided `stats` between each other).
2. overwrite the existing stats in `filepath` with the provided `stats`.

If I understood correctly how `sinter` works, merging two stats is equivalent to writing both of them in the same file. So, if the previous is right, it should be sufficient to simply append the stats to the existing file.

```suggestion
        elif if_file_exists == "overwrite":
            filemode = "w"
        elif if_file_exists == "merge":
            filemode = "a"
            
    with open(filepath, filemode) as f:
        _write_sinter_stats(f, stats)
```

To be tested though, I am not 100% sure of that.

src/tqec/simulation/io_utils/csv_files.py
nelimee, line 36: Maybe use `TQECError` or `RuntimeError` instead?

src/tqec/simulation/io_utils/csv_files.py
nelimee: I missed the two filepaths in `stats = sinter.read_stats_from_csv_files(filepath, in_memory_file)`. You can disregard my comment :)

nelimee (changes requested):
A few minor comments or suggestions, but once addressed this should be good :)

src/tqec/simulation/io_utils/csv_files.py
rainij: Using a `TQECException` now.

rainij:
Thanks for the quick feedback. I did some changes in line with your suggestions.

nelimee (approved):
LGTM

inmzhang:
Seems that Reinhard does not have the permission to merge the PR.

nelimee:
I forgot that! I'll merge then :)

Issue #441, opened 2025-01-13
`BlockGraph.get_abstract_observables` creates duplicates in presence of labels


rainij:
## Describe the bug
It seems that `BlockGraph.get_abstract_observables` can create the same observables twice *if labels are present (at non-port nodes)*. The underlying correlation surfaces are different but seem to contain duplicate edges.

## Steps to reproduce the behavior:
The following code creates a block graph which should only have one correlation surface and hence one observable. It represents a logical qubit which gets moved to another place (see also pictures below). It is initialized and measured in the X-basis.

IMPORTANT: The error goes away (the asserts fail) if you remove the labels. They are there to indicate that "normally" there should go the ports. Only for simulation I fix to some particular kind of initialization and measurement.

```python
from tqec import Position3D, ZXNode, ZXGraph, ZXKind

# NOTE: the label changes the behaviour of the code (asserts fail if you remove both of them).
_in = ZXNode(Position3D(0, 0, 0), ZXKind.Z, "In")
_a = ZXNode(Position3D(0, 0, 1), ZXKind.Z)
_b = ZXNode(Position3D(1, 0, 1), ZXKind.Z)
_out = ZXNode(Position3D(1, 0, 2), ZXKind.Z, "Out")

g_move_zx = ZXGraph(name="move")
g_move_zx.add_edge(_in, _a)
g_move_zx.add_edge(_a, _b)
g_move_zx.add_edge(_b, _out)

g_move_block = g_move_zx.to_block_graph()

observables, correlation_surfaces = (
    g_move_block.get_abstract_observables()
)

# The following asserts (wrongly) succeed:

assert len(observables) == 2, "Actuall I would expect == 1."
assert observables[0] == observables[1]

assert len(correlation_surfaces) == 2, "Actuall I would expect == 1."
assert correlation_surfaces[0] != correlation_surfaces[1]
```

The following shows that both correlation surfaces have a duplicate edge:

```python
for s in correlation_surfaces:
    for e in s.span:
        print(e)
    print()
```

Output:

```
X(1,0,1)-X(1,0,2)
X(0,0,1)-X(1,0,1)
X(0,0,0)-X(0,0,1)
X(0,0,0)-X(0,0,1)

X(0,0,1)-X(1,0,1)
X(1,0,1)-X(1,0,2)
X(0,0,0)-X(0,0,1)
X(1,0,1)-X(1,0,2)
```

## Expected behavior
[...]

PR #452, opened 2025-01-19
Fix duplicate correlation surfaces

rainij:
Resolves https://github.com/tqec/tqec/issues/441.

I added a test which is sensitive to this bug. Thereby I also refactored the (test) code a little bit. 

I solved the bug by initializing the frontier with a fresh instantiation of the leaf without the label. This solution fits best with the rest of the code because this is actually done too when adding new nodes to the frontier. It also makes sense in regard of the fact that the span just reuses `ZXNode` in a way where a label has actually no meaning.

### DIFF
--- a/src/tqec/computation/correlation.py
+++ b/src/tqec/computation/correlation.py
@@ -39,7 +39,7 @@ class CorrelationSurface:
     represented as a mapping from the the input/output labels to the logical operator type there.
 
     Attributes:
-        nodes: A set of ``ZXNode`` representing the logical operators appeared in the correlation surface.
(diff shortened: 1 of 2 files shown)


rainij:
@inmzhang resolved the issue essentially as planned. Would you like to review it?

inmzhang:
Thanks for the PR, I'll review it tomorrow.

inmzhang (approved):
LGTM

rainij:
Cool, can be merged IMO. Can you do it @inmzhang ?

PR #455, opened 2025-01-24
Add back `F401` for unused imports

nelimee:
The linting rule `F401` was ignored a long time ago in #164. Now that the `__init__.py` files correctly re-export, we can enable it back.


inmzhang (approved):
LGTM.

PR #456, opened 2025-01-29
Introduce `tqec.utils`

nelimee:
*Note: this PR seems large, but should be rather easy to review by doing a commit-by-commit check.*

In the interest of having clear dependencies between `tqec` sub-modules, I feel like it is needed to have a dedicated sub-module for classes and functions that do not clearly belong to any other of the existing sub-modules of `tqec` and that also do not depend on them.

This PR introduces `tqec.utils` for that purpose, and basically moves a lot of code to this sub-module. Additionally, it:

- removes the `RAPNG` data-structure as 1) it is not currently used anywhere in the code base and 2) `RPNG`-based notation will not be as central as we thought initially*.
- removes `tqec.circuit.detectors` that only contained an empty `__init__.py` file.

*: the plaquettes needed to implement the extended stabilizers needed for spatial junctions are not representable using the `RPNG` notation. That means that we cannot use the RPNG representation as our central and only way to represent `Plaquette` instances. Following the thought process, that means that `RPNG` will become one of the means to create a `Plaquette`, but `Plaquette` instances will be the object that will be used internally in the code. This will be enforced in another refactor PR.


inmzhang (approved):
LGTM.

PR #457, opened 2025-01-29
Refactor `tqec.plaquette` and `tqec.templates`

nelimee:
Following #456, this PR refactors `tqec.plaquette` and `tqec.templates`. The main changes are:

1. Isolation of the `RPNG`-related code in `tqec.plaquettes.rpng`. That includes the `RPNGDesciption` data-structure and the `translator` module that was used to get `Plaquette` instances from `RPNGDescription` instances.
2. Include a few improvements to the `translator` module from #437 (fix for the ancilla qubit reset basis and filtering out unused qubits).
3. Move `tqec.circuit.generation` to `tqec.compile.generation` because:
    - with the exception of `tqec.circuit.generation`, the `tqec.circuit` module only requires `tqec.utils`, so moving `tqec.circuit.generation` out makes the dependency of `tqec.circuit` simple.
    - `tqec.circuit.generation` needed `Plaquettes`, which needs `Plaquette`, which needs `ScheduledCircuit`, which is part of `tqec.circuit`, which resulted in a cyclic import.
4. Move back `tqec.templates.indices.*` to `tqec.templates`, because the `RPNGTemplate` approach is not generic enough for our use case (plaquettes needed by spatial junctions are not `RPNG`-representable).


inmzhang (approved):
LGTM.

PR #459, opened 2025-01-30
Tentative implementation of spatial junctions

nelimee:
This is a third attempt at implementing spatial junctions, based on #457.

At the time of opening, this PR:

- extracts most of the from #446,
- moves `tqec.templates.library` to `tqec.compile.specs.library.generators`,
- make `RPNG` a mean of creating `Plaquettes` instances instead of the default representation used in the code base.

A few things are still missing:

- [ ] Just like in #446, one of the test is now failing. This test has been skipped in 54c949ff7bb1d315c75f0702b814c98f80483ed3, but should be fixed and re-integrated before merging.
- [ ] The new plaquettes definition to implement the extended stabilizers of spatial junction have not been implemented.
- [ ] It is currently unclear how we will implement the needed alternation between the plaquettes needed to implement extended stabilizers. This might require some hacks, or a larger refactor. (#460 is trying to solve that)

Fixes #420 because `RPNGTemplate` is now only used internally to ease the implementation of tests.
Fixes #387 with the new `BaseSubstitutionBuilder` implementation.
Fixes #384 and #417 by implementing both Hadamard pipes and spatial cubes.

PR #460, opened 2025-02-05
Introduce `tqec.blocks` 

nelimee:
*Note: this PR should eventually be merged in #459, but is not ready to be merged yet as it lacks tests and integration into the existing code base.*

--- 

This PR introduces the module `tqec.blocks` (that might be better placed in `tqec.compile.blocks`, to be confirmed).

The goal of this module is to replace `CompiledGraph` by a data-structure (`TopologicalComputationGraph`) that will allow more "scheduling schemes" to be returned by the `compile_block_graph` function.

### Why is `CompiledGraph` not enough?

One way of implementing spatial junctions requires to alternate two sets of plaquettes. This is currently impossible to encode using `CompiledGraph`.

As a side note, `CompiledGraph` is inherently limited to computation that can be represented using `Template`/`Plaquettes` instances. The next goal(s) for `tqec` is to support such computations (e.g., Y-basis state preparation / measurement).

### What is the replacement?

The architecture introduced in this PR revolves around the core concept of a layer:

- a `BaseLayer` (and all its subclasses) represents the equivalent of one round of the QEC code. 
- a `BaseComposedLayer` (and all its subclasses) represents 2 or more `BaseLayer` instances applied one after the other.

`BaseLayer` instances can be:

- `PlaquetteLayer` that uses the usual `Template` / `Plaquettes` combo to represent a layer,
- `RawCircuitLayer` that uses a user-provided circuit construction function to represent a layer,
- `LayoutLayer` that represents several `BaseLayer` instances executed in parallel on a 2-dimensional grid.

`BaseComposedLayer` instances can be:

- `RepeatedLayer` that repeats a given number of times (that can scale with `k`) an internal circuit that is represented either as a `BaseLayer` instance or as a `BaseComposedLayer` instance,
- `SequencedLayers` that represents a sequence of layers that are applied one after the other in time.

I think that this architecture is capable of representing most of the quantum circuits we will care about in the near future.


nelimee:
This PR will have to change a lot, so wait a few days before starting a review. First, I want to rebase it on `main` instead of `feat/spatial_junctions`. Then, I'll likely have to change a little bit the approach to avoid changing too much code. Finally, the code introduced is currently not used at all and will have to be integrated into the main code eventually, which will come with the new specs implementation in `feat/spatial_junctions`.

nelimee:
Replaced by #476

PR #465, opened 2025-02-07
Support taking PyZX `GraphS` with 3D positions as inputs

inmzhang:
This PR adds interoperability code for `PyZX`. The ultimate goal is to completely remove the existing `ZXGraph` data structure, which will be addressed in follow-up PRs.

In this PR, the `PositionedZX` data structure is introduced as a container class for `GraphS` from `PyZX`, along with attached 3D positions for each vertex. Additionally, the code from `tqec/computation/conversion.py` has been adapted into `tqec/interop/pyzx/synthesis/positioned.py` to support the conversion of a `PositionedZX` into a `BlockGraph`. In the future, additional block synthesis strategies should be added under the same directory.


src/tqec/interop/pyzx/synthesis/strategy.py
```diff
@@ -0,0 +1,40 @@
...
+def block_synthesis(
+    zx_graph: GraphS,
+    strategy: SynthesisStrategy = SynthesisStrategy.POSITIONED,
+    *,
+    positions: Mapping[int, Position3D] | None = None,
+    **_kwargs: dict[str, Any],
+) -> BlockGraph:
+    match strategy:
+        case SynthesisStrategy.POSITIONED:
+            g = PositionedZX(zx_graph, positions or {})
```
nelimee, line 39: I would add code here to check that `positions` is not `None` and throw a specific exception if it is instead of letting the constructor of `PositionedZX` throw an exception that will not be descriptive of the real issue.

nelimee (approved):
This looks good to me. I made a few formatting-nly adjustements. I only had a quick look at `pyzx/synthesis/positioned.py` because at first glance it seems like the algorithm is strongly inspired from the one that was used for `ZXGraph`. If that is not the case then please tell me, and I will review it in more details.

I will review the following PRs in the next few days, hoping to review them all before Wednesday without any certainty.

PR #466, opened 2025-02-07
Remove `ZXGraph` data structure from the code base

inmzhang:
This is a follow-up PR of #465 and tries to remove `ZXGraph` data structure completely.

Currently, the algorithm to find correlation surface is kept and adapted to take a `GraphS` as input, instead of utilizing  `compute_pauli_webs` from `PyZX`. That’s because:

1. I don't fully understand the algorithm in `compute_pauli_webs`.
2. `compute_pauli_webs` requires specifying `inputs`/`outputs` boundary vertices in ZX graph and does not support computing on no-port graphs.


src/tqec/gallery/memory.py
```diff
@@ -0,0 +1,21 @@
...
+    """Create a block graph with a single cube that represents a logical memory experiment.
+
+    Args:
+        observable_basis: The logical observable basis for the memory experiment.
+
+    Returns:
+        A :py:class:`~tqec.computation.block_graph.BlockGraph` instance.
+    """
+    g = BlockGraph(f"Logical {observable_basis} Memory Experiment")
+    node_kind = "ZXZ" if observable_basis == Basis.Z else "ZXX"
```
nelimee, line 19: Can be 
```suggestion
    node_kind = f"ZX{observable_basis.value}"
```

src/tqec/gallery/cz.py
```diff
@@ -91,33 +92,17 @@ def _resolve_ports(flows: list[str]) -> list[ZXKind]:
...
+            return ZXCube.from_str("XZX")
 
-    By default, the Hadamard edge in the CZ gate will be horizontal. If you
-    want to use vertical Hadamard edges, you can rotate the graph by 90 degrees
-    by calling `~tqec.computation.block_graph.BlockGraph.rotate`.
+    def _hkind(p: str) -> ZXCube:
+        if p == "Z":
+            return ZXCube.from_str("XZZ")
+        else:
+            return ZXCube.from_str("XXZ")
```
nelimee, line 106: If I understood correctly, `p` can only be `X` or `Z`, and so these can likely be replaced by
```py
def _vkind(p: str) -> ZXCube:
    return ZXCube.from_str(f"XZ{p}")
def _hkind(p: str) -> ZXCube:
    return ZXCube.from_str(f"X{p}Z")
```

src/tqec/gallery/cnot_test.py
nelimee, line 1: The previous tests were also checking the observables of the generated structure. Is there a specific reason to remove that check?

nelimee (commented):
Initial (very small) review for the "easy" parts.

src/tqec/gallery/cnot_test.py
inmzhang: I forgot to add them back... It reminds me that I also forget to verify the construction at the ZX graph level with the help of pyzx. I'll try to improve the tests.

nelimee (commented):
This is only part of a full review, I'll have to re-review a few things.

src/tqec/computation/correlation.py
```diff
@@ -1,15 +1,78 @@
...
+
+    Correlation surface is represented by a set of edges in the ZX graph. Each edge
+    consists of two nodes. The edge can be decomposed into two half edges, each
+    incident to a node. The half edges are labeled with the Pauli operator.
+    If the edge has no hadamard effect, the Pauli operators on the two half edges
+    should be the same. Otherwise, the Pauli operators should be flipped.
+
+    Note that the Pauli operators are represented by `Basis` enum, which only
+    contains `X` and `Z` operators. The `Y` operator is represented by two
+    edges with `X` and `Z` operators respectively.
```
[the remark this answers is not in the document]
inmzhang: If a correlation surface includes a Pauli Y edge—which can be viewed as both an X surface and a Z surface present in a pipe—we represent it using two `ZXEdge` instances instead of one. In principle, we could introduce a Pauli enum and represent it as a single edge. However, this decomposition offers some benefits when writing the code that visualizes the correlation surface and constructs abstract observables.

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -223,3 +234,49 @@ def has_obs_include(cube: Cube, correlation: ZXKind) -> bool:
...
+
+
+def _check_correlation_surface_validity(
+    correlation_surface: CorrelationSurface, g: GraphS
+) -> None:
+    """Check the ZX graph can support the correlation surface."""
+    # 1. Check the vertices in the correlation surface are in the graph
+    if missing_vertices := (correlation_surface.span_vertices() - g.vertex_set()):
+        raise TQECException(
+            f"The following vertices in the correlation surface are not in the graph: {missing_vertices} "
```
nelimee, line 246:
```suggestion
            "The following vertices in the correlation surface are not in the "
            f"graph: {missing_vertices} "
```

src/tqec/computation/correlation.py
```diff
@@ -19,356 +82,90 @@ class CorrelationSurface:
...
     Note:
-        We use the term "correlation surface" and "observable" interchangeably in the library.
-
-    Here we represent the correlation surface in terms of the ZX graph. The insight is that the spiders
-    pose parity constraints on the operators supported on the incident edges. The flow of the logical
-    operators through the ZX graph, respecting the parity constraints, forms the correlation between the
-    inputs and outputs. However, the sign of measurement outcomes is neglected in this representation.
-    And we need to recover the measurements when instantiating an explicit logical observable from the
+        We use the term "correlation surface", "pauli web" and "observable" interchangeably in
+        the library.
```
nelimee, line 86:
```suggestion
        We use the term "correlation surface", "pauli web" and "observable" 
        interchangeably in the library.
```

src/tqec/computation/correlation.py
```diff
@@ -19,356 +82,90 @@ class CorrelationSurface:
...
+
+    Here we represent the correlation surface by the set of edges it spans in the ZX graph.
+    The insight is that the spiders pose parity constraints on the operators supported on
+    the incident edges. The flow of the logical operators through the ZX graph, respecting
+    the parity constraints, forms the correlation between the inputs and outputs. However,
+    the sign of measurement outcomes is neglected in this representation. And we need to
+    recover the measurements when instantiating an explicit logical observable from the
     correlation surface.
```
nelimee, line 94:
```suggestion
    Here we represent the correlation surface by the set of edges it spans in 
    the ZX graph. The insight is that the spiders pose parity constraints on 
    the operators supported on the incident edges. The flow of the logical 
    operators through the ZX graph, respecting the parity constraints, forms 
    the correlation between the inputs and outputs. However, the sign of 
    measurement outcomes is neglected in this representation. And we need to
    recover the measurements when instantiating an explicit logical observable 
    from the correlation surface.
```

src/tqec/interop/pyzx/correlation.py
```diff
@@ -0,0 +1,315 @@
...
+    at different nodes are produced to form the correlation surface.
+
+    Args:
+        g: The ZX graph to find the correlation surfaces.
+        roots: The set of leaf nodes to start the correlation surface finding. If not provided,
+            all the leaf nodes in the graph are used.
+
+    Returns:
+        A list of `CorrelationSurface` in the graph.
+    """
```
nelimee, line 108:
```suggestion
    """Find the correlation surfaces in a ZX graph.

    Starting from each leaf node in the graph, the function explores how can the
    X/Z logical observable move through the graph to form a correlation surface:

    - For a X/Z type leaf node, it can only support the logical observable with 
      the opposite type. Only a single type of logical observable is explored 
      from the leaf node.
    - For a Y type leaf node, it can only support the Y logical observable, i.e. 
      the presence of both X and Z logical observable. Both X and Z type logical 
      observable are explored from the leaf node. And the two correlation 
      surfaces are combined to form the Y type correlation surface.
    - For the BOUNDARY node, it can support any type of logical observable. Both
      X and Z type logical observable are explored from it.

    The function uses a flood fill like recursive algorithm to find the 
    correlation surface in the graph. Firstly, we define two types of nodes in 
    the graph:

    - *broadcast node:* A node that has seen logical observable with basis 
      opposite to its own basis. A logical observable needs to be broadcasted to 
      all the neighbors of the node.
    - *passthrough node:* A node that has seen logical observable with the same 
      basis as its own basis. A logical observable needs to be only supported on
      an even number of edges connected to the node.

    The algorithm starts from a set of frontier nodes and greedily expands the 
    correlation surface until no more broadcast nodes are in the frontier. Then 
    it explore the passthrough nodes, and select even number of edges to be 
    included in the surface. If no such selection can be made, the search is
    pruned. For different choices, the algorithm recursively explores the next 
    frontier until the search is completed. Finally, the branches at different 
    nodes are produced to form the correlation surface.

    Args:
        g: The ZX graph to find the correlation surfaces.
        roots: The set of leaf nodes to start the correlation surface finding. 
            If not provided, all the leaf nodes in the graph are used.

    Returns:
        A list of `CorrelationSurface` in the graph.
    """
```

src/tqec/interop/pyzx/positioned.py
```diff
@@ -111,11 +107,128 @@ def positions(self) -> dict[int, Position3D]:
...
+        1. For each cube in the block graph, convert it to a ZX vertex.
+        2. For each pipe in the block graph, add an edge to the ZX graph with the corresponding endpoints and Hadamard flag.
+
+        Args:
+            block_graph: The block graph to be converted to a ZX graph.
+
+        Returns:
+            The :py:class:`~tqec.interop.pyzx.positioned_zx.PositionedZX` object converted from the block
+            graph.
+        """
```
nelimee, line 136:
```suggestion
        2. For each pipe in the block graph, add an edge to the ZX graph with 
           the corresponding endpoints and Hadamard flag.

        Args:
            block_graph: The block graph to be converted to a ZX graph.

        Returns:
            The :py:class:`~tqec.interop.pyzx.positioned_zx.PositionedZX` object 
            converted from the block graph.
        """
```

nelimee (approved):

nelimee:
Only minor comments on long lines. Except that, the code looks fine to me. I did not checked fully `tqec.interop.pyzx.correlation` because the algorithm seems very similar to what was already working before and the tests seems reasonable. If you think that I should take the time to review that part in detail before merging, please let me know.

inmzhang:
> Only minor comments on long lines. Except that, the code looks fine to me. I did not checked fully `tqec.interop.pyzx.correlation` because the algorithm seems very similar to what was already working before and the tests seems reasonable. If you think that I should take the time to review that part in detail before merging, please let me know.

I think it's fine because the algorithm stays the same as before and the tests should be enough.

PR #468, opened 2025-02-07
Rotated imports

jbolns:
The cumulative commits in this draft PR cover the following features (typing checks pending):
- Enable import rotations of simple structures
- Enable import rotations of complex structures without hadamards
- Enable import of rotations of structures with hadamards.

The commits also include a number of assets for testing. 
- Files contain different structures with different rotations
- Files have been rotated using different versions of SketchUp
  - Rotations from SketchUp pro have fairly clean matrices with integer positions
  - Rotations from SketchUp desktop have ugly matrices with long floats ("snap" to grid unavailable in SketchUp Desktop).

There is also a Jupyter file to test the pure imports
- This file is ready and all operations should work (give or take a path adjustment or two)
- The file does NOT cover simulations.

There is also a work-in-progress Python script to test simulations.
- Blockgraph builds seemingly fine
- Correlation surfaces build seemingly fine
- There *is* an issue with the compiled computation (has the right shape, but flows in the wrong direction).
  - Changes needed to address this seem unrelated to the operations in any of the commits in this PR
  - Probably best handled separately to not over-complicate a single PR.


inmzhang (commented):
It looks great! Here are some general comments:

- Unit tests for rotations should be added. However, I don't think we need the demo notebook or the integration test (`rotation_test`).
- A single `.dae` file containing various rotated blocks should be sufficient for testing. This approach will also help keep the test footprint lightweight.
- The `Y`-type cube should be handled as well. It has an additional constraint: it can only connect to time-direction pipes, and its rotation must preserve this invariant.

Additionally, we also need to support the direct rotation of the `BlockGraph` data structure. Given the functions introduced in this PR, implementing this should require minimal effort. If you're interested, you can include it in this PR or address it in a follow-up.

afowler:
Personally, I think we should forbid rotation of the Y and future state
cultivation blocks about any access other than the blue/Z/time axis. If we
detect that these have been rotated around any other axis, we should throw
an error.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,56 @@
...
+cubes if only part of the structure was rotated. But this becomes harder
+as the size of the computation grows.
+
+
+"""
+
+import numpy as np
+
+
+def calc_rotation_angles(M: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
jbolns: Yup

src/tqec/interop/collada/read_write.py
```diff
@@ -63,43 +64,106 @@ def read_block_graph_from_dae_file(
...
+                    any(
+                        [
+                            abs(int(angle)) not in [0, 90, 180]
+                            for angle in rotation_angles
+                        ]
+                    )
+                    or sum([abs(angle) for angle in rotation_angles]) < 180
+                ):
+                    raise TQECException(
+                        f"There is a non-identity rotation for {kind} block at position {translation}."
```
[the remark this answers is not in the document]
jbolns: Yup

src/tqec/interop/collada/read_write.py
```diff
@@ -63,43 +64,106 @@ def read_block_graph_from_dae_file(
...
+            # - If matrix YES rotated: check closer & make necessary adjustments
             if not np.allclose(transformation.rotation, np.eye(3), atol=1e-9):
-                raise TQECException(
-                    f"There is a non-identity rotation for {kind} block at position {translation}."
+                # Calculate rotation
+                rotation_angles = calc_rotation_angles(transformation.rotation)
+
+                # Reject invalid rotations:
+                # - Any != 0 or != 90 (degrees) rotation: partially rotated block/pipe
+                # - Less than 2 valid rotations: dimensional collapse.
```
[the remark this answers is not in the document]
jbolns: Yes, this seems fair. Actually, it can be 1 * 180-degree rotation or 2 * 90-degree rotation. The reason being that a single 90-degree rotation would put the vector for *this* or *that* axis on the plane made by the other two axes.

Not sure where to add explanation, though. Maybe I'll include the explanation at the top of the rotation file and reference it elsewhere.

src/tqec/utils/rotations.py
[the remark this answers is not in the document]
jbolns: Done. Also changed the type hint for the numpy array in the symbolic multiplication function.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Returns:
+        rotations (np.ndarray): the rotation angle for each of the three vectors in M
+    """
+    rotations = np.array([])
+    ID = np.identity(3, dtype=int)
+    for i, row in enumerate(M):
+        cos_theta = np.dot(ID[i], row) / (np.linalg.norm(ID[i]) * np.linalg.norm(row))
+        angle_rad = np.arccos(np.clip(cos_theta, -1.0, 1.0))
+        angle_deg = sum(row) * np.degrees(angle_rad)
```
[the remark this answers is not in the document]
jbolns: Yes, there was no need for it any more. Took it away, and also got rid of the abs() statements in related operations in read_write.py since all angles will now be positive

jbolns:
I think all comments by @inmzhang and @afowler are considered in the additional commits. 

I took one liberty and one liberty only. I left one of the rotated CNOTs in the assets folder for any quick testing anyone might want to do, and uploaded a SketchUp template with several rotations focusing not on correctness but on using as many primitives as possible. It's almost what Yiming suggested.

jbolns:
Forgot to delete demo notebook and integration tests.
- Last commit deletes demo notebook.
- Need to keep integration test at least momentarily: will need it if addressing blockgraph/simulation-specific rotations in this PR.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,99 @@
...
+Additionally, since the names of blocks/pipes in TQEC are tied to the face of the unrotated blocks,
+name equivalences can be calculated algebraically using the transformation matrix directly.
+
+"""
+
+import numpy as np
+import numpy.typing as npt
+
+
+def calc_rotation_angles(M: npt.NDArray) -> npt.NDArray:
```
inmzhang, line 39: `mypy` errror needs to be fixed. You can run `mypy src/tqec` to check for mypy errors.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,99 @@
...
+    """Calculates the angle between the vectors of a matrix (M) and the vectors of identity matrix (ID).
```
inmzhang, line 40: Maybe more explicit on what it means by "angle between the vectors and vectors". My understand is that it calculates the rotation angles of the three row vectors of matrix M from the original X/Y/Z axis.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Args:
+        M (array): 3x3 matrix.
```
inmzhang, line 43: By convention in `tqec`, there is no need to type annotate the arguments and returned value in docstring. It will be automatically generated by the sphinx docs with the function type annotation.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,99 @@
...
+    for i, row in enumerate(M):
+        cos_theta = np.dot(ID[i], row) / (np.linalg.norm(ID[i]) * np.linalg.norm(row))
+        angle_rad = np.arccos(np.clip(cos_theta, -1.0, 1.0))
+        angle_deg = np.degrees(angle_rad)
+        rotations = np.append(rotations, [round(angle_deg)])
+
+    return rotations
+
+
+def symbolic_multiplication(M: npt.NDArray, name: str) -> tuple[str, dict[str, int]]:
```
inmzhang, line 68: A name like `rotate_block_kind_by_matrix` might be more self-explanatory. This function will likely be used in the implementation of `BlockGraph` rotation. So it's better to move the error handling for Y cube here and remove the `!` mark. Then I think there is no need to return the `axes_directions` as well and the function signature can be more concise as `rotate_block_name_by_matrix(block_kind: BlockKind, rotate_matrix: npt.NDArray[float]) -> BlockKind`. Adding unit test for this function will be helpful.

assets/rotations.dae
inmzhang, line 1: I don't think both files need to exist. We aim to minimize the number of `assets` files. Maybe adding them to the Google Driver in tqec google group is a good option.

jbolns: Forgot to delete that one. Deleted locally. Will be updated in next commit/push.

src/tqec/utils/rotations.py
jbolns: Yup, that's actually clearer.

src/tqec/utils/rotations.py
jbolns: Ok

src/tqec/utils/rotations.py
jbolns: Now in line with the type in the incoming transformation matrix. Uploading in next commit/push

src/tqec/utils/rotations.py
jbolns: Right, this one's done and included in the commit I just pushed, with some caveats.

I took axes_directions into a function of its own to enable the rotate_block_... function to return only the new kind. This does create some redundancy during import, as the function is now called by the import and by the rotate_block_... function. It's a very inexpensive operation, and that way the rotate_block_... function can be used elsewhere to rotate blocks, so that should be fine. 

I had to take a function originally in read_write.py and move it to rotations. That's hardly the best place for that specific function, I just couldn't find a fitting /utils file to dump it. I'll leave it to you to define final destination.

&& 

Unit test .dae for "Y" blocks added. I cannot get a .dae to be more condensed than that. Not sure if that's me, or not having a Premium version, but even my smallest memory test has almost 400 lines of code.

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,158 @@
...
+        - rotate_matrix is NOT rotated: block_kind untouched
+        - rotate_matrix is rotated: block_kind rotated accordingly
+
+    Args:
+        rotate_matrix: rotation matrix for node.
+        block_kind: original kind.
+
+    Returns:
+        rotated_kind: rotated kind for the node.
+        axes_directions: up/down multipliers for each axis
```
inmzhang, line 117:
```suggestion
```

src/tqec/utils/rotations.py
```diff
@@ -0,0 +1,158 @@
...
+import numpy as np
+import numpy.typing as npt
+from tqec.computation.block_graph import BlockKind
+from tqec.utils.exceptions import TQECException
+
+from tqec.computation.pipe import PipeKind
+from tqec.computation.cube import YCube, ZXCube
+
+
+def block_kind_from_str(string: str) -> BlockKind:
```
inmzhang, line 44: This function can be placed at `src/tqec/computation/block_graph`.

inmzhang (commented):
Hi J, thanks for you fast iteration on this.

The included `.dae` file can be used for tests, but no tests are currently added in the code. We need to ensure that the code tests itself rather than relying on user input. Specifically, we should include a test case for rotated imports in `read_write_test`. It would be best to place the test `.dae` file alongside the test file, for example in `src/tqec/interop/collada/test_files/rotation_*.dae`. 

I'm also unclear about the necessity of the `rotation_combo.skp` file, as we can't import or test it in the code. My initial suggestion is to replace it with a rotation combo `.dae` file that can be imported and tested in the code, and then implement the corresponding test. If it is too hard to make such a `.dae` file, I think a rotated CNOT is enough for the test.

Finally, the unit tests for `utils/rotations` should be added to `utils/rotations_test.py`.

src/tqec/utils/rotations.py
jbolns: Done

jbolns:
Alright, that last commit should be more along those lines. 
- "Unit tests" for all functional units in `src\tqec\utils\rotations_test.py`.
  - No assets involved. All code-based. 
- "Unit tests" for all methods involved, i.e., `read_block_graph_from_dae_file`, in `src\tqec\interop\collada\read_write_test.py`.
  - Using a .dae with a rotated CNOT.

I'm glad you offered more details @inmzhang. Perhaps I took you too literally on the "removing assets" front. I was here scratching my head trying to figure out how to test an import without importing stuff. =D

inmzhang (approved):
With the final two comments addressed, we're ready to merge this. Thanks!

assets/rotations_cnot.dae
inmzhang, line 1: I suggest to put this under `src/tqec/interop/collada/test_files/` and resolve the file path in test by `Path(__file__).parent / "test_files/rotated_cnot.dae"`.

src/tqec/gallery/cnot.py
```diff
@@ -46,3 +46,47 @@ def cnot(observable_basis: Basis | None = None) -> BlockGraph:
     elif observable_basis == Basis.X:
         g.fill_ports(ZXCube.from_str("ZXX"))
     return g
+
+
+def rotated_cnot(observable_basis: Basis | None = None) -> BlockGraph:
```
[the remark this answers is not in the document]
jbolns: Done.

assets/rotations_cnot.dae
jbolns: I thought I had successfully ignored this. I really do not like it, but it's done.

inmzhang:
Thanks J, you can go ahead to merge this :-)

PR #475, opened 2025-02-11
Update the docs

inmzhang:
After removing the `ZXGraph` from the code base and refactor some methods on `BlockGraph`, some docs need to be updated.


nelimee (approved):
LGTM

Issue #458, opened 2025-01-30
Enable import of rotated SketchUp 3D structures


jbolns:
**Problem**
You can rotate structures in SketchUp, but you cannot import rotated structures to TQEC. This is inconvenient if thinking about whole structures (currently: go back to SketchUp, rotate, and re-export) and more than inconvenient when only parts of a structure are rotated (currently: manually delete each rotated element in SketchUp, replace with the unrotated equivalent, re-export).

*[Labelled as a good first issue because it does not require full knowledge of operations taking place after import. It is also, well, my first issue.]*

**General proposed approach**
The general proposed solution is relatively straightforward:
1. Modify existing error handling [here](https://github.com/tqec/tqec/blob/64c0031cee9457d1162bac5b0a9cf798eecb0b3e/src/tqec/interop/collada/read_write.py#L96) to let rotations pass while still catching illegal rotations. 
2. Create a new function (I like functions) or class to get rotational information for each node and:
  2.1 If the node is not rotated: carry on.
  2.2 If the node is rotated: exchange node for the un-rotated equivalent.
3. Handle cases when a rotated block/pipe fails to clear TQEC's "round to integer" checks or adjust the said checks. 

**Syntax challenges**
I think understanding syntax is key for this issue. 

SketchUp encodes shapes and rotations into a matrix which, as far as I can tell, follows a standard [OpenGL syntax](https://community.khronos.org/t/get-direction-from-transformation-matrix-or-quat/65502/2). I like to think of it as having a box in front of me, with a corner designated as the origin (position given in 4th column) and 3 vectors projecting shape right/left, up/down, front/back (the first 3x3 bit).

```
[X.x] [X.y] [X.z] [POS.x]
[Y.x] [Y.y] [Y.z] [POS.y]
[Z.x] [Z.y] [Z.z] [POS.Z]
[    ] [    ] [    ] [US   ]
```
You don't need degree rotations to rotate anything. You just change the 3x3 vectors to get the box to "grow" in a different direction. So, for instance, below, a "zxo" block unrotated, and rotated. Note that the "url/ID" remains. This *is** the essence of the challenge. When creating the block/pipe, what matters for blockgraph is the ID/name, which does NOT change when you rotate things in SketchUp.

``` xml

<node id="ID2" name="instance_0">
    <matrix>  <!-- TQEC uses this to confirm shape is valid and get origin (4th column) information. -->
        1 0 0 0
        0 1 0 0
        0 0 1 0
        0 0 0 1
[...]

PR #477, opened 2025-02-13
Shift correlation surface during compilation

inmzhang:
Fix https://github.com/tqec/tqec/issues/458#issuecomment-2656115898.


jbolns (approved):
Looks merge-able. Pulled locally, merged both this and mine into a joint test, and got a successful computation for the CNOT rotated 180.

PR #478, opened 2025-02-14
Correlation surface makes 90 degree turn in the junction

inmzhang:
As suggested by Austin, the visualization of the correlation at junction should have an angle of 90 degree to make it consistent with the measurements in the observable. After the change, the correlation surface should look like this:


nelimee (approved):
LGTM

PR #481, opened 2025-02-19
Implement `rotate` method on block graph

inmzhang:
With #468, we can now import a `.dae` file with rotation transformations as a `BlockGraph`. This PR introduces direct rotation functionality to the `BlockGraph` data structure, allowing us to rotate by multiples of 90 degrees using `block_graph.rotate`.

### DIFF
--- a/src/tqec/computation/block_graph_test.py
+++ b/src/tqec/computation/block_graph_test.py
@@ -3,6 +3,7 @@ import pytest
 from tqec.computation.block_graph import BlockGraph
 from tqec.computation.cube import ZXCube
 from tqec.computation.pipe import PipeKind
+from tqec.utils.enums import Basis
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Direction3D, Position3D
 
@@ -206,3 +207,17 @@ def test_graph_rotation() -> None:
     rg = g.rotate(Direction3D.Z)
     assert Position3D(-1, 0, 0) in rg
     assert str(rg.cubes[0].kind) == "Y"
+
+
+@pytest.mark.parametrize("obs_basis", [Basis.Z, Basis.X, None])
+def test_cnot_graph_rotation(obs_basis: Basis | None) -> None:
+    from tqec.interop.collada.read_write_test import rotated_cnot
+    from tqec.gallery.cnot import cnot
+
+    g = cnot(obs_basis)
+    rg = g.rotate(Direction3D.X, False)
+
+    rg_from_scratch = rotated_cnot(obs_basis)
+
+    # We need to shift the rotated graph in Z direction to match the two
+    assert rg.shift_by(dz=2) == rg_from_scratch
--- a/src/tqec/computation/block_graph_test.py
+++ b/src/tqec/computation/block_graph_test.py
@@ -198,3 +198,11 @@ def test_graph_rotation() -> None:
     assert str(rg[Position3D(-1, 1, 0)].kind) == "XZZ"
     assert str(rg.pipes[0].kind) == "XOZ"
     assert rg.rotate(Direction3D.Z, num_90_degree_rotation=3) == g
+
+    g = BlockGraph()
+    g.add_cube(Position3D(0, 0, 0), "Y")
+    with pytest.raises(TQECException):
+        g.rotate(Direction3D.X)
+    rg = g.rotate(Direction3D.Z)
+    assert Position3D(-1, 0, 0) in rg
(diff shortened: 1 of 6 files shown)


src/tqec/utils/rotations.py
```diff
@@ -34,9 +34,12 @@
...
+from scipy.spatial.transform import Rotation as R
+
+from tqec.computation.block_graph import BlockKind, block_kind_from_str
 from tqec.utils.exceptions import TQECException
-from tqec.computation.block_graph import block_kind_from_str
+from tqec.utils.position import Direction3D, Position3D
+from tqec.utils.scale import round_or_fail
 
 
 def calc_rotation_angles(M: npt.NDArray[np.float32]) -> npt.NDArray[np.float32]:
```
jbolns, line 45: I think we're calling this "rotation_matrix" elsewhere. Might as well match.

src/tqec/utils/rotations.py
```diff
@@ -142,3 +147,50 @@ def rotate_block_kind_by_matrix(
...
+        rotation_axis: axis to rotate around.
+        counterclockwise: whether to rotate counterclockwise.
+        angle: rotation angle in radians.
+
+    Returns:
+        The rotation matrix.
+    """
+    rot_vec = np.array([0, 0, 0])
+    rot_vec[rotation_axis.value] = 1 if counterclockwise else -1
+    return np.array(R.from_rotvec(rot_vec * angle).as_matrix(), dtype=np.float32)
```
jbolns, line 169: Did not check operation. Since this is SciPy, I'm just going to trust it does what it "says on the tin". I mean, if we can't trust SciPy, we have bigger problems than this function, haha!

src/tqec/utils/rotations.py
```diff
@@ -142,3 +149,50 @@ def rotate_block_kind_by_matrix(
...
+
```
jbolns, line 172: Right, this one is abstraaaaact. Please confirm. So, in theory, thanks to SciPy, this would return the position of any incoming element entirely into the new axis. So, a block at 0,0,6 would end at 0,6,0 if we rotate on X (I think clockwise). Is this correct? If so, yeah, exactly what it should happen.

src/tqec/computation/block_graph_test.py
```diff
@@ -186,3 +186,23 @@ def test_single_connected() -> None:
 
     g.add_pipe(n1, n2)
     assert g.is_single_connected()
+
+
+def test_graph_rotation() -> None:
```
jbolns, line 191: Can we not use the CNOT and rotated CNOT we use interop/collada/read_write_test.py to test this more ambitiously?

jbolns:
@inmzhang . Completely unrelated, but are you getting warnings on block_graph.py (line 63)?

src/tqec/utils/rotations.py
inmzhang: > So, a block at 0,0,6 would end at 0,6,0 if we rotate on X (I think clockwise). Is this correct? If so, yeah, exactly what it should happen.

A "point" at (0,0,6) will rotate to (0,6,0). For a block, as explained in the docstring of `rotate_position_by_matrix`, the center of the cube would be rotate so there might be some offset in calculation. In such case, a cube at (0,0,6) would rotate to (0,6,-1) I believe and the offset is handled in `rotate_position_by_matrix`.

src/tqec/computation/block_graph_test.py
inmzhang: Indeed, we can. I'll add that to the test.

inmzhang:
> @inmzhang . Completely unrelated, but are you getting warnings on block_graph.py (line 63)?

I think it's because you did not install the dev dependencies? There is the `types-networkx` which is the type stub file for `networkx`.

src/tqec/utils/rotations.py
jbolns: Yeah, then all good.

src/tqec/computation/block_graph_test.py
jbolns: Cool

jbolns:
@inmzhang Maybe. Altho, I think I did. Will check what's going on my side then, since it's local then. Thanks!

jbolns (approved):
All seems to check out. 
Ps. Had to do some stuff so I couldn't check them immediately, but it was impressive how quickly you turned around the two last commits. I couldn't turn back something that quickly even if I tried. You're legit a genius!

PR #265, opened 2024-07-07
Fill template from backend

giangiac:
This is the second step of the 4-step plan:

- [x] merge #250 so that the frontend v2 [what we called the alternative frontend and should be the official frontend moving forward] returns to the main TQEC repo in folder `<repo>/ggg_frontend`
- [ ] merge latest development to the frontend v2 [from Gian's fork]
- [ ] create another PR to move frontend v1 to `<repo>/old_frontend`, and move frontend v2 to `<repo>/frontend`
- [ ] find a way to update the tqec website

Specifically, what are the features added to the frontend v2?

- add tab 4 in which one can fill templates using the plaquette from tab 2
- interface between the JS frontend and a simple Python program acting as the backend
- the frontend request data from the backend in the form of a JSON file containing a library of templates
- the user can select the template to fill from a drop-down menu on top of tab 4
- drag-and-drop one plaquette-type per cell with different label, then click on the "fill the template" button to complete the template

PR #483, opened 2025-02-21
Removing the frontend

### LINKED ISSUE #413: Moving `ggg_frontend` to its own repository
> Concerning moving the frontend to a separate repo, I think that's possible since the codebases are largely independent.
> I would just like to discuss what workflow would take advantage of both tools (frontend and backend):
> For example one could use the former to create the library used by the latter, but I am not up to date with the I/O of the backend.
>
> _Originally posted by @giangiac in https://github.com/tqec/tqec/issues/325#issuecomment-2400361154_

The `ggg_frontend` would be better placed in its own repository because:

- its development might be stale for a while (we are not sure it is needed anymore),
- it helps having the `tqec` repository only about the Python library, and other tools around `tqec` in their own repository.

nelimee:
This PR removes the `frontend` (in `ggg_frontend`) from the main `tqec` repository and moves it to https://github.com/tqec/frontend.

Closes #265 (whose changes are included in https://github.com/tqec/frontend).
Closes #413.


inmzhang (approved):
```sh
❯ rg -i "frontend"
CONTRIBUTING.md
5:If you found a bug or would like a missing feature to be implemented, [create an issue](https://github.com/tqec/tqec/issues/new/choose). In order for your issue to be efficiently treated, please tag it with either of the `backend` or `frontend` tags and include as much information as possible.
```

Can be merged after addressing the above issue.

PR #486, opened 2025-02-24
Create a specs library

nelimee:
This PR cherry-picks a few of the important commits from #459.

The few notable changes are organised by commits. 3fcc5a1ca5c665de275a64da03098a875a366aa0 is the only commit that does not come directly from #459. It removes the module `tqec.compile.specs.library.base` introduced by 07d8ca8fcefafedcfd974bad9fd9d0f5a4bb726a because this module is out of the scope of this PR.


src/tqec/compile/specs/library/generators/__init__.py
Zhaoyilunnn, line 1: This PR looks good to me. I just have one question: Why adding so many seemingly redundant `impot xx as xx` 
For example, why not just

```python
from .hadamard import get_spatial_horizontal_hadamard_plaquettes, get_spatial_horizontal_hadamard_raw_template, xxx, ...
```

nelimee: You have a better explanation than what I can write here: https://docs.astral.sh/ruff/rules/unused-import/.

Basically, if not "re-exported" (i.e., without `as ZZZ`), the `from XXX import YYY` is seen by linters as a unused import, even in `__init__.py` files, and that is annoying.

inmzhang (approved):
LGTM.

PR #488, opened 2025-02-27
Implement Plaquette for spatial junctions

Zhaoyilunnn:
This PR extracts commits from https://github.com/tqec/tqec/pull/487 and aims to provide the actual circuit implementation of plaquettes required for the stretched stabilizer in spatial junction.

Will open another PR to implement the `get_spatial_cube_arm_plaquettes` method


src/tqec/plaquette/library/spatial.py
```diff
@@ -0,0 +1,198 @@
...
+        f ----- g
+
+    This is split into two plaquettes, with "UP" being (a, b, c, d) and "DOWN" being (d, e, f, g).
+
+    Args:
+        basis: the basis of the plaquette.
+        plaquette_kind: the kind of the plaquette.
+        reset: the logical basis for data initialization.
+        measurement: the logical basis for data measurement.
+        is_reverse: whether the plaquette has controlled-A reversed.
```
nelimee, line 47: I would add a precision on what you mean by "reversed" as it was unclear to me at first glance. I *think* you mean "reverse the schedules", I'll likely understand it better when reading the implementation.

src/tqec/plaquette/library/spatial.py
```diff
@@ -0,0 +1,198 @@
...
+        is_corner_trimmed: whether the plaquette has corner trimmed, for "UP" plaquette
+            the left top corner is trimmed, for "DOWN" plaquette the right bottom corner is trimmed.
+
+    Returns:
+        A plaquette for spatial cube arms.
+
+    Notes:
+        This method could generate 8 different plaquettes.
```
nelimee, line 55:
```suggestion
        This method can generate 8 different plaquettes.
```

src/tqec/plaquette/library/spatial.py
```diff
@@ -0,0 +1,198 @@
...
+        targ_order = [4, 3] if self._is_reverse else [3, 4]
+        self._append_ctrl_op_to_data_qubit(circuit, targ_order[0])
+        circuit.append("TICK")
+        circuit.append("TICK")
+        self._append_ctrl_op_to_data_qubit(circuit, targ_order[1])
+        circuit.append("TICK")
+        circuit.append("CX", [0, 1], [])
+        circuit.append("TICK")
+        circuit.append("MX", [0], [])
+        return list(iter_stim_circuit_without_repeat_by_moments(circuit))
```
nelimee, line 186: These two methods would benefit from a docstring where you explicitly recall the numbering of each qubit. For example, for the first one, having something like
```
1 ----- 2
|       |
|   0   |
|       |
3 ----- 
```
would help a lot any reader interested in the implementation.

src/tqec/plaquette/library/spatial.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def _build_memory_moments_down(self) -> list[Moment]:
+        circuit = stim.Circuit()
+        circuit.append("RZ", [1], [])
+        circuit.append("TICK")
+        circuit.append("RZ", [0], [])
+        circuit.append("TICK")
+        circuit.append("CX", [1, 0], [])
+        circuit.append("TICK")
+        targ_order = [4, 3] if self._is_reverse else [3, 4]
```
nelimee, line 177: I do not understand how this qubit indexing can work. When trying to do the circuit by myself to check your implementation, I came up with `[3, 2] if self._is_reverse else [2, 3]`. Could you explain your indexing here?

nelimee (commented):
The implementation seems correct according to the tests. I do not understand your indexing on the DOWN plaquette, but it seems to work, could you try to explain me? 
Also, a few comments on documentation.

src/tqec/plaquette/library/spatial.py
Zhaoyilunnn: Yes your understanding is correct.  will improve the description to make it clear.

src/tqec/plaquette/library/spatial.py
Zhaoyilunnn: Thanks. Will add this

src/tqec/plaquette/library/spatial.py
nelimee: Yes! I simply skipped (2) in my head, and that was my mistake. Your indexing is the correct one, thanks for the explanation.

src/tqec/plaquette/library/spatial.py
nelimee: I think this is the only point left to address. Once addressed, I'll accept the PR.

Zhaoyilunnn: I have updated the docstring in the last commit. Could you elaborate what's left for this point, maybe I missed sth

nelimee: I do not see any change in the docstring of the `is_reverse` parameter. Did I miss something?

Zhaoyilunnn: image

At my [side](https://github.com/tqec/tqec/pull/488/files#diff-235a66b7fd49769effd70125689da67573a81da3db8a788bf91024d1048be280), the newest code looks like above. 

And in this page, it is already outdated like below

image

I modified the docstring from 

"whether the plaquette has controlled-A reversed."

to 

"whether the schedules of controlled-A gates are reversed."

Is it because this two sentence look too similar? 😄

nelimee: It is my fault. I was expecting a longer description and checked too quickly. Reading again the docstring, I think it is perfect! I'll accept the PR and merge :)

nelimee (approved):
LGTM

PR #492, opened 2025-02-28
Add cube/pipe naming convention

purva-thakre:
This PR adds a sub-section discussing the naming convention for the objects discussed in the `Terminology` section of the user guide. This section had previously skipped a discussion about the RGB --> XYZ axes, open wall, Hadamard band, etc. 

Let me know if I have missed anything that ought to be included here. 

Edit: Also replaces `YCube` with `YHalfCube`.


afowler:
Minor comment that it is really a green half cube, not a full cube. This represents Y basis measurement or initialization. It's also worth mentioning that this block must be oriented such that the thin direction is along the time axis.

docs/user_guide/terminology.rst
```diff
@@ -19,6 +19,38 @@ This is the essence of fault-tolerant quantum computation.
 
 In ``tqec``, a block is either a :ref:`Cube <cube>` or a :ref:`Pipe <pipe>`.
 
+The naming convention for a cube or a pipe is as follows:
+
+- The axes used for labeling are as shown in the figure below. ``RGB`` axes are synonymous to ``XYZ`` axes.
+- We begin by labeling the boundary that's facing the X-axis first, then the one that's facing the Y-axis followed by the one facing the Z-axis.
```
nelimee, line 25: Minor nitpick
```suggestion
- We begin by labeling the boundary that is facing the X-axis first, then the one that is facing the Y-axis followed by the one facing the Z-axis.
```

docs/user_guide/terminology.rst
```diff
@@ -19,6 +19,38 @@ This is the essence of fault-tolerant quantum computation.
...
+The labels based on color of the boundaries are provided in the table below.
+
+.. list-table::
+   :header-rows: 1
+   :align: center
+
+   * - Color
+     - Boundary
+   * - Red Wall
+     - x
```
nelimee, line 43: The labels are more often used in upper-case I think, it might be less confusing to have upper case here too.

docs/user_guide/terminology.rst
```diff
@@ -19,6 +19,38 @@ This is the essence of fault-tolerant quantum computation.
...
+   * - Green Cube
```
nelimee, line 44:
```suggestion
   * - Green Half-Cube
```

purva-thakre: Would it make sense to relabel this class to `YHalfCube`? 

https://tqec.github.io/tqec/_autosummary/tqec.computation.YCube.html#tqec.computation.YCube

It is called the `YCube` in the terminology a couple of sections later. 
https://tqec.github.io/tqec/user_guide/terminology.html#ycube

nelimee: I think that would make sense yes.

nelimee: See #493

nelimee (approved):
LGTM, renaming `YCube` is left to #493

purva-thakre:
Don't merge this PR yet. I will add a few commits later today. This will also include the fix for #493

purva-thakre:
@nelimee Apologies for the delay. This PR can be merged now.

nelimee:
The command
```sh
find src/tqec -name "*.py" -exec grep -Ei "Y(-| )cube"  {} \;
```
returns
```
            # Do not add surfaces in ports or Y-cubes
        # 1 y cube block
        """Geometry for the y cube."""
                f"The pipe must connect two nearby cubes in direction {self.kind.direction}."
        - **Time-like Y:** Y cubes can only have time-like pipes connected to them.
                    f"Y cube at {cube.position} does not have exactly one pipe connected."
                    f"Y cube at {cube.position} has non-timelike pipes connected."
        Exception, match="The pipe must connect two nearby cubes in direction Y."
```
on my computer. 
References to "Y-cube" or "Y cube" should also be changed in this PR. You can ignore a few lines above that are false matches ("nearb**y cube**s" for example).

nelimee (approved):
LGTM

PR #495, opened 2025-03-06
Minor docstring fix

### LINKED ISSUE #474: Fix formatting in docstring of `RPNGDescription`
The figure in the docstring of `RPNGDescription` appears weird.

**What it is supposed to look like:** 

https://github.com/tqec/tqec/blob/b41d4b18f925de04a14a5a54c2579e1661965fca/src/tqec/plaquette/rpng/rpng.py#L163-L170 

**What it looks like:**

Image

purva-thakre:
Fixes #474 

image


nelimee:
Note that the amount of documented modules in the API reference was reduced in a previous PR. Here, you add back a few modules to the API documentation.

My personal opinion is that the API documentation should include everything, and I cannot remember why we limited that previously. @inmzhang do you remember by any chance?

inmzhang:
> Note that the amount of documented modules in the API reference was reduced in a previous PR. Here, you add back a few modules to the API documentation.
> 
> My personal opinion is that the API documentation should include everything, and I cannot remember why we limited that previously. @inmzhang do you remember by any chance?

I remember that we only list the 2nd level modules at `api.rst` because we were trying to keep the modules tree appeared at https://tqec.github.io/tqec/api.html and sidebar as clean as possible. The functions/classes are intended to be included in the documentation need to be imported at the module level `__init__.py`.

If you think it's better to list some sub-modules at the module tree, I'm quite open to the change.

purva-thakre:
Well, my issue with the current version of `api.rst` is that not all submodules get listed under a module. Indeed this does not look clean. 

image

purva-thakre:
Let's go with options 1, 2, and 3.

PR #504, opened 2025-03-10
SketchUp import failure: delete spaces

angelaelisa:
Fix SketchUp import failure by adjusting face popping logic: delete spaces. Before, the import of the DAE file failed because SketchUp does not support spaces in node's ID. Just delete the spaces in the string returned by def __str__ and it's done (@inmzhang gave me the solution, actually :)


nelimee (approved):
Looks good to me :)

Congratulations for your first contribution! :tada:

PR #510, opened 2025-03-13
Implement spatial junctions, replacing the "odd parity" invariant 

nelimee:
The goal of this PR is to explore the implementation of spatial junctions without extended stabilizers. This is possible by changing one invariant from the code base.

Before, we ensured that logical qubits where all odd-parity. That means that "in clockwise order, the first plaquette being set on the boundary has an odd index when using ~~0~~1-based indexing"*. Basically, logical qubits should always look like the following:

image

(from the [Terminology section of the documentation](https://tqec.github.io/tqec/user_guide/terminology.html)).

That invariant can be changed to "the top-left plaquette in the bulk is always measuring a Z stabilizer (i.e., a blue plaquette)".

That simple change remove the need to implement extended stabilizers to implement spatial junctions (but there is no free lunch: spatial hadamard would now require such extended stabilizers).

*: ~~that definition may change if it turns out to be wrong, because the definition was never written anywhere from what I remember.~~ It changed.

PR #511, opened 2025-03-15
Fix shadowed faces to match pipe color

inmzhang:
Fixes https://github.com/tqec/tqec/pull/510#issuecomment-272544649.

In the following `.dae` file, the highlighted cube is incorrectly defined as `ZZX` instead of `ZXZ`. Although it appears correct because its front and back faces are shadowed by connected pipes, this discrepancy can affect cube kinds and the compilation pipeline.

image

This PR introduces a new method, `fix_shadowed_faces`, in `BlockGraph` to detect and correct such shadowed faces by adjusting their basis to match the pipe color. For example, the cube above will be properly corrected to `ZXZ`. The fix procedure has also been integrated into `BlockGraph.from_dae_file`, allowing users to make such fixable mistakes in SketchUp without affecting the final compilation.

### DIFF
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -641,7 +641,7 @@ class BlockGraph:
                 pos_map[pipe.v.position],
                 cast(PipeKind, rotated_kind),
             )
-        return rotated
+        return rotated.fix_shadowed_faces()
 
     def fix_shadowed_faces(self) -> BlockGraph:
         """Fix the basis of those shadowed faces of the cubes in the graph.
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -676,11 +676,6 @@ class BlockGraph:
             pipes_by_direction: dict[Direction3D, list[Pipe]] = {}
             for pipe in self.pipes_at(cube.position):
                 pipes_by_direction.setdefault(pipe.direction, []).append(pipe)
-            shadowed_directions = {
-                d for d, ps in pipes_by_direction.items() if len(ps) == 2
-            }
-            if not shadowed_directions:
-                continue
             # No need to handle the case `len(pipes_by_direction) == 0` as there
             # is no pipes connected to the cube.
             # No need to handle the case `len(pipes_by_direction) == 3` as it's
@@ -688,8 +683,14 @@ class BlockGraph:
             # Spatial pass-through, ensure that the cube is not a spatial cube
             if len(pipes_by_direction) in [0, 3]:
                 continue
+            shadowed_directions = {
+                d for d, ps in pipes_by_direction.items() if len(ps) == 2
+            }
+            if not shadowed_directions:
+                continue
             new_kind = cube.kind
             for shadowed_direction in shadowed_directions:
+                # Spatial pass-through, ensure that the cube is not a spatial cube
                 if (
                     len(pipes_by_direction) == 1
                     and shadowed_direction != Direction3D.Z
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -670,40 +670,53 @@ class BlockGraph:
         """
         fixed_cubes: dict[Cube, Cube] = {}
         for cube in self.cubes:
+            if not isinstance(cube.kind, ZXCube):
+                continue
+            # Group connected pipes by direction
             pipes_by_direction: dict[Direction3D, list[Pipe]] = {}
             for pipe in self.pipes_at(cube.position):
                 pipes_by_direction.setdefault(pipe.direction, []).append(pipe)
-            # Regularize the cube in a spatial pass-through
-            if len(pipes_by_direction) == 1:
-                direction = next(iter(pipes_by_direction))
-                if len(pipes_by_direction[direction]) != 2:
-                    continue
-                if direction == Direction3D.Z or not cube.is_spatial:
-                    continue
-                kind = cube.kind
-                assert isinstance(kind, ZXCube)
-                basis = kind.get_basis_along(direction)
-                new_kind = kind.with_basis_along(direction, basis.flipped())
-                new_cube = Cube(cube.position, new_kind, cube.label)
-                fixed_cubes[cube] = new_cube
-            # Fix the shadowed faces
-            if len(pipes_by_direction) != 2:
+            shadowed_directions = {
+                d for d, ps in pipes_by_direction.items() if len(ps) == 2
+            }
+            if not shadowed_directions:
                 continue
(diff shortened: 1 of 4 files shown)


src/tqec/interop/collada/read_write.py
```diff
@@ -214,7 +218,15 @@ def offset_y_cube_position(pos: FloatPosition3D) -> FloatPosition3D:
             graph.add_cube(tail_pos, Port(), label=f"Port{port_index}")
             port_index += 1
         graph.add_pipe(head_pos, tail_pos, pipe_kind)
-    return graph
+
+    graph_fixed = graph.fix_shadowed_faces()
+    if graph_fixed != graph:
+        warnings.warn(
+            "There are some shadowed faces have been fixed. "
+            "Check `BlockGraph.fix_shadowed_faces` for more details.",
```
nelimee, line 226:
```suggestion
            "Some shadowed faces have been fixed. "
            "Check `BlockGraph.fix_shadowed_faces` for more details.",
```

src/tqec/computation/block_graph.py
```diff
@@ -641,7 +641,79 @@ def rotate(
...
+                if len(pipes_by_direction[direction]) != 2:
+                    continue
+                if direction == Direction3D.Z or not cube.is_spatial:
+                    continue
+                kind = cube.kind
+                assert isinstance(kind, ZXCube)
+                basis = kind.get_basis_along(direction)
+                new_kind = kind.with_basis_along(direction, basis.flipped())
+                new_cube = Cube(cube.position, new_kind, cube.label)
+                fixed_cubes[cube] = new_cube
```
nelimee, line 688: It is still unclear to me if that following case is handled in the next parts of the code, but a T-shaped junction would not be caught here, but may have to be corrected. In order to catch potential pass-through nodes, I would check for the two spatial dimensions. Something like:
```py
for spatial_dir in [Direction3D.X, Direction3D.Y]:
    if spatial_dir in pipes_by_direction and len(pipes_by_direction[spatial_dir]) == 2:
        # Fix the cube
```

I might amend that comment later when understanding the second part of the function.

src/tqec/computation/block_graph.py
```diff
@@ -641,7 +641,79 @@ def rotate(
...
+            # Fix the shadowed faces
+            if len(pipes_by_direction) != 2:
+                continue
```
nelimee, line 691: If I understood correctly:

- `len(pipes_by_direction) == 0` can only be seen for an isolated cube (memory experiment or stability experiment),
- `len(pipes_by_direction) == 1` is a spatial pass-through that might needs to be corrected,
- `len(pipes_by_direction) == 2` may be a spatial pass-through, that might need to be corrected, or a regular spatial junction, or a regular junction with one pipe in space and one pipe in time,
- `len(pipes_by_direction) == 3` is impossible, because that would make a spatial junction with a pipe in time, which is disallowed.

If that is effectively the case, maybe this `if` can be accompanied with a comment explaining why all the other cases are not concerned by the method.

src/tqec/computation/block_graph.py
```diff
@@ -641,7 +641,79 @@ def rotate(
...
+                pipe_basis = need_match_pipe.kind.get_basis_along(
+                    d_this, need_match_pipe.at_head(cube.position)
+                )
+                assert pipe_basis is not None
+                if pipe_basis == basis_this:
+                    continue
+                new_kind = new_kind.with_basis_along(d_this, pipe_basis)
+            if new_kind != cube.kind:
+                new_cube = Cube(cube.position, new_kind, cube.label)
+                fixed_cubes[cube] = new_cube
```
nelimee, line 709: I find this part quite difficult to understand. For me, the code is trying to be generic on the directions `d1` and `d2` without a clear need. I might be missing something, but:

- if `Direction3D.Z not in (d1, d2)`, there is nothing to do (it is a legit spatial junction).
- else, if the other direction (i.e., not the one that is `Direction3D.Z`) has `1` pipe, there is nothing to do,
- else, the other direction has `2` pipes, this is a `T`-shaped junction, the cube should be corrected.

Am I missing something here?

nelimee (commented):
A few comments, I feel like the main method can be simplified, but might have missed some intricacies or have wrong pre-conditions on the `BlockGraph` input.

src/tqec/computation/block_graph.py
inmzhang: When `len(pipes_by_direction)==1`, it will only handle the straight-line-passthrough case like this:

image

It will try to make it a regular cube instead of spatial cube by fixing the shadowed face. For example, a `XXZ` cube in the figure will be turned into a `ZXZ`.

src/tqec/computation/block_graph.py
inmzhang: It's correct. I'll see how to simplify this.

nelimee: If the above is correct then isn't 

```py
for spatial_dir in [Direction3D.X, Direction3D.Y]:
    if spatial_dir in pipes_by_direction and len(pipes_by_direction[spatial_dir]) == 2:
        # Fix the cube
```

sufficient to implement the whole method? Taking all the restrictions (the list of cases in another comment) into account.

nelimee: With an addition to avoid checking if we have a real spatial junction.

inmzhang: Now I remember that, when I implemented this, I also tried to fix a four-way spatial junction, which have two pairs of the faces that are shadowed.

inmzhang:
Don't know how to simplify further, but I added more comments and did some refactoring.

src/tqec/computation/block_graph.py
```diff
@@ -641,7 +641,92 @@ def rotate(
...
+            }
+            if not shadowed_directions:
+                continue
+            # No need to handle the case `len(pipes_by_direction) == 0` as there
+            # is no pipes connected to the cube.
+            # No need to handle the case `len(pipes_by_direction) == 3` as it's
+            # a 3D corner that cannot be a valid structure.
+            # Spatial pass-through, ensure that the cube is not a spatial cube
+            if len(pipes_by_direction) in [0, 3]:
+                continue
```
nelimee, line 690: This can be moved before the computation of `shadowed_directions`.

src/tqec/computation/block_graph.py
```diff
@@ -641,7 +641,92 @@ def rotate(
...
+            new_kind = cube.kind
+            for shadowed_direction in shadowed_directions:
+                if (
```
nelimee, line 693: Maybe add a one-line comment to explicitly write that this is for a pass-through node?

nelimee (approved):

PR #513, opened 2025-03-15
Using detector database in `start_simulation_using_sinter`

nelimee:
This small PR improves by a lot the efficiency of `start_simulation_using_sinter` by implementing a smarter circuit generation method.

Before this PR, performing simulations for `K` different values of `k` and `P` different values for the noise parameter `p` called the **whole** `tqec` compilation pipeline `K * P` times. To avoid very long runtime, this was paralellised. 

It turns out that 2 optimisations can be applied:

- Use `DetectorDatabase`, because we know that the detector computation part is costly, and we know that the computed detectors for each quantum circuit are largely similar (because the same computation is created several times, for different values of `k`). Note that, for the moment, that requires the circuit generation to be fully sequential (see below for notes on that part).
- Avoid re-generating `P` times the same circuit for each values of `k` by generating the noiseless circuit for `k` once, and then applying the noise model `P` times on the generated noiseless circuit.

On my laptop (Intel i7-10610U (8 cores) @ 4.900GHz), the parallel "dumb" generation took 5m30s. The sequential generation with a detector database takes 45s.

Note on parallelism: the only thing that makes the method implemented in this PR sequential is the fact that the detector database is not made for parallel access. It might be possible to implement that feature at low cost later with [`multiprocessing.Manager`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Manager).


afowler:
Question: do we package all of the compatible observables together so we do
a single set of simulations? For example, for the Steane code, we should
only need to run two sets of simulations, one for all the X observables and
one for all the Z observables.

inmzhang:
> Question: do we package all of the compatible observables together so we do a single set of simulations? For example, for the Steane code, we should only need to run two sets of simulations, one for all the X observables and one for all the Z observables.

No, issue #364 is still open.

inmzhang (approved):

PR #514, opened 2025-03-16
Update the description of label attribute

angelaelisa:
@inmzhang 

Clarified that any str can be used as a label, but must be unique:

The label of the cube. It’s mainly used for annotating the input/output ports of the block graph. If the cube is a port, the label must be non-empty **and unique within the block graph. The label can be any string, but duplicate labels are not allowed**. Default is an empty string.


inmzhang:
Hi Angela, you need to fix [pre-commit hooks](https://pre-commit.com/) error. There is trailing whitespace in the docstring.

angelaelisa:
> Hi Angela, you need to fix [pre-commit hooks](https://pre-commit.com/) error. There is trailing whitespace in the docstring.

Sorry, this won't happen again, I'll check before, I promise :)

inmzhang (approved):

Issue #517, opened 2025-03-16
Look for a cube using its label


angelaelisa:
## Method to get cubes by label

Labeling could be a good way to identify cubes in the `BlockGraph`. Maybe it could return the `position` and the `CubeKind` of a `Cube` searching by `label`. It could be really useful to operate over cubes without thinking in which position they are. For example, numbering cubes with 'label' arg makes easier to create pipes.

## Possible solution

`get_cubes_by_label(label)`

With a given str, the label, returns the 3D integer position, and the `CubeKind` (ZXCube | Port | YHalfCube).

## Alternatives

Maybe it could be better to return, instead of `CubeKind` type, the string. I'm not sure about this.

## Additional context

See #512

inmzhang:
Closed by #519.

PR #519, opened 2025-03-16
Get cubes by label

angelaelisa:
Look for a cube using its label #517

get_cubes_by_label(label): With a given str, the label, returns the 3D integer position, and the CubeKind (ZXCube | Port | YHalfCube)

I checked using this:
```
    result = block_graph.get_cubes_by_label("P3")

    if result:
        position, kind = result
        print(f"Cube found at {position} with kind {kind}")
    else:
        print("No cube found with that label")
```
Result:
```
python cubes_label_test.py
Cube found at (-1,0,3) with kind PORT
```

### DIFF
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -729,6 +729,18 @@ class BlockGraph:
             new_graph.add_pipe(pipe.u.position, pipe.v.position, pipe.kind)
         return new_graph
 
+    def get_cubes_by_label(self, label: str) -> list[Cube]:
+        """
+        Find cubes with the specified label in the BlockGraph.
+
+        Args:
+            label: The label of the cubes.
+
+        Returns:
+            The cube instances that have the specified label.
+        """
+        return [cube for cube in self.cubes if cube.label == label]
+
 
 def block_kind_from_str(string: str) -> BlockKind:
     """Parse a block kind from a string."""


src/tqec/computation/block_graph.py
```diff
@@ -729,6 +729,22 @@ def fix_shadowed_faces(self) -> BlockGraph:
             new_graph.add_pipe(pipe.u.position, pipe.v.position, pipe.kind)
         return new_graph
 
+    def get_cubes_by_label(self, label: str):
```
inmzhang, line 732: Return type annotation needs to be added.

src/tqec/computation/block_graph.py
```diff
@@ -729,6 +729,22 @@ def fix_shadowed_faces(self) -> BlockGraph:
...
+
+        Args:
+            label: The label (string) of the cube.
+
+        Returns:
+            A tuple (position, kind) if found, otherwise None.
+        """
+        for cube in self.cubes:  # Using the existing cubes property
+            if cube.label == label:
+                return (cube.position, cube.kind)
```
inmzhang, line 744: We can return `Cube` instead of its attributes.

inmzhang (commented):
Hi Angela, thanks for the PR. What if some normal cubes have the same label? It makes sense to me that we collect all the cubes with the same label and return a `list[Cube]`.

angelaelisa:
> Hi Angela, thanks for the PR. What if some normal cubes have the same label? It makes sense to me that we collect all the cubes with the same label and return a `list[Cube]`.

Yes, absolutely. I had that doubt. I think Ports should be unique because of `def compose(self, other: BlockGraph, self_port: str, other_port: str)` but not ZXCube which should be possible to group by label. For example label with the same str all the ZXCube that are "connectors", as pipes, that just gives more volume. 

Another thing related... Could we label ports automatically while the DAE file is imported? What do you think?

inmzhang:
> Another thing related... Could we label ports automatically while the DAE file is imported? What do you think?

It should already be labeled by "Port0", "Port1", ... See code below:

https://github.com/tqec/tqec/blob/3b96e7d898a4afc8561a056a2decbfc795884de4/src/tqec/interop/collada/read_write.py#L214-L219

angelaelisa:
> It should already be labeled by "Port0", "Port1",

Sorry, I see, it's perfect:

```
ZXX(0,-1,1) 
ZXX(0,1,0) 
XZX(1,0,0) 
XZX(1,0,1) 
XZX(1,0,2) 
XZX(1,0,3) 
ZZX(0,0,1) 
ZZX(0,0,3) 
ZZX(0,0,0) 
ZXX(0,-1,2) 
ZXX(0,-1,3) 
ZXX(0,1,1) 
ZXX(0,1,2) 
ZZX(0,0,2) 
PORT(-1,0,0) Port0
PORT(1,0,4) Port1
PORT(0,-1,4) Port2
PORT(-1,0,3) Port3
PORT(0,1,3) Port4
PORT(-1,0,1) Port5
PORT(-1,0,2) Port6
```

src/tqec/computation/block_graph.py
angelaelisa: @inmzhang thanks for being so patient :) How do you see this version?

```
    def get_cubes_by_label(self, label: str) -> Cube | None:
        """
        Find a cube by its label in the BlockGraph.

        Args:
            label (str): The label of the cube.

        Returns:
            Cube | None: The Cube instance if found, otherwise None.
        """
        for cube in self.cubes:
            if cube.label == label:
                return cube

        return None  # No cube with the given label found

```

angelaelisa: I tested:

```
    result = block_graph.get_cubes_by_label("P3")

    if result:
        cube = result
        print(f"Cube found at {cube.position} with kind {cube.kind}")
    else:
        print("No cube found with that label")
```
Printed:

`Cube found at (-1,0,3) with kind PORT`

inmzhang: Hi Angela, what do you think about the following? We can simply return an empty list if the label is not present:

```py
def get_cubes_by_label(self, label: str) -> list[Cube]:
    """
    Find cubes with the specified label in the BlockGraph.

    Args:
        label: The label of the cubes.

    Returns:
        The cube instances that have the specified label.
    """
    return [cube for cube in self.cubes if cube.label == label]
```

As a side note, we do not require type annotation in docstring in tqec. As the signature has included type annotation. `Sphinx` will take care of that when generating documentation pages.

inmzhang: Great! Could you make the commit? Then we'll go ahead merging this :-)

angelaelisa: Many thanks Yiming! I'd love to do it to get used, but I have a deadline in a legal task. Could the commit wait until 16:00 CET? Sorry...

inmzhang: Take your time and no worries.

inmzhang (approved):

PR #522, opened 2025-03-17
Automate ports filling to minimize the number of simulation runs

### LINKED ISSUE #521: Automate ports filling to minimize the number of simulation runs

*A clear and concise description of what the problem is.*

Austin: https://github.com/tqec/tqec/issues/512#issuecomment-2727545303

To summarize: given a computation with open ports, we want to find an automatic way to fill in the ports and generate as few as `BlockGraph` instances that covers all possible observables. Therefore, we can minimize the number of simulation runs that generate statistics for all the observables, assuming #364 is solved.

As described in https://github.com/tqec/tqec/issues/512#issuecomment-2727989613, my initial proposal for the problem is:

1. Based on the correlation surface, compute the set of stabilizer generators.  
2. A set of stabilizers forms a clique if they are mutually compatible. Stabilizers A and B are compatible if and only if, for each Pauli term index `i`, `A[i] = I`, `B[i] = I`, or `A[i] = B[i]`. Stabilizers within the same clique can be simulated in a single run.  
3. This reduces to a minimum clique cover problem. While NP-hard, it can be approximated heuristically or solved exactly for small problem sizes.

However, the choice of stabilizer generators may impact the results. We can choose the generators with the minimum correlation surface area but there might be better options.

None

inmzhang:
Fixes #521.

This is still WIP. Tests and docs need to be added. Currently the [canonical stabilizer generators](https://github.com/quantumlib/Stim/blob/main/doc/python_api_reference_vDev.md#stim.Tableau.to_stabilizers) are used for observables, which might not have the minimum correlation surface area. However, the current state of this PR can already generate good ports filling results.

The following code:

```py
from tqec.computation.open_graph import fill_ports_for_simulation
from tqec.gallery import cnot, three_cnots, cz

for factory in [cnot, three_cnots, cz]:
    g = factory()
    filled_graphs = fill_ports_for_simulation(g)
    for i, fg in enumerate(filled_graphs):
        fg.graph.view_as_html(f"filled_{factory.__name__}_{i}.html")
        print(f"Simulation Run {i} for {factory.__name__} will support observables: {fg.stabilizers}")
    print("=====================================")
```
prints:

```txt
Simulation Run 0 for cnot will support observables: ['XIXX', 'IXIX']
Simulation Run 1 for cnot will support observables: ['IZZZ', 'ZIZI']
=====================================
Simulation Run 0 for three_cnots will support observables: ['XIIXXI', 'IXIIXX', 'IIXIIX']
Simulation Run 1 for three_cnots will support observables: ['IIZZZZ', 'IZIZZI', 'ZIIZII']
=====================================
Simulation Run 0 for cz will support observables: ['XIXZ', 'IZIZ']
Simulation Run 1 for cz will support observables: ['IXZX', 'ZIZI']
=====================================
```

which yields the correct and optimal ports filling.

EDITED:

I ordered all possible correlation surfaces (the entire stabilizer group) by area and added generators to a set from smallest to largest until the set generated the whole group. It seems to work but needs more testing and refactoring.


src/tqec/computation/correlation.py
```diff
@@ -169,3 +170,31 @@ def external_stabilizer(self, io_ports: list[int]) -> str:
...
+        port_positions = [graph.ports[p] for p in port_labels]
+        zx = graph.to_zx_graph()
+        zx_ports = [zx.p2v[p] for p in port_positions]
+        return self.external_stabilizer(zx_ports)
+
+    def area(self) -> int:
+        """Return the area of the correlation surface.
+
+        The area of the correlation surface is the number of nodes it spans.
+        An X node and a Z node with the same id are counted as two nodes.
```
nelimee, line 194:
```suggestion
        A X node and a Z node with the same id are counted as two nodes.
```

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,220 @@
...
+from tqec.utils.position import Direction3D
+
+
+@dataclass(frozen=True)
+class FilledGraph:
+    """A block graph without open ports that can be used for simulation.
+
+    Attributes:
+        graph: The block graph with all ports filled.
+        stabilizers: The external stabilizers simulatable on the filled graph.
```
nelimee, line 25:
```suggestion
        stabilizers: The external stabilizers simulable on the filled graph.
```
? Not sure.

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,220 @@
...
+from tqec.utils.exceptions import TQECException
+from tqec.utils.position import Direction3D
+
+
+@dataclass(frozen=True)
+class FilledGraph:
+    """A block graph without open ports that can be used for simulation.
+
+    Attributes:
+        graph: The block graph with all ports filled.
```
nelimee, line 24: That should be checked in the `__post_init__` method along with the other class invariants (e.g., the provided `stabilizers` are valid).

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Returns:
+        A list of :class:`~tqec.computation.open_graph.FilledGraph` instances, each
+        containing a block graph with all ports filled and a set of correlation
+        surfaces that can be used as logical observables for the simulation on that
+        block graph.
+    """
+    num_ports = graph.num_ports
+    if num_ports == 0:
+        raise TQECException("The provided graph has no open ports.")
```
nelimee, line 56: Should this raise? At first glance it seems like
```suggestion
    if num_ports == 0:
        return FilledGraph(graph, [], [])
```
would be semantically valid.

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,220 @@
...
+    }
+    identity = "I" * num_ports
+    # Need to collect all the possible correlation surfaces because we want
+    # to find the generators with the smallest correlation surface area
+    for stabilizer, comb in _iter_stabilizer_group(ext_stabs):
+        if stabilizer != identity and stabilizer not in stab_to_surface:
+            correlation_surface = reduce(
+                lambda a, b: a + b,
+                [stab_to_surface[s] for s in comb],
+            )
```
nelimee, line 70:
```suggestion
            correlation_surface = sum(
                [stab_to_surface[s] for s in comb], start=CorrelationSurface()
            )
```
should work fine also. It is a matter of preference. I find the above version more readable, I leave the final decision to you.

nelimee: Also, maybe overloading the `^` operator for `CorrelationSurface` make more sense (see other comment). If you think `^` is better, this suggested change becomes invalid.

src/tqec/computation/correlation.py
```diff
@@ -169,3 +170,31 @@ def external_stabilizer(self, io_ports: list[int]) -> str:
...
+        """
+        span_nodes = {node for edge in self.span for node in edge}
+        return len(span_nodes)
+
+    def __add__(self, other: CorrelationSurface) -> CorrelationSurface:
+        return CorrelationSurface(self.span.symmetric_difference(other.span))
```
nelimee, line 200: For me, it would make more sense to have the `xor` (`^`) operator doing that:
```suggestion
    def __xor__(self, other: CorrelationSurface) -> CorrelationSurface:
        return CorrelationSurface(self.span.symmetric_difference(other.span))
```

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,220 @@
...
+            stab_to_surface[stabilizer] = correlation_surface
+
+    # Sort the stabilizers by its corresponidng correlation surface's area to
```
nelimee, line 73:
```suggestion
    # Sort the stabilizers by its corresponding correlation surface's area to
```

nelimee (commented):
I need more time to understand the main algorithm. One other remark that did not fit anywhere else: why are you using `list[str]` to represent pauli strings (see `_multiply_unsigned_paulis` or `_iter_stabilizer_group`) instead of directly `stim.PauliString`?

inmzhang:
> I need more time to understand the main algorithm. One other remark that did not fit anywhere else: why are you using `list[str]` to represent pauli strings (see `_multiply_unsigned_paulis` or `_iter_stabilizer_group`) instead of directly `stim.PauliString`?

It's because 1. `stim.PauliString` is not hashable; 2. The `__str__` implementation of `stim.PauliString` includes the sign.

src/tqec/computation/open_graph.py
inmzhang: The `FilledGraph` includes the external stabilizer attribute that can only be computed before filling the ports. To be consistent, we should specify that even for zero-port input, but it requires some unnecessary complexity.

src/tqec/computation/correlation.py
```diff
@@ -169,3 +170,31 @@ def external_stabilizer(self, io_ports: list[int]) -> str:
...
+        Args:
+            g: The block graph to consider.
+
+        Returns:
+            The Pauli operator supported on the input/output ports of the graph.
+        """
+        port_labels = graph.ordered_ports
+        port_positions = [graph.ports[p] for p in port_labels]
+        zx = graph.to_zx_graph()
+        zx_ports = [zx.p2v[p] for p in port_positions]
```
nelimee, line 187: Note that here `p2v` is re-computed at each call. Maybe cache it in a variable to avoid that? 
```suggestion
        p2v = zx.p2v
        zx_ports = [p2v[p] for p in port_positions]
```

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,228 @@
...
+        if self.graph.num_ports != 0:
+            raise TQECException("The filled graph should not have open ports.")
+        if len(self.stabilizers) != len(self.observables):
+            raise TQECException(
+                "The number of stabilizers and observables should match."
+            )
+
+    def get_external_stabilizers(self) -> list[str]:
+        """Return the external stabilizers of the correlation surfaces."""
+        return [s.external_stabilizer_on_graph(self.graph) for s in self.observables]
```
nelimee, line 45: Minor:
```suggestion
        return [obs.external_stabilizer_on_graph(self.graph) for obs in self.observables]
```

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,228 @@
...
+    #     [
+    #         stim.PauliString("ZZ_"),
+    #         stim.PauliString("_ZZ"),
+    #     ],
+    #     allow_redundant=False,
+    #     allow_underconstrained=True,
+    # ) will map `Z__` to `ZZ_`, `_Z_` to `_ZZ`
+    # If a Pauli string can be generated by two stabilizers, then the inverse
+    # application of the tableau will map the Pauli string to a Pauli string
+    # only has `Z` operators on the first two qubits.
```
nelimee, line 187:
```suggestion
    # that only has `Z` operators on the first two qubits.
```
?

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,228 @@
...
+    # If there are `X`/`Y` appearing, then the Pauli string is anti-commuting
+    # with the stabilizers.
+    # If there are non-trivial operators on the last qubit, then the Pauli string
+    # is not in the stabilizer group.
```
nelimee, line 191: I do not really understand the explanation here. A few points that I cannot grasp:

- why do we only consider `Z` as input?
- why only the first two qubits in the example?

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,228 @@
...
+            pipe_kind = pipe_at_port.kind
+            at_head = pipe_at_port.u == graph[port_pos]
+            fill_kind = ZXCube(
+                *[
+                    pipe_kind.get_basis_along(dir, at_head) or Basis(basis)
+                    for dir in Direction3D.all_directions()
+                ]
+            )
+            fg.fill_ports({port: fill_kind})
+        assert fg.num_ports == 0
```
nelimee, line 153: Not strictly needed anymore, but might still be interesting to keep in order to remind the expected invariant.

nelimee (commented):
Added a few comments, I still need one or two more passes to understand the algorithm fully.

src/tqec/computation/open_graph.py
inmzhang: Yes, I'd like to keep this.

src/tqec/computation/open_graph.py
inmzhang: > why do we only consider Z as input?

Because it's how [`stim.Tableau.from_stabilizers`](https://github.com/quantumlib/Stim/blob/cae4e9fcde59f9f62650c5cfa7e28c94ad444795/doc/python_api_reference_vDev.md#stim.Tableau.from_stabilizers) is implemented. Quoting from the stim doc: **Guarantees that result.z_output(k) will be equal to the k'th independent stabilizer from the `stabilizers` argument.**

> why only the first two qubits in the example?

We only provides two independent stabilizers in this case.

inmzhang: To expand more on this:

- `stim.Tableau.from_stabilizers` constructs a Tableau, which represents a Clifford circuit mapping the `Z` Pauli on the nth input to the nth stabilizer specified in the argument at the output.  
- This is equivalent to the inverse circuit or Tableau mapping the nth stabilizer to `Z` on the nth output.  
- If a Pauli operator is a product of the stabilizers, it can be decomposed into them. As a result, after applying the Tableau, the Pauli operator will have `Z` operators only on the corresponding outputs.

nelimee: The above explanations are enough for me, I understood. You could maybe include them in the function documentation?

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,230 @@
...
+    identity = "I" * num_ports
+    # Need to collect all the possible correlation surfaces because we want
+    # to find the generators with the smallest correlation surface area
+    for stabilizer, comb in _iter_stabilizer_group(ext_stabs):
+        if stabilizer != identity and stabilizer not in stab_to_surface:
+            correlation_surface = reduce(
+                lambda a, b: a ^ b,
+                [stab_to_surface[s] for s in comb],
+            )
+            stab_to_surface[stabilizer] = correlation_surface
```
nelimee, line 81: Here, is it guaranteed that if `stabilizer in stab_to_surface` then the correlation_surface associated to the stabilizer is the same? Basically, I am wondering if the following code

```suggestion
        if stabilizer != identity:
            correlation_surface = reduce(
                lambda a, b: a ^ b,
                [stab_to_surface[s] for s in comb],
            )
            stab_to_surface[stabilizer] = (
                correlation_surface
                if stabilizer not in stab_to_surface else
                min(
                    (correlation_surface, stab_to_surface[stabilizer]), 
                    key=lambda cs: cs.area()
                )
            )
```

should be rather considered or if we know for sure that the computed `correlation_surface` will always be equal to `stab_to_surface[stabilizer]`.

src/tqec/computation/open_graph.py
```diff
@@ -0,0 +1,230 @@
...
+        cliques.setdefault(color, []).append(generators[node])
+
+    def ports_basis_for_clique(
+        supported_stabilizers: list[str],
+    ) -> list[str]:
+        port_ops = ["I"] * num_ports
+        for s in supported_stabilizers:
+            for i, p in enumerate(s):
+                if p != "I":
+                    port_ops[i] = p
```
nelimee, line 127: Just to check that I understood the algorithm, the following should be valid right?

```suggestion
            for i, p in enumerate(s):
                if p != "I":
                    assert port_ops[i] in ["I", p]
                    port_ops[i] = p
```

src/tqec/computation/open_graph.py
inmzhang: There are two cases to consider:

1. If the initial generating set of the correlation surface from `find_correlation_surface` is exact-complete, the unique decomposition of stabilizers into these generators ensures that no stabilizer will be repeated during iteration through the group.
2. If the generating set is over-complete and a stabilizer has two different correlation surfaces => the identity stabilizer could have a non-empty correlation surface. That is possible but leads to an unusual structure, such as loops in the diagram.

Since we do not guarantee that the generators are exact-complete, you are right—we need to include this check. However, I would prefer to first reduce the generating set to an exact-complete form and then iterate over the group generated by them. This approach not only ensures correctness but also reduces the total number of iterations.

src/tqec/computation/open_graph.py
inmzhang: That's correct.

inmzhang:
> The only question I have left is whether or not the algorithm you implement returns the expected output. Do you have any low-effort (for you, do not write down an entire explanation just for that here) way of explaining me the algorithm?

The key idea is that stabilizers within the same clique should have compatible Paulis at their ports. This allows a single simulation run to simultaneously simulate for all stabilizers in a clique. Minimizing the number of simulations, therefore, translates to minimizing the number of cliques while ensuring full coverage of all stabilizers. This leads to solving the minimum clique cover problem on a graph, which in turn reduces to a graph coloring problem on its complement.

Here, I use the term "stabilizer" interchangeably with the stabilizer flow supported by the computation. However, due to space-time symmetry, a flow is not well-defined for a block graph. By leveraging state-channel duality (i.e., bending the wires), we can interpret the entire computation as a stabilizer state.

src/tqec/computation/open_graph.py
inmzhang: > I would prefer to first reduce the generating set to an exact-complete form and then iterate over the group generated by them

This has been addressed in the last commit.

nelimee:
> The key idea is that stabilizers within the same clique should have compatible Paulis at their ports. This allows a single simulation run to simultaneously simulate for all stabilizers in a clique. Minimizing the number of simulations, therefore, translates to minimizing the number of cliques while ensuring full coverage of all stabilizers. This leads to solving the minimum clique cover problem on a graph, which in turn reduces to a graph coloring problem on its complement.

This part is OK for me, I am used to manipulate graph algorithms and I already used such a reformulation, so this is fine. Thinking more about it, I did not get the idea behind taking a minimal-area generating set, but now I think I start to understand. This is basically a heuristic to try to get the smallest-possible areas in the end, to reduce error rate?

nelimee (approved):
I am a little worried about the performance of `_iter_stabilizer_group` that generates `2^[number of correlation surfaces]`, which can quickly explode (the `steane` example has 22 correlations surfaces if I remember correctly, that ~4 millions items).

Maybe adding a warning in `_iter_stabilizer_group` if the length of the input is larger than a certain number would be a good idea?

Else, the implementation seems OK.

inmzhang:
>  This is basically a heuristic to try to get the smallest-possible areas in the end, to reduce error rate?

That's correct.

> Maybe adding a warning in _iter_stabilizer_group if the length of the input is larger than a certain number would be a good idea?

Actually, I think obtaining the generators with the smallest surface area is not crucial for simulation. By default, we can optimize for small-area initial generators and only search for the global minimum across the entire group when a specific flag is set.

nelimee:
> Actually, I think obtaining the generators with the smallest surface area is not crucial for simulation. By default, we can optimize for small-area initial generators and only search for the global minimum across the entire group when a specific flag is set.

That seems to be a reasonable behaviour.

inmzhang:
Please check out the last commit, if you are ok with this, I'll merge the PR.

nelimee (approved):
LGTM

KabirDubey:
This PR seems fine, but just want to make sure I understand your algorithm.

> The key idea is that stabilizers within the same clique should have compatible Paulis at their ports.

I would like to confirm the notion of compatible Paulis with an example.

> Here, I use the term "stabilizer" interchangeably with the stabilizer flow supported by the computation. However, due to space-time symmetry, a flow is not well-defined for a block graph. By leveraging state-channel duality (i.e., bending the wires), we can interpret the entire computation as a stabilizer state.

Given a computation and one of its stabilizer flows, let's interpret it as a stabilizer state. From there, how can we study the Paulis at its ports? I know what it means for a block graph to have a [Port](https://tqec.github.io/tqec/_autosummary/tqec.computation.Port.html), but I'm not sure what it means for a stabilizer to have a port. 

Thank you

inmzhang:
> I would like to confirm the notion of compatible Paulis with an example.

Stabilizers A and B are compatible if and only if, for each Pauli term index `i`, `A[i] = I`, `B[i] = I`, or `A[i] = B[i]`. For example `IZ` is compatible with `YZ` but incompatible with `YX`.

> Given a computation and one of its stabilizer flows, let us interpret it as a stabilizer. From there, how can we study the Paulis at its ports?

A stabilizer is exactly the Paulis at ports.

PR #526, opened 2025-03-19
Single simulation run for multiple observable statistics

### LINKED ISSUE #364: Improve simulation for multiple observables

inmzhang:
Fixes #364.

The solution is mainly based on [my previous comments](https://github.com/tqec/tqec/issues/364#issuecomment-2574720834).

For the `custom_error_key` used in `sinter`, I use the following heuristic algorithm, copied from the docstring of `heuristic_custom_error_key`.

```txt
    Currently, we determine the most likely observable error combinations by
    analyzing their correlation surfaces. For each combination, we perform an
    AND on the corresponding correlation surfaces and calculate the area of the
    resulting span. We then rank these combinations according to the computed
    AND-area metric. To derive a reasonable error metric, we select
    ``ordered_combs[-len(observables)]`` for two reasons: first, errors should
    be easy to sample so as not to require excessive shots; and second, once an
    error reaches a preset MAX value, there should be enough statistical data for
    all individual observables.
```

The heuristic might need improvement in the future, or not.

For the logical CNOT, I have tested for the Z basis observables and get results:

css_logical_cnot_result_Z_observable_0
css_logical_cnot_result_Z_observable_1
css_logical_cnot_result_Z_observable_2

### DIFF
--- a/src/tqec/simulation/simulation.py
+++ b/src/tqec/simulation/simulation.py
@@ -1,4 +1,5 @@
 import multiprocessing
+from pathlib import Path
 from typing import Callable, Iterable, Sequence
 
 import sinter
@@ -34,6 +35,9 @@ def start_simulation_using_sinter(
     decoders: Iterable[str] = ("pymatching",),
     print_progress: bool = False,
     custom_decoders: dict[str, sinter.Decoder | sinter.Sampler] | None = None,
+    save_resume_filepath: str | Path | None = None,
+    existing_data_filepaths: Iterable[str | Path] = (),
+    split_observable_stats: bool = True,
 ) -> list[list[sinter.TaskStats]]:
     """Helper to run `stim` simulations using `sinter`.
 
@@ -95,13 +99,33 @@ def start_simulation_using_sinter(
             used if requested by name by a task or by the decoders list.
             If not specified, only decoders with support built into sinter, such
             as 'pymatching' and 'fusion_blossom', can be used.
-
+        save_resume_filepath: Defaults to None (unused). If set to a filepath,
+            results will be saved to that file while they are collected. If the
+            python interpreter is stopped or killed, calling this method again
+            with the same save_resume_filepath will load the previous results
+            from the file so it can resume where it left off. The stats in this
+            file will be counted in addition to each task's previous_stats field
+            (as opposed to overriding the field). Notes that the stats for each
+            individual observable are not split but saved in the `custom_counts`
+            field of the stats.
+        existing_data_filepaths: CSV data saved to these files will be loaded,
+            included in the returned results, and count towards things like
+            max_shots and max_errors.
+        split_observable_stats: Defaults to True. If True, the results are
+            post-processed to get individual statistics for each observable in
+            `observables`. If False, the results are returned as they are
+            collected.
     Returns:
         one simulation result (of type `list[sinter.TaskStats]`) per provided
         observable in `observables`.
     """
     if observables is None:
         observables = block_graph.find_correlation_surfaces()
+    custom_error_count_key: str | None = None
+    if not split_observable_stats or len(observables) <= 1:
+        custom_error_count_key = None
+    else:
+        custom_error_count_key = heuristic_custom_error_key(observables)
 
     compiled_graph = compile_block_graph(
         block_graph,
@@ -119,6 +143,8 @@ def start_simulation_using_sinter(
             manhattan_radius,
             detector_database,
         ),
+        existing_data_filepaths=existing_data_filepaths,
+        save_resume_filepath=save_resume_filepath,
         progress_callback=progress_callback,
         max_shots=max_shots,
         max_errors=max_errors,
@@ -127,6 +153,8 @@ def start_simulation_using_sinter(
         custom_decoders=custom_decoders,
         hint_num_tasks=len(ks) * len(ps),
         count_observable_error_combos=True,
-        custom_error_count_key=heuristic_custom_error_key(observables),
+        custom_error_count_key=custom_error_count_key,
     )
-    return split_stats_for_observables(stats, len(observables))
+    if split_stats_for_observables:
+        return split_stats_for_observables(stats, len(observables))
+    return [stats]
--- a/src/tqec/simulation/simulation.py
+++ b/src/tqec/simulation/simulation.py
@@ -125,9 +125,7 @@ def start_simulation_using_sinter(
     if observables is None:
         observables = block_graph.find_correlation_surfaces()
     custom_error_count_key: str | None = None
-    if not split_observable_stats or len(observables) <= 1:
-        custom_error_count_key = None
-    else:
+    if split_observable_stats and len(observables) > 1:
         custom_error_count_key = heuristic_custom_error_key(observables)
 
     compiled_graph = compile_block_graph(
--- a/src/tqec/simulation/simulation.py
+++ b/src/tqec/simulation/simulation.py
@@ -1,5 +1,5 @@
 import multiprocessing
-from typing import Callable, Iterable, Iterator, Sequence
+from typing import Callable, Iterable, Sequence
 
 import sinter
 
@@ -10,6 +10,10 @@ from tqec.compile.specs.library.css import CSS_BLOCK_BUILDER, CSS_SUBSTITUTION_B
 from tqec.computation.block_graph import BlockGraph
 from tqec.computation.correlation import CorrelationSurface
 from tqec.simulation.generation import generate_sinter_tasks
+from tqec.simulation.split import (
+    heuristic_custom_error_key,
+    split_stats_for_observables,
+)
 from tqec.utils.noise_model import NoiseModel
 
 
@@ -30,7 +34,7 @@ def start_simulation_using_sinter(
     decoders: Iterable[str] = ("pymatching",),
     print_progress: bool = False,
     custom_decoders: dict[str, sinter.Decoder | sinter.Sampler] | None = None,
-) -> Iterator[list[sinter.TaskStats]]:
+) -> list[list[sinter.TaskStats]]:
     """Helper to run `stim` simulations using `sinter`.
 
     This function is the preferred entry-point to run `sinter` computations using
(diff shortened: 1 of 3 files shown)


src/tqec/simulation/split.py
```diff
@@ -0,0 +1,115 @@
...
+    all individual observables.
+
+    Args:
+        observables: The list of observables.
+
+    Returns:
+        The heuristic custom error key.
+    """
+    key_area: dict[str, int] = {}
+    span_union: set[ZXEdge] = reduce(
```
KabirDubey, line 103: Just so I understand, the span of observables is a set of ZXEdge objects. So what's going on here is that you've generated a single set containing all the edges that appear in any of the observables spans. Then, you iterate through each edge to determine the most frequent edge patterns in each observable.

KabirDubey (commented):
LGTM. There seems to be a risk of a bias for observables with larger spans.

src/tqec/simulation/split.py
inmzhang: > Then, you iterate through each edge to determine the most frequent edge patterns in each observable.

For each edge, we identify the observables that include it and obtain the observable mistake mask for those observables, such as `E_E_`. Then, we increment the count as a proxy for the frequency of that observable combination mistake.

src/tqec/simulation/simulation.py
```diff
@@ -91,42 +99,65 @@ def start_simulation_using_sinter(
...
+        If `split_observable_stats` is False, the outer list has only one element,
+        containing the raw statistics collected.
     """
     if observables is None:
         observables = block_graph.find_correlation_surfaces()
+    custom_error_count_key: str | None = None
+    if not split_observable_stats or len(observables) <= 1:
+        custom_error_count_key = None
+    else:
+        custom_error_count_key = heuristic_custom_error_key(observables)
```
nelimee, line 131:
```suggestion
    custom_error_count_key: str | None = None
    if split_observable_stats and len(observables) > 1:
        custom_error_count_key = heuristic_custom_error_key(observables)
```

src/tqec/simulation/split.py
```diff
@@ -0,0 +1,115 @@
...
+                            for k, v in task_stats.custom_counts.items()
+                            if not k.startswith("obs_mistake_mask=")
+                        }
+                    ),
+                )
+            )
+    return stats_by_observables
+
+
+def heuristic_custom_error_key(observables: list[CorrelationSurface]) -> str:
```
nelimee, line 82: I needed quite some time to understand that the goal of this function was to try (heuristically) to find the key that should be used to compare the number of errors to the maximum number of errors allowed for a simulation. At first, I thought that this function was defining the format "obs_mistake_mask=_EE__" (obviously without understanding it). 

I think the docstring would benefit from remembering a little bit that `sinter` works with both

- a maximum number of shots,
- a maximum number of errors,

and that there are several types of errors we could choose from.

nelimee (commented):
I've had a hard time understanding the purpose of the heuristic (see comment), but that looks pretty nice. I was expecting something way more complex to make that feature work. I still need some time to understand the heuristic.

src/tqec/simulation/split.py
inmzhang: When I first worked on this issue, I thought we could simply select the single observable with the smallest area as the count key (formatted like `obs_mistake_mask=_E_`), assuming it would have the lowest error rate. The idea was that if we sampled enough errors for that observable, the others would also be sufficiently covered.

However, I later realized that correlation surfaces overlap. For example, `ZI -> ZI` and `ZZ -> IZ` share the entire vertical pipes at the control qubit. This means the error rate for `E_` is significantly lower than that for `EE`—roughly an order of magnitude lower. If we were to use `E_` as the count key, it would lead to an unnecessarily large number of shots.

The heuristic aims to find an error combination that is reasonably "easy" to sample. I use the AND correlation surface area of an observable combination as a proxy for the probability of that error combination. If we simply use the most probable error combination as the key, it might be too easy to sample, resulting in insufficient statistics for each individual observable. To balance this, I use the `len(observables)`th largest combination as the key. 

If the correlation surface area is too imbalanced across observables, this approach might be problematic. However, in general, I believe the metric should work well. Additionally, I added the `save_resume_filepath` argument so we can increase the number of shots if needed.

nelimee: The more you can document that in the code, the better in my opinion.

nelimee (commented):
For me this looks good. I am impressed by the low quantity of code that was needed to solve that issue, I was expecting the need for more changes, and even maybe the need to change `sinter`. Well done!

Even though that is not blocking, I would strongly advise adding all the details you can to the heuristic used. Your comment in review is a good start, and the more information you can add the better. 

One note that comes to my mind just now, and I will accept the PR as soon as this is solved: you mentioned that the heuristic may fail for imbalanced observables. Would it be possible/desirable to check if that is the case and output a warning?

inmzhang:
> Even though that is not blocking, I would strongly advise adding all the details you can to the heuristic used. Your comment in review is a good start, and the more information you can add the better.

I'll add more docs at tomorrow.

> Would it be possible/desirable to check if that is the case and output a warning?

I need to think more about it but I believe we can.

inmzhang:
It turns out we can use a simpler and more deterministic key, which should always work well as long as we aim to sample good statistics for all observables. The updated docstring explains this choice in detail—please take a look. If you have any questions, feel free to ask here, and I can expand the documentation further.

src/tqec/simulation/split.py
```diff
@@ -80,36 +79,60 @@ def split_stats_for_observables(
...
-        key = "obs_mistake_mask=" + "".join(key_list)
-        key_area[key] = key_area.get(key, 0) + 1
-    most_frequent_keys = sorted(key_area.items(), key=lambda x: x[1])
-    return most_frequent_keys[-len(observables)][0]
+            obs_comb.append("E" if fragment in obs.span else "_")
+        key = "obs_mistake_mask=" + "".join(obs_comb)
+        obs_comb_to_prob[key] = obs_comb_to_prob.get(key, 0) + 1
+    # order the observable combinations by probability
+    ordered_combs = sorted(obs_comb_to_prob.keys(), key=lambda x: obs_comb_to_prob[x])
+    return ordered_combs[-1]
```
nelimee, line 138: There is not need to sort here, a `max` should be enough right?

```py
return max(obs_comb_to_prob.keys(), key=lambda x: obs_comb_to_prob[x])
```

nelimee (commented):
The new heuristic looks good. Have you tried it in practice to check that the results are sensibly similar to the previous ones?

inmzhang:
> The new heuristic looks good. Have you tried it in practice to check that the results are sensibly similar to the previous ones?

I've only run for logical CNOT and it worked fine. If you think it's necessary, I can test it against the three_cnots and steane examples at next Monday.

nelimee:
> I've only run for logical CNOT and it worked fine. If you think it's necessary, I can test it against the three_cnots and steane examples at next Monday.

It would help being sure that neither of us missed something. As this will be used for all simulations once merged, we want to avoid any mistake as that would impact all the results.

inmzhang:
@nelimee I've run simulation for `three cnots` and `steane volume 12 encoding` and the results look good to me. Please check out the simulation results at https://drive.google.com/drive/folders/1QIdLV3TdQT6ZygahbDc-2skPV76EhHXL.

The simulation was run on branch https://github.com/tqec/tqec/tree/feat/change-plaquette-invariant and I've locally included the changes of this PR and #522. Then you should be able to replicate the results by running [this script](https://drive.google.com/file/d/15AKRab3ngPHM9CL-t7DrAHljoJnXThar/view?usp=drive_link).

afowler:
Wonderful, the graphs look spot on. Would love to run this too, but I think
for me it would be better to wait until all this code is in a single branch.

Best,
Austin.

nelimee (approved):
Nice! If it works for the steane and 3-cnots example, I think this is ready to be merged!

afowler:
Is this code likely to be in a single branch soon? Would very much like to
demonstrate this on Wednesday. A single branch capable of doing everything
you have demonstrated. If that isn't likely to be possible, it would be
great to elaborate on exactly what needs to be done to "locally included
the changes of this PR and #522 <https://github.com/tqec/tqec/pull/522>" in
a safe manner. Are you able to give a sequence of git commands that
achieves this?

Best,
Austin.

inmzhang:
> Is this code likely to be in a single branch soon?

Sure, I'll update branch `feat/change-plaquette-invariant` to include the changes in a few minutes.

inmzhang:
@afowler Please pull the latest changes on the `feat/change-plaquette-invariant` branch. I've also added an example script, `main.py`, to help you get started. Let me know if you run into any issues.

afowler:
Excellent, I got a warning about k=1, but it is currently running on my
computer.

inmzhang:
> I got a warning about k=1

You can ignore the warning for now. 

@nelimee would it be possible to eliminate the warning for `k=1` spatial cube template?

afowler:
I'm done for tonight, I got stuck trying to install pymatching, for some
reason when it tries to compile it is unable to create temporary files on
my system and fails...

nelimee:
> @nelimee would it be possible to eliminate the warning for `k=1` spatial cube template?

That makes a lot of sense, the warning is not even up-to-date. I'll do that.

inmzhang:
> I got stuck trying to install pymatching, for some reason when it tries to compile it is unable to create temporary files on my system and fails...

That's strange — pymatching should have pre-built wheels for Windows with Python 3.7 to 3.12, so no compilation should be needed on your end. Are you using Python 3.13?

nelimee:
> > @nelimee would it be possible to eliminate the warning for `k=1` spatial cube template?
> 
> That makes a lot of sense, the warning is not even up-to-date. I'll do that.

See #540

afowler:
I was indeed using Python 3.13, uninstalling this and installing 3.12 fixed
the issue. Gathering stats now :-) Thanks!

PR #530, opened 2025-03-21
Implement a few post-processing functions for `stim.Circuit`

### LINKED ISSUE #462: Allow `Plaquette` instances to overlap in time
Spatial junctions require `Plaquette` instances that will be implemented in #461 and that can overlap in time:

Image

In the image above, the measurement can be scheduled at the same time as the next round resets.
This is not doable with the current code, and it would be nice to be able to do so.

Note that this is not required for spatial junctions to work, but is a nice optimisation that would allow less time-steps in QEC circuits (and so less noise due to idling errors).

There are two main ways I can see for fixing this issue:

1. Allow `Plaquette` instances to overlap in time, literally. That would likely lead to large structural changes because overlaps in the time dimension were never included in the initial design.
2. Do not allow overlapping in time (i.e., keep the code as it is right now) but provide a post-processing utility that will merge the measurement layer and the reset layer.

I really think 2. is the best option at the moment. It is unintrusive, does not require large changes to the code base, fits best with the task at hand: `tqec` is generating a valid quantum circuit, and if you want some specific optimisations to be performed on the output then you can do so using the utilities provided.

nelimee:
This PR implements two post-processing functions for `stim.Circuit`:

- `remove_empty_moments`
- `merge_adjacent_moments`

The implementation of `merge_adjacent_moments` should fix #462.

### DIFF
--- a/src/tqec/circuit/moment.py
+++ b/src/tqec/circuit/moment.py
@@ -10,6 +10,7 @@ instead of using ``cirq`` data-structures.
 
 from __future__ import annotations
 
+from copy import deepcopy
 from typing import Any, Callable, Iterable, Iterator, cast
 
 import stim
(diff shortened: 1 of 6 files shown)


inmzhang (approved):
LGTM.

Issue #527, opened 2025-03-19
[Question]: Is there a method to change labels?


angelaelisa:
**Question**: Is there a method to change labels? 

I looked for it, but I don't find it.

I need to relabel to use `tqec.computation.BlockGraph.compose` whit two imported .dae files. The issue is that ports are automatically labeled, so if you import two .dae files to merge them, you must change the labels of one of the BlockGraphs. If you don't relabel it raises an exception because there are ports with the same label, and they cannot merge into a new one. 

```
    block_graph = BlockGraph.from_dae_file("slide67_open.dae")
    block_graph2 = BlockGraph.from_dae_file("cnot_4_test.dae")
    block_graph.view_as_html("composeBG1.html")
    block_graph2.view_as_html("composeBG2.html")
    
    print(f"Cubes BG1:")
    for cube in block_graph.cubes:
        print(cube, cube.label)
    print(f"Cubes BG2:")

    block_graph3 = block_graph.compose(block_graph2, "Port4", "Port2_0")
    block_graph3.view_as_html("composeBGfinal.html")

```
If you don't relabel, it fails. If relabel, works perfect. @inmzhang if it doesn't already exist, I would like to code it.

Related: **could we create a method to cap open ports by label?** I'd also like to code it. Maybe now it look unnecessary, but if we have later a frontend for users, could be nice they don't have to create new models in SketchUp to cap ports.

nelimee:
I do not think there is any method to rename labels. A function to do that will likely look like the existing 

https://github.com/tqec/tqec/blob/c0a5a34f554446bd202d45af92d0929294fcb482/src/tqec/computation/block_graph.py#L495-L524

Same remark for the capping by label, we do not have that, but it can be done using the above method quite easily already.

angelaelisa:
Thanks Adrien, you're right I forgot about fill_ports

nelimee:
> [@nelimee](https://github.com/nelimee) I've been testing and I think there is a bug: `fill_ports` allows me to use any string as kind of cube:

The typing information explicitly say that the provided `fill` argument should be either `CubeKind` or `Mapping[str, CubeKind]`. It is expected that users respect the typing information. I am on the side of not over-checking the inputs, I think I remember that Python's philosophy goes this way too. 

Nevertheless, I agree that's quite confusing, I would expect things to break, maybe an additional check would be nice? 

And yes, for renaming ports, a new method will have to be created, but can likely be inspired by `fill_ports`.

[...]

PR #531, opened 2025-03-21
Introduce BlockGraph.relabel_cubes

angelaelisa:
See issue #527 (there will be a second PR for `fill_ports` but need to have `relabel_cubes` solved first)

Relabel cubes searching the cubes by `position` or by `label` (allows to relabel a group of cubes). It raises an exception if the name given is been used in an existing port. I've tested it with:

```
from tqec import BlockGraph, Position3D

def main():    
    block_graph = BlockGraph.from_dae_file("slide67_open.dae")
    
    # let's see the list of cubes in the blockgraph
    print(f"Cubes:")
    for cube in block_graph.cubes:
        print(cube, cube.label)
    
    block_graph.relabel_cubes(Position3D(-1,0,0), "PortNew1")

    print(f"New Cubes:")
    for cube in block_graph.cubes:
        print(cube, cube.label)
    block_graph.relabel_cubes("Port1", "PortNew2")

    print(f"Cubes:")
    for cube in block_graph.cubes:
        print(cube, cube.label)

    block_graph.view_as_html("open_ports.html")

if __name__ == "__main__":
    main()
```

### DIFF
--- a/src/tqec/computation/block_graph.py
+++ b/src/tqec/computation/block_graph.py
@@ -894,6 +894,65 @@ class BlockGraph:
             obj_dict = json.loads(json_text)
         return BlockGraph.from_dict(obj_dict)
 
+    def relabel_cubes(self, label_mapping: Mapping[Position3D | str, str]) -> None:
+        """Relabel cubes in the block graph.
+
+        This method updates the labels of cubes in the graph, based on a mapping
+        from either a cube position or its existing label to a new label.
+
+        Args:
+            label_mapping: A mapping from either Position3D or current cube label (str)
+                to the new label to assign.
+
+        Raises:
+            TQECException: If a cube is not found for the given key, if a port label
+                is reused, or if the new label conflicts with existing port labels.
+        """
+        port_labels = {cube.label for cube in self.cubes if cube.is_port}
+        assigned_new_labels: set[str] = set()
+
+        for key, new_label in label_mapping.items():
+            if not new_label:
+                raise TQECException("New label must be non-empty.")
+
+            if isinstance(key, Position3D):
+                matching_cubes = [cube for cube in self.cubes if cube.position == key]
+            elif isinstance(key, str):
+                matching_cubes = [cube for cube in self.cubes if cube.label == key]
+            else:
+                raise TQECException(
+                    f"Invalid identifier '{key}'. Must be Position3D or str."
+                )
+
+            if not matching_cubes:
+                raise TQECException(f"No cube found for identifier '{key}'.")
+
+            is_port_relabel = any(cube.is_port for cube in matching_cubes)
+
+            if is_port_relabel:
+                if new_label in port_labels:
+                    raise TQECException(
+                        f"Port label '{new_label}' is already assigned to another port."
+                    )
+                if new_label in assigned_new_labels:
+                    raise TQECException(
+                        f"Port label '{new_label}' is reused multiple times."
+                    )
+                assigned_new_labels.add(new_label)
+            else:
+                if new_label in port_labels:
+                    raise TQECException(
+                        f"The label '{new_label}' belongs to a port and cannot be reused by a non-port cube."
+                    )
+
+            for cube in matching_cubes:
+                updated_cube = Cube(
+                    position=cube.position, kind=cube.kind, label=new_label
+                )
+                self._graph.add_node(
+                    cube.position, **{self._NODE_DATA_KEY: updated_cube}
+                )
+
 
 def block_kind_from_str(string: str) -> BlockKind:
     """Parse a block kind from a string."""
--- a/src/tqec/computation/block_graph_test.py
+++ b/src/tqec/computation/block_graph_test.py
@@ -296,7 +296,7 @@ def test_block_graph_relabel_cubes() -> None:
     n3 = g.add_cube(Position3D(2, 0, 0), "P", "Out")
     g.add_pipe(n2, n3, "OXZH")
 
-    label_mapping = {
+    label_mapping: dict[Position3D | str, str] = {
         Position3D(0, 0, 0): "InputPortByPos",
         "Out": "OutputPortByLabel",
     }
--- a/src/tqec/computation/block_graph_test.py
+++ b/src/tqec/computation/block_graph_test.py
@@ -9,7 +9,6 @@ from tqec.gallery import cz, memory, cnot
 from tqec.utils.enums import Basis
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Direction3D, Position3D
-from tqec.interop.collada import read_block_graph_from_dae_file
 
 
 def test_block_graph_construction() -> None:
@@ -289,32 +288,43 @@ def test_block_graph_to_from_json() -> None:
     os.remove(temp_file.name)
 
 
-def test_relabel_cubes():
-    block_graph = read_block_graph_from_dae_file("slide67_open.dae")
+def test_block_graph_relabel_cubes_success() -> None:
+    g = BlockGraph()
+    g.add_cube(Position3D(0, 0, 0), "ZXZ", label="A")
+    g.add_cube(Position3D(1, 0, 0), "P", label="In")
+    g.add_pipe(Position3D(0, 0, 0), Position3D(1, 0, 0), "OXZ")
 
-    print("\nCubes before relabeling:")
-    for cube in block_graph.cubes:
-        print(f"  {cube} — label: '{cube.label}'")
+    assert g["In"].is_port
+    assert g["A"].label == "A"
 
-    # Update the label of a cube by old label or position
     label_mapping = {
-        "Port3": "Port1110",
-        Position3D(-1, 0, 0): "Port_relabeled_by_position",
-        "Port4": "Port40",
+        "In": "Input",
+        "A": "Alpha",
+        Position3D(1, 0, 0): "InputPortByPos",
     }
 
-    try:
-        block_graph.relabel_cubes(label_mapping)
-        print("\n✅ Relabeling completed successfully.")
-    except TQECException as e:
-        print(f"\n❌ Relabeling failed: {e}")
-        return
+    g.relabel_cubes(label_mapping)
 
-    print("\nCubes after relabeling:")
-    for cube in block_graph.cubes:
-        print(f"  {cube} — label: '{cube.label}'")
+    new_labels = {cube.label for cube in g.cubes}
+    assert "Alpha" in new_labels
+    assert "InputPortByPos" in new_labels or "Input" in new_labels
+    assert "In" not in new_labels
+    assert "A" not in new_labels
 
-    block_graph.view_as_html("relabeled_ports.html")
-    print(
-        "\n Visualization saved to 'relabeled_ports.html'. You can open it in your browser."
-    )
+
+def test_block_graph_relabel_cubes_conflict() -> None:
+    g = BlockGraph()
+    g.add_cube(Position3D(0, 0, 0), "P", label="P0")
+    g.add_cube(Position3D(1, 0, 0), "P", label="P1")
+    g.add_pipe(Position3D(0, 0, 0), Position3D(1, 0, 0), "OZX")
+
+    # Attempt to rename two cubes to the same label
+    conflict_mapping = {
+        "P0": "P1",
+        "P1": "P1",  # Already exists
+    }
+
+    with pytest.raises(
+        TQECException, match="already assigned to another port|reused multiple times"
+    ):
+        g.relabel_cubes(conflict_mapping)
--- a/src/tqec/computation/block_graph_test.py
+++ b/src/tqec/computation/block_graph_test.py
@@ -9,6 +9,7 @@ from tqec.gallery import cz, memory, cnot
 from tqec.utils.enums import Basis
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Direction3D, Position3D
+from tqec.interop.collada import read_block_graph_from_dae_file
 
 
 def test_block_graph_construction() -> None:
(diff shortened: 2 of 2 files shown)


src/tqec/computation/block_graph.py
```diff
@@ -414,6 +414,57 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
...
+            block_graph (BlockGraph): The block graph in which to search for cubes.
+            identifier (Position3D | str): A position (to rename one cube) or a label (to rename all cubes
+            found with that label).
+            new_label (str): The new label to assign.
+
+        Raises:
+            TQECException: If the new label is already used by a port.
+
+        Returns:
+            BlockGraph: The updated block graph with the relabeled cubes.
```
nelimee, line 432:
```suggestion
            block_graph: The block graph in which to search for cubes.
            identifier: A position (to rename one cube) or a label (to rename all cubes
            found with that label).
            new_label: The new label to assign.

        Raises:
            TQECException: If the new label is already used by a port.

        Returns:
            The updated block graph with the relabeled cubes.
```

src/tqec/computation/block_graph.py
```diff
@@ -414,6 +414,57 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
 
         return read_block_graph_from_dae_file(filename, graph_name)
 
+    def relabel_cubes(
+        block_graph: BlockGraph, identifier: Position3D | str, new_label: str
```
nelimee, line 418: According to the method name, I would rather see the following interface:

```suggestion
        block_graph: BlockGraph, label_mapping: Mapping[Position3D | str, str]
```

nelimee (commented):
A few comments, the main one being about the interface. I think it would be beneficial to be able to change multiple labels (or all of them) with one call to the method, hence the proposed interface. Let me know what you think about it.

src/tqec/computation/block_graph.py
angelaelisa: > According to the method name, I would rather see the following interface:

Yep, sorry. I was thinking on relabel more than one cube depending on the label, but I didn't realize I should name it `relabel_cube`, so yes, your solution make sense :)

> A few comments, the main one being about the interface. I think it would be beneficial to be able to change multiple labels (or all of them) with one call to the method, hence the proposed interface. Let me know what you think about it.

The purpose was to allow to relabel more than one cube, especially in case of ZXCubes, because the ones that are not a node in the ZXGraph, I think normally will be labeled with a simple name. For cubes that doesn't have label, you can rename all of them at once `block_graph.relabel_cubes("", "Cubes")` but for Ports always has to be one by one because they are unique labels. Remember that I'm a really beginner at this, so anything is good for you, it's going to be good for me :)

angelaelisa: > block_graph: BlockGraph, label_mapping: Mapping[Position3D | str, str]

Now I see, sorry. Definitely yes, this allows all at once, perfect

angelaelisa:
@nelimee it now has `label_mapping`. I've tested to ensure that it correctly throws the TQECException and changes as many labels as desired. This is the code used for testing:

```
from tqec import BlockGraph, Position3D

def main():    
    block_graph = BlockGraph.from_dae_file("slide67_open.dae")
    
    # let's see the list of cubes in the blockgraph
    print(f"Cubes:")
    for cube in block_graph.cubes:
        print(cube, cube.label)

    label_mapping = {
        "Port3": "Port30",
        "": "LabelCubes",
        Position3D(0, 0, 0): "Label000",
        "Port3": "Port40",

    }

    block_graph.relabel_cubes(label_mapping)

    print(f"New Cubes:")
    for cube in block_graph.cubes:
        print(cube, cube.label)

    block_graph.view_as_html("open_ports.html")

if __name__ == "__main__":
    main()
```

nelimee:
> @nelimee it now has `label_mapping`. I've tested to ensure that it correctly throws the TQECException and changes as many labels as desired. This is the code used for testing:
> 
> ```
> from tqec import BlockGraph, Position3D
> 
> def main():    
>     block_graph = BlockGraph.from_dae_file("slide67_open.dae")
>     
>     # let's see the list of cubes in the blockgraph
>     print(f"Cubes:")
>     for cube in block_graph.cubes:
>         print(cube, cube.label)
> 
>     label_mapping = {
>         "Port3": "Port30",
>         "": "LabelCubes",
>         Position3D(0, 0, 0): "Label000",
>         "Port3": "Port40",
> 
>     }
> 
>     block_graph.relabel_cubes(label_mapping)
> 
>     print(f"New Cubes:")
>     for cube in block_graph.cubes:
>         print(cube, cube.label)
> 
>     block_graph.view_as_html("open_ports.html")
> 
> if __name__ == "__main__":
>     main()
> ```
> 

I'll review the code tomorrow :)

src/tqec/computation/block_graph.py
```diff
@@ -452,6 +452,56 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
...
 
+    def relabel_cubes(
+        block_graph: BlockGraph, label_mapping: Mapping[Position3D | str, str]
+    ) -> BlockGraph:
+        """
+        Relabel multiple cubes in a BlockGraph using a mapping from position or label to new label.
+
+        Args:
+            block_graph (BlockGraph): The graph containing the cubes to be relabeled.
+            label_mapping (Mapping[Position3D | str, str]): Keys are either Position3D or str (old label),
```
nelimee, line 463:
```suggestion
            block_graph: The graph containing the cubes to be relabeled.
            label_mapping: Keys are either Position3D or str (old label),
```

src/tqec/computation/block_graph.py
```diff
@@ -452,6 +452,56 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
...
+                values are the new labels to apply.
+
+        Raises:
+            TQECException: If a target label already exists as a port label.
+
+        Returns:
+            BlockGraph: The updated block graph with relabeled cubes.
```
nelimee, line 470:
```suggestion
            The updated block graph with relabeled cubes.
```

src/tqec/computation/block_graph.py
```diff
@@ -452,6 +452,56 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
 
         return read_block_graph_from_dae_file(filename, graph_name)
 
+    def relabel_cubes(
+        block_graph: BlockGraph, label_mapping: Mapping[Position3D | str, str]
```
nelimee, line 456: If that is a method, the first argument should be `self` which represents the instance this method is applied on (and is basically what you called `block_graph`):
```suggestion
        self, label_mapping: Mapping[Position3D | str, str]
```

nelimee (commented):
I added 2 inline comments, but I also have 2 more general comments.

First, I think that `relabel_cubes` should mutate the instance it is applied on, inline. Any user that need to not modify the instance inline just have to copy + mutate the copy, which is easy to do (and can also be implemented in `tqec`).

Second, it seems (and I am unsure about that, please correct me if I am wrong) that you are not using some of the invariant of the `BlockGraph` structure that could help you. Namely, `BlockGraph` ensures that labels are unique. That means that, in your loop, you know for sure that `targets` is either empty or contains one entry.

angelaelisa:
> First, I think that `relabel_cubes` should mutate the instance it is applied on, inline. Any user that need to not modify the instance inline just have to copy + mutate the copy, which is easy to do (and can also be implemented in `tqec`).

Ok!

> Second, it seems (and I am unsure about that, please correct me if I am wrong) that you are not using some of the invariant of the `BlockGraph` structure that could help you. Namely, `BlockGraph` ensures that labels are unique. That means that, in your loop, you know for sure that `targets` is either empty or contains one entry.

I'm not sure if you mean this... The problem I got while testing a simpler version for `relabel_cubes` is that, as now is a mapping, if you introduced the same label for two ports, it was not always detected and the BlockGraph finally had two ports with same label. It's like sometimes, as when you want to fill cubes with label "" (empty string), it *jumps* the control of that labels are unique. In the afternoon I'll try to find the test I made that causes that error. Next time I'll take note of the tests and the issues detected in each one, sorry

angelaelisa:
@nelimee I modify it to fit the style too. I need three `TQECException`, to avoid this failure:

```
    label_mapping = {
        "Port3": "Port30",
        Position3D(0, 0, 0): "Label000",
        "Port4": "Port40",
        "": "Port30"
    }
```

src/tqec/computation/block_graph.py
```diff
@@ -414,6 +414,58 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
...
+                raise TQECException("New label must be non-empty.")
+
+            if new_label in port_labels:
+                raise TQECException(
+                    f"The label '{new_label}' is already used by a port and cannot be reassigned."
+                )
+            if new_label in assigned_new_labels:
+                raise TQECException(
+                    f"The label '{new_label}' is reused multiple times in this relabeling."
+                )
```
nelimee, line 445: These might be all checked before the loop. It is up to preference, here is an example of code that might be used instead.

```py
from collections import Counter
from itertools import chain

class BlockGraph:
    # ....
    def relabel_cubes(self, label_mapping: Mapping[Position3D | str, str]) -> None:
        # ....
        labels = Counter(chain.from_iterables((port_labels, label_mapping.values())))
        duplicated_labels = {label for label, count in labels.items() if count > 1}
        if duplicated_labels:
            raise TQECException(f"Found duplicated label(s): {duplicated_labels}.")
```

It is up to you to decide what you prefer, I am OK with both approaches.

src/tqec/computation/block_graph.py
```diff
@@ -414,6 +414,58 @@ def from_dae_file(filename: str | pathlib.Path, graph_name: str = "") -> BlockGr
...
+                matching_cubes = [cube for cube in self.cubes if cube.label == key]
+            else:
+                raise TQECException(
+                    f"Invalid identifier '{key}'. Must be Position3D or str."
+                )
+
+            if not matching_cubes:
+                raise TQECException(f"No cube found for identifier '{key}'.")
+
+            for cube in matching_cubes:
```
nelimee, line 461: There can be only one cube at a given position, and only one cube with a specific label, so `matching_cubes` is either empty or contains one item. Also, it might be nice to replace that with a single new method:

```py
def get_cube(self, pos_or_label: Position3D | str) -> Cube:
    # Implementation here
```

and call it here like that:

```py
cube = self.get_cube(key)
if cube is None:
    raise TQECException(f"No cube found for identifier '{key}'.")
```

nelimee (commented):
Sorry for the delay, I added a few comments :)

src/tqec/computation/block_graph.py
angelaelisa: > There can be only one cube at a given position, and only one cube with a specific label, so `matching_cubes` is either empty or contains one item. Also, it might be nice to replace that with a single new method:

@nelimee I totally agree that it would be useful to have a method like that. In fact, I think it could be helpful not only for this specific use case but also for managing `BlockGraph` from the frontend layer. Do you mind if I make a new commit in this PR that includes the `get_cube()` method and integrates it into `relabel_cubes`? Or would you prefer I submit it as a separate PR, so it's easier to review in isolation? :)

nelimee: Commits in this PR are fine, it is not a huge PR yet :)

angelaelisa: @nelimee I was thinking that since we already have:

`def get_cubes_by_label(self, label: str) -> list[Cube]`

…maybe we could also introduce a complementary method like:

`def get_cube_by_position(self, pos: Position3D) -> Cube`

To make it more explicit the differences between:

- `label` → possibly multiple cubes
- `position` → exactly one cube

And we won't have to change the methods that are already using `def get_cubes_by_label`.

What do you think?

nelimee: I am fine with the introduction of two methods with explicit names. I think that "since **only** ports are required to have unique labels" should be discussed further. Initially I thought that labels were only used on ports (and so unique in the whole `BlockGraph`). What are the use-cases to label non-port cubes? And to have several non-port cubes with the same label?

angelaelisa: I feel that labeling non-port cubes—even with repeated labels—can be quite useful, especially when managing or creating block graphs programmatically, outside of SketchUp.

In the frontend layer, for example, if I want to add elements to a specific region of interest, having labels helps me quickly identify those cubes without relying on exact positions. Being able to search by label and perform operations—like inserting or linking blocks—adds flexibility for users building or editing graphs through a UI.

So, although not strictly necessary for simulation or compilation, supporting labeled non-port cubes could empower alternative workflows that avoid SketchUp entirely and make frontend development smoother. What do you think?

nelimee:
@angelaelisa I am sorry for not re-reviewing that PR, I forgot. Do you want me to 1) merge changes from `main` here and 2) re-review?

angelaelisa:
> @angelaelisa I am sorry for not re-reviewing that PR, I forgot. Do you want me to 1) merge changes from `main` here and 2) re-review?

Yes, please, thanks @nelimee 

After using `fill_ports_for_minimal_simulation()` I think solving relabeling with this PR is enough and we could close  #527 succeeding it. Thanks again :)

nelimee:
Seems like I was able to push my modifications but they are not taken into account here. You might have something to validate @angelaelisa. Not sure though.

angelaelisa:
@nelimee could you review it? I fetched the latest version from the current state of TQEC and reintroduce `def relabel_cubes` in `block_graph.py`

nelimee (commented):
Looks good to me. Could you add one or two simple tests in `block_graph_test.py`?

nelimee (commented):
In tests:

- prints are not needed.
- things that should be verified should be verified with `assert [thing that should be true]`.
- things that should raise an exception should be checked with

```py
with pytest.raises([expected exception type, for example TQECException]):
    [code that should raise]
```

angelaelisa:
@nelimee I'm gonna have a look at it carefully in the afternoon, sorry

angelaelisa:
@nelimee sorry for the delay, it was a difficult day :)

nelimee (approved):
LGTM.

Just confirm that you want that PR merged now (and that you do not have anything not pushed yet that you wish to include here), and I will do it.

angelaelisa:
@nelimee I think is enough. The flask app is working perfect with minimal fill ports. So, just need this to compose two blockgraphs into one, relabeling first the ports of one of them :)

PR #537, opened 2025-03-23
Add explanations for `-e` option to `pip` when installing

nelimee:
This PR adds more details on the use of the `-e` option when installing `tqec` for contributors.

### DIFF
--- a/docs/contributor_guide.rst
+++ b/docs/contributor_guide.rst
@@ -35,14 +35,14 @@ compatibility with Python 3.13.
     See https://stackoverflow.com/a/71585691 for more info.
 
 .. warning::
-    The `-e` option to the `python -m pip install` call is **important**. Without that option, changes made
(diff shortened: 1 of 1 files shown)


docs/contributor_guide.rst
```diff
@@ -33,6 +34,16 @@ compatibility with Python 3.13.
     You will have to install ``pandoc`` separately as the instructions above only install a ``pandoc`` wrapper.
     See https://stackoverflow.com/a/71585691 for more info.
 
+.. warning::
+    The ``-e`` option to the ``python -m pip install`` call is **important**. Without that option, changes made
```
purva-thakre, line 38:
```suggestion
    The ``-e`` option to the ``python -m pip install`` call is **important** as it installs an editable version of ``tqec``. Without that option, changes made
```

minor nit

inmzhang (approved):

PR #538, opened 2025-03-23
Add note about how to import assets

purva-thakre:
The example in the readme can result in an `ImportError`. This PR adds a line below the example on how to check if the path exists or not.


README.md
```diff
@@ -56,6 +56,13 @@ circuit = compiled_computation.generate_stim_circuit(
 )
 ```
 
+If the ``assets/logical_cnot.dae`` file cannot be imported, check if the path exists on your local device.
+
+```py
+import os
+os.path.isfile('/local_path/tqec/assets/logical_cnot.dae')
```
nelimee, line 63: Maybe?
```suggestion
assert os.path.isfile('/local_path/tqec/assets/logical_cnot.dae')
```

README.md
```diff
@@ -56,6 +56,13 @@ circuit = compiled_computation.generate_stim_circuit(
 )
 ```
 
+If the ``assets/logical_cnot.dae`` file cannot be imported, check if the path exists on your local device.
+
+```py
+import os
+assert os.path.isfile('/local_path/tqec/assets/logical_cnot.dae')
+```
```
nelimee, line 64: I'm not sure this is strictly needed. Just the sentence above should be sufficient. If we want to add a check, maybe include it directly in the code above with something like:

```py
from pathlib import Path

from tqec import BlockGraph, compile_block_graph, NoiseModel

# 0. Path to the logical_cnot.dae file
cnot_dae_filepath = Path.cwd() / "assets" / "logical_cnot.dae"
if not cnot_dae_filepath.exists():
    print(f"The file '{cnot_dae_filepath}' does not exists.")

# 1. Construct the logical computation
block_graph = BlockGraph.from_dae_file(cnot_dae_filepath)
# ...
```

That would also allow the example to be cross-platform (it is not right now due to Windows using... backslashs `\` as delimiters).

purva-thakre: The `Path.cwd()` depends on where the `tqec` repo is stored and where a user is running their code. 

For example, if I have my jupyter notebook in a separate location, then `Path.cwd()` tries to find the assets in this folder.

nelimee: > `Path.cwd()` depends on where the `tqec` repo is stored and where a user is running their code.

Nope, only on where a user is running their code (mostly, in which folder they called `python`). And that's expected, because relative filepaths are relative to `cwd()` if I am right. Am I mistaken somewhere here?

purva-thakre: let me check. Maybe my understanding is incorrect...

nelimee: From the [documentation of `open`](https://docs.python.org/3/library/functions.html#open):

> file is a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object) giving the pathname (absolute or **relative to the current working directory**)

purva-thakre: Thanks. My understanding was based on what the current working directory is for my jupyter notebook.

nelimee (approved):

PR #545, opened 2025-03-25
Implement threshold-finding algorithm

nelimee:
This PR adds a function to estimate the threshold of any given `BlockGraph`/`CorrelationSurface` combo.


src/tqec/simulation/threshold.py
```diff
@@ -0,0 +1,218 @@
...
+    return result
+
+
+def binary_search_threshold(
+    block_graph: BlockGraph,
+    observable: CorrelationSurface,
+    noise_model_factory: Callable[[float], NoiseModel],
+    minp: float = 10**-5,
+    maxp: float = 0.1,
+    ks: tuple[int, ...] = (1, 2),
```
inmzhang, line 70:
```suggestion
    ks: Sequence[int] = (1, 2),
```

src/tqec/simulation/threshold.py
```diff
@@ -0,0 +1,218 @@
...
+    atol: float = 10**-8,
+    rtol: float = 10**-4,
+    manhattan_radius: int = 2,
+    block_builder: BlockBuilder = CSS_BLOCK_BUILDER,
+    substitution_builder: SubstitutionBuilder = CSS_SUBSTITUTION_BUILDER,
+    detector_database: DetectorDatabase | None = None,
+    num_workers: int = multiprocessing.cpu_count(),
+    max_shots: int | None = None,
+    max_errors: int | None = None,
```
inmzhang, line 79: It seems that the two must both be not None. Otherwise, `sinter` will raise exception.

src/tqec/simulation/threshold.py
```diff
@@ -0,0 +1,218 @@
...
+        num_hits=stat.errors,
+        max_likelihood_factor=max_likelihood_factor,
+    )
+    if stat.errors == 0:
+        result = sinter.Fit(low=result.low, high=result.high, best=float("nan"))
+
+    return result
+
+
+def binary_search_threshold(
```
inmzhang, line 64: This function searchs for the threshold of "logical error rate per `d` rounds for distance `d` code". Usually, we refer the threshold in the sense "logical error rate per round". The two thresholds will have some difference and the former will be smaller. We should note that in the docstring in my opinion.

src/tqec/simulation/threshold.py
```diff
@@ -0,0 +1,218 @@
...
+    block_graph: BlockGraph,
+    observable: CorrelationSurface,
+    noise_model_factory: Callable[[float], NoiseModel],
+    minp: float = 10**-5,
+    maxp: float = 0.1,
+    ks: tuple[int, ...] = (1, 2),
+    atol: float = 10**-8,
+    rtol: float = 10**-4,
```
inmzhang, line 72: This is a very strict tolerance. I didn’t expect the surface code to have such good crossing, especially with finite-size effects at small code distances. However, when I tested it in the memory experiment, it worked well. So, I think I'm okay with it.

nelimee: This was for testing, and I never changed it back. In practice the algorithm here:

- may return before that tolerance is reached (the last branch in the while loop, which can only be entered if the user provided 3 or more values to `ks`),
- will unconditionally decrease the precision, exponentially fast (halving at each loop iteration), even if the numbers do not make sense, that is something I would like to take into account.

src/tqec/simulation/threshold.py
nelimee: Does it? My understanding of `sinter` output is that this function rather computes the "shot error-rate", or in other words the probability that the provided computation fails on one trial. I agree that this may warrant more explanations in the docstring. Also, even though I think it does make sense here, there is a possibility that I misunderstood what people usually call "threshold" and that word is not the best one to describe what this function is doing.

inmzhang: Oh, I was referring to the `d*d*d` memory experiment—sorry for the confusion. For a memory experiment, logical error rate per `d` rounds is equivalent to the shot error rate. There can be different definition of threshold regarding Monte Carlo simulation. But most papers use the logical error rate per round as the metric for threshold computation. I think a simple note here should suffice.

inmzhang: I realized that for general logical computation there is no "per round" concept at all... Then I think there is no need to add the explanation.

inmzhang (approved):

nelimee:
Before merging I'll have a try at taking into account error bars to stop the iteration when it makes no sense anymore.

afowler:
Not to overemphasize the importance of the threshold error rate, but it
should be a property of the code, not whether it's round by round or
whether it's part of a computation or memory. The threshold error rate is
the error rate below which one set of logically equivalent corrections is
exponentially favored over the other as the code distance increases.

Now finite size boundary effects do certainly have an impact, but just want
to be clear on the strict definition of threshold error rate. It's an open
question whether we can simulate situations at sufficient scale to observe
a uniform threshold error rate.

nelimee:
> Not to overemphasize the importance of the threshold error rate, but it
> should be a property of the code, not whether it's round by round or
> whether it's part of a computation or memory. The threshold error rate is
> the error rate below which one set of logically equivalent corrections is
> exponentially favored over the other as the code distance increases.
> 
> Now finite size boundary effects do certainly have an impact, but just want
> to be clear on the strict definition of threshold error rate. It's an open
> question whether we can simulate situations at sufficient scale to observe
> a uniform threshold error rate.


That makes perfect sense. That means that the threshold simulations for anything else than a single qubit memory experiment are a waste of resources, right?

afowler:
I think it is still a very useful feature, but perhaps not something you
want to turn on by default. For example, when a structure contains some
unusual features, such as stretched stabilizers, or Y basis initialization
or measurement, it could still be very interesting to check whether the
threshold error rate has been impacted. You typically wouldn't study the
threshold error rate of a large computation, rather just memory or small
pieces of computation with specific interesting structures in them.

Issue #551, opened 2025-03-29
Unable to build Sphinx docs


smburdick:
## Describe the bug

I'd like to test a change to the docs locally, but I cannot export the Sphinx docs to HTML.

## Steps to reproduce the behavior

After setting up a local development environment, I get the following in `docs/`

```bash
sphinx-build -M html . out/
Running Sphinx v7.2.6

Extension error:
Could not import extension nbsphinx (exception: No module named 'nbsphinx')
```

## Expected behavior

Having a local HTML copy of the docs would be nice to view, before pushing it out.

## Python dependency versions

Run `pip freeze` and put the output in the `<details>` element below.

<details>
accessible-pygments==0.0.5
alabaster==1.0.0
anyio==4.9.0
appnope==0.1.4
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==3.0.0
async-lru==2.0.5
attrs==25.3.0
babel==2.17.0
beautifulsoup4==4.13.3
bleach==6.2.0
certifi==2025.1.31
cffi==1.17.1
cfgv==3.4.0
charset-normalizer==3.4.1
comm==0.2.2
contourpy==1.3.1
coverage==7.7.1
cycler==0.12.1
debugpy==1.8.13
decorator==5.2.1
defusedxml==0.7.1
distlib==0.3.9
docutils==0.21.2
executing==2.2.0
fastjsonschema==2.21.1
filelock==3.18.0
fonttools==4.56.0
fqdn==1.5.1
h11==0.14.0
httpcore==1.0.7
httpx==0.28.1
identify==2.6.9
idna==3.10
imagesize==1.4.1
iniconfig==2.1.0
ipykernel==6.29.5
ipython==9.0.0
ipython_pygments_lexers==1.1.1
ipywidgets==8.1.5
isoduration==20.11.0
jedi==0.19.2
Jinja2==3.1.6
json5==0.10.0
jsonpointer==3.0.0
jsonschema==4.23.0
jsonschema-specifications==2024.10.1
jupyter-events==0.12.0
jupyter-lsp==2.2.5
jupyter_client==8.6.3
jupyter_core==5.7.2
jupyter_server==2.15.0
jupyter_server_terminals==0.5.3
jupyterlab==4.3.6
jupyterlab_pygments==0.3.0
jupyterlab_server==2.27.3
jupyterlab_widgets==3.0.13
kiwisolver==1.4.8
lark==1.2.2
[...]

PR #552, opened 2025-03-29
Update installation docs to show developer options for advanced use c…

smburdick:
Austin and I have discussed a use case linked in this [slide](https://docs.google.com/presentation/d/1oqUCnXWEe0NPKJyh_9cp2_sXLfJdgzEcby_VWz1BqNA/edit#slide=id.g33b2ade785c_0_2), which we'd like to have reflected in the actual documentation. He suggested including the step that installs the dev dependencies as well, so I'm updating the docs to include that. Can also replicate this use case in one of the tutorials. 

Related issue: #551

### DIFF
--- a/docs/user_guide/installation.rst
+++ b/docs/user_guide/installation.rst
@@ -62,3 +62,13 @@ Else, a message like
     ModuleNotFoundError: No module named 'tqec'
 
 should appear.
+
+Install dev dependencies
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+For some applications, you will need the development dependencies of the ``tqec`` package.
+To install them, run
+
+.. code-block:: bash
+
+    pip install -e . -r requirements-dev.txt


nelimee:
Hi @smburdick !

The mention of the `-e` option for `pip install` is already in the [Contributor's guide](https://tqec.github.io/tqec/contributor_guide.html). The workflow you mention is using code that has not been thoroughly tested and that won't be merged into `main`. 

I think the `-e` option should stay only for contributors. As soon as we will publish a version on PyPI (which should come soon), the `-e` option won't make any sense anymore for regular users.

afowler:
Thanks Adrien for clarifying, you can blame me for the suggestion to Sam to
include this in the documentation. What you've said below makes sense. Look
forward to seeing this resolved in the right way as described.

nelimee:
> Thanks Adrien for clarifying, you can blame me for the suggestion to Sam to include this in the documentation. What you've said below makes sense. Look forward to seeing this resolved in the right way as described.

Hope to get the layer approach merged in the next few days, then fix a few issues with plaquette schedules and we should be able to reproduce on `main` the spatial junction experiments done on the specific branch :)

afowler:
Excellent :-) looking forward to it :-)

PR #553, opened 2025-03-30
Fix local docs build errors

purva-thakre:
This PR makes sure documentation build errors are resolved in their respective PRs. 

I have noticed some build errors multiple times when building the documentation locally. I made a couple of PRs to fix them. 

To avoid this cycle from repeating, I feel it would be better to fix these errors in the PR where these changes are introduced.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,13 +3,15 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+# -W will make the workflow fail if there are any build errors/warnings
+SPHINXOPTS    ?= -W
 SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 
 # Put it first so that "make" without argument is like "make help".
 help:
+	rm -rfv "$(SOURCEDIR)/_autosummary"
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 .PHONY: help Makefile
@@ -17,4 +19,5 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	rm -rfv "$(SOURCEDIR)/_autosummary"
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -12,7 +12,7 @@ BUILDDIR      = _build
 # Put it first so that "make" without argument is like "make help".
 clean:
 	rm -rfv "$(SOURCEDIR)/_autosummary"
-	rm -rfv "$(SOURCEDIR)/_build"
+	@$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -10,8 +10,11 @@ SOURCEDIR     = .
 BUILDDIR      = _build
 
 # Put it first so that "make" without argument is like "make help".
-help:
+clean:
 	rm -rfv "$(SOURCEDIR)/_autosummary"
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 .PHONY: help Makefile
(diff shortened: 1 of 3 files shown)


purva-thakre:
@nelimee Do you mind approving the workflow? I want to check it fails as expected due to build errors.

nelimee:
Not sure this is the problem though, I do not think that the documentation is cached in CI, and so the `make clean` is basically is noop. What's the rationale behind it?

purva-thakre:
> What's the rationale behind it?

I am seeing some build errors locally. Trying to find the source of them.

nelimee:
> > What's the rationale behind it?
> 
> I am seeing some build errors locally on the `main` branch. Trying to find the source of them. And how to spot them in a github workflow run. 

I do not see any error or warning in CI. Maybe that's from your local installation? What are the errors?

purva-thakre:
> I do not see any error or warning in CI. 

Yeah, this is the thing that's confusing me.

nelimee:
Maybe try to remove the _apidocs (or a name like that) folder that is not cleaned automatically by make clean?

purva-thakre:
> Maybe try to remove the _apidocs (or a name like that) folder that is not cleaned automatically by make clean

I feel we have had a similar conversation about removing the build directory. 

I will check the sphinx configuration settings.

purva-thakre:
The issue was `make clean` was not removing autosummary generated files. The issue was fixed locally in https://github.com/tqec/tqec/pull/553/commits/c4c9799db948e1cf4e1f149d0e38225dca458e1d

docs/Makefile
```diff
@@ -3,18 +3,21 @@
...
-SPHINXOPTS    ?=
+# -W will make the workflow fail if there are any build errors/warnings
+SPHINXOPTS    ?= -W
 SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 
 # Put it first so that "make" without argument is like "make help".
 help:
+	rm -rfv "$(SOURCEDIR)/_autosummary"
```
nelimee, line 14: That seems strange to perform a destructive action before displaying an help message. I really think this should rather be `clean`:

*(warning: code below contains space instead of the required tabs)*
```
clean: 
    rm -rfv "$(SOURCEDIR)/_autosummary"
    @$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

docs/Makefile
```diff
@@ -3,18 +3,21 @@
...
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 .PHONY: help Makefile
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	rm -rfv "$(SOURCEDIR)/_autosummary"
```
nelimee, line 22: This should not be unconditionally executed in my opinion. I think we should just execute it on already destructive actions (I only see `clean`).

purva-thakre: Well, the issue with both of your suggestions is that without cleaning up the build directory locally, build errors are observed.

nelimee: There is already a `clean` target in the `Makefile`. At least `make help` shows it on my side:
```
Sphinx v8.1.3
Please use `make target' where target is one of
  html        to make standalone HTML files
  dirhtml     to make HTML files named index.html in directories
  singlehtml  to make a single large HTML file
  pickle      to make pickle files
  json        to make JSON files
  htmlhelp    to make HTML files and an HTML help project
  qthelp      to make HTML files and a qthelp project
  devhelp     to make HTML files and a Devhelp project
  epub        to make an epub
  latex       to make LaTeX files, you can set PAPER=a4 or PAPER=letter
  latexpdf    to make LaTeX and PDF files (default pdflatex)
  latexpdfja  to make LaTeX files and run them through platex/dvipdfmx
  text        to make text files
  man         to make manual pages
  texinfo     to make Texinfo files
  info        to make Texinfo files and run them through makeinfo
  gettext     to make PO message catalogs
  changes     to make an overview of all changed/added/deprecated items
  xml         to make Docutils-native XML files
  pseudoxml   to make pseudoxml-XML files for display purposes
  linkcheck   to check all external links for integrity
  doctest     to run all doctests embedded in the documentation (if enabled)
  coverage    to run coverage check of the documentation (if enabled)
  clean       to remove everything in the build directory
```
(see the last command)

nelimee: I think that `_autobuild` should only be rebuilt when existing code was removed or moved, it should remain valid when adding new code.

purva-thakre: > I think that _autosummary should only be rebuilt when existing code was removed or moved

There's no way to verify this locally unless we expect a person to go through all the previous PRs. 

I changed the target name to `autosummary_clean`. So, to build the docs locally in addition to cleaning up the build directory, we will have to use `make autosummary_clean html`.

nelimee: I still do not get why adding the removal to the existing `make clean` is not a suitable solution, sorry for that.

purva-thakre: > I still do not get why adding the removal to the existing make clean is not a suitable solution

Because that's what I interpreted from this https://github.com/tqec/tqec/pull/553#discussion_r2021053857 

Apologies!

I'll work on adding the removal to the clean target.

docs/Makefile
```diff
@@ -10,8 +10,9 @@ SOURCEDIR     = .
 BUILDDIR      = _build
 
 # Put it first so that "make" without argument is like "make help".
-autosummary_clean:
+clean:
 	rm -rfv "$(SOURCEDIR)/_autosummary"
+	rm -rfv "$(SOURCEDIR)/_build"
```
nelimee, line 15: Why not
```suggestion
	@$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
?

nelimee: Also, the `clean` target should be after the `help` one to keep `make` being equivalent to `make help`.

PR #558, opened 2025-04-04
Implement builders for the fixed-bulk invariant

nelimee:
This PR implements a `CubeBuilder` and a `PipeBuilder` following the fixed-bulk invariant that the top-left plaquette of the bulk is always measuring a `Z` stabilizer.

PR #559, opened 2025-04-07
Construct fixed-bulk observables

inmzhang:
Add changes from #510. This PR introduces the `PatchStyle` enum, which determines the `ObservableBuilder` used to construct observables.

It looks like `PatchStyle` is essential for building observables correctly. However, passing it around the entire compilation pipeline doesn’t seem ideal. Since specific `PatchStyle` values are tied to specific block builders, would it make sense to remove the builder arguments from `compile_block_graph` and instead only support high-level options like `CompileOption.FixedBulk` or `CompileOption.FixedBoundaryParity`? While this limits the flexibility for users to define custom block builders, I believe in 99.999...% of cases, users won't need to implement their own anyway.

### DIFF
--- a/src/tqec/compile/compile_test.py
+++ b/src/tqec/compile/compile_test.py
@@ -6,7 +6,7 @@ from tqec.compile.compile import compile_block_graph
 from tqec.compile.specs.library import ALL_SPECS
 from tqec.computation.block_graph import BlockGraph
 from tqec.computation.pipe import PipeKind
-from tqec.utils.enums import Basis
+from tqec.utils.enums import Basis, PatchStyle
 from tqec.utils.noise_model import NoiseModel
 from tqec.utils.position import Position3D
 
@@ -181,7 +181,11 @@ def test_compile_logical_cnot(spec: str, obs_basis: Basis, k: int) -> None:
     correlation_surfaces = g.find_correlation_surfaces()
     assert len(correlation_surfaces) == 2
     compiled_graph = compile_block_graph(
-        g, cube_builder, pipe_builder, correlation_surfaces
+        g,
+        cube_builder,
+        pipe_builder,
+        correlation_surfaces,
+        patch_style=PatchStyle.FixedBoundaryParity,
     )
     circuit = compiled_graph.generate_stim_circuit(
         k, noise_model=NoiseModel.uniform_depolarizing(0.001), manhattan_radius=2
--- a/src/tqec/compile/deprecated/compile.py
+++ b/src/tqec/compile/deprecated/compile.py
@@ -8,6 +8,10 @@ from typing import Final, Literal, Sequence, cast
 
 import stim
 
+from tqec.compile.observables.builder import (
+    compute_observable_qubits,
+    get_observable_with_measurement_records,
+)
 from tqec.utils.coordinates import StimCoordinates
 from tqec.circuit.measurement_map import MeasurementRecordsMap
 from tqec.circuit.qubit_map import QubitMap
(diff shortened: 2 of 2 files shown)


nelimee:
> Add changes from #510. This PR introduces the `PatchStyle` enum, which determines the `ObservableBuilder` used to construct observables.
> 
> It looks like `PatchStyle` is essential for building observables correctly. However, passing it around the entire compilation pipeline doesn’t seem ideal. Since specific `PatchStyle` values are tied to specific block builders, would it make sense to remove the builder arguments from `compile_block_graph` and instead only support high-level options like `CompileOption.FixedBulk` or `CompileOption.FixedBoundaryParity`? While this limits the flexibility for users to define custom block builders, I believe in 99.999...% of cases, users won't need to implement their own anyway.

Nice!

I think such options would make sense, because the block/pipe builders provided should match anyway and that is already an API issue to require the user to provide both while not ensuring they have the same convention.

Another option, that I prefer, would be to have a datastructure representing the triplet (cube builder, pipe builder, patch style) and require that data-structure to be passed around. That enforces the "the cube and pipe builders should follow the same convention" while also removing linked arguments. Does that make sense to you?

src/tqec/compile/compile.py
```diff
@@ -34,6 +35,7 @@ def compile_block_graph(
     cube_builder: CubeBuilder = STANDARD_CUBE_BUILDER,
     pipe_builder: PipeBuilder = STANDARD_PIPE_BUILDER,
     observables: list[CorrelationSurface] | Literal["auto"] | None = "auto",
+    patch_style: PatchStyle = PatchStyle.FixedBulk,
```
nelimee, line 38: If that is `FixedBulk` by default, the `STANDARD_*` options above for the cube/pipe builders should also be `FixedBulk` by default in my opinion.

src/tqec/compile/tree/annotators/observables.py
```diff
@@ -35,10 +38,25 @@ def annotate_observable(
...
+    )
+    from tqec.compile.observables.fixed_bulk_builder import (
+        FIXED_BULK_OBSERVABLE_BUILDER,
+    )
+
+    obs_builder = (
+        FIXED_PARITY_OBSERVABLE_BUILDER
+        if patch_style == PatchStyle.FixedBoundaryParity
+        else FIXED_BULK_OBSERVABLE_BUILDER
+    )
```
nelimee, line 54: The third member of the triplet proposed above can very much be an ObservableBuilder. That would solve the issue right?

inmzhang:
> Another option, that I prefer, would be to have a datastructure representing the triplet (cube builder, pipe builder, patch style) and require that data-structure to be passed around. That enforces the "the cube and pipe builders should follow the same convention" while also removing linked arguments. Does that make sense to you?

That makes sense. Can this be addressed in #558?

src/tqec/compile/compile.py
inmzhang: Yes. It can be changed to `FixedBoundaryParity` for now as the fixed bulk builders have not been implemented in the main branch.

src/tqec/compile/tree/annotators/observables.py
inmzhang: yes I think so.

nelimee (approved):
This PR is based on code that was already written (for both conventions), so I did not review again the whole logic behind the computation of measurements.

With the exception of the introduction of a triplet to represent the current convention, this looks good. I'll introduce the triplet in #558 

You can merge, or ask me to merge, up to you :)

Issue #566, opened 2025-04-10
Update cnot_all_observables.py


BSchelpe:
The dependency in benchmarks/cnot_all_observables.py was missed when the deprecated folder in tqec/compile was deleted. 

When I try to run cnot_all_observables.py , it doesn't run because it is looking for the deprecated.compile file. 

benchmarks/cnot_all_observables.py needs to be updated to use the new pipeline.

BSchelpe:
Solved by the above commit.

purva-thakre:
@BSchelpe If you use one of the keywords provided at [this link](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) right before an issue number, the issue will be automatically closed after the PR is merged.

BSchelpe:
Ah, thank you! I didn't know that. If only I hadn't put 'issue' between 'fixes' and my issue number :)

PR #568, opened 2025-04-11
Fix cnot_all_observables.py

BSchelpe:
This fixes issue #566.

The code now runs. The output looks plausible. It is difficult to confirm that it is the same as that output by the old compile pipeline as there have been several commits since the deprecated compile script was deleted. 

My changes to the script are minimal. I have replaced the default arguments of CSS_BLOCK_BUILDER and CSS_SUBSTITUTION_BUILDER in compile_block_graph(), with the new default parameter of FIXED_BULK_CONVENTION in the new compile_block_graph function. It would be good if someone more familiar with the new pipeline could confirm that the new default parameters are intended to produce the same result as the old default parameters.

### DIFF
--- a/benchmarks/cnot_all_observables.py
+++ b/benchmarks/cnot_all_observables.py
@@ -4,8 +4,9 @@ from pathlib import Path
 import stim
 from tqecd.construction import annotate_detectors_automatically
 
-from tqec.compile.deprecated.compile import CompiledGraph, compile_block_graph
-from tqec.compile.specs.library.css import CSS_BLOCK_BUILDER, CSS_SUBSTITUTION_BUILDER
+from tqec.compile.compile import compile_block_graph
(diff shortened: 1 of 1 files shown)


nelimee:
> It would be good if someone more familiar with the new pipeline could confirm that the new default parameters are intended to produce the same result as the old default parameters.

They are not producing the same result, because for the moment `tqec` can only generate circuits using the "standard" plaquettes ("standard" is how we call it, not because it is the standard in the community), which is not the same as "CSS". This is not a problem though, as long as the circuit generated in correct and the logical error-rate decreases as expected, it is fine :)

nelimee:
@BSchelpe if you are fine with it, I can merge this PR :)

BSchelpe:
Yes please! :)

PR #569, opened 2025-04-14
Fix outdated documentation

inmzhang:
Fix the observable counts in the user guide and update the simulation scripts/results in gallery.


nelimee (approved):
For the memory experiment, there is a note, it would be better to use the 

```rst
.. note::
    Note content here.
```
construct. 

Else, this looks good to me.

nelimee (approved):
LGTM

Issue #509, opened 2025-03-12
Use `DetectorDatabase` transparently


nelimee:
## Is your feature request related to a problem? Please describe.

The default and recommended usage of `tqec` leaves out a nice performance improvement by not using `DetectorDatabase`. That leads to performance issues (e.g., #494, even though this issue have another source).

## Describe the solution you'd like

`DetectorDatabase` should be used transparently when computing detectors. Ideally, it should be saved somewhere for future reuse. Here are the possible solutions:

1. Do not save anything, and use a fresh database at each computation. That can already save a lot of computation for large topological computations, but there will be an upfront cost of re-computing once each detector and that cost is removed in the other solutions below.
2. Transparently save (when finishing a detector computation) and reload (when computing new detectors) the `DetectorDatabase` instance to a unique path on the user home directory. That means that the first detector computation will be costly (computing new detectors), but subsequent detector computations will re-use the saved database and will take way less time to run.
3. Ship an already populated database along with `tqec`.

First and foremost, which of the above points should be prioritised in your opinion?

afowler:
I'm very supportive of a precomputed database since that leads to the
maximum potential performance improvement.

nelimee:
> We have a small

If I remember correctly, for the logical CNOT, that around 800 different structures. I would have to check again.

> , well-defined, and finite set of structures that we want to simulate, so the size of a pre-populated database with common detector patterns is bounded.

but it is bounded, yes.

> Perhaps in the long-term we can take a 2 & 3 hybrid approach where we ship a pre-populated db with common detector patterns and add interfaces to load/save from the users home directory.

Ideally that's what I had in mind. 

Something like a CLI to call (e.g., `tqec download`) to download a pre-populated database to a user-writable directory (something like `$XDG_DATA_HOME` or `~/.local/share/`) and then tqec can enrich that database when the user sees situations not in it.

BSchelpe:
I would like to try this issue, if no one else is working on it. 

In similar situations, I have usually seen option 2 in Adrien's list implemented.
My worry about option 3 would be that it would be tricky to enforce that the shipped database stays up to date as the 
code base changes (ie it is rebuilt every time the code changes in a way that changes the detectors). 
 
My proposal would be to implement option 2 for the near term now, that would already improve performance
now. And then the addition of a precomputed shipped database could be left open to be added in the future. I don't 
think this would cause much tech debt. 

Is this an acceptable way forward?

nelimee:
[...]

PR #573, opened 2025-04-23
Implement detector database transparently

BSchelpe:
This implements #509, specifically option 2. Ie if the user does not pass in any preferences about the detector database the code under the hood uses the local default database, both reading from (if it already exists) and writing to it. 

Some other options have also been added: do_not_use_database means no database is used anywhere (in case the user wants a completely clean computation with no interaction with the database) and only_use_database - this was a previously existing code path, that was cutoff by the change from CompiledGraph to TopologicalCompiledGraph, but has now been reinstated.  

Notes about my code changes:
1. the detector compute functions now pass the database as well as the detectors back up, so that it can be saved to file at the end of the computation.
2. I added a detector_database property to AnnotateDetectorsOnLayerNode and LayerNode. The former because it belongs there, the latter because the updated database needs to persist between walks along the children in order to then be saved to file, LayerNode is also where the save_database_to_file() function sits. At the end of LayerTree._annotate_detectors() is where the database is saved to file. 
3. I have put the conversion of the user arguments into values for detector_database and database_path at the beginning of LayerTree.generate_circuit() as this looked to be the latest one might expect a user to be calling in to the code path.

### DIFF
--- a/src/tqec/compile/detectors/database.py
+++ b/src/tqec/compile/detectors/database.py
@@ -22,8 +22,11 @@ from tqec.templates.subtemplates import SubTemplateType
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Shift2D
 
-DEFAULT_DETECTOR_DATABASE_PATH: Final[Path] = Path(
-    user_data_dir(appname="TQEC") + "\\my_detector_database.pkl"
+DEFAULT_DETECTOR_DATABASE_PATH: Final[Path] = (
+    # Note the use of the "*_path" variant that returns directly a Path instance.
+    # Operator "/" is overloaded by the Path class to be the correct path separator depending on the OS,
+    # so the below line should be a valid path independently of the OS.
+    user_data_path(appname="TQEC") / "my_detector_database.pkl"
 )
 
 
--- a/src/tqec/compile/detectors/database.py
+++ b/src/tqec/compile/detectors/database.py
@@ -5,8 +5,7 @@ import pickle
 from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
-from platformdirs import user_data_dir
-from typing import Final, Sequence
+from typing import Sequence
 
 from tqec.circuit.measurement_map import MeasurementRecordsMap
 from tqec.circuit.moment import Moment
@@ -21,13 +20,6 @@ from tqec.plaquette.plaquette import Plaquettes
 from tqec.templates.subtemplates import SubTemplateType
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Shift2D
-
-DEFAULT_DETECTOR_DATABASE_PATH: Final[Path] = (
-    # Note the use of the "*_path" variant that returns directly a Path instance.
-    # Operator "/" is overloaded by the Path class to be the correct path separator depending on the OS,
-    # so the below line should be a valid path independently of the OS.
-    user_data_path(appname="TQEC") / "my_detector_database.pkl"
-)
 
 
 @dataclass(frozen=True)
--- a/src/tqec/compile/detectors/database.py
+++ b/src/tqec/compile/detectors/database.py
@@ -5,7 +5,8 @@ import pickle
 from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
-from typing import Sequence
+from platformdirs import user_data_dir
+from typing import Final, Sequence
 
 from tqec.circuit.measurement_map import MeasurementRecordsMap
 from tqec.circuit.moment import Moment
@@ -21,7 +22,9 @@ from tqec.templates.subtemplates import SubTemplateType
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Shift2D
 
-DEFAULT_DETECTOR_DATABASE_PATH = "./my_detector_database.pkl"
+DEFAULT_DETECTOR_DATABASE_PATH: Final[Path] = Path(
+    user_data_dir(appname="TQEC") + "\\my_detector_database.pkl"
+)
 
 
 @dataclass(frozen=True)
--- a/src/tqec/compile/detectors/database.py
+++ b/src/tqec/compile/detectors/database.py
@@ -21,6 +21,8 @@ from tqec.templates.subtemplates import SubTemplateType
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Shift2D
 
+DEFAULT_DETECTOR_DATABASE_PATH = "./my_detector_database.pkl"
+
 
 @dataclass(frozen=True)
 class _DetectorDatabaseKey:
--- a/src/tqec/compile/graph.py
+++ b/src/tqec/compile/graph.py
@@ -472,8 +472,8 @@ class TopologicalComputationGraph:
                 argument is negative.
            detector_database: an instance to retrieve from / store in detectors
                 that are computed as part of the circuit generation. If not given,
-                the detectors are retrieved from/stored in the default location of
-                /my_detector_database.
+                the detectors are retrieved from/stored in the provided 
+                ``database_path``.
             database_path: specify where to save to after the calculation, when augmented.
                 As for detector_database, this defaults to /my_detector_database if
                 not specified. If detector_database is not passed in, the code attempts to
--- a/src/tqec/compile/graph.py
+++ b/src/tqec/compile/graph.py
@@ -464,7 +464,7 @@ class TopologicalComputationGraph:
 
         Args:
             k: scale factor of the templates.
-            noise_models: noise models to be applied to the circuit.
+            noise_model: noise model to be applied to the circuit.
             manhattan_radius: radius considered to compute detectors.
                 Detectors are not computed and added to the circuit if this
                 argument is negative.
@@ -472,8 +472,8 @@ class TopologicalComputationGraph:
                 that are computed as part of the circuit generation. If not given,
                 the detectors are retrieved from/stored in the provided
                 ``database_path``.
-            database_path: specify where to save to after the calculation, when
-                augmented. This defaults to :data:`.DEFAULT_DETECTOR_DATABASE_PATH`
+            database_path: specify where to save to after the calculation. This
+                defaults to :data:`.DEFAULT_DETECTOR_DATABASE_PATH`
                 if not specified. If detector_database is not passed in, the code
                 attempts to retrieve the database from this location. The user
                 may pass in the path either in str format, or as a Path instance.
@@ -492,7 +492,7 @@ class TopologicalComputationGraph:
             database_path=database_path,
             do_not_use_database=do_not_use_database,
             only_use_database=only_use_database,
-        )  # pass through the variables
+        )
         # If provided, apply the noise model.
         if noise_model is not None:
             circuit = noise_model.noisy_circuit(circuit)
--- a/src/tqec/compile/graph.py
+++ b/src/tqec/compile/graph.py
@@ -477,7 +477,7 @@ class TopologicalComputationGraph:
                 if not specified. If detector_database is not passed in, the code
                 attempts to retrieve the database from this location. The user
                 may pass in the path either in str format, or as a Path instance.
-            do_not_use_database: if True, even the default database will not be used.
+            do_not_use_database: if ``True``, even the default database will not be used.
             only_use_database: if ``True``, only detectors from the database
                 will be used. An error will be raised if a situation that is not
                 registered in the database is encountered.
--- a/src/tqec/compile/graph.py
+++ b/src/tqec/compile/graph.py
@@ -62,10 +62,7 @@ from tqec.compile.blocks.positioning import (
     LayoutPosition2D,
     LayoutPosition3D,
 )
-from tqec.compile.detectors.database import (
-    DetectorDatabase,
-    DEFAULT_DETECTOR_DATABASE_PATH,
-)
+from tqec.compile.detectors.database import DetectorDatabase
 from tqec.compile.observables.abstract_observable import AbstractObservable
 from tqec.compile.observables.builder import ObservableBuilder
 from tqec.compile.tree.tree import LayerTree
@@ -73,6 +70,7 @@ from tqec.templates.enums import TemplateBorder
 from tqec.utils.exceptions import TQECException
 from tqec.utils.noise_model import NoiseModel
 from tqec.utils.position import BlockPosition3D, Direction3D, SignedDirection3D
+from tqec.utils.paths import DEFAULT_DETECTOR_DATABASE_PATH
 from tqec.utils.scale import PhysicalQubitScalable2D
 
 
@@ -472,10 +470,10 @@ class TopologicalComputationGraph:
                 argument is negative.
            detector_database: an instance to retrieve from / store in detectors
                 that are computed as part of the circuit generation. If not given,
-                the detectors are retrieved from/stored in the provided 
+                the detectors are retrieved from/stored in the provided
                 ``database_path``.
             database_path: specify where to save to after the calculation, when augmented.
-                As for detector_database, this defaults to /my_detector_database if
+                This defaults to tqec.utils.paths.DEFAULT_DETECTOR_DATABASE_PATH if
                 not specified. If detector_database is not passed in, the code attempts to
                 retrieve the database from this location. The user may pass in the path
                 either in str format, or as a Path instance.
--- a/src/tqec/compile/graph.py
+++ b/src/tqec/compile/graph.py
@@ -102,6 +102,12 @@ class TopologicalComputationGraph:
         """Represents a topological computation with
         :class:`~tqec.compile.blocks.block.Block` instances."""
         self._blocks: dict[LayoutPosition3D, Block] = {}
+        # For fixed-bulk convention, temporal Hadamard pipe has its on space-time
+        # extent. We need to keep track of the temporal pipes that are at the
+        # same layer of at least one temporal Hadamard pipe.
+        # We use the bottom cube position `z` to store the temporal pipe, s.t.
+        # the pipe is actually at the position `z+0.5`
+        self._temporal_pipes_at_hadamard_layer: dict[LayoutPosition3D, Block] = {}
         self._scalable_qubit_shape: Final[PhysicalQubitScalable2D] = (
             scalable_qubit_shape
         )
@@ -345,13 +351,13 @@ class TopologicalComputationGraph:
         self._replace_temporal_border(
             source,
             TemporalBlockBorder.Z_POSITIVE,
-            block.get_temporal_border(TemporalBlockBorder.Z_POSITIVE),
+            block.get_temporal_border(TemporalBlockBorder.Z_NEGATIVE),
         )
         # Sink
         self._replace_temporal_border(
             sink,
             TemporalBlockBorder.Z_NEGATIVE,
-            block.get_temporal_border(TemporalBlockBorder.Z_NEGATIVE),
+            block.get_temporal_border(TemporalBlockBorder.Z_POSITIVE),
         )
 
     def add_pipe(
@@ -377,6 +383,19 @@ class TopologicalComputationGraph:
         if block.is_temporal_pipe:
             self._check_block_spatial_shape(block)
             self._replace_temporal_borders(source, sink, block)
+            block_trimmed_temporal_borders = block.with_temporal_borders_replaced(
+                {
+                    TemporalBlockBorder.Z_NEGATIVE: None,
+                    TemporalBlockBorder.Z_POSITIVE: None,
+                }
+            )
+            if block_trimmed_temporal_borders:
+                u_pos = LayoutPosition3D.from_block_position(source)
+                # We use the bottom cube position `z` to store the temporal pipe, s.t.
+                # the pipe is actually at the position `z+0.5`
+                self._temporal_pipes_at_hadamard_layer[u_pos] = (
+                    block_trimmed_temporal_borders
+                )
         else:  # block is a spatial pipe
             self._trim_cube_spatial_borders(source, sink)
             key = LayoutPosition3D.from_pipe_position((source, sink))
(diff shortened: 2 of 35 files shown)


nelimee:
One preliminary comment without looking (yet) at the code: initially, I planned to change the `DetectorDatabase` in-place, hence removing the need to return it because the instance provided is modified in-place. I think this is the most reasonable way of doing things, as the detectors in the database should be unconditionally valid*.

It turns out that this was not working correctly with `AnnotateDetectorsOnLayerNode` due to an implicit copy of the database. I fixed that in 3e9c62a6588f2eaa3e767755c2662672d4b17dfa. 

I'll do a full review later, thanks for the PR :)

*: as long as the detector computation procedures and the link `Plaquette.name <-> Plaquette.circuit` are left untouched.

BSchelpe:
Thanks for looking at it!

On modifying the detector database in place, I'm happy to change back to doing that if you prefer it:) 
Just to explain my motivation: I feel more comfortable with functions being explicit about what they've modified 
if the modified object/variable is going to be used again outside the function, by explicitly returning the object/variable. I feel if it isn't explicit it potentially leads to errors when other people modify the code who aren't aware of the implicit behaviour. I agree though that adding the updated database to be returned by the function makes the code clunkier. Could an alternative be to add the functions which modify the detector database as functions on a class? Eg compute_detectors_for_fixed_radius() could become a function on AnnotateDetectorsOnLayerNode and then it would have access to self._database during the computation? I am happy though to follow whichever structure you prefer.

I also see some tests have failed. I'm sorry I ran all the tests I thought were relevant and saw explicitly labelled in the code base before pushing, but didn't see these. I will look into why they're failing.

nelimee:
> I feel more comfortable with functions being explicit about what they've modified if the modified object/variable is going to be used again outside the function, by explicitly returning the object/variable.

I agree with you in general, but for cache/store objects I feel like that is not as much a necessity as for other objects. Let's see what other think about that: should we always return a `DetectorDatabase` instance, or rather modify it in-place and only return the circuit/detectors? 

------

In general, #509 was only about making the use of `DetectorDatabase` transparent to users. I feel like most of the changes here (returning instead of modifying in-place, adding the database as an attribute of `NodeLayer`, ...) are not really solving directly that issue. I am OK discussing these changes, but they should likely rather be introduced in another PR related to another (to be opened and discussed) issue.

src/tqec/compile/detectors/database.py
```diff
@@ -21,6 +21,8 @@
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Shift2D
 
+DEFAULT_DETECTOR_DATABASE_PATH = "./my_detector_database.pkl"
+
```
nelimee, line 25: This should be an absolute path. Else, it will depend on where the user launched the `python` executable, and the user will have several `my_detector_database.pkl` all over its folders.

I think the best approach would be to use https://github.com/tox-dev/platformdirs to get a reliable and sensible absolute path to store the database.

src/tqec/compile/tree/tree.py
```diff
@@ -238,13 +250,23 @@ def _generate_annotations(
         k: int,
         manhattan_radius: int = 2,
         detector_database: DetectorDatabase | None = None,
+        database_path: Path = Path(DEFAULT_DETECTOR_DATABASE_PATH),
```
nelimee, line 253: The `DEFAULT_DETECTOR_DATABASE_PATH` should already be a `Path` instance. It is better to declare it as a `Path` and not instantiate a new instance each time this method is called.

src/tqec/compile/detectors/database.py
nelimee: Also, the type should be `Final[Path]`:

```py
from typing import Final
from pathlib import Path 

DEFAULT_DETECTOR_DATABASE_PATH: Final[Path] = Path(...)
```

BSchelpe:
> I vote for modifying it in-place for reducing the complexity.

Well, now we have a majority vote :) I'll revert the code to that behaviour.

BSchelpe:
> A few more precise comments.
> 
> I am really not convinced by the addition of a `_detector_database` attribute to `LayerNode`. Why do you need that again? It seems to me that storing the database in the `LayerTree` instance make way more sense, and I am also doubtful about the need of doing that.

This was intended to reduce the number of functions that had to pass the database back up, because then the functions 
on the classes have access to the ._database property on the class. If we're reverting to changing the database in place then this is no longer needed.

src/tqec/compile/detectors/database.py
BSchelpe: > This should be an absolute path. Else, it will depend on where the user launched the `python` executable, and the user will have several `my_detector_database.pkl` all over its folders.
> 
> I think the best approach would be to use https://github.com/tox-dev/platformdirs to get a reliable and sensible absolute path to store the database.

Yes, I agree. I'll look into making this change.

src/tqec/compile/graph.py
```diff
@@ -463,15 +470,30 @@ def generate_stim_circuit(
...
                 argument is negative.
-            detector_database: an instance to retrieve from / store in detectors
-                that are computed as part of the circuit generation.
-
+           detector_database: an instance to retrieve from / store in detectors
+                that are computed as part of the circuit generation. If not given,
+                the detectors are retrieved from/stored in the default location of
+                /my_detector_database.
+            database_path: specify where to save to after the calculation, when augmented.
+                As for detector_database, this defaults to /my_detector_database if
```
nelimee, line 478:
```suggestion
                As for detector_database, this defaults to :data:`tqec.utils.paths.DEFAULT_DATABASE_PATH` if
```

src/tqec/compile/tree/tree.py
```diff
@@ -238,13 +253,23 @@ def _generate_annotations(
...
-        self._annotate_detectors(k, manhattan_radius, detector_database, lookback)
+        self._annotate_detectors(
+            k,
+            manhattan_radius,
+            detector_database,
+            database_path,
+            only_use_database,
+            lookback,
+        )
+        # This function will also update the detector_database and save it to disk at database_path.
```
nelimee, line 272:
```suggestion
        # This method will also update the detector_database and save it to disk at database_path.        
        self._annotate_detectors(
            k,
            manhattan_radius,
            detector_database,
            database_path,
            only_use_database,
            lookback,
        )
```

src/tqec/compile/tree/tree.py
```diff
@@ -273,22 +301,46 @@ def generate_circuit(
...
+                the detectors are retrieved from/stored in the default location of
+                /my_detector_database.
+            database_path: specify where to save to after the calculation, when augmented.
+                As for detector_database, this defaults to /my_detector_database if
+                not specified. If detector_database is not passed in, the code attempts to
+                retrieve the database from this location.
+            do_not_use_database: if True, even the default database will not be used.
+            only_use_database: if ``True``, only detectors from the database
+                will be used. An error will be raised if a situation that is not
+                registered in the database is encountered.
```
nelimee, line 315: See the other review comment changing a few things in the descriptions of the parameters.

src/tqec/compile/tree/tree.py
```diff
@@ -273,22 +301,46 @@ def generate_circuit(
...
+            database_path = Path(database_path)
+        # If the user has passed a database in, use that, otherwise:
+        if detector_database is None:  # Nothing passed in,
+            if database_path.exists():  # look for an existing database at the path.
+                detector_database = DetectorDatabase.from_file(database_path)
+            else:  # if there is no existing database, create one.
+                detector_database = DetectorDatabase()
+        if (
+            do_not_use_database
+        ):  # override the above code and reset the database to None.
```
nelimee, line 334: Minor styling change
```suggestion
        # Override the above code and reset the database to None.
        if do_not_use_database:  
```

nelimee (commented):
A few more comments. 

I realise now that we might need to be able to version the detector database if we want that feature to work as expected, because we need to be able to tell `tqec` "the database format changed, you cannot rely on previous database formats, please re-generate it from scratch". 

For the moment I do not know how often we will have to invalidate the database, but any of the following event should invalidate any previously computed database:

- the `DetectorDatabase` class changed,
- the `_DetectorDatabaseKey` class changed (e.g., how we compute its hash),
- an existing "plaquette circuit <=> plaquette name" link has been broken (e.g., changing the implementation of a plaquette without changing its name),

This is outside of the scope of this PR for now, but if you want to have a look at it you can also include it in this PR. If you prefer to stay focused on the main goal, I can do that in a separate PR. Please tell me what you prefer :)

src/tqec/compile/detectors/database.py
```diff
@@ -21,6 +22,10 @@
 from tqec.utils.exceptions import TQECException
 from tqec.utils.position import Shift2D
 
+DEFAULT_DETECTOR_DATABASE_PATH: Final[Path] = Path(
+    user_data_dir(appname="TQEC") + "\\my_detector_database.pkl"
+)
```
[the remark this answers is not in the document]
BSchelpe: Thank you for catching that! I wasn't thinking about unix and windows having different \ vs /!

I'll coordinate with Yilun about making a paths.py file.

src/tqec/compile/tree/tree.py
BSchelpe: I'll expand the wording to make this a bit clearer in order to move the comment to above the line.

src/tqec/compile/graph.py
BSchelpe: Just to check, I've never seen :data: used in a comment before - what does it mean?

src/tqec/compile/graph.py
```diff
@@ -463,15 +470,30 @@ def generate_stim_circuit(
             manhattan_radius: radius considered to compute detectors.
                 Detectors are not computed and added to the circuit if this
                 argument is negative.
-            detector_database: an instance to retrieve from / store in detectors
-                that are computed as part of the circuit generation.
-
+           detector_database: an instance to retrieve from / store in detectors
+                that are computed as part of the circuit generation. If not given,
+                the detectors are retrieved from/stored in the default location of
+                /my_detector_database.
```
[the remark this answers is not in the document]
BSchelpe: nice catch, thank you!

BSchelpe:
> A few more comments.
> 
> I realise now that we might need to be able to version the detector database if we want that feature to work as expected, because we need to be able to tell `tqec` "the database format changed, you cannot rely on previous database formats, please re-generate it from scratch".
> 
> For the moment I do not know how often we will have to invalidate the database, but any of the following event should invalidate any previously computed database:
> 
> * the `DetectorDatabase` class changed,
> * the `_DetectorDatabaseKey` class changed (e.g., how we compute its hash),
> * an existing "plaquette circuit <=> plaquette name" link has been broken (e.g., changing the implementation of a plaquette without changing its name),
> 
> This is outside of the scope of this PR for now, but if you want to have a look at it you can also include it in this PR. If you prefer to stay focused on the main goal, I can do that in a separate PR. Please tell me what you prefer :)

I think I'd find it less complicated to do this as a separate PR if that would be ok. But I would like to have a go at it :) I'll make a issue for it once I'm done with this PR.

BSchelpe:
I think this is now ready for re-review. I have just two questions:
1. generate_stim_circuits_with_detectors() in simulation/generation.py seems to have a similar interface. I'm not quite sure when this is called. Should it also have its parameters changed to match the new ones of .generate_stim_circuit() on TopologicalCompiledGraph (and then pass those through to that function)? Is that right, or is this called in a different way? 
2. I've looked at the failing tests and I'm afraid I don't understand them. Firstly, coverage_python seems to be complaining about 'No module named platformdirs', but it runs fine on my machine? And secondly build_and_test_python is complaining about 'No such file or directory /home/runner/.local/share/TQEC', which doesn't seem obviously related to my changes.

nelimee:
> I think this is now ready for re-review. I have just two questions:

I will do a full review later this week :)

>     1. generate_stim_circuits_with_detectors() in simulation/generation.py seems to have a similar interface. I'm not quite sure when this is called. Should it also have its parameters changed to match the new ones of .generate_stim_circuit() on TopologicalCompiledGraph (and then pass those through to that function)? Is that right, or is this called in a different way?

Yes, nice catch, I think it would also benefit from the new parameters :)

>     2. I've looked at the failing tests and I'm afraid I don't understand them. Firstly, coverage_python seems to be complaining about 'No module named platformdirs', but it runs fine on my machine? And secondly build_and_test_python is complaining about 'No such file or directory /home/runner/.local/share/TQEC', which doesn't seem obviously related to my changes.

I will shortly push 2 changes that should in theory fix those bugs. Might push more if I see that the errors are not fully fixed.

src/tqec/compile/graph.py
nelimee: According to https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#automatically-document-attributes-or-data, `:data:` is a way to automatically link variables or attributes.

src/tqec/compile/graph.py
```diff
@@ -464,14 +469,30 @@ def generate_stim_circuit(
                 Detectors are not computed and added to the circuit if this
                 argument is negative.
             detector_database: an instance to retrieve from / store in detectors
-                that are computed as part of the circuit generation.
+                that are computed as part of the circuit generation. If not given,
+                the detectors are retrieved from/stored in the provided
+                ``database_path``.
+            database_path: specify where to save to after the calculation, when augmented.
```
nelimee, line 475:
```suggestion
            database_path: specify where to save to after the calculation.
```
I think the database is unconditionally saved, isn't it?

src/tqec/compile/graph.py
```diff
@@ -464,14 +469,30 @@ def generate_stim_circuit(
...
+                This defaults to tqec.utils.paths.DEFAULT_DETECTOR_DATABASE_PATH if
```
nelimee, line 476: I think that `:data` would be better:

```suggestion
                This defaults to :data:`.DEFAULT_DETECTOR_DATABASE_PATH` if
```
I'll commit that change.

src/tqec/compile/graph.py
```diff
@@ -464,14 +469,30 @@ def generate_stim_circuit(
...
+                not specified. If detector_database is not passed in, the code attempts to
+                retrieve the database from this location. The user may pass in the path
+                either in str format, or as a Path instance.
+            do_not_use_database: if True, even the default database will not be used.
```
nelimee, line 480:
```suggestion
            do_not_use_database: if ``True``, even the default database will not be used.
```

src/tqec/compile/graph.py
```diff
@@ -464,14 +469,30 @@ def generate_stim_circuit(
...
         circuit = self.to_layer_tree().generate_circuit(
-            k, manhattan_radius=manhattan_radius, detector_database=detector_database
-        )
+            k,
+            manhattan_radius=manhattan_radius,
+            detector_database=detector_database,
+            database_path=database_path,
+            do_not_use_database=do_not_use_database,
+            only_use_database=only_use_database,
+        )  # pass through the variables
```
nelimee, line 495: This comment is not really needed I think.

src/tqec/compile/tree/annotators/detectors.py
```diff
@@ -183,6 +184,9 @@ def __init__(
                 avoid computing detectors if the database already contains them.
                 Default to `None` which result in not using any kind of database
                 and unconditionally performing the detector computation.
+            only_use_database: if True, only detectors from the database will be
+                used. An error will be raised if a situation that is not registered
+                in the database is encountered. Default to False.
```
nelimee, line 189:
```suggestion
            only_use_database: if ``True``, only detectors from the database will be
                used. An error will be raised if a situation that is not registered
                in the database is encountered. Default to ``False``.
```

src/tqec/compile/tree/tree.py
```diff
@@ -143,15 +145,26 @@ def _annotate_detectors(
...
         )
+        # The database will have been updated inside the above function, and here at the end of the computation we save it
+        # to file:
+        if detector_database is not None:
+            detector_database.to_file(database_path)
+
+        # NB if database is None when passed in to .walk(AnnotateDetectorsOnLayerNode), then no database will be created
+        # in this function, the database will still be None afterwards and so nothing will be saved to file.
+        # Therefore, the 'do not use database' code path (which was initialised by setting the database to None) will
+        # produce nothing at the end, as desired.
```
nelimee, line 167: I would simplify that comment to:

```suggestion
        # Save the database to a file if it was not None.
        if detector_database is not None:
            detector_database.to_file(database_path)
```

Even that comment might not be needed.

src/tqec/compile/tree/tree.py
```diff
@@ -273,22 +299,45 @@ def generate_circuit(
...
-            detector_database: existing database of detectors that is used to
-                avoid computing detectors if the database already contains them.
-                Default to `None` which result in not using any kind of database
-                and unconditionally performing the detector computation.
+            detector_database: an instance to retrieve from / store in detectors
+                that are computed as part of the circuit generation. If not given,
+                the detectors are retrieved from/stored in the the provided
+                ``database_path``.
+            database_path: specify where to save to after the calculation, when augmented.
+                This defaults to tqec.utils.paths.DEFAULT_DETECTOR_DATABASE_PATH if
```
nelimee, line 307:
```suggestion
                This defaults to :data:`.DEFAULT_DETECTOR_DATABASE_PATH` if
```

src/tqec/compile/tree/tree.py
```diff
@@ -273,22 +299,45 @@ def generate_circuit(
...
+                not specified. If detector_database is not passed in, the code attempts to
+                retrieve the database from this location.
+            do_not_use_database: if True, even the default database will not be used.
```
nelimee, line 310:
```suggestion
            do_not_use_database: if ``True``, even the default database will not be used.
```

src/tqec/simulation/generation.py
```diff
@@ -51,7 +56,18 @@ def generate_stim_circuits_with_detectors(
             negative, detectors are not computed automatically and are not added
             to the generated circuits.
         detector_database: an instance to retrieve from / store in detectors
-            that are computed as part of the circuit generation.
+            that are computed as part of the circuit generation. If not given,
+            the detectors are retrieved from/stored in the provided
+            ``database_path``.
+        database_path: specify where to save to after the calculation, when augmented.
+            This defaults to tqec.utils.paths.DEFAULT_DETECTOR_DATABASE_PATH if
```
nelimee, line 63:
```suggestion
            This defaults to :class:`.DEFAULT_DETECTOR_DATABASE_PATH` if
```

src/tqec/simulation/generation.py
```diff
@@ -51,7 +56,18 @@ def generate_stim_circuits_with_detectors(
...
+            not specified. If detector_database is not passed in, the code attempts to
```
nelimee, line 64:
```suggestion
            not specified. If ``detector_database`` is not passed in, the code attempts to
```

src/tqec/simulation/generation.py
```diff
@@ -51,7 +56,18 @@ def generate_stim_circuits_with_detectors(
...
+            retrieve the database from this location. The user may pass in the path
+            either in str format, or as a Path instance.
+        do_not_use_database: if True, even the default database will not be used.
```
nelimee, line 67:
```suggestion
        do_not_use_database: if ``True``, even the default database will not be used.
```

src/tqec/simulation/generation.py
```diff
@@ -64,7 +80,10 @@ def generate_stim_circuits_with_detectors(
             k,
             manhattan_radius=manhattan_radius,
             detector_database=detector_database,
-        )
+            database_path=database_path,
+            do_not_use_database=do_not_use_database,
+            only_use_database=only_use_database,
+        )  # pass through the variables
```
nelimee, line 86: Comment not really needed.

src/tqec/simulation/generation.py
```diff
@@ -105,7 +127,18 @@ def generate_sinter_tasks(
             negative, detectors are not computed automatically and are not added
             to the generated circuits.
         detector_database: an instance to retrieve from / store in detectors
-            that are computed as part of the circuit generation.
+            that are computed as part of the circuit generation. If not given,
+            the detectors are retrieved from/stored in the provided
+            ``database_path``.
+        database_path: specify where to save to after the calculation, when augmented.
+            This defaults to tqec.utils.paths.DEFAULT_DETECTOR_DATABASE_PATH if
```
nelimee, line 134:
```suggestion
            This defaults to :data:`.DEFAULT_DETECTOR_DATABASE_PATH` if
```

src/tqec/simulation/generation.py
```diff
@@ -105,7 +127,18 @@ def generate_sinter_tasks(
...
+            not specified. If detector_database is not passed in, the code attempts to
```
nelimee, line 135:
```suggestion
            not specified. If ``detector_database`` is not passed in, the code attempts to
```

src/tqec/simulation/generation.py
```diff
@@ -105,7 +127,18 @@ def generate_sinter_tasks(
...
+            retrieve the database from this location. The user may pass in the path
+            either in str format, or as a Path instance.
+        do_not_use_database: if True, even the default database will not be used.
```
nelimee, line 138:
```suggestion
        do_not_use_database: if ``True``, even the default database will not be used.
```

src/tqec/simulation/generation.py
```diff
@@ -122,5 +155,8 @@ def generate_sinter_tasks(
             noise_model_factory,
             manhattan_radius,
             detector_database,
-        )
+            database_path=database_path,
+            do_not_use_database=do_not_use_database,
+            only_use_database=only_use_database,
+        )  # pass through the variables
```
nelimee, line 161: Comment not needed.

nelimee (commented):
A lot of very small comments about the docstrings to make them appear nicely in the HTML pages generated. Overall, this PR is fine to me. Once the comments will be addressed, I am ok to merge. Please let me know when you finished and want me to merge :)

src/tqec/compile/tree/tree.py
BSchelpe: I agree with removing the second comment - it made sense to me in my head, but perhaps wouldn't to someone outside my head :) However, I think the first one is important, for others who like I did initially saw nothing happening in this level of the function and would then be confused by why an unchanged database would be being saved back to disc.

BSchelpe: thanks!

BSchelpe:
Done! I think it is ready to merge now,please :)

nelimee (approved):

PR #580, opened 2025-05-08
Large performance improvement for `import tqec`

nelimee:
This PR changes a few imports to try to limit to a maximum what is imported by `import tqec`.

With 
```py
# test.py
import tqec
```
and benchmarking with
```sh
time python -m pyinstrument -o assets/benchmark_import.html -r html test.py
```

--------

Before this PR:

```
real    0m2.090s
user    0m6.648s
sys     0m0.282s
```
and 
image

-----

After this PR:

```
real    0m0.431s
user    0m2.814s
sys     0m0.050s
```
and 
image

----

The following external packages are no more imported automatically by `import tqec` (but are imported when the modules in `tqec` that require them are imported):

- `pyzx`
- `matplotlib`
- `scipy`

These were 3 out of the 5 main offenders. The 2 remaining ones are:

- `numpy`, that is extensively used everywhere so I have little hope to be able to avoid importing it.
- `networkx`, that is used by `BlockGraph`, the main user data-structure that we cannot avoid importing with `tqec`.


src/tqec/compile/blocks/__init__.py
```diff
@@ -32,16 +32,3 @@
...
-from .enums import SpatialBlockBorder as SpatialBlockBorder
-from .enums import TemporalBlockBorder as TemporalBlockBorder
-from .layers import BaseComposedLayer as BaseComposedLayer
-from .layers import BaseLayer as BaseLayer
-from .layers import LayoutLayer as LayoutLayer
-from .layers import PlaquetteLayer as PlaquetteLayer
-from .layers import RawCircuitLayer as RawCircuitLayer
-from .layers import RepeatedLayer as RepeatedLayer
-from .layers import SequencedLayers as SequencedLayers
-from .positioning import LayoutPosition2D as LayoutPosition2D
```
Zhaoyilunnn, line 47: Now all components have a long import path. For example `from tqec.compile.blocks.layers.atomic.layout import LayoutLayer`. 

I can understand that these internal components are commonly not used by users, but I still wonder whether deleting them from this `__init__.py` is necessary.

Specifically, I'd like to know how much does these imports contribute to the time consumption of `import tqec` (similarly questions for other modules like `specs`)

If these imports does not introduce external packages that harm performance , like `pyzx`, I vote for reserve them in the `__init__.py` of their submodules

nelimee: For `tqec` internals, only long and explicit import paths are used. In other words, `tqec` does not need these aliases (or re-exports) at all. 

That means that these are uniquely for users. 

With this fact, my rationale was the following: regular users will mostly use what's in the public documentation, and in theory that should be enough to get a quantum circuit from a topological computation, which is the goal of `tqec`. Any user needing more than what is imported in the main `src/tqec/__init__.py` file is either:

- doing something that `tqec` is not made for, but provide helpers for (e.g., `tqec.simulation`),
- going at least a little bit in `tqec` internals.

For both cases, long imports seem adequate.

Now about `__init__.py` files that are not the main one, I wanted to reduce the time spent trying to find *the* problematic import, so I mostly removed the whole `__init__.py` file. That also has the benefit of consistency: we do not have some objects of the module that are re-exported while others are not.

I am willing to see on a case-by-case basis though, and you are free to play around with the `__init__.py` in this PR to try to include back things without harming too much the performance of `import tqec`. (tell me if you cannot push on this PR, not sure about the rights here).

Zhaoyilunnn: Ok, make sense to me. Thanks!

nelimee: Also, if some imports are missing, we can add them back in another PR. This one was just because I found the import time to be large when benchmarking `tqec` for small values of `k` and wanted to see if it could be improved easily or not.

PR #582, opened 2025-05-10
Reduce the size of `DetectorDatabase` on disk

### LINKED ISSUE #577: Reduce the size of `DetectorDatabase` on disk

It would be nice to include a pre-populated detector database along with the `tqec` package (when it will be on `pypi`). To avoid requiring users to have to download too much data to get `tqec`, it would be nice to reduce the size of the database on disk.

A database that is optimised for low size when saved. Can be tested by

```py
import os
from pathlib import Path

from tqec.compile.convention import FIXED_PARITY_CONVENTION

DATABASE_PATH = Path(__file__).parent / "database.pkl"
k = 3

database = DetectorDatabase()

block_graph = steane_encoding(Basis.X)
compiled_graph = compile_block_graph(block_graph, FIXED_PARITY_CONVENTION)
    k, manhattan_radius=2, detector_database=database
)

database.to_file(DATABASE_PATH)
database_size = os.path.getsize(DATABASE_PATH)


# From https://stackoverflow.com/a/1094933
def sizeof_fmt(num, suffix="B"):
    for unit in ("", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"):
        if abs(num) < 1024.0:
            return f"{num:3.1f}{unit}{suffix}"
        num /= 1024.0
    return f"{num:.1f}Yi{suffix}"


print("Size of the database on disk:", sizeof_fmt(database_size))
```
which currently prints 
```
Size of the database on disk: 4.1MiB
```
on my system.

Zhaoyilunnn:
WIP, will add more details soon

close #577 

---

original: a224a2e0c36b4346dd8c2a152041338c20c1754b
opt_0: compact data structure for `Detector` 73f023100ce59306e4f600fa8db3e0735838ebf5
opt_1: compact data structure for `_DetectorDatabaseKey` a33d6d4bfe5f473217ff688d2522de23e10bf26b

- [ ] now opt_1 does not reduce size, need to understand why
- [x] fix the mypy type error later


nelimee:
>     * [ ]  now opt_1 does not reduce size, need to understand why

That might be due to an optimisation I was not aware of in `pickle`: 

> The [pickle](https://docs.python.org/3/library/pickle.html#module-pickle) module keeps track of the objects it has already serialized, so that later references to the same object won’t be serialized again.

From [this subsection of the `pickle` documentation](https://docs.python.org/3/library/pickle.html#comparison-with-marshal).

At first sight, I *think* that, with how `tqec` works, the same plaquette should be represented by differrent `Plaquette` instances (basically, we re-instantiate the instance each time we need it). If that is not the case, then `pickle` is likely already storing a unique instance to represent several equal plaquettes. If you have the time, that would be interesting to check.

If that is the case, the last optimisation I see requires to change the database: in practice, we do not need the subtemplates and plaquettes in the database, we only need the result of `_DetectorDatabaseKey.plaquette_names`, which can even be optimised as a 3-dimensional array of indices into a "store" of plaquette names. The only place where the subtemplates/plaquettes are needed it to get the circuit of a database key, that in turns is used to generate the crumble URL representing the database entry. My opinion is that this is a valuable information for the future (to be able to navigate with a future tool the stored detectors), but it is not strictly needed for `tqec` to operate correctly.

So *maybe* have a new data-structure that is a detector database, but optimised in size thanks to the above procedure and unable to export to a crumble URL? Or we simply say that the range 1-10 MiB is acceptable to download, and do not optimise the database.

Zhaoyilunnn:
> > ```
> > * [ ]  now opt_1 does not reduce size, need to understand why
> > ```
> 
> That might be due to an optimisation I was not aware of in `pickle`:
> 
> > The [pickle](https://docs.python.org/3/library/pickle.html#module-pickle) module keeps track of the objects it has already serialized, so that later references to the same object won’t be serialized again.
> 
> From [this subsection of the `pickle` documentation](https://docs.python.org/3/library/pickle.html#comparison-with-marshal).
> 
> At first sight, I _think_ that, with how `tqec` works, the same plaquette should be represented by differrent `Plaquette` instances (basically, we re-instantiate the instance each time we need it). If that is not the case, then `pickle` is likely already storing a unique instance to represent several equal plaquettes. If you have the time, that would be interesting to check.

I have not yet started investigating this issue. Thanks a lot for providing this information! I'll take a look and try to confirm whether this is the root cause.


> If that is the case, the last optimisation I see requires to change the database: in practice, we do not need the subtemplates and plaquettes in the database, we only need the result of `_DetectorDatabaseKey.plaquette_names`, which can even be optimised as a 3-dimensional array of indices into a "store" of plaquette names. The only place where the subtemplates/plaquettes are needed it to get the circuit of a database key, that in turns is used to generate the crumble URL representing the database entry. My opinion is that this is a valuable information for the future (to be able to navigate with a future tool the stored detectors), but it is not strictly needed for `tqec` to operate correctly.

This seems to be a promising and interesting solution to me. Thanks!

> So _maybe_ have a new data-structure that is a detector database, but optimised in size thanks to the above procedure and unable to export to a crumble URL? Or we simply say that the range 1-10 MiB is acceptable to download, and do not optimise the database.

Maybe we can have a try and do some experiment to see whether this really improves users experience?

nelimee:
> Maybe we can have a try and do some experiment to see whether this really improves users experience?

I am all-in into trying to optimize the amount of data that needs to be downloaded prior to using `tqec` efficiently, and the current detector database is part of that. Nevertheless, a 3MiB or even 10MiB database is not significant compared to the size of all the dependencies we require in `tqec`.

If you would like to explore that and do not mind potentially reaching the conclusion that such a size optimisation is not worth it or not attainable without sacrificing important features (which might be discouraging for some), then yes please go ahead :)

Zhaoyilunnn:
I did an experiment on top of this PR using [my own branch](https://github.com/Zhaoyilunnn/tqec/tree/perf/detect-db-size-test)

The test code is 

```python
import os
from pathlib import Path

from tqec.compile.convention import FIXED_BULK_CONVENTION, FIXED_PARITY_CONVENTION

DATABASE_PATH = Path(__file__).parent / "database.pkl"
k = 3

database = DetectorDatabase()
# database = DetectorDatabase.from_file(DEFAULT_DETECTOR_DATABASE_PATH)

block_graph = steane_encoding(Basis.X)
# block_graph = cnot(Basis.X)
    k, manhattan_radius=2, detector_database=database
)

# database.to_file(DATABASE_PATH)
database_size = os.path.getsize(DEFAULT_DETECTOR_DATABASE_PATH)


# From https://stackoverflow.com/a/1094933
def sizeof_fmt(num, suffix="B"):
    for unit in ("", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"):
        if abs(num) < 1024.0:
            return f"{num:3.1f}{unit}{suffix}"
        num /= 1024.0
    return f"{num:.1f}Yi{suffix}"


print("Size of the database on disk:", sizeof_fmt(database_size))
```

Some additional pickle files are stored for analysis purpose. 
using `du /home/zhaoyilun/.local/share/TQEC/detector_database*`
The results are

```
SIZE     FILE                                                                                                                                     DESC

1312    /home/zhaoyilun/.local/share/TQEC/detector_database.compact_plaquettes_all             # list of all "plaquettes" in the form of `dict[int, int]`, in which the value is the index of corresponding plaquette in the uniq plaquettes
1088    /home/zhaoyilun/.local/share/TQEC/detector_database.compact_subtemplates             # list of subtemplates
2512    /home/zhaoyilun/.local/share/TQEC/detector_database.mapping                                    # the optimized mapping
48      /home/zhaoyilun/.local/share/TQEC/detector_database.uniq_plaquettes                           # list of unique plaquettes

=========================================================================

1976    /home/zhaoyilun/.local/share/TQEC/detector_database.plaquettes_all                            # list of original plaquettes
1080    /home/zhaoyilun/.local/share/TQEC/detector_database.subtemplates                            # list of original subtemplates
2536    /home/zhaoyilun/.local/share/TQEC/detector_database.original_mapping                      # list of original mapping
```

- I didn't optimize subtemplates, so the sizes are almost unchanged
- The "plaquettes" size is reduced from 1916 to 1312
- However the total size of mapping remains similar
    - before: total = 2536, smaller than 1976 (plaquettes) + 1080 (subtemplates) 
    - after: total = 2512, larger than 1312 (plaquettes) + 1088 (subtemplates)

To summarize, the optimization to `_DetectorDatabaseKey` seems to work if I save them separately (but the improvement is not very remarkable I think). But when I save the whole mapping, no improvement is observed.

At this moment, I don't understand why

Zhaoyilunnn:
> At first sight, I think that, with how tqec works, the same plaquette should be represented by differrent Plaquette instances (basically, we re-instantiate the instance each time we need it). If that is not the case, then pickle is likely already storing a unique instance to represent several equal plaquettes. If you have the time, that would be interesting to check.

Now I doubt this is the case, I think `plaquettes_by_timestamp` saves different plaquette instances.

Because I observed that, for different computation the sizes of `plaquettes_by_timestamp` differ significantly, if only references are stored I guess they won't differ a lot

nelimee:
With the size improvements you show, it seems like there is only little room to optimize size and `pickle` is already doing a excellent job at trying to reduce the size.

With the small potential gains and the added code complexity, I feel like the best would be to keep the `pickle` version for its simplicity. One drawback is security: it is possible to perform arbitrary code execution by unpickling a file, that means that users would have to trust us (and trust the database we will provide by default) to not include malicious code. That also means that sharing database between people requires this trust. 

To summarize, here is what I think we should do:

- keep the original `pickle` version as the default. User-created database is trusted by default (because created via `DetectorDatabase.to_file` that does not try to inject malicious code into the resulting `pickle` file), and any user that does not want to trust us will be able to remove the pre-included database and generate its own.
- add another serialization procedure, that does not use `pickle`, in order to let users share databases with others without needing to trust the other. That can very much be JSON for example.

What do you think about that?

Zhaoyilunnn:
> * keep the original `pickle` version as the default. User-created database is trusted by default (because created via `DetectorDatabase.to_file` that does not try to inject malicious code into the resulting `pickle` file), and any user that does not want to trust us will be able to remove the pre-included database and generate its own.

I now agree that size optimization is not necessary.  Now I can confirm that many `Plaquettes` in `_DetectorDatabaseKey` are just references to existing ones base on [an additonal experiment ](https://github.com/Zhaoyilunnn/tqec/blob/1ecb15e1cdd32feef961bab5f2f537288c9806f3/src/tqec/compile/detectors/database.py#L425-L429)

> That might be due to an optimisation I was not aware of in pickle:
> The [pickle](https://docs.python.org/3/library/pickle.html#module-pickle) module keeps track of the objects it has already serialized, so that later references to the same object won’t be serialized again.

So, it's very likely that you're right and that's the root cause

> * add another serialization procedure, that does not use `pickle`, in order to let users share databases with others without needing to trust the other. That can very much be JSON for example.
> 
> What do you think about that?

I agree. I am aware of some other serialization library such as `msgpack` and `cbor`. Not sure whether they also have security issue or suitable for us. And JSON is also a good choice to me. I'll do some investigation and open another PR for this point. 

I'll close this PR, perhaps #577 can also be closed

Zhaoyilunnn:
I didn't see any special advantage to introduce additional dependencies like `msgpack` or `cbor`.

So I plan to use json. BUT, I met an issue when trying to jsonfy stim Circuit (required to jsonfy plaquettes). Actually, I *think* it is not possible to jsonfy it? @nelimee do you have any suggestions?

Zhaoyilunnn:
> Note that, using the `PlaquetteDebugInformation`, you might be able to format most of the plaquettes only using their RPNG notation, only leaving the extended stabilizers to be serialized with their full circuit. We can discuss that more in details here, or during the meeting if you plan to come :)

I *think* I've got your idea, maybe I can go ahead, and let you review when I finish

PR #584, opened 2025-05-14
Support json format when saving detector database to disk

Zhaoyilunnn:
As discussed in #582, it would be nice to have an option other than pickle to suppress the security concern

---

- [x] Only save unique plaquettes and refer to them using indices (should we consider readability?) This effectively reduces the db size (for a single logical cnot) from >30 MB to 1.3 MB
~~- [x] Jsonfy a plaquette only using its name (RPNG description string) for most cases. (`PlaquetteMapper` introduce circular import if used in `plaquette.py`)~~

---

The following test program seems to work as expected

```python
import os
from pathlib import Path

from tqec.compile.convention import FIXED_BULK_CONVENTION, FIXED_PARITY_CONVENTION

DATABASE_PATH = Path(__file__).parent / "database.json"
k = 3

database = DetectorDatabase()
# database = DetectorDatabase.from_file(DEFAULT_DETECTOR_DATABASE_PATH)
# database = DetectorDatabase.from_file(DATABASE_PATH, format="json")

# block_graph = steane_encoding(Basis.X)
block_graph = cnot(Basis.X)
    k, manhattan_radius=2, detector_database=database
)

database.to_file(DATABASE_PATH, format="json")
# database_size = os.path.getsize(DEFAULT_DETECTOR_DATABASE_PATH)
database_size = os.path.getsize(DATABASE_PATH)


# From https://stackoverflow.com/a/1094933
def sizeof_fmt(num, suffix="B"):
    for unit in ("", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"):
        if abs(num) < 1024.0:
            return f"{num:3.1f}{unit}{suffix}"
        num /= 1024.0
    return f"{num:.1f}Yi{suffix}"


print("Size of the database on disk:", sizeof_fmt(database_size))
```


src/tqec/plaquette/plaquette.py
```diff
@@ -132,6 +132,62 @@ def is_empty(self) -> bool:
...
+        #     return {
+        #         "name": self.name,
+        #         "qubits": self.qubits.to_dict(),
+        #         "circuit": self.circuit.to_dict(),
+        #         "mergeable_instructions": list(self.mergeable_instructions),
+        #         "debug_information": self.debug_information.to_dict(),
+        #     }
+        # return {
+        #     "name": self.name,
+        # }
```
nelimee, line 154: If `self.debug_information.rpng` is not `None`, that means that the plaquette can be represented as a RPNG. In that case, the following should be enough:

```suggestion
if self.debug_information.rpng is not None:
    return {"rpng": str(self.debug_information.rpng)}
```

With modification to the `from_dict` method below.

src/tqec/circuit/qubit_map.py
```diff
@@ -116,6 +116,30 @@ def to_circuit(self) -> stim.Circuit:
...
         return self.q2i[index]
 
+    def to_dict(self) -> dict[str, Any]:
+        """Return a dictionary representation of the qubit map.
+
+        The dictionary is intended to be used as a JSON object.
+        """
+        return {
+            "i2q": [[qi, q.to_dict()] for qi, q in self.i2q.items()],
+            "q2i": [[q.to_dict(), qi] for qi, q in self.i2q.items()],
```
nelimee, line 126: This is duplicated information. You can store only `i2q` and that should be enough.

src/tqec/compile/detectors/database.py
```diff
@@ -270,16 +313,84 @@ def to_crumble_urls(
...
-    def from_file(filepath: Path) -> DetectorDatabase:
+    def from_file(
+        filepath: Path, format: Literal["pickle", "json"] = "pickle"
+    ) -> DetectorDatabase:
         with open(filepath, "rb") as f:
-            database = pickle.load(f)
+            if format == "pickle":
+                database = pickle.load(f)
+            else:
+                data = json.load(f)
```
nelimee, line 392: For JSON, shouldn't the file be opened in mode `r` instead of `rb`?

src/tqec/plaquette/plaquette.py
```diff
@@ -208,6 +264,62 @@ def without_plaquettes(self, indices: Collection[int]) -> Plaquettes:
...
+
+        The dictionary is intended to be used as a JSON object.
+        """
+
+        def convert(value: Plaquette) -> Any:
+            return (
+                plaquettes_to_indices[value]
+                if plaquettes_to_indices
+                else value.to_dict()
+            )
```
nelimee, line 280: Why do you need that? Is there cases where `plaquettes_to_indices` does not contain all the plaquettes that might be explored?

nelimee: Just got it, disregard my above comment

src/tqec/plaquette/rpng/rpng.py
```diff
@@ -16,6 +17,22 @@ def __str__(self) -> str:
     def to_extended_basis(self) -> ExtendedBasis:
         return ExtendedBasis(self.value)
 
+    def to_dict(self) -> dict[str, Any]:
+        return {"value": self.value}
```
nelimee, line 21: Is that needed? Can't we just represent a `PauliBasis` by its value instead of having another indirection?

src/tqec/plaquette/rpng/rpng.py
```diff
@@ -117,6 +134,35 @@ def is_null(self) -> bool:
...
+        """Return a dictionary representation of the RPNG object.
+
+        The dictionary is intended to be used as a JSON object.
+        """
+        return {
+            "r": self.r.value if self.r else None,
+            "p": self.p.value if self.p else None,
+            "n": self.n,
+            "g": self.g.value if self.g else None,
+        }
```
nelimee, line 147: Same as `PauliBasis`: can't we just use `str(rpng)` in the calling context?

src/tqec/plaquette/rpng/rpng.py
```diff
@@ -150,6 +196,31 @@ def from_string(cls, rg_string: str) -> RG:
...
 
+    def to_dict(self) -> dict[str, Any]:
+        """Return a dictionary representation of the RG object.
+
+        The dictionary is intended to be used as a JSON object.
+        """
+        return {
+            "r": self.r.value if self.r else None,
+            "g": self.g.value if self.g else None,
+        }
```
nelimee, line 207: Same as `RPNG` and `PauliBasis`

src/tqec/plaquette/rpng/rpng.py
```diff
@@ -285,3 +356,37 @@ def view_as_svg(
...
+
+    def to_dict(self) -> dict[str, Any]:
+        """Return a dictionary representation of the RPNG description.
+
+        The dictionary is intended to be used as a JSON object.
+        """
+        return {
+            "corners": [rpng.to_dict() for rpng in self.corners],
+            "ancilla": self.ancilla.to_dict(),
+        }
```
nelimee, line 368: Same as `RPNG`, `PauliBasis`, ... Using `str(rpng_description)` and `RPNGDescription.from_string` should be enough, isn't it?

nelimee (commented):
That's a nice addition, a few comments for improvements left, but that seems very nice.

src/tqec/plaquette/plaquette.py
Zhaoyilunnn: Yes, that's right! But I encounter an issue. I know there's already a `PlaquetteMapper` exist that could translate the rpng description to a `Plaquette` instance. 

However, importing `PlaqutteMapper` in the `plaquette.py` file introduce "circular import" error that I haven't started trying to resolve.
Some alternatives in my mind now are (1) do not use `Plaquettes.from_dict()`, rather, implement utility functions to create plaquttes from rpng strings using `PlaqutteMapper` directly in the `database.py`. (2) Add something like `Plaquette.from_rpng_desc`

src/tqec/plaquette/rpng/rpng.py
Zhaoyilunnn: Absolutely right for these simple classes that can be easily converted between strings! I just write it in a hurry and have't check it, thank you!

Zhaoyilunnn:
Hi @nelimee I *think* I have addressed all comments. Except that I use `PlaquetteMapper` by a local import (to avoid circular import error, not elegant to me but I didn't find better solution)

I feel this PR is ready and you can merge it if you are fine with it

nelimee:
I'll have a look at the tests and re-do a review. If I do not have any significant changes to request, should I merge directly this PR?

Zhaoyilunnn:
> I'll have a look at the tests and re-do a review. If I do not have any significant changes to request, should I merge directly this PR? 

Yes please. And I'll definitely address any additional comments.

src/tqec/plaquette/plaquette.py
```diff
@@ -132,6 +133,55 @@ def is_empty(self) -> bool:
...
+            a new instance of :class:`Plaquette` with the provided
+            ``name``, ``qubits``, ``circuit``, ``mergeable_instructions`` and
+            ``debug_information``.
+        """
+        if "rpng" in data:
+            from tqec.compile.specs.library.generators.utils import PlaquetteMapper
+
+            return PlaquetteMapper().get_plaquette(
+                RPNGDescription.from_string(data["rpng"])
+            )
```
nelimee, line 169: So, that's an issue in my opinion. First, I think you should directly use [`DefaultRPNGTranslator`](https://github.com/tqec/tqec/blob/b9e8c83490f8bd9689eb861d3a06168686394462/src/tqec/plaquette/rpng/translators/default.py#L21) and the compiler that was used by the plaquette when serialized.

Now, that would require to have a way to find which translator/compiler were used. That calls for additional properties to be added in `PlaquetteDebugInformation` to host these variables, and a way to serialize/deserialize them. I think this would be the best option. 

Does that seem like a good idea to you, or do you have another way of doing it?

src/tqec/plaquette/plaquette.py
```diff
@@ -208,6 +258,62 @@ def without_plaquettes(self, indices: Collection[int]) -> Plaquettes:
...
+        }
+
+    @staticmethod
+    def from_dict(
+        data: dict[str, Any], plaquettes: Sequence[Plaquette] | None = None
+    ) -> Plaquettes:
+        """Return a collection of plaquettes from its dictionary representation.
+
+        Args:
+            data: dictionary with the keys ``plaquettes`` and ``default``.
```
nelimee, line 293: Missing docstring for `plaquettes`.

src/tqec/plaquette/plaquette.py
```diff
@@ -132,6 +133,55 @@ def is_empty(self) -> bool:
             self.circuit.get_circuit(include_qubit_coords=False) == stim.Circuit()
         )
 
+    def to_dict(self) -> dict[str, Any]:
+        """Return a dictionary representation of the plaquette.
+
+        The dictionary is intended to be used as a JSON object.
+        """
+        if self.debug_information.rpng is not None:
+            return {"rpng": str(self.debug_information.rpng)}
```
nelimee, line 142: I made a mistake: this line is not enough, because the default `str` representation of a `RPNGDescription` does not include the `RG` of the ancilla qubit, which might be different from the default values. 

The safe way would be to re-use the `to_dict`/`from_dict` methods that are less space-efficient but correct. Another way would be to implement `to_extended_str` and use `from_extended_str` in `from_dict`. 

I guess the safe way is better for the moment, and we can try to optimise once the serialization is correct.

src/tqec/utils/coordinates.py
```diff
@@ -59,3 +60,31 @@ def __eq__(self, value: object) -> bool:
...
+        Args:
+            data: dictionary with the keys ``x``, ``y`` and ``t``.
+
+        Returns:
+            a new instance of :class:`StimCoordinates` with the provided
+            ``x``, ``y`` and ``t``.
+        """
+        x = cast(float, data["x"])
+        y = cast(float, data["y"])
+        t = cast(float | None, data.get("t"))
```
nelimee, line 89: Why are these casts required here and nowhere else? Does that has something to do with the use of `Mapping` instead of `dict`?

nelimee (commented):
A lot of comments, the most important ones are about the use of `str(rpng)` that is not enough (sorry for that wrong comment before) and about the serialization of plaquette with their RPNG description and how we should also serialize the translator/compiler used to get back the plaquette from the RPNG description.

Zhaoyilunnn:
> A lot of comments, the most important ones are about the use of `str(rpng)` that is not enough (sorry for that wrong comment before) and about the serialization of plaquette with their RPNG description and how we should also serialize the translator/compiler used to get back the plaquette from the RPNG description.

Thank you!  Before addressing this issue, I am wondering whether it is significant to save plaquettes in rpng format.

Without this optimization, i.e. if I save the whole plaquette, the size of database in the cnot example only grows from 1.2MB to 1.3 MB

Consider now we even need more info for correct serialization/deserialization, it seems that saving a plaquette in rpng   string may not improve a lot.

What do you think?

nelimee:
That make sense. You can remove that optimisation for the moment, and we might implement it later correctly if it makes sense.

src/tqec/utils/coordinates.py
Zhaoyilunnn: I also used `cast` in `Measurement`. I initially thought that it is necessary for static analysis due to the use of `Any` (for `mypy`) , but it turns out that it is not necessary and mypy runs without error. I removed them.

src/tqec/compile/detectors/database.py
```diff
@@ -270,20 +313,89 @@ def to_crumble_urls(
...
-                    "expected."
-                )
+    def from_file(
+        filepath: Path, format: Literal["pickle", "json"] = "pickle"
+    ) -> DetectorDatabase:
+        if format == "pickle":
+            with open(filepath, "rb") as f:
+                database = pickle.load(f)
+        else:
+            with open(filepath) as f:
```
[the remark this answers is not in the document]
Zhaoyilunnn: I have applied your suggestion, but the parameter always disappear. Finally I found that it will be automatically removed by the `pyupgrade` pre-commit hook....  😄

Zhaoyilunnn:
Hi @nelimee, now I removed the RPNG optimization. I think I have addressed most of other comments, one minor comment left is about explicitly setting the file open mode. I wasn't able to fix it because it will be overwridden by `pyupgrade`.

I think this PR it's ready for another last review, please merge it if you are fine with it, thanks

src/tqec/compile/detectors/database.py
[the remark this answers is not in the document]
nelimee: I would prefer an explicit default, but let's keep it that way for the moment, that's really a detail and should not block this PR.

Issue #547, opened 2025-03-26
SketchUp Imports


jbolns:
## Describe the bug

Not a bug with TQEC, per se, but nonetheless a bug that we need to address on our end due to its importance.

SketchUp's importer (the feature you use in SketchUp to import a .dae with a blockgraph) has something of a bug that leads to the names of blocks/pipes being lost in the process. This breaks the export/import cycle needed to move 3D structures back and forth between SketchUp and TQEC. 

Please note. The issue does not affect SketchUp exports. Those still work just fine. 

## Steps to reproduce the behavior

Export a blockgraph from TQEC as dae and try importing it to SketchUp using the Collada importer. Check the names of pipes. They won't be there. Any subsequent export will not have the attributes needed for TQEC.

## Expected behavior

Imports into SketchUp need to preserve the name of each of the blocks/pipes in the structure, so that they can be exported again in a way that can be used in TQEC. 

## Solution

**Long-term**
We contacted SketchUp and are in contact with their engineers. They seem to agree it is a desired fix and have moved the matter into their backlog. In theory, a full solution will eventually be available. We are also hoping to convince them of making something similar available in free and web versions. 

There is no *better* solution. Nothing beats simply being able to click a button to import a file. 

**Short-term**
We still need alternatives in case the fix takes too long. 

We have tested a way to trigger imports via SketchUp Ruby console. It requires some set up and is not a solution for web versions (no console on web). But, in principle, it enables meaningful import of blockgraphs into SketchUp.

A [branch has been created](https://github.com/tqec/tqec/tree/lean_exports) for the following:
1. Adaptations to our current exports so the user has a choice between exporting a full .dae or an array with spatial information (needed to import via the console). 
2. Fully developing and adding the Ruby script to our assets, so anyone can download and use it. 
3. Tests for all the above.
4. Guidance.

## How to help
Item 1 above is in progress. I am taking care of this together with @inmzhang .

Item 2 will be addressed by @angelaelisa, who is a SketchUp Pro user and therefore knows the tool well, has followed the conversation with SketchUp, and knows Ruby.

We need additional hands for item 3 and 4.
- Item 3 is ideal for with sufficient knowledge of Python testing as to make tests. Since TQEC has a strict (don't ask me, I hate testing) approach to testing, this would also help you familiarise with something very useful across the entire code base.
- Item 4 would be perfect for someone completely new that wants to start experimenting with how to use TQEC.

[...]

PR #587, opened 2025-05-16
Lean exports

jbolns:
Cumulative commits in this PR cover:
1. Split COLLADA exports into (a) core export that outputs a list of cubes and pipes and (b) COLLADA exporter for final `.dae`. 
2. JSON importer and exporter on the TQEC end.
3. Extracted rotation logics from COLLADA importer so it can be used in both COLLADA and JSON importers.
4. Ruby script to use inside SketchUp to import JSONs from (1).
5. New template.skp with additional primitive for ports. 

Worth keeping in mind:
- Rotations implemented on JSON importer in `read_write.py` but not tested. We would need to create a SketchUp JSON exporter for that, which makes no sense currently being that SketchUp exports to COLLADA do work fine. We can revisit if/when another system integration makes itself available.
- The SketchUp importer requires an .skp component per primitive (saved to SketchUp's component folder). These components are easy to create: place ONE primitive at origin, save with name/kind of primitive, repeat for all primitives. It's boring, but it took me only 15 minutes. We can also host pre-made component files somewhere.

Closes: #547


assets/template.skb
inmzhang, line 1: What's the purpose of using `skb` over `skp`?

assets/tqec_blockgraph_loader.rb
inmzhang, line 1: Need to document what problem the plugin is trying to solve. Maybe in the new repository's readme file?

inmzhang (commented):
The code looks good to me. Some pre-commit hooks need to be run and fixed. And there are some mypy errors need to be fixed too.

inmzhang:
With the component files provided by J and the ruby extension in this PR, I successfully import the CZ block graph (`.json`) into SketchUp. Well Done!

And I suggest to make a `.rbz` file out of `.rb` file, which makes it easy to install the plugin directly in SketchUp, without manually placing the extension file in the nested Plugins directory. You can zip the single ruby file and rename it from `.zip` to `.rbz`. You can also optionally add icon to it.

assets/template.skb
jbolns: Oops. Typo.

assets/tqec_blockgraph_loader.rb
jbolns: Yes, on TQEC's JSON import/export is fairly agnostic but detailed guidance for the ruby/SketchUp specific aspects is helpful. Will do as suggested.

jbolns:
Since the `.rbz` matter is specific to the Ruby side of this issue, will think about it for a bit and address in the new repository. There are commercial reasons to not want to do this.

Other than that. Will check for mypy errors and issues with pre-commit and will push an improved version later today or tomorrow.

jbolns:
@inmzhang you'll notice a significant re-arrange of `read_write.py` in the last commit. While checking for errors and running tests I found out that, unless I'm missing something, we never actually used `core_exports`. It was also creating issues with a couple tests, so got rid of it altogether.

inmzhang (approved):
LGTM.

jbolns:
Sweet! Will hit button tomorrow. Never push on a Friday, nor 5 mins before having to go pick up your kids from Kinder. 😂

angelaelisa:
Nice @jbolns 🎉

PR #588, opened 2025-05-20
Add a version attribute to the DetectorDatabase class

### LINKED ISSUE #575: The detector database needs to be invalidated for relevant code changes
This came out of PR #573. 

@nelimee : "We need to be able to version the detector database if we want that feature [the transparent use of the detector database] to work as expected, because we need to be able to tell tqec "the database format changed, you cannot rely on previous database formats, please re-generate it from scratch"."

Previously, it would be up to the user to regenerate their own detector databases if a relevant part of the code changed and invalidated them. Now that the database is generated and used 'under the hood' by default, the code also needs to be responsible for regenerating the default database 'under the hood' when necessary.   

@nelimee  provided an initial list of events which should invalidate any previously computed database:

- the DetectorDatabase class changed,

- the _DetectorDatabaseKey class changed (e.g., how we compute its hash),

- an existing "plaquette circuit <=> plaquette name" link has been broken (e.g., changing the implementation of a plaquette without changing its name),

BSchelpe:
This should close #575 

Two comments about the code:
1. I'm not sure what numbering system to use for the version. It looks more like a version number in the format a.b, although it seems a bit overkill to distinguish changes warranting an increment in a vs an increment in b. On the other hand using a single number, a, looks odd. My slight preference is for a.b, leaving it up to the coder to choose whether their change is major or minor. There is also a question of what number to start from. I set it to 1.0, but possibly 0.1 would be better, more in keeping with the TQEC version being 0.01 at the moment?
2. One suboptimality in the code is that ideally I would like database instances saved before this code change (ie with no version number) to be treated as having the initial version number (1.0 at the moment). However, due to the way pickle loading works they are loaded with the current version number of the code.  This is only problematic if someone tries to load an old database instance (pre code change) after a long time when the database version has increased from 1.0 (which we don't anticipate happening very often). I hope it is enough to warn the user about this in the documentation, as I don't see a neat way to avoid it.

### DIFF
--- a/src/tqec/compile/detectors/database.py
+++ b/src/tqec/compile/detectors/database.py
@@ -208,11 +208,6 @@ class DetectorDatabase:
     - MINOR when the content of the database is invalidated (e.g. by changing a plaquette
       implementation without changing its name).
 
-    ``None`` is a special value which can be set in testing mode and translates into the
-    database not being used (it is equivalent to the 'do_not_use_database' parameter
-    in the functions which use the database). NB it must be set in __init__. If it is set
-    directly on .version, __init__ will overwrite it.
-
     Old databases generated prior to the introduction of a version attribute will be
     loaded with the default value of .version, without passing through __init__,
     ie (0,0,0).


nelimee:
In my opinion, the most pythonic way of handling versions is to use [`semver`](https://pypi.org/project/semver/). From the package documentation:

> The module follows the `MAJOR.MINOR.PATCH` style:
> 
> - MAJOR version when you make incompatible API changes,
> - MINOR version when you add functionality in a backwards compatible manner, and
> - PATCH version when you make backwards compatible bug fixes.

For this specific case, and in my opinion:

- `MAJOR` changes when the *format* of the file changes (i.e., when the attributes of `DetectorDatabase` change),
- `MINOR` changes when the *content* of the database is invalidated (e.g., by changing a plaquette implementation without changing its name),
- `PATCH` does not need to change.

BSchelpe:
> In my opinion, the most pythonic way of handling versions is to use [`semver`](https://pypi.org/project/semver/). From the package documentation:
> 
> > The module follows the `MAJOR.MINOR.PATCH` style:
> > 
> > * MAJOR version when you make incompatible API changes,
> > * MINOR version when you add functionality in a backwards compatible manner, and
> > * PATCH version when you make backwards compatible bug fixes.
> 
> For this specific case, and in my opinion:
> 
> * `MAJOR` changes when the _format_ of the file changes (i.e., when the attributes of `DetectorDatabase` change),
> * `MINOR` changes when the _content_ of the database is invalidated (e.g., by changing a plaquette implementation without changing its name),
> * `PATCH` does not need to change.

That makes sense to me. I'll add your guidance on what constitutes a major/minor change to the doc string. Given that does starting the version at 0.1 make sense?

nelimee:
> That makes sense to me. I'll add your guidance on what constitutes a major/minor change to the doc string. Given that does starting the version at 0.1 make sense?

I may be wrong, but for me the `0` major version in semantic versioning is slightly special. It basically means "this is still an alpha/beta, so each minor version change may be incompatible". As such, and because the `DetectorDatabase` currently works, I would start by `1.0`.

BSchelpe:
> > That makes sense to me. I'll add your guidance on what constitutes a major/minor change to the doc string. Given that does starting the version at 0.1 make sense?
> 
> I may be wrong, but for me the `0` major version in semantic versioning is slightly special. It basically means "this is still an alpha/beta, so each minor version change may be incompatible". As such, and because the `DetectorDatabase` currently works, I would start by `1.0`.

Ah that makes sense I was a bit unclear on the 0/1 difference.

nelimee (commented):
Initial comments. These are large texts, but should not generate large changes.

In order to add `semver` to the requirements of `tqec`, you should also add a line `semver` to `requirements.txt` in the root folder of tqec.

src/tqec/compile/detectors/database.py
```diff
@@ -197,12 +197,27 @@ class DetectorDatabase:
...
+    increased past 1.0 when this happens the database will need to be force regenerated
+    manually (by deleting and then rerunning - simply running the default code will
+    not add a version attribute to the saved database) to avoid errors.
     """
 
     mapping: dict[_DetectorDatabaseKey, frozenset[Detector]] = field(
         default_factory=dict
     )
     frozen: bool = False
+    version: float | None = 1.0
```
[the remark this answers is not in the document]
BSchelpe: Luckily your code suggestions work fine, thank you! At least, the code runs fine - you weren't thinking of any specific checks to check the ClassVar and Final are having the desired effect, were you?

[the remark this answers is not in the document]
BSchelpe: Ah no, it doesn't. It's fine locally for me, but the mypy checking after I pushed failed, with the error message you predicted (Final can only be used as an outermost qualifier). I'll change to your second best suggestion.

src/tqec/compile/detectors/database.py
```diff
@@ -197,12 +201,31 @@ class DetectorDatabase:
...
+
+    ``None`` is a special value which can be set in testing mode and translates into the
+    database not being used (it is equivalent to the 'do_not_use_database' parameter
+    in the functions which use the database).
+
+    Old databases generated prior to the introduction of a version attribute will be
+    loaded with the current version number. If the version number has increased past
+    1.0 when this happens the database will need to be force regenerated manually
+    (by deleting and then rerunning - simply running the default code will not add a
+    version attribute to the saved database) to avoid errors.
```
nelimee: Thinking about that again, wouldn't it make more sense to default to `None` here? That means that if you load a database without a version number (i.e., prior to this PR), the database is considered invalid and should be re-computed once.

As that is a one-time cost, I think that is reasonable. What's your opinion?

src/tqec/compile/tree/tree.py
```diff
@@ -317,15 +318,40 @@ def generate_circuit(
...
             else:  # if there is no existing database, create one.
                 detector_database = DetectorDatabase()
+        else:
+            user_defined = True
+        if database_path != DEFAULT_DETECTOR_DATABASE_PATH:
+            user_defined = True
         # If do_not_use_database is True, override the above code and reset the database to None
-        if do_not_use_database:
+        # Also do this if the database version is set to None in the TQEC code (ie in testing mode)
+        current_version = DetectorDatabase.version
```
nelimee: Isn't that line rather supposed to be
```suggestion
        current_version = detector_database.version
```
?

nelimee: I start to understand this one. Then shouldn't it be `CURRENT_DATABASE_VERSION`?

BSchelpe: Ah, no we can't. Your suggestion would work normally (and be more readable), but not when the version has been set to None on DetectorDatabase in the code for testing purposes. The subsequent code needs current_version to be None in this case.

src/tqec/compile/detectors/database.py
BSchelpe: I'm not sure it will work unfortunately, unless I have misunderstood. `None` translates into the database not being used, rather than regenerated. Even if we set some other value to mean "invalid and needs regenerating", I think the default still needs to be `CURRENT_DATABASE_VERSION`, so that any newly generated DetectorDatabase object will have this as its version number.

nelimee:
@BSchelpe looking again at the code, I think the `None` trick for the database makes everything more complex. Removing the "developers should set XXXX to None when they want to regenerate the database each time" requirement for this PR makes it way simpler, and that requirement can be addressed differently.

What is your opinion on the above?

BSchelpe:
> @BSchelpe looking again at the code, I think the `None` trick for the database makes everything more complex. Removing the "developers should set XXXX to None when they want to regenerate the database each time" requirement for this PR makes it way simpler, and that requirement can be addressed differently.
> 
> What is your opinion on the above?

That's a good point. I think I agree. In the short term, the do_not_use_database flag can be used for developers wanting the behaviour, so the functionality is there, if not as convenient as would be ideal. I'll strip out the None option from the PR.

src/tqec/compile/detectors/database.py
```diff
@@ -197,12 +197,27 @@ class DetectorDatabase:
...
+    format of the file changes (i.e. when the attributes of DetectorDatabase change);
+    MINOR when the content of the database is invalidated (e.g. by changing a plaquette
+    implementation without changing its name).``None`` is a special value which can be
+    set in testing mode and translates into the database not being used (it is
+    equivalent to the 'do_not_use_database' parameter in the functions which use the
+    database). NB old databases generated prior to the introduction of a version
+    attribute will be loaded with the current version number. If the version number has
+    increased past 1.0 when this happens the database will need to be force regenerated
+    manually (by deleting and then rerunning - simply running the default code will
+    not add a version attribute to the saved database) to avoid errors.
```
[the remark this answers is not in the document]
BSchelpe: I think I have found a way to solve this by adding an `init` method to DetectorDatabase which will be called when a new instance is initialised, but not when a pickled database is reloaded. This allows us to separate the default for the reloaded database from that for a new instance and so distinguish their behaviours. However, this seems to affect the behaviour of the other class variables, which I don't yet understand and need to dig into more.

src/tqec/compile/detectors/database.py
```diff
@@ -197,12 +201,38 @@ class DetectorDatabase:
...
+        self,
+        mapping: dict[_DetectorDatabaseKey, frozenset[Detector]] = None,
+        frozen: bool = False,
+    ):
+        self.version = CURRENT_DATABASE_VERSION
+        if mapping is None:
+            self.mapping = dict()
+        else:
+            self.mapping = mapping
+        self.frozen = frozen
```
BSchelpe: @nelimee I haven't removed the None option from the version number yet, because I wanted to check you were happy with this change to the initialisation of DetectorDatabase first. To recap, making the init explicit was necessary to enable the default value for the version which old pre-code-change databases with no saved version number would be loaded with, to be different from the default for currently created instances. This then also required moving mapping and frozen into init. I believe this way of dealing with the mapping default works and does not run into the potential bugs of mutable defaults (which as I understand it the previous field construction was trying to guard against). I have checked that initialising two DetectorDatabases and adding mapping content to one does not affect the other, whether the second initialisation is done before or after the adding.

nelimee: I will have a look later today :)

src/tqec/compile/tree/tree.py
```diff
@@ -317,15 +318,40 @@ def generate_circuit(
...
             detector_database = None
+        if detector_database is not None:
+            loaded_version = detector_database.version
+            if loaded_version != current_version:  # type:ignore
+                if user_defined:
+                    raise TQECException(
+                        f"The detector database on disk you have specified is incompatible with the version"
+                        f"in the TQEC code you are running. The version of the disk database is {loaded_version}"
+                        f", while the version in the TQEC code is {current_version}"
+                    )
```
nelimee:
```suggestion
                    raise TQECException(
                        "The detector database on disk you have specified is incompatible with "
                        "the version in the TQEC code you are running. The version of the disk "
                        f"database is {loaded_version}, while the version in the TQEC code is "
                        f"{current_version}"
                    )
```

src/tqec/compile/tree/tree.py
```diff
@@ -317,15 +318,40 @@ def generate_circuit(
...
+                else:  # ie using the default
+                    warnings.warn(
+                        f"The default detector database that you have saved on your system is out of date"
+                        f"(version {loaded_version}). The version in the TQEC code you are running is newer"
+                        f"(version {current_version}). The database will be regenerated.",
+                        TQECWarning,
```
nelimee:
```suggestion
                        "The default detector database that you have saved on your system is out "
                        f"of date (version {loaded_version}). The version in the TQEC code you are "
                        f"running is newer (version {current_version}). The database will be regenerated.",
                        TQECWarning,
```

nelimee (commented):
The current code seems correct. 

Just remove the `None` possibility in the database version, and I be able to accept and merge that PR after testing one last time.

nelimee:
Disregard my previous message, I reviewed the wrong commit and did not pull your changes. Doing the review again :)

src/tqec/compile/detectors/database.py
```diff
@@ -197,12 +201,38 @@ class DetectorDatabase:
...
+        self,
+        mapping: dict[_DetectorDatabaseKey, frozenset[Detector]] | None = None,
+        frozen: bool = False,
+    ):
+        self.version = CURRENT_DATABASE_VERSION
+        if mapping is None:
+            self.mapping = dict()
+        else:
+            self.mapping = mapping
+        self.frozen = frozen
```
nelimee: I find the following easier to read:

```suggestion
        if mapping is None:
            mapping = dict()
        self.mapping = mapping
        self.frozen = frozen
        self.version = CURRENT_DATABASE_VERSION
```

src/tqec/compile/detectors/database.py
```diff
@@ -197,12 +201,38 @@ class DetectorDatabase:
...
-        default_factory=dict
-    )
-    frozen: bool = False
+    version: semver.Version | None = semver.Version(0, 0, 0)
+
+    def __init__(
+        self,
+        mapping: dict[_DetectorDatabaseKey, frozenset[Detector]] | None = None,
+        frozen: bool = False,
+    ):
```
nelimee, line 222: I understand now the need for a separate `__init__`. I am unsure how that works out with `@dataclass` that is supposed to generate its own `__init__`. The best would be to remove the `@dataclass` decorator for clarity.

src/tqec/compile/tree/tree.py
```diff
@@ -317,15 +318,40 @@ def generate_circuit(
         # First, before we start any computations, decide which detector database to use.
         if isinstance(database_path, str):
             database_path = Path(database_path)
+        # Start user_defined from False, override with True later if certain conditions are met:
+        user_defined = False
```
nelimee: I prefer to set that explicitly at initialisation, I find that easier to understand:
```suggestion
        # We need to know for later if the user explicitly provided a database or
        # not to decide if we should warn or raise.
        user_defined = (
            detector_database is not None
            or database_path != DEFAULT_DETECTOR_DATABASE_PATH
        )
```

src/tqec/compile/tree/tree.py
```diff
@@ -317,15 +318,40 @@ def generate_circuit(
...
         # If the user has passed a database in, use that, otherwise:
         if detector_database is None:  # Nothing passed in,
             if database_path.exists():  # look for an existing database at the path.
                 detector_database = DetectorDatabase.from_file(database_path)
             else:  # if there is no existing database, create one.
                 detector_database = DetectorDatabase()
+        else:
+            user_defined = True
+        if database_path != DEFAULT_DETECTOR_DATABASE_PATH:
+            user_defined = True
```
nelimee: If set at initialisation, there is no need for these branches anymore.

src/tqec/compile/detectors/database.py
BSchelpe: good point! Will do

src/tqec/compile/tree/tree.py
BSchelpe: much clearer, thank you!

BSchelpe:
@nelimee I think this is all done now - your suggestions are incorporated and I have removed the None option. I think this is mergeable once you're happy with it. Thank you!

PR #590, opened 2025-05-21
Test ruff failures

purva-thakre:


src/tqec/__init__.py
```diff
@@ -1,5 +1,5 @@
-from ._version import __version__ as __version__
-from .compile import compile_block_graph as compile_block_graph
-from .computation.block_graph import BlockGraph as BlockGraph
-from .utils import Basis as Basis
-from .utils import NoiseModel as NoiseModel
```
purva-thakre, line 5: @nelimee Is there a particular reason why the project utilizes this type of import pattern?

purva-thakre:
Looks like we might have to temporarily disable `pre-commit` to allow adding `ruff` in separate PRs.

Issue #579, opened 2025-05-08
Performance improvements: Task Checklist


nelimee:
This is an evolving checklist of the different performance improvement opportunities in TQEC and the respective issues tracking it.

- [ ] #494 
- [ ] #482

Zhaoyilunnn:
https://github.com/tqec/tqec/issues/696 Maybe we should also add this issue in this checklist?

SMC17:
Contributes to #579 via topologiq PR tqec/topologiq#17

This work adds performance infrastructure to the topologiq component of the TQEC ecosystem:

**Performance Improvements:**
- **1.78x measured speedup** with A* pathfinding algorithm
- Statistically validated on complex quantum circuits
- Environment-based algorithm switching for research

**Performance Research Infrastructure:**
- Comprehensive benchmark framework with parametrized tests
- Statistical validation patterns (mean speedup + significance testing)
- Custom pytest markers for performance tests
- Reproducible benchmarks via forced parameters

**Research Enablement:**
- Algorithm comparison framework (BFS vs A*)
- Exploration tracking for algorithmic analysis
- Performance metrics collection patterns

These tools can serve as a model for performance optimization efforts in other TQEC components. The benchmark framework demonstrates how to:
1. Measure performance improvements rigorously
2. Validate statistical significance
3. Enable reproducible performance research
4. Integrate performance tests into CI/CD

Happy to discuss how these patterns might apply to other TQEC performance work!

purva-thakre:
making a note that we need to look into why the unit tests in `src/tqec/compile/compile_test.py` are extremely slow. If benchmarking tqec can help us find the bottleneck here, it will be great to further improve the performance of the tool. 

https://github.com/tqec/tqec/discussions/698

smburdick:
Can we add #823 as a subtask

PR #592, opened 2025-05-30
Executable examples and tutorials

purva-thakre:
To do:

- [x]  Ensure all gallery examples show plots
- [x] Add references section to the examples that use `refs.bib`
- [x] Add cache options to reduce docs build time: https://github.com/actions/cache?tab=readme-ov-file#restoring-and-saving-cache-using-a-single-action
- [x] Add `jupyter-sphinx` for tutorials
- [ ] Add an issue on how to update the user guide tutorials and gallery examples to the contributing guide

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -36,9 +36,10 @@ jobs:
         id: cache-docs
         with:
           path: docs/_examples_database/
-          key: ${{ runner.os }}-docs_cache
+          key: docs_database_and_stim_results_cache
+          enableCrossOsArchive: true
       - name: Generate the documentation
-        if: steps.${{ runner.os }}-docs_cache.outputs.cache-hit != 'true'
+        if: steps.docs_database_and_stim_results_cache.outputs.cache-hit != 'true'
         run: |
           make html
         working-directory: docs/


docs/gallery/cnot.ipynb
```diff
@@ -1384,15 +1384,83 @@
...
-    "</details>\n",
+    "</details>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "\"hide-input\""
```
purva-thakre, line 1395: To do - this only works locally. Need to find a way to hide the code block in `nbsphinx` while also executing it. 

image

The code block above does not execute the code

docs/gallery/cnot.ipynb
```diff
@@ -1401,13 +1469,13 @@
...
     "\n",
-    "- Horsman, D., Fowler, A. G., Devitt, S., & Van Meter, R. (2012). Surface code quantum computing by lattice surgery. New Journal of Physics, 14(12), 123011."
+    ".. footbibliography::"
    ]
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "tqec312",
```
purva-thakre, line 1478: Todo: revert this change. It is due to my local environment.

.github/workflows/gh-pages.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs/_build/html/
           destination_dir: ./pull/${{github.event.number}}/
-        continue-on-error: true
+        # continue-on-error: true
```
purva-thakre, line 49:
```suggestion
        continue-on-error: true
```
undo this

purva-thakre:
I still can't find the source for why the [code block in the examples gallery is not executed](https://tqec.github.io/tqec/pull/592/gallery/cnot.html#Example-Circuit), whereas the [`.rst` test is executed successfully.](https://tqec.github.io/tqec/pull/592/user_guide/cnot_test.html#example-circuit) 

I cannot tell if [`nbgallery`](https://github.com/nbgallery/nbgallery), `nbsphinx`, or `myset_parser` is the culprit. The latter two are added in `conf.py`, and the gallery page is built with the former. 

https://github.com/tqec/tqec/blob/8f2f79adf02cc3b71fae1849091e824311846f6a/docs/conf.py#L45-L50

https://github.com/tqec/tqec/blob/8f2f79adf02cc3b71fae1849091e824311846f6a/docs/gallery/index.rst?plain=1#L3-L8

nelimee:
> I still can't find the source for why the [code block in the examples gallery is not executed](https://tqec.github.io/tqec/pull/592/gallery/cnot.html#Example-Circuit), whereas the [`.rst` test is executed successfully.](https://tqec.github.io/tqec/pull/592/user_guide/cnot_test.html#example-circuit)

Isn't that because the notebook has some cells that are executed? The rule is the following:

- if a notebook has no executed cell, it is run by Sphinx,
- else, if at least one cell is marked as executed in the notebook, it is left as-is.

purva-thakre:
> else, if at least one cell is marked as executed in the notebook, it is left as-is.

Ooh. Thanks for this! I thought if at least one cell is unexecuted, then that one gets executed, but not the others that have an output.

docs/gallery/cnot.ipynb
```diff
@@ -1367,32 +203,25 @@
...
+    "        ax.set_ylabel(\"Logical Error Rate\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "generate_graphs(Basis.Z)"
```
purva-thakre, line 318: To do: Hide the pymatching progress output from stim

https://tqec.github.io/tqec/pull/592/gallery/cnot.html#Example-Circuit

purva-thakre:
**Total docs build time with executable tutorials and gallery examples (without cache)** = 2h 21m 53s 

Worst build times are shown in the screenshot below:

image

I guess once the caching action is added, we should expect the first workflow run to take longer. Subsequent runs shouldn't as they would be relying on the cached data.

purva-thakre:
To do: Build warnings

image

Edit: Done ✅

.github/workflows/gh-pages.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt -r requirements-dev.txt
           python -m pip install '.[all]'
+      - uses: actions/cache@v4
+        with:
+          path: docs/_build
+          key: ${{ runner.os }}-sphinx
```
purva-thakre: At the first run, no cache was found. But according to [this page](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#using-the-cache-action), a new cache should be created if the job runs successfully. 

> On a cache miss, the action automatically creates a new cache if the job completes successfully

image

purva-thakre: A cache has been added:

image

Now, the question is how to use this to reduce the docs build time. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#using-the-output-of-the-cache-action

https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d

.github/workflows/gh-pages.yml
```diff
@@ -29,10 +29,12 @@ jobs:
...
+        id: cache-docs
         with:
           path: docs/_build
           key: ${{ runner.os }}-sphinx
-      - name: Generate the documentation
+      - if: ${{ steps.cache-docs.outputs.cache-hit != 'true' }}
+        name: Generate the documentation
         run: |
           cd docs/
           make html
```
purva-thakre: This is not working as intended. These two were skipped in the [latest run](https://github.com/tqec/tqec/actions/runs/15452216197/job/43496802675?pr=592)

image

.github/workflows/gh-pages.yml
```diff
@@ -28,7 +28,13 @@ jobs:
...
+      - uses: actions/cache@v4
+        id: cache-docs
+        with:
+          path: docs/_build
+          key: ${{ runner.os }}-sphinx
       - name: Generate the documentation
+        if: steps.${{ runner.os }}-sphinx.outputs.cache-hit != 'true'
         run: |
           cd docs/
           make html
```
purva-thakre: @nelimee Do you mind taking a quick look at the changes? [The cache was added successfully](https://github.com/tqec/tqec/actions/caches), but I can't seem to make sphinx use the cache. The latest build is about to [take > 2 hours](https://github.com/tqec/tqec/actions/runs/15452845946/job/43498873970?pr=592). 

Edit: Maybe it worked? I made an empty commit, and it took approximately [2 minutes](https://github.com/tqec/tqec/actions/runs/15453840775/job/43501988662). Perhaps the build was taking longer earlier because I had made changes to the workflow itself. 

image

Should we expect the docs build to take longer if we add a new example/tutorial or make changes to the existing documents? I made changes to the metadata of a jupyter notebook in https://github.com/tqec/tqec/pull/592/commits/b141ffa8fecbb60081fc57ce6bc528e36fa207ad, but it still took approx 1 min.

docs/gallery/cnot.ipynb
```diff
@@ -1401,13 +330,13 @@
...
     "\n",
-    "- Horsman, D., Fowler, A. G., Devitt, S., & Van Meter, R. (2012). Surface code quantum computing by lattice surgery. New Journal of Physics, 14(12), 123011."
+    ".. footbibliography::"
    ]
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "tqec312",
```
purva-thakre:
```suggestion
   "display_name": "Python 3 (ipykernel)",
```

.github/workflows/gh-pages.yml
nelimee: From what I see in the CI action, you are caching the `docs` folder, so the documentation will never be updated and will only be retrieved from cache. So yes, the docs should take longer to build. Probably around 3 to 5 minutes with a database.

What you should do (in my opinion) is the following:

1. Use a database with a known path (setting `database_path` when generating quantum circuits),
2. Cache **only** the database with this known path.
3. You will likely need [restore/save actions](https://github.com/actions/cache?tab=readme-ov-file#using-a-combination-of-restore-and-save-actions) rather than a single cache action.

purva-thakre: Just to double-check, are you referring to this? Specifically, to the `to_file` and `from_file` methods. 

https://github.com/tqec/tqec/blob/d6f90192269b96651a62f5957b193308ff1bc2c5/src/tqec/compile/detectors/database.py#L191

nelimee: Yes, but you should not have to dig that much. Just setting `database_path` in 
https://github.com/tqec/tqec/blob/d6f90192269b96651a62f5957b193308ff1bc2c5/src/tqec/compile/graph.py#L453-L461
should be enough to generate the database there.

purva-thakre: But since the gallery simulation examples rely on `start_simulation_using_sinter`, `database_path` from the function you linked is the same as `save_resume_filepath`, right?

https://github.com/tqec/tqec/blob/fefc16849f6e773cea5f666da32b3cce32177f0f/src/tqec/simulation/simulation.py#L36

This function does not have a `database_path` option. 

Edit: I do not think they are the same. I just got some results from an example computation for `database_path`. Looks like we need to add an option for the database path in `start_simulation_using_sinter`.

BTW What format should the `database_path` file supposed to be?

docs/user_guide/detectors/detector_computation.ipynb
```diff
@@ -24,213 +24,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
```
purva-thakre, line 27: To do: This tutorial is out of date. 

image

image

image

image

purva-thakre: @nelimee Would it be preferable to move this tutorial to the `tqecd` repo? In https://github.com/tqec/tqec/pull/592/commits/90f253087bbebe568ece72aa75dcd257a0e1450d, I installed `tqecd` in the jupyter notebook. It ran locally without any issues. 

I noticed there isn't an examples gallery in the `tqecd` documentation. 

Edit: it looks like the same tutorial exists in the tqecd docs. So, do we want to keep this in the tqec docs too? https://tqec.github.io/tqecd/media/detectors/detector_finding_illustration.html

.github/workflows/gh-pages.yml
purva-thakre: Noticed an interesting issue with the plots, when I use `save_resume_filepath`, the same X-axis data point is plotted twice. 

Compare the [Z basis plots](https://tqec.github.io/tqec/pull/592/gallery/cnot.html#Z-Basis) for the CNOT gate to those of the [X basis plots](https://tqec.github.io/tqec/pull/592/gallery/cnot.html#X-Basis). I think the latter is using the saved data while the former is used to save the data. 

| Z-Basis      | X-Basis |
| ----------- | ----------- |
| image| image|

docs/user_guide/detectors/detector_computation.ipynb
nelimee: It can be removed from `tqec` docs, I agree.

.github/workflows/gh-pages.yml
nelimee: `database_path` is a path to a file that can have any extension (but I prefer `.pkl` as it is saved using `pickle`) and that is used at circuit generation time, **not** at simulation time. 

There are two steps there:

1. Generate the quantum circuit using `tqec`, that's where the `database_path` should matter,
2. Simulate the resulting `.stim` quantum circuit using `stim`, that's where the `save_resume_filepath` should matter.

In theory, `save_resume_filepath` saves results along with a hash of the quantum circuit that has been used to generate those results, so that is strange that both points are plotted. I am suspecting that it is rather an issue with how we plot, I will have a quick look.

nelimee: Seems like the code of `generate_graph` has not been updated in [the documentation](https://tqec.github.io/tqec/pull/592/gallery/cnot.html#Z-Basis): 
image

nelimee: I forgot how `start_simulation_using_sinter` was working, and in particular that it was taking a `BlockGraph` in. Then yes, you are right, the database_path argument should be added.

nelimee: I cannot replicate that locally: I get an error from `sinter` when trying to execute the cell. Do you have it too?

purva-thakre: > Seems like the code of generate_graph has not been updated in [the documentation](https://tqec.github.io/tqec/pull/592/gallery/cnot.html#Z-Basis)

It is not for the code block that a user will have to collapse. There is a hidden code block right below this to get the results. Scroll to the bottom https://github.com/tqec/tqec/blob/e505dc0ab358b2660162e5e8a9a7d7d52d817fd6/docs/gallery/cnot.ipynb

nelimee: Yes, I have seen the executed code block, hence my "not updated" comment. 

Will have a deeper look on Monday.

purva-thakre: Ah, gotcha. I was a bit confused when I read that line and the next comment, where you get an error from sinter. I do not get any error locally. I see the graphs as described in the table in my previous comment. 

I plan to update the code block the user will have to collapse after I have figured out how to reduce the docs build time etc.

purva-thakre:
@nelimee Do I need to unpack the saved database as a `DetectorDatabase`, or `only_use_database` is supposed to do this?

.github/workflows/gh-pages.yml
nelimee: Yes I do have an error when executing the notebook, but that is using a fresh environment, so brand new versions for everything, that might be the issue.

purva-thakre:
@nelimee Realistically, how much time should the simulation take with the `database_path` option? I tried this locally, but I do not see a major change. The database is saved with the first run for the 3 CNOTs simulation example. It utilizes & saves it to the path I provided. I expected the second run to use the database from the first run... 

I think `save_resume_filepath` works much better than `database_path` in terms of the time taken by the simulation, based on what I have noticed locally. 

I added the option directly in `start_simulation_using_sinter` as it relies on `generate_sinter_tasks`, which has the `database_path` option. You can see the changes in https://github.com/tqec/tqec/pull/592/commits/c7795ce963a3e1787ade16f4fcb00ed9aa486a82 

https://github.com/tqec/tqec/blob/16f26d633969935e3de8123b6f191680a9e1c57e/src/tqec/simulation/generation.py#L104

.github/workflows/gh-pages.yml
```diff
@@ -32,7 +32,13 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt -r requirements-dev.txt
           python -m pip install '.[all]'
+      - uses: actions/cache@v4
+        id: cache-docs
+        with:
+          path: docs/_examples_database/
+          key: ${{ runner.os }}-sphinx
```
purva-thakre: to do: need to add `restore` and `save` actions 

https://github.com/actions/cache?tab=readme-ov-file#using-a-combination-of-restore-and-save-actions

docs/gallery/steane_encoding.ipynb
```diff
@@ -883,22 +179,103 @@
...
+    "#### X Basis"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "generate_graphs(Basis.X)"
```
purva-thakre, line 287: This code block led to a build error. https://tqec.github.io/tqec/pull/592/gallery/steane_encoding.html#X-Basis

I think this might be related to the wonky X basis graphs for the other examples...

nelimee:
> @nelimee Realistically, how much time should the simulation take with the `database_path` option? I tried this locally, but I do not see a major change. The database is saved with the first run for the 3 CNOTs simulation example. It utilizes & saves it to the path I provided. I expected the second run to use the database from the first run...
> 
> I think `save_resume_filepath` works much better than `database_path` in terms of the time taken by the simulation, based on what I have noticed locally.
> 
> I added the option directly in `start_simulation_using_sinter` as it relies on `generate_sinter_tasks`, which has the `database_path` option. You can see the changes in [c7795ce](https://github.com/tqec/tqec/commit/c7795ce963a3e1787ade16f4fcb00ed9aa486a82)
> 
> https://github.com/tqec/tqec/blob/16f26d633969935e3de8123b6f191680a9e1c57e/src/tqec/simulation/generation.py#L104

So in theory, both options can be used together. The `database_path` will be used to speed-up circuit generation, and the `save_resume_filepath` for circuit simulation.

With both options set, and considering that GitHub runners are a little under-powered, I think that we should be able to get everything in less than 2 minutes of computation. But that is just an estimation. 5 minutes would be acceptable. And if its more, we can start looking at the CI, for example to only generate the documentation when explicitly asked for. 

If you want, I can have a look at the performance of CI on monday. 

As a guideline, on my computer, executing the CNOT notebook for the first time (with database and stats not saved) is done in ~17 minutes (mostly due to simulation) and the plots already look abnormal (meaning that we will have to fix this eventually). The second run (so with database and stats already saved) takes ~17 seconds for the Z-basis run, but strangely re-do some simulations for the X-basis run. The Z-basis run is approximately what I would expect, so let's say from 1 to 2 minutes to execute the whole CNOT notebook seems reasonable.

purva-thakre:
> The second run (so with database and stats already saved) takes ~17 seconds for the Z-basis run, but **strangely re-do some simulations for the X-basis run.**

Yes, I have noticed this too.

purva-thakre:
Just tested an empty commit. Managed to shave off 40 mins from the build time. Yesterday, the build took 2h and 19 mins to save the stats and database. Today, the build took ~> 45 mins so far~ 1h 39m 22s . Something else during the simulation process is taking more time. Perhaps related to #579?

Most of the time is/was taken up by the Steane encoding example. Maybe, we could keep this jupyter notebook with executed outputs to reduce the build time. Without it, the build should take approximately 20 - 30 mins. 

If we want to further reduce the time, I think it might be better to cache the entire docs build. It probably works similarly to `make html` locally, where only the new example is executed, and the rest remains the same. We could add a workflow that automates the process of cleaning & restoring the docs cache on a regular basis. This is similar to Adrien's idea of only running the docs when we need to.

nelimee:
I'm having a look at this PR.

First, the issue about multiple points being plotted. That's normal, because we just "dumbly" iterate over all the statistics gathered, without filtering. The best solution in my opinion would be to have a statistic file for each run. For example:

```py

def generate_graphs(support_observable_basis: Basis) -> None:
    block_graph = cnot(support_observable_basis)
    zx_graph = block_graph.to_zx_graph()

    correlation_surfaces = block_graph.find_correlation_surfaces()

    stats = start_simulation_using_sinter(
        block_graph,
        range(1, 4),
        list(numpy.logspace(-4, -1, 10)),
        NoiseModel.uniform_depolarizing,
        manhattan_radius=2,
        num_workers=25,
        max_shots=1_000_000,
        max_errors=5_000,
        decoders=["pymatching"],
        # CHANGES BELOW
        save_resume_filepath=Path(f"../_examples_database/cnot_stats_{basis.value}.csv"),
        database_path=Path("../_examples_database/database.pkl"),
    )

    for i, stat in enumerate(stats):
        fig, ax = plt.subplots()
        sinter.plot_error_rate(
            ax=ax,
            stats=stat,
            x_func=lambda stat: stat.json_metadata["p"],
            group_func=lambda stat: stat.json_metadata["d"],
        )
        plot_observable_as_inset(ax, zx_graph, correlation_surfaces[i])
        ax.grid(axis="both")
        ax.legend()
        ax.loglog()
        ax.set_title("Logical CNOT Error Rate")
        ax.set_xlabel("Physical Error Rate")
        ax.set_ylabel("Logical Error Rate")
```

The above seem to also fix the issue with `X`-basis statistics being re-computed.

Then, for performance purposes, it is better to not create too much processes. The number of workers `num_workers` should be rather set to `multiprocessing.cpu_count` rather than a fixed `20` that is way too much for a GitHub runner.

Then, the `DetectorDatabase` instance should be unconditionally valid. There is no need to use different databases for different computations, you can use the same database for each one.

Finally, it might be worth reducing the number of shots we take in the documentation. Maybe using a million shots is enough?

For reference, on my computer (32 cores, so when 20 workers are used, that means that all workers are executing in parallel):

- 10M shots, 20 cores, 1st run: 20m45s
- 10M shots, 20 cores, 2nd run: 13m49s (for a reason I ignore, steane and 3-cnot re-did the `stim` simulations)

purva-thakre:
> First, the issue about multiple points being plotted. That's normal, because we just "dumbly" iterate over all the statistics gathered, without filtering.

Thanks for this. I figured the issue was some issue of 'dumbly' iterating. But I thought it would be with `start_simulation_using_sinter` or `plot_observable_as_inset` and not about filtering the stats. 

> The above seem to also fix the issue with X-basis statistics being re-computed.

Why does filtering fix this issue? I couldn't understand why the stats were recomputed for the second basis except maybe the file was being replaced with the stats of the second basis?

purva-thakre:
> Then, for performance purposes, it is better to not create too much processes. The number of workers num_workers should be rather set to multiprocessing.cpu_count rather than a fixed 20 that is way too much for a GitHub runner.

Ok, I agree with this. cpu_count was a great suggestion!

> Then, the DetectorDatabase instance should be unconditionally valid. There is no need to use different databases for different computations, you can use the same database for each one.

I am confused. Shouldn't each example have a different detector database?

nelimee:
> Why does filtering fix this issue? I couldn't understand why the stats were recomputed for the second basis except maybe the file was being replaced with the stats of the second basis?

For the moment, I do not understand either. But the behaviour seems fine now. 

> I am confused. Shouldn't each example have a different detector database? 

Not really. The detector database contains pieces of computations (subtemplates + plaquettes) that are always valid, whatever the computation they come from. So yes, the same database can be used for any computation, that is one of the strength of that approach :)

purva-thakre:
> > Why does filtering fix this issue? I couldn't understand why the stats were recomputed for the second basis except maybe the file was being replaced with the stats of the second basis?

> For the moment, I do not understand either. But the behaviour seems fine now.

I think this is probably because the stats file for the Z basis was being replaced & regenerated for the X basis. With the changes you made, two separate files get added for each example. 

image

purva-thakre:
The graphs still have a few weird data points. 

https://tqec.github.io/tqec/pull/592/gallery/steane_encoding.html#Z-Basis

image

nelimee:
> The graphs still have a few weird data points.

Last possibility I see: `tqec` does not generate deterministic circuits, and so the two/more points at the same physical error are due to 2 different circuits being executed. I will investigate that, as well as try to optimise as much as possible the documentation runs.

nelimee:
> > The graphs still have a few weird data points.
> 
> Last possibility I see: `tqec` does not generate deterministic circuits, and so the two/more points at the same physical error are due to 2 different circuits being executed. I will investigate that, as well as try to optimise as much as possible the documentation runs.

The code 
```py
from pathlib import Path

from tqec import compile_block_graph
from tqec.compile.convention import FIXED_BULK_CONVENTION
from tqec.compile.graph import TopologicalComputationGraph
from tqec.gallery import cnot
from tqec.utils.noise_model import NoiseModel

ASSETS_FOLDER = Path(__file__).parent / "assets"

k = 2
block_graph = cnot(Basis.Z)
correlation_surfaces = block_graph.find_correlation_surfaces()
different_instances: dict[int, TopologicalComputationGraph] = {}
for i in range(10):
    compiled_computation = compile_block_graph(
        block_graph, FIXED_BULK_CONVENTION, observables=[correlation_surfaces[0]]
    )
    circuit = compiled_computation.generate_stim_circuit(k, NoiseModel.uniform_depolarizing(0.001))
    print(hash(str(circuit)))
```
that outputs
```
665142247272478203
-1565201332388988843
665142247272478203
-1565201332388988843
665142247272478203
-1565201332388988843
665142247272478203
-1565201332388988843
665142247272478203
-1565201332388988843
```
on my computer (hashes change at each new run because of how Python's `hash` works, but should be reliable on a single run, so the important factor here is that the hashes change within a single run, you will have different hashes when running that circuit locally) seems to support my claim. This does not seem to be the case on `move_rotation` though. Digging further.

EDIT: the issue seems to come from the fact that we forgot to sort detector targets. In the CNOT example above, line 788 is alternating between
```
DETECTOR(6, 12, 0) rec[-132] rec[-125] rec[-120] rec[-49]
```
and 
```
DETECTOR(6, 12, 0) rec[-132] rec[-49] rec[-120] rec[-125]
```
which represent the same computation, but do not have the same hashes.

EDIT2: I'll push the modification that should fix that, and test again performances.

nelimee:
A few timings (32 cores processor, 1,000,000 shots), by keeping the database (e.g., the first run of memory uses the database generated from the CNOT):

| Example        | Saved results? | Time |
| --------------- | -------------- | ----- |
| CNOT                           | no                           | 1m15s      |
| CNOT                           | yes                          | 20s        |
| Memory                         | no                           | <10s       |
| Memory                         | yes                          | <5s        |
| Move rotation                  | no                           | 45s        |
| Move rotation                  | yes                          | <15s       |
| Steane encoding                | no                           | 5m15s      |
| Steane encoding                | yes                          | <1m        |
| Three cnots                    | no                           | 2m45s      |
| Three cnots                    | yes                          | <45s       |

An important note about these timings: the first run (with no saved results) is particularly sensitive to the compute power available (due to simulations using `stim`), and so my timings are not representative of the CI timings, whereas the second run is mostly sequential and should be quite representative of CI timings (with a >1 pre-factor as I expect CI CPUs to not be able to increase their frequency as much as my processor).

That means that, with the current code, we will likely need more than 2m30s to re-generate the documentation, provided that the generated circuits did not change.

I'll try this out on CI and report the timings in another comment.

nelimee:
According to the output of the last two actions that ran and their runtime, the cache system does not seem to work correctly. 

I think the restore/backup scheme is needed here:

- if there is a cache, restore it, else do nothing,
- independently of the first point, unconditionally save the content of the directory containing simulation results and the database to the cache.

The only issue I see with this approach is that out-dated simulation results will never be removed from the `.csv` files. That is, in my opinion, a very minor problem because these files are relatively small, and if we see that their size is too large we can still remove the cache to force a fresh-regeneration of the simulation results, effectively removing by hand all out-of-date results.

purva-thakre:
> I think the restore/backup scheme is needed here

Yes, I have been procrastinating on adding this, as I still don't know how the cache is restored with the `restore` action. Working on it today and tomorrow.

purva-thakre:
@nelimee Sorry, it took me a while, but I still don't understand how the docs build time is not reduced when there is a saved cache. The following correctly saved a cache, but it is not utilized by `make html` to reduce the build time? 

```
      - uses: actions/cache@v4
        id: cache-docs
        with:
          path: docs/_examples_database/
          key: ${{ runner.os }}-docs_cache
      - name: Generate the documentation
        if: steps.${{ runner.os }}-docs_cache.outputs.cache-hit != 'true'
        run: |
          cd docs/
          make html
```

nelimee:
> @nelimee Sorry, it took me a while

No problem at all, do not worry.

> but I still don't understand how the docs build time is not reduced when there is a saved cache. The following correctly saved a cache, but it is not utilized by `make html` to reduce the build time?
> 
> ```
>       - uses: actions/cache@v4
>         id: cache-docs
>         with:
>           path: docs/_examples_database/
>           key: ${{ runner.os }}-docs_cache
>       - name: Generate the documentation
>         if: steps.${{ runner.os }}-docs_cache.outputs.cache-hit != 'true'
>         run: |
>           cd docs/
>           make html
> ```

From what I understood (which might be inaccurate) from the [cache action documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#using-the-cache-action), and in particular from the below paragraph:

> If there is an exact match to the provided key, this is considered a cache hit. If no cache exactly matches the provided key, this is considered a cache miss. On a cache miss, the action automatically creates a new cache if the job completes successfully. The new cache will use the key you provided and contains the files you specify in path. For more information about how this is handled, see [Cache hits and misses](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#cache-hits-and-misses).
> 
> You cannot change the contents of an existing cache. Instead, you can create a new cache with a new key.

- On a cache hit, the only thing that is happening is "load the existing cache".
- On a cache miss, a new cache is created, and saved only if the action is successful.

In our case, the cache already exist, so this is a cache hit. That means that it is loaded, used, but never updated. 

What we want is:

- at the beginning of the action, load the cache if it exists, else do nothing.
- at the end of the action, unconditionally save the cache back, overriding the potentially existing entry.

That seems to be doable with [restore](https://github.com/actions/cache/blob/main/restore/README.md)/[save](https://github.com/actions/cache/blob/main/save/README.md) actions:

- Start by using a restore action to get the cache (if it exists),
- End by using a save action only if the workflow succeeded (i.e., do **not** use the [always](https://github.com/actions/cache/blob/main/save/README.md#always-save-cache) keyword) because we do not want to override an existing correct cache with a faulty one.

Does that helps?

nelimee:
PS: you might want to disable every example but one in the gallery to speed-up CI for testing purpose. For example, only keep memory and disable everything else.

purva-thakre:
> PS: you might want to disable every example but one in the gallery to speed-up CI for testing purpose. For example, only keep memory and disable everything else.

no worries. I used the option of `working-directory` instead of changing the directory to docs in https://github.com/tqec/tqec/pull/592/commits/09b93a7ddfc2ecee6c80a340cd14bbf96fe3d824, and it looks like the cache was used!! 💃🏾 

https://github.com/tqec/tqec/actions/runs/15854860172/job/44697484696?pr=592#step:6:78

Total build time with cache: approx 8 min

Will spend some time trying to understand if we need `restore` and `save` as well.

nelimee:
> > PS: you might want to disable every example but one in the gallery to speed-up CI for testing purpose. For example, only keep memory and disable everything else.
> 
> no worries. I used the option of `working-directory` instead of changing the directory to docs in [09b93a7](https://github.com/tqec/tqec/commit/09b93a7ddfc2ecee6c80a340cd14bbf96fe3d824), and it looks like the cache was used!! 💃🏾
> 
> https://github.com/tqec/tqec/actions/runs/15854860172/job/44697484696?pr=592#step:6:78
> 
> Total build time with cache: approx 8 min
> 
> Will spend some time trying to understand if we need `restore` and `save` as well.

If the cache works now, I guess I misunderstood the documentation of action/cache. No need to change anything if it works. 

We should check that it works by removing the cache, run the action once to check that the cache is populated correctly, and re-run it a second time to check that it works.

I think #624 might be needed to avoid surprises with the detector database in CI. For the moment that's not blocking, but we should have a check for #624.

purva-thakre:
image

> We should check that it works by removing the cache, run the action once to check that the cache is populated correctly, and re-run it a second time to check that it works.

Good point. When the cache is not found, the action should run the docs build and then save the cache. The build is going to take its time in https://github.com/tqec/tqec/pull/592/commits/1c74e36c1fc2663471493aacee767de9e124e16d because I deleted the cache in `Actions`.

purva-thakre:
It looks like the `cache` action also ends up caching what we want after the cache was deleted and we ran the docs build again. 

image

Triggering an empty commit to check.

nelimee:
> It looks like the `cache` action also ends up caching what we want after the cache was deleted and we ran the docs build again.
> 
> image
> 
> Triggering an empty commit to check.

I just triggered the action again, no need for a commit :)

purva-thakre:
> I just triggered the action again, no need for a commit :)

Sorry, just saw this. :(

nelimee:
> > I just triggered the action again, no need for a commit :)
> 
> Sorry, just saw this. :(

Seems like it is working! Let's wait for CI to finish :)

.github/workflows/gh-pages.yml
```diff
@@ -32,10 +32,16 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt -r requirements-dev.txt
           python -m pip install '.[all]'
+      - uses: actions/cache@v4
+        id: cache-docs
+        with:
+          path: docs/_examples_database/
+          key: ${{ runner.os }}-docs_cache
```
nelimee: Why include the OS in the cache key? Is there a specific reason for that?

.github/workflows/gh-pages.yml
```diff
@@ -32,10 +32,16 @@ jobs:
...
       - name: Generate the documentation
+        if: steps.${{ runner.os }}-docs_cache.outputs.cache-hit != 'true'
         run: |
-          cd docs/
           make html
+        working-directory: docs/
```
nelimee, line 45: Could you add some comments to explain why this was needed?

docs/Makefile
```diff
@@ -15,6 +15,7 @@ help:
 
 clean:
 	rm -rfv "$(SOURCEDIR)/_autosummary"
+	rm -rfv "$(SOURCEDIR)/_examples_database"
```
nelimee: Maybe, because this is costly to re-generate, add a new `cleanall` target that performs `clean` + that line?

purva-thakre:
> Seems like it is working! Let's wait for CI to finish :)

Ok. I still have to make some changes to the code blocks a user has to collapse https://github.com/tqec/tqec/pull/592#discussion_r2142105321. Will mark this PR ready for review and request a review from you after I am done.

docs/conf.py
```diff
@@ -171,6 +176,8 @@ def setup(app):
     "gallery/steane_encoding": "_static/media/gallery/steane_encoding.png",
 }
 
+nbsphinx_allow_errors = True
```
nelimee: Why do you need that? If there is an error building the docs, we want the CI to fail.

docs/user_guide/logical_cnot.dae
nelimee, line 1: Duplicated file with `docs/logical_cnot.dae`?

docs/user_guide/quick_start.rst
```diff
@@ -139,8 +142,9 @@ The compilation of the block graph is done automatically based on the inputs.
         decoders=["pymatching"],
         num_workers=cpu_count(),
         max_shots=10_000_000,
-        max_errors=5_000,
-        print_progress=True,
+        max_errors=5_00,
```
nelimee: Missing a `0` here. Or remove the `_`: `500`.

docs/user_guide/quick_start.rst
```diff
@@ -139,8 +142,9 @@ The compilation of the block graph is done automatically based on the inputs.
...
+        save_resume_filepath=Path(f"./_examples_database/quick_start.csv"),
```
nelimee: No need for a f-string here.
```suggestion
        save_resume_filepath=Path("./_examples_database/quick_start.csv"),
```

docs/user_guide/quick_start.rst
nelimee:
```suggestion
        max_errors=500,
```

docs/user_guide/quick_start.rst
```diff
@@ -188,7 +192,8 @@ Simulation results can be plotted with ``matplolib`` using the
     ax.legend()
     ax.loglog()
     ax.set_title("Logical CNOT Error Rate")
-    fig.savefig(f"logical_cnot_result_x_observable_{1}.png")
+    plt.show()
+    # fig.savefig(f"logical_cnot_result_x_observable_{1}.png")
```
nelimee: Remove the commented line?
```suggestion
```

pyproject.toml
```diff
@@ -221,10 +222,49 @@ select = ["I", "E", "W", "D", "F", "PL", "NPY", "N", "PERF", "RUF", "UP"]
...
+  "RUF005",
+  "RUF015",
+  "RUF002",
+  "RUF003",
+  "N806",
+  "N817",
+  "N818",
+  "N803",
+  "UP038",
+  "UP035",
```
nelimee: That is quite inconvenient. I could not find a way to avoid reformatting that part automatically. I prefer the original formatting, way more readable.

nelimee (commented):
A few comments, I did not review the correspondence between the old `.ipynb` and the new `.rst`.

.github/workflows/gh-pages.yml
purva-thakre: That's what the example workflow in the docs did. So, I used it as is. 

https://github.com/actions/cache?tab=readme-ov-file#restoring-and-saving-cache-using-a-single-action

docs/conf.py
purva-thakre: Yes, this is an old change when I wanted to check the build time etc.

pyproject.toml
purva-thakre: This was introduced by you in 668f082dea0875ef34994b6a75607b77a0dca02c and e3a5e06beb2e4a0dea9563292b501144bf5f51f1. What did you use to format them this way?

.github/workflows/gh-pages.yml
nelimee: Ok. I *think* the cache can be unique across OSes. That would allow us to always generate the cache on Linux (the cheaper runner) and re-use it on Windows and Mac (that are up to 10x more expensive).

pyproject.toml
nelimee: Arf... sorry! 
I have ruff configured to run at each file save. I should likely just tell to ruff to ignore that file entirely. 
Sorry for the inconvenience.

docs/user_guide/logical_cnot.dae
purva-thakre: This was added when I ran `make clean html` locally due to `collada_interop.rst`. A code block in this exports the file. I guess it gets saved in the user_guide folder.

pyproject.toml
nelimee: Just reverted.

purva-thakre:
I think this PR is ready. I will take a quick look through the diff tomorrow before requesting a review.

docs/Makefile
```diff
@@ -17,6 +17,11 @@ clean:
 	rm -rfv "$(SOURCEDIR)/_autosummary"
 	@$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+cleanall:
+	rm -rfv "$(SOURCEDIR)/_autosummary"
+	rm -rfv "$(SOURCEDIR)/_examples_database"
+	@$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
nelimee: I think the code below should work and is cleaner:
```suggestion
cleanall: clean
	rm -rfv "$(SOURCEDIR)/_examples_database"
```

docs/user_guide/quick_start.rst
```diff
@@ -192,6 +192,8 @@ Simulation results can be plotted with ``matplolib`` using the
     ax.legend()
     ax.loglog()
     ax.set_title("Logical CNOT Error Rate")
+    ax.set_xlabel("Physical Error Rate")
+    ax.set_ylabel("Logical Error Rate")
     plt.show()
```
purva-thakre: Interestingly, if we make changes to a cached page, they do not show up. We will have to delete the cache for changes like these such that they show up and are cached again. 

image

https://tqec.github.io/tqec/pull/592/user_guide/quick_start.html#plot-the-results

nelimee: That is not normal in my opinion. The cache should not impact sphinx process at all, except when `tqec` and `stim` are called. Unresolving to check that later.

nelimee: One potential idea to look after: wait a few minutes before checking the actual page, and empty the local browser cache to be sure you have the last online version. Changes to GitHub pages may take a few minutes to propagate, so we should check that.

purva-thakre: you were right. I should have waited a few more minutes for the changes to propagate. I made a test change in https://github.com/tqec/tqec/pull/592/commits/221bb2fcb68a838dbd88980f90d04ce4899e5af2, and it showed up. 

image

Testing it again with the axes labels.

.github/workflows/gh-pages.yml
nelimee:
```suggestion
          key: docs_database_and_stim_results_cache
          enableCrossOsArchive: true
```

.github/workflows/gh-pages.yml
```diff
@@ -32,10 +32,18 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt -r requirements-dev.txt
           python -m pip install '.[all]'
+      - uses: actions/cache@v4
+        id: cache-docs
+        with:
+          path: docs/_examples_database/
+          key: ${{ runner.os }}-docs_cache
       - name: Generate the documentation
+        if: steps.${{ runner.os }}-docs_cache.outputs.cache-hit != 'true'
```
nelimee: If the previous proposition is accepted
```suggestion
        if: steps.docs_database_and_stim_results_cache.outputs.cache-hit != 'true'
```

.github/workflows/gh-pages.yml
```diff
@@ -32,10 +32,18 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt -r requirements-dev.txt
           python -m pip install '.[all]'
+      - uses: actions/cache@v4
+        id: cache-docs
+        with:
+          path: docs/_examples_database/
+          key: ${{ runner.os }}-docs_cache
```
nelimee: Add `enableCrossOsArchive` to allow the cache to be shared across Linux, MacOS and Windows (see the documentation: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cross-os-cache)
```suggestion
          key: ${{ runner.os }}-docs_cache
          enableCrossOsArchive: true
```

purva-thakre:
The new cache was saved after the build took approximately 50 mins. Retriggering the docs build to ensure the new cache is used. 

image

Edit: The new cache was used by the retriggered build. As expected, it took approximately 8 mins. I think it might be fine to delete the previous cache that used Linux for the key.

purva-thakre:
Merging this does not use the cache. 

https://github.com/tqec/tqec/actions/runs/15882215387/job/44785294659#step:6:78

The cache exists but it can't be found on main

image

image

nelimee:
> Merging this does not use the cache. 
> 
> https://github.com/tqec/tqec/actions/runs/15882215387/job/44785294659#step:6:78
> 
> The cache exists but it can't be found on main
> 
> image
> 
> image
> 

That's expected according to cache sharing rules I think. That should not be a problem for the next runs, as branches should be able to use the cache on main.

purva-thakre:
It is strange to see two different caches of the same name. One is on main and the other one is specific to the PR. 

Can you update your other PR and check if the `main` cache is used in your draft PR? I am worried a new cache might need to be generated for each PR. 

https://github.com/tqec/tqec/actions/caches

nelimee:
> It is strange to see two different caches of the same name. One is on main and the other one is specific to the PR. 
> 
> Can you update your other PR and check if the `main` cache is used in your draft PR? I am worried a new cache might need to be generated for each PR. 
> 
> https://github.com/tqec/tqec/actions/caches

Only tomorrow :/

According to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache, cache on main can be used by other branches.

nelimee:
Cache hit in #562, everything seems to work correctly :)

nelimee:
It might be necessary to version the cache according to the current detector database version. In any case, we will need to make sure that the detector database is correctly handled.

PR #593, opened 2025-05-30
Add option to show lines missing unit test coverage

purva-thakre:
This PR adds the option to show which lines are not covered by unit tests. I noticed two modules are missing proper unit test coverage, but there wasn't enough information. 

image

This PR adds the option to show such lines. 

https://github.com/tqec/tqec/actions/runs/15336433395/job/43154562414?pr=593#step:6:167

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,7 +137,7 @@ module = ["tqec.interop.pyzx.*"]
 disallow_untyped_calls = false
 
 [tool.coverage.run]
-omit = ["*/*_test.py"]
+omit = ["*/*_test.py", "*/simulation/plotting/*"]
 
 [tool.coverage.report]
 exclude_also = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,7 +137,7 @@ module = ["tqec.interop.pyzx.*"]
 disallow_untyped_calls = false
 
 [tool.coverage.run]
-omit = ["*/*_test.py", "*/simulation/plotting/*"]
+omit = ["*/*_test.py", "*/simulation/*"]
 
 [tool.coverage.report]
 exclude_also = [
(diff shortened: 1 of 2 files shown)


nelimee (approved):

purva-thakre:
@nelimee Any other submodules I need to ignore from the coverage report?

nelimee:
> @nelimee Any other submodules I need to ignore from the coverage report? Would it be preferable to ignore the entire `simulation` module?

I do not understand why tqec.simulation is still reported at 27% although we omitted tqec.simulation.plotting. The simulation module might probably be omitted entirely yes.

purva-thakre:
> I do not understand why tqec.simulation is still reported at 27% although we omitted tqec.simulation.plotting. 

Probably related to the options of the workflow that adds the PR comment. 
https://github.com/tqec/tqec/actions/runs/15355154990/job/43212513900?pr=593#step:8:3

image

pyproject.toml
```diff
@@ -137,7 +137,7 @@ module = ["tqec.interop.pyzx.*"]
 disallow_untyped_calls = false
 
 [tool.coverage.run]
-omit = ["*/*_test.py"]
+omit = ["*/*_test.py", "*/simulation/plotting/*"]
```
purva-thakre, line 140:
```suggestion
omit = ["*/*_test.py", "*/simulation/*"]
```

nelimee:
> Not sure why this workflow is failing on main. https://github.com/tqec/tqec/actions/runs/15364967186/job/43243897140
> 
> image

I could not find an obvious fix for that, could you open an issue? 

The problem lies with `python-sat` failing to build. First, it seems like a cache got invalidated, which requires all packages to be rebuilt. Then, when building `python-sat`, a UNIX-only header file is missing and plenty of message related to windows subsystem for linux are printed. I wonder if the build process is not done through WSL without the base headers.

Issue #482, opened 2025-02-21
Support parallel execution when generating stim circuits


Zhaoyilunnn:
**Is your feature request related to a problem? Please describe.**
Hi Team, thanks for the great work. I was building a similar tool to translate lattice surgery to stim circuit util I found this awesome project. I am very interested in this project and I found its design philosophy insightful and scalable. Thus I wish to contribute to this project instead of buiding my own.
Currently I am catching up all previous recordings and walking through the code base and would like to contribute to resolve more issues in the future. 
Before I could start working on some functional enhancement, I found a performance issue, that is, it time consuming to translate a CNOT to stim circuit. 
After profiling the performance bottleneck corresponds to `_inplace_add_detectors_to_circuits` which iteratively process circuit one by one.
Thus I wonder if we could accelerate this procedure via parallel processing.
I am not sure whether it is appropriate to consider performance issue at this stage of the project, If possible, I could open a PR to enhance this performance.

**Describe the solution you'd like**
Use sth like `joblib` to leverage multiprocessing to accelerate this

I did a simple perf comparison using the following script

```python
from tqec.gallery import cnot, move_rotation
from tqec import Basis
from tqec import compile_block_graph, NoiseModel
import time

# block_graph = move_rotation(Basis.Z)
block_graph = cnot(Basis.Z)

# block_graph.view_as_html(write_html_filepath="temp.html")

correlation_surfaces = block_graph.find_correlation_surfaces()
compiled_computation = compile_block_graph(block_graph, observables="auto")

# 3. Generate the `stim.Circuit` of target code distance
start = time.perf_counter()
circuit = compiled_computation.generate_stim_circuit(
    # k = (d-1)/2 is the scale factor
    # Large values will take a lot of time.
    k=2,
    # The noise applied and noise levels can be changed.
    noise_model=NoiseModel.uniform_depolarizing(0.001),
    parallel=True,
)
print("Parallel: Generate stim circuit consumes: ", time.perf_counter() - start)

# 3. Generate the `stim.Circuit` of target code distance
start = time.perf_counter()
circuit = compiled_computation.generate_stim_circuit(
    # k = (d-1)/2 is the scale factor
    # Large values will take a lot of time.
    k=2,
    # The noise applied and noise levels can be changed.
    noise_model=NoiseModel.uniform_depolarizing(0.001),
    parallel=False,
)
print("Serial: Generate stim circuit consumes: ", time.perf_counter() - start)

```

[...]

PR #595, opened 2025-06-01
Improving parallelism when computing detectors

Zhaoyilunnn:
#482 points out that the process of generating stim circuit from graph can be accelerated using parallel processing. This PR implements the solution according to [this comment ](https://github.com/tqec/tqec/issues/482#issuecomment-2705827343).

## Benchmarking Results

All numbers are in seconds

### Benchmarking program

```py
import os
import time
from pathlib import Path

from tqec.compile.convention import FIXED_BULK_CONVENTION

k = 3

# database = None # w/o database
database = DetectorDatabase() # w/ database

# Generate block graph
# block_graph = cnot(Basis.X)
block_graph = steane_encoding(Basis.X)

# Compile block graph

# Measure runtime of generate_stim_circuit
start_time = time.time()
    k,
    manhattan_radius=2,
    detector_database=database,
    # do_not_use_database=True, # w/o database
    do_not_use_database=False, # w/ database
    parallel=True,
    # parallel=False,
)
end_time = time.time()

runtime = end_time - start_time

print(f"Runtime of generate_stim_circuit: {runtime:.4f} seconds")


```

### Environment

OS: Ubuntu 22.04.4 LTS
Python: 3.12.9
Hardware: A server with 2 AMD 48-Core Processors

### Without using database

parallel: 25.4626
non-parallel: 134.1192
speedup: 5.26X

#### database does not exist

parallel: 26.5886
non-parallel: 93.2584
speedup: 3.51X

#### database already exist

parallel: very large memory footprint (>20G) and very slow (I killed it after running several minutes), possibly because when database already exist, copying database objects across many processes and sync them is very very costly
non-parallel: 3.9597
speedup:  null

---

- [x] ~~Handle the cases when database is not None~~
- [x] ~~Add a logic to handle cases when database already exist, in this case, automatically set parallel to be False~~


nelimee:
With
```py
import itertools
import time

from tqec.compile.convention import FIXED_BULK_CONVENTION

for do_not_use_database, parallel in itertools.product((True, False), (True, False)):
    k = 3
    database = DetectorDatabase()

    # Generate block graph
    block_graph = steane_encoding(Basis.X)
    # Compile block graph

    # Measure runtime of generate_stim_circuit
    start_time = time.time()
        k,
        manhattan_radius=2,
        detector_database=database,
        do_not_use_database=do_not_use_database,
        parallel=parallel,
    )
    end_time = time.time()

    runtime = end_time - start_time

    print(
        f"Runtime of generate_stim_circuit ({parallel=}, "
        f"{do_not_use_database=}): {runtime:.4f} seconds"
    )
```
on a AMD Ryzen 9 5950X (32 cores), I get
```
Runtime of generate_stim_circuit (parallel=True, do_not_use_database=True): 11.8246 seconds
Runtime of generate_stim_circuit (parallel=False, do_not_use_database=True): 85.6813 seconds
Runtime of generate_stim_circuit (parallel=True, do_not_use_database=False): 12.1149 seconds
Runtime of generate_stim_circuit (parallel=False, do_not_use_database=False): 58.6104 seconds
```

Zhaoyilunnn:
That's a much better test script :)

Mine is 

```
Runtime of generate_stim_circuit (parallel=True, do_not_use_database=True): 28.5132 seconds
Runtime of generate_stim_circuit (parallel=False, do_not_use_database=True): 139.6295 seconds
Runtime of generate_stim_circuit (parallel=True, do_not_use_database=False): 30.4452 seconds
Runtime of generate_stim_circuit (parallel=False, do_not_use_database=False): 97.7685 seconds
```

Currently I don't understand why the last run does not reuse the database created by the third one (the third run should save a database right?)

In my previous tests, when there exist a database (the situation of the fourth run), the time cost is much lower. I will check this.

---

Oh I see, if explicitly create an empty database (`database = DetectorDataBase()`) then we'll not try to re-instantiate it from the db path. That makes sense. 
I think our experimental results perfectly align

src/tqec/compile/tree/tree.py
```diff
@@ -337,6 +335,16 @@ def generate_circuit(
         if do_not_use_database:
             detector_database = None
 
+        # Decide whether to enable parallel processing when computing detectors
+        # If not using a database, or if the database is empty, we can use parallel processing.
+        # to accelerate the computation.
+        if detector_database is None or len(detector_database.mapping) == 0:
```
nelimee:
```suggestion
        if detector_database is None or len(detector_database) == 0:
```

src/tqec/compile/tree/tree.py
```diff
@@ -337,6 +335,16 @@ def generate_circuit(
...
+            parallel = True
+        # Otherwise we disable parallel processing to avoid overhead of copying
+        # the database to multiple processes.
```
nelimee: Are you sure that is the issue? If not, do not present it as a possibility and not a certainty.

Zhaoyilunnn: You're absolutely correct. I'll remove this comment. Not very urgent/relevant for now but I did some experiments to understand the overhead.

Using the gallery cnot with parallel processing enabled and with an existing database being reused. 

```
Runtime of generate_stim_circuit (parallel=True, do_not_use_database=False): 252.2647 seconds
         18498152 function calls (15923496 primitive calls) in 176.943 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       12    0.025    0.002  244.032   20.336 compute.py:550(compute_detectors_for_fixed_radius)
     3940    0.045    0.000  196.853    0.050 connection.py:246(recv)
     1934    0.018    0.000  148.736    0.077 managers.py:815(_callmethod)
     1922    0.015    0.000  147.864    0.077 managers.py:1127(__getattr__)
     3964   28.829    0.007  110.224    0.028 {built-in method _pickle.loads}
       12    0.001    0.000   92.346    7.695 pool.py:738(__exit__)
       12    0.002    0.000   92.345    7.695 pool.py:654(terminate)
     1867    0.025    0.000   85.179    0.046 pool.py:500(_wait_for_updates)
4144/4120    0.039    0.000   77.298    0.019 connection.py:429(_recv_bytes)
     3782    0.068    0.000   68.550    0.018 connection.py:1122(wait)
     5068    0.960    0.000   62.558    0.012 connection.py:202(send)
     1200    0.005    0.000   62.262    0.052 util.py:208(__call__)
       12    0.003    0.000   60.283    5.024 pool.py:680(_terminate_pool)
     5308    0.094    0.000   59.894    0.011 connection.py:406(_send_bytes)
     7242    0.016    0.000   58.995    0.008 connection.py:381(_send)
     7242    2.187    0.000   58.978    0.008 {built-in method posix.write}
       12    0.001    0.000   57.141    4.762 pool.py:671(_help_stuff_finish)
       12    0.326    0.027   57.139    4.762 {method 'acquire' of '_multiprocessing.SemLock' objects}
    36/12    0.002    0.000   56.813    4.734 threading.py:1018(_bootstrap)
    36/12    0.002    0.000   56.813    4.734 threading.py:1058(_bootstrap_inner)
    36/12    0.001    0.000   56.810    4.734 threading.py:1001(run)
       12    0.001    0.000   56.810    4.734 pool.py:527(_handle_tasks)
       12    0.002    0.000   56.781    4.732 pool.py:573(_handle_results)
       12    0.001    0.000   56.749    4.729 pool.py:362(map)
8288/8240    1.985    0.000   48.258    0.006 connection.py:390(_recv)
11313/11265   38.220    0.003   45.216    0.004 {built-in method posix.read}
     5128   40.629    0.008   40.662    0.008 {method 'dump' of '_pickle.Pickler' objects}

```

Obviously `SyncManager` introduces significant overhead, after re-sorting the method by their `tottime`, the top four methods are `pickle.dump/loads` and `posix.read/fork`. So we may safely conclude that the processes are passing some serialized objects between each other which becomes very costly. 

```
Sorted by tottime:
    5128    40.6290    40.6620 {method 'dump' of '_pickle.Pickler' objects}
   11313    38.2200    45.2160 {built-in method posix.read}
    1164    31.5500    31.5980 {built-in method posix.fork}
    3964    28.8290   110.2240 {built-in method _pickle.loads}
 1134059     6.5040     6.7200 detector.py:28(__hash__)
    3782     5.0410    25.1090 selectors.py:402(select)

```

Reading the docs of SyncManager make me feel that only the manager process (the server) holds the real database objects and the client processes only hold a proxy object that should be lightweight. And initially I think reading/writing attributes from/to the server should not be that expensive. But turned out it is expensive. Need to understand why but I'll move forward for now.

I just feel that python is really not good at parallel processing

Zhaoyilunnn:
Hi @nelimee, this PR is ready for review. I'll address any new comments, and when all comments are addressed, please merge this PR directly

src/tqec/compile/tree/tree.py
nelimee: Using https://docs.python.org/3/library/multiprocessing.shared_memory.html might be a path to avoid back-and-forth data transfers. Pretty sure it won't be trivial to code correctly though, so not sure it is worth exploring.

Zhaoyilunnn: > Using https://docs.python.org/3/library/multiprocessing.shared_memory.html might be a path to avoid back-and-forth data transfers. Pretty sure it won't be trivial to code correctly though, so not sure it is worth exploring.

Thanks! I am also looking into this method. Maybe we can address the non/empty-db cases first. I'll explore this to see if we can accelerate the reuse-db case later. The issue will not be close by this PR so I think this might be fine.

src/tqec/compile/detectors/compute.py
```diff
@@ -494,13 +496,69 @@ def _compute_superimposed_template_instantiations(
...
+
+
+def _compute_detector_for_subtemplate(
+    args: tuple[
+        tuple[int, ...],
+        npt.NDArray[numpy.int_],
+        Sequence[Plaquettes],
+        Shift2D,
+        DetectorDatabase | None,
+        bool,
```
nelimee: I think the following would make sense here:
```suggestion
        tuple[int, ...],  # indices
        npt.NDArray[numpy.int_],  # s3d
        Sequence[Plaquettes],  # plaquettes
        Shift2D,  # increments
        DetectorDatabase | None,  # database
        bool,  # only_use_database
```

src/tqec/compile/detectors/compute.py
```diff
@@ -494,13 +496,69 @@ def _compute_superimposed_template_instantiations(
...
+
+
 def compute_detectors_for_fixed_radius(
     templates: Sequence[Template],
     k: int,
     plaquettes: Sequence[Plaquettes],
     fixed_subtemplate_radius: int = 2,
     database: DetectorDatabase | None = None,
     only_use_database: bool = False,
+    parallel: bool = False,
```
nelimee: I would replace that argument with something like `parallel_processes_count: int = 1` with the following behaviour:

- If `parallel_processes_count` is 1, that is a sequential run.
- If `parallel_processes_count > 1`, that is a parallel run with the provided number of processes.
- If `parallel_processes_count == -1`, that means "use all the possible processes, so `cpu_count()`.
- If anything else, that is an invalid value.

src/tqec/compile/detectors/compute_test.py
```diff
@@ -428,3 +428,10 @@ def test_compute_detectors_for_fixed_radius(
         [template, template], k, [init_plaquettes, memory_plaquettes]
     )
     assert len(detectors) == d**2
+
+    # Test the parallel parameter
+    # Should get the same results with parallelization as without
+    parallel_detectors = compute_detectors_for_fixed_radius(
+        [template, template], k, [init_plaquettes, memory_plaquettes], parallel=True
+    )
+    assert len(parallel_detectors) == len(detectors)
```
nelimee: You can rather compare that using `set` comparison:
```suggestion
    assert set(parallel_detectors) == set(detectors)
```

That should be a better test.

src/tqec/compile/tree/annotators/detectors.py
```diff
@@ -164,6 +164,7 @@ def __init__(
         detector_database: DetectorDatabase | None = None,
         only_use_database: bool = False,
         lookback: int = 2,
+        parallel: bool = False,
```
nelimee: Same comment as above, maybe change that to be an `int` to give more freedom to the user?

src/tqec/compile/tree/tree.py
```diff
@@ -149,12 +149,18 @@ def _annotate_detectors(
         database_path: Path = DEFAULT_DETECTOR_DATABASE_PATH,
         only_use_database: bool = False,
         lookback: int = 2,
+        parallel: bool = False,
```
nelimee: Same comment on making that an `int`

src/tqec/compile/tree/tree.py
```diff
@@ -327,13 +335,23 @@ def generate_circuit(
...
             detector_database = None
 
+        # Decide whether to enable parallel processing when computing detectors
+        # If not using a database, or if the database is empty, we can use parallel processing.
+        # to accelerate the computation.
+        if detector_database is None or len(detector_database) == 0:
+            parallel = True
+        # Otherwise we disable parallel processing
+        else:
+            parallel = False
```
nelimee:
```suggestion
        parallel = detector_database is None or len(detector_database) == 0
```

nelimee (commented):
Initial set of comments, will do a deeper review with tests on my computer later.

src/tqec/compile/tree/tree.py
Zhaoyilunnn: with this logic, we are enabling parallel processing *implicitly*, as you can see that currently `parallel` argument is not exposed to user in `generate_stim_circuit` method.

I agree that we should make the number of cores to be configurable, this thing is also in my mind, but I also do not want users to completely control whether or not enabling parallel processing because it incurs very high overhead (at least in current implementation) when reusing a database.

As such, my opinion is that maybe we add back the parallel (i.e., the parallel_process_count option suggest by you) to users in the `generate_stim_circuit` method.
But if a non-empty database is provided, we automatically set the `parallel_process_count` to 1 and print a message to tell users that we are not using parallel processing in this case

What do you think?


---

or we add the `parallel_process_count` argument in `generate_stim_circuit` with clear docstring illustrating when should users enable parallel processing, and internally we respect users' setting and remove this automatic checking logic

nelimee: Right. With that in mind, I think the current behaviour is fine. Maybe just add a comment explaining why it is not (yet?) controllable?

Zhaoyilunnn: Fine for me. I prefer to change the name `parallel` to `parallel_process_count` and default to "cpu_count()//2 +1". Plus a clear comment illustrating the current status of this feature. Hopefully, we can accelerate the database-reuse cases in the future and expose `parallle_process_count` to users at that time.
Is this fine for you?

nelimee: Yes, BTW the comment you started a discussion on is not related to the discussion, and is simply the code transformation:
```py
flag: bool
if condition:
    flag = True
else:
    flag = False
```
to
```py
flag: bool = condition
```

Zhaoyilunnn: 😄 Yeah, here the code transformation is what I called the "logic of automatically determining whether to enable parallel processing", so I started the discussion here... I'll definitely apply the code transformation change, thanks

Zhaoyilunnn:
> Initial set of comments, will do a deeper review with tests on my computer later.

I am done with these initial comments

src/tqec/compile/tree/tree.py
```diff
@@ -257,7 +258,7 @@ def _generate_annotations(
         only_use_database: bool = False,
         lookback: int = 2,
         add_polygons: bool = False,
-        parallel: bool = False,
+        parallel_process_count: int = 1,
```
nelimee, line 256: Either:

- add an explicit note (or even warning) in the docstring saying that this setting is automatically reset to 1 when a detector database is provided or,
- "raise" a warning when the database is provided and this is not `1`.

I personally prefer option 1

Zhaoyilunnn: The `_generate_annotations` function is a private function currently without docstring. I think this is because we do not expect users to use this function directly. For the public method `generate_circuit`, there's no `parallel_process_count` parameter. It is set automatically here: https://github.com/Zhaoyilunnn/tqec/blob/2aafd7bd442dc17d3ee3011f85cba4ecd7f48350/src/tqec/compile/tree/tree.py#L339-L347 with a comment exlaining the details.


If we also need to explain in a docstring, maybe in `AnnotateDetectorsOnLayerNode`'s `__init__` is better?

Or, we add a docstring for all arguments in the `_generate_annotations` function, although it is private 

What do you think?

nelimee: Ho right, I missed the fact that this was a private function. Then disregard my comment, the current status is perfectly fine for me, sorry for missing that detail.

I'll try to find time to test the PR from a user perspective today, and if everything is alright I'll approve it.

src/tqec/compile/detectors/compute.py
```diff
@@ -558,16 +620,73 @@ def compute_detectors_for_fixed_radius(
...
-            database,
-            only_use_database,
+    detectors_by_subtemplate: dict[tuple[int, ...], frozenset[Detector]]
+
+    # If parallel_process_count > 1 or parallel_process_count == -1, we will
+    # enable parallel processing to compute detectors in parallel.
+    if parallel_process_count > 1 or parallel_process_count == -1:
+        parallel_process_count = (
+            cpu_count() if parallel_process_count == -1 else parallel_process_count
         )
```
nelimee: Maybe

```suggestion
    # Handle the special case of parallel_process_count == -1
    if parallel_process_count == -1:
        parallel_process_count = cpu_count()
    # If parallel_process_count > 1 we will enable parallel processing to 
    # compute detectors in parallel.
    if parallel_process_count > 1:
```

src/tqec/compile/detectors/compute.py
```diff
@@ -558,16 +620,73 @@ def compute_detectors_for_fixed_radius(
...
-        for indices, s3d in unique_3d_subtemplates.subtemplates.items()
-    }
+        # If database is not None, using a shared database via SyncManager
+        if database is not None:
```
nelimee: To check that I understood correctly: in current TQEC pipeline, that is dead code that will never be executed, because TQEC never calls that function with `parallel_process_count != 1 and database is not None`, is that correct?

Zhaoyilunnn:
> There is an issue somewhere with this PR.
> 
> The following code
> 
> ```python
> from tqec.compile.compile import compile_block_graph
> from tqec.compile.convention import FIXED_BULK_CONVENTION
> from tqec.compile.detectors.database import DetectorDatabase
> from tqec.gallery import steane_encoding
> from tqec.utils.enums import Basis
> 
> 
> k = 3
> database = DetectorDatabase()
> print(database)
> # Generate block graph
> block_graph = steane_encoding(Basis.X)
> # Compile block graph
> compiled_graph = compile_block_graph(block_graph, FIXED_BULK_CONVENTION)
> 
> circuit = compiled_graph.generate_stim_circuit(
>     k, manhattan_radius=2, detector_database=database
> )
> ```
> 
> does not save a full database to the expected path but rather an empty one.
> 
> I do not have the time to investiguate at the moment.

I'll fix this issue. Thanks!

Zhaoyilunnn:
I am doing some experiments regarding this [issue](https://github.com/tqec/tqec/pull/595#pullrequestreview-2909422449), please ignore any updated commits until I give a summary of my observations and my options on follow up actions

Zhaoyilunnn:
First of all, it's my fault that I didn't find this [issue ](https://github.com/tqec/tqec/pull/595#pullrequestreview-2909422449) before. I assume that `SyncManager` automatically handles objects synchronization across processes, but it does not. It turns out that the modifications of child processes do not affect the original object in the parent process, which is why the database is empty on disk.

## Status before this summary

- With `SyncManager`, we achieve effective speedup through parallel processing when
  1. **not-using-db**: The input database is `None` and `do_not_use_database` is `True`. Which goes to [this branch](https://github.com/Zhaoyilunnn/tqec/blob/f9bd635c6f15a5b71d1c17bfe1e9c34bea01039a/src/tqec/compile/detectors/compute.py#L675-L686).
  2. **empty-db**: The input database is `None` or empty (`DetectorDatabase()`) and `do_not_use_database` is `False`. Which goes to [this branch](https://github.com/Zhaoyilunnn/tqec/blob/f9bd635c6f15a5b71d1c17bfe1e9c34bea01039a/src/tqec/compile/detectors/compute.py#L637-L673). Note that when `do_not_use_database` is `False`, we create an empty database internally if the input is `None`, so it also goes to this branch.

The issue is that for case-2, the database stored on disk is empty.

## Why empty database?

[`SyncManager`](https://docs.python.org//3/library/multiprocessing.html#multiprocessing.managers.SyncManager) creates and returns  [`ProxyObjects`](https://docs.python.org//3/library/multiprocessing.html#multiprocessing-proxy-objects) for (but maybe only for) some commonly used *built-in* types such as lists and dictionaries.

However, it does not automatically support user-defined classes. (experiments showed that it does not. I didn't find the relevant documentation though.

As a result, although we have shared the `DetectorDatabase` class across processes using `SyncManager`, they do not actually synchronize with each other, that is the original object is not changed at all. That's why it stores an empty database on disk.

## How to ensure the database is updated?

### With `SyncManager`

From my understanding, we need to create our own proxy objects for `DetectorDatabase` and create a custom `Manager` inheriting from `BaseManager`, which seems to be tricky and I need to more time to investigate the feasibility. See [here](https://docs.python.org//3/library/multiprocessing.html#customized-managers)

A workaround is to manually do the modifications to the database in the parent process, this is tricky as `compute_detectors_at_end_of_situation` returns [*shifted* detectors](https://github.com/tqec/tqec/blob/d6f90192269b96651a62f5957b193308ff1bc2c5/src/tqec/compile/detectors/compute.py#L346), while the database is updated using [non-shifted detectors](https://github.com/tqec/tqec/blob/d6f90192269b96651a62f5957b193308ff1bc2c5/src/tqec/compile/detectors/compute.py#L326). As such, I modified the return types of `compute_detectors_at_end_of_situation`, see [here](https://github.com/Zhaoyilunnn/tqec/blob/f9bd635c6f15a5b71d1c17bfe1e9c34bea01039a/src/tqec/compile/detectors/compute.py#L315-L353), but this way is not elegant.

After this modification, the database is not empty on disk. However, there's no speedup in the computation time anymore. This is reasonable because previous I already observed that [when the database is not empty, there is significant overhead.](https://github.com/tqec/tqec/issues/482#issuecomment-2933163741), with more values stored in the database, it behaves more like the non-empty-database case.

### Without `SyncManager`

To solve this problem, I simply fall back to using [vanilla multiprocessing and always pass a `None` database to child processes](https://github.com/Zhaoyilunnn/tqec/blob/2749da10728e144600d7c83b9b88cb8e336352e0/src/tqec/compile/detectors/compute.py#L634-L644). To ensure that the database is correctly updated and stored on disk, I also [collect results and update the database in parent process](https://github.com/Zhaoyilunnn/tqec/blob/809d2b2bcd7afc3d9413d05683b1abb89be1cffc/src/tqec/compile/detectors/compute.py#L641-L647). This way, the speed up is notable.

| computation | time (s) parallel | time (s) sequential | case |
|-------------|--------------------|----------------------|------|
| `cnot` |  9.65 | 60.23 | not-using-db |
| `cnot` | 9.22  | 27.82  | empty-db |
| `steane_encoding` | 18.82 | 136.12  | not-using-db |
| `steane_encoding` | 17.52 | 93.93  | empty-db |

**To summarize**, for not-using-db case and empty-db case, we can achieve speedup through parallel processing by simply **avoid** reading and writing to the same database in different processes, we just compute detectors in parallel and update database in the parent process, which ensures that the database is correctly updated and stored on disk.

## My opinions

We have two options.

1. Only use parallel processing for not-using-db case.
2. Or we still use parallel processing for both empty-db and not-using-db cases, by adopting the "Without `SyncManager`" approach described above, with detailed comments illustrating the current implementation.

For me, I prefer option 2. What do you think? @nelimee 
If this is fine for you, I will do a final pass to the code changes and do more tests and finalize this PR. Note that the current failure of CI is due to a mypy error that seems to be false-positive. I will check it later.

src/tqec/compile/detectors/compute.py
Zhaoyilunnn: This is not a dead code. Please also refer to this [comment](https://github.com/tqec/tqec/pull/595#issuecomment-2956264430). When database is empty (`DetectorDatabase()`), it will go into this branch. If the database is not empty, the parallel count will be set to 1 and it will goto the sequential branch

src/tqec/compile/tree/tree.py
Zhaoyilunnn: I did a quick investigation on `shared_memory`, it *seems* that it works best for fixed-size objects as one need pass `size` during initialization. Pre-allocate a large enough memory seems ok but it may be difficult to predict a maximum size. Potentially we can allocate memory dynamically but this get things complicated. I still need more time to come up with a solution for database-reuse case. 
For not-using-db and empty-db case, my final solution is illustrated in this [comment](https://github.com/tqec/tqec/pull/595#issuecomment-2956264430).

nelimee:
A few comments (will be able to provide more Wednesday):

- The `DetectorDatabase` is basically a `dict`. It should not be that hard to allow a `DetectorDatabase` instance to be created from an existing `MutableMapping[_DetectorDatabaseKey, frozenset[Detector]]`. You can then create that existing thing with `SyncManager.dict()`. That can also be done with an existing database, for example (not valid code, some methods do not exist):
  ```py
  existing_database: DetectorDatabase
  with SyncManager() as manager:
      managed_dict = manager.dict(existing_database.get_dict())
      managed_database = DetectorDatabase(managed_dict)
      # Use managed_database across processes.
      # Once finished, get back a normal dictionary.
      existing_database = DetectorDatabase(dict(managed_dict))
  ```
- If the above is not fine (not as efficient, too hard to implement, not working because I missed something, ...), option 2 seems like the best way.

nelimee:
PS: the mypy error is likely due to numpy pushing version 2.3 with new typings. I am investiguating.

Zhaoyilunnn:
> A few comments (will be able to provide more Wednesday):
> 
> * The `DetectorDatabase` is basically a `dict`. It should not be that hard to allow a `DetectorDatabase` instance to be created from an existing `MutableMapping[_DetectorDatabaseKey, frozenset[Detector]]`. You can then create that existing thing with `SyncManager.dict()`. That can also be done with an existing database, for example (not valid code, some methods do not exist):
>   ```python
>   existing_database: DetectorDatabase
>   with SyncManager() as manager:
>       managed_dict = manager.dict(existing_database.get_dict())
>       managed_database = DetectorDatabase(managed_dict)
>       # Use managed_database across processes.
>       # Once finished, get back a normal dictionary.
>       existing_database = DetectorDatabase(dict(managed_dict))
>   ```
> * If the above is not fine (not as efficient, too hard to implement, not working because I missed something, ...), option 2 seems like the best way.

Using manager.dict() seems promising, but I doubt whether it will be efficient as expected. I'll give a try and get back here later

nelimee:
For the mypy errors: https://github.com/tqec/tqec/pull/599#discussion_r2136160768

Zhaoyilunnn:
> A few comments (will be able to provide more Wednesday):
> 
> * The `DetectorDatabase` is basically a `dict`. It should not be that hard to allow a `DetectorDatabase` instance to be created from an existing `MutableMapping[_DetectorDatabaseKey, frozenset[Detector]]`. You can then create that existing thing with `SyncManager.dict()`. That can also be done with an existing database, for example (not valid code, some methods do not exist):

> * If the above is not fine (not as efficient, too hard to implement, not working because I missed something, ...), option 2 seems like the best way.

I did an experiment on another branch: https://github.com/Zhaoyilunnn/tqec/commit/2566df94622ee7c0090b63978feb8a71ce636429
using this script

```py
import time
import os
from tqec.compile.convention import FIXED_BULK_CONVENTION
from tqec.gallery import steane_encoding, cnot


k = 3
test_cases = [
    # (graph_func, do_not_use_database)
    # (steane_encoding, True),
    # (steane_encoding, False),
    (cnot, True),
    (cnot, False),
]


file_to_delete = DEFAULT_DETECTOR_DATABASE_PATH

for graph_func, do_not_use_database in test_cases:
    if os.path.exists(file_to_delete):
        os.remove(file_to_delete)

    print(f"Testing {graph_func.__name__} with do_not_use_database={do_not_use_database}")

    database = DetectorDatabase()
    # database = None
    block_graph = graph_func(Basis.X)
    start = time.perf_counter()
        k,
        manhattan_radius=2,
        detector_database=database,
        do_not_use_database=do_not_use_database,
    )
    elapsed = time.perf_counter() - start
    print(f"Time taken: {elapsed:.2f} seconds\n")

```

The output is 

```
Testing cnot with do_not_use_database=True
Time taken: 8.80 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' does not exist
Testing cnot with do_not_use_database=False
Time taken: 138.05 seconds

```

The 138.05-seconds-case is only 9.22 second as reported in this [comment](https://github.com/tqec/tqec/pull/595#issuecomment-2956264430)

It seems that using `manager.dict()` is also inefficient, need to dig into the internals of python multiprocessing library to understand the reason, or maybe I did sth wrong with my implementation.

nelimee:
So in your opinion we should keep the current implementation, because it successfully speeds up `tqec` in some situations, and you made sure that it does not slow `tqec` when computing detectors in parallel is not really efficient?

If that is the case, I'll re-review that PR :)

Zhaoyilunnn:
> So in your opinion we should keep the current implementation, because it successfully speeds up `tqec` in some situations, and you made sure that it does not slow `tqec` when computing detectors in parallel is not really efficient?

Yes. I'll finalize this PR soon (just add some necessary comments, if everything is fine)

Zhaoyilunnn:
I am done with this PR at my side. 

Some notes of the current state of the PR

1. Automatically enable parallel processing for `not-using-db` and `empty-db` cases.
1. when using parallel processing, always pass a `None` database to `compute_detectors_at_end_of_situation`, so that we always compute detectors rather than trying to read and modify database. This is the most efficient way I found to speedup the computation using parallel processing. If the database is not `None`, sync the db between processes incurs high overhead (at least based on my experiments)
1. The `compute_detectors_at_end_of_situation` function is split into two: the shifting coordinates part is extracted because in parallel processing case, we need sync all child processes and then update the database, when updating the database, the detectors should not be shifted, so we need to shift them separately and later.


I've checked the speedup, and whether databases are correctly stored, using the following program

```py
import time
import os
from tqec.compile.convention import FIXED_BULK_CONVENTION
from tqec.gallery import steane_encoding, cnot


k = 3
test_cases = [
    # (graph_func, do_not_use_database)
    # (steane_encoding, True),
    # (steane_encoding, False),
    (cnot, True, False),
    (cnot, False, False),
    (cnot, False, True),
    (steane_encoding, True, False),
    (steane_encoding, False, False),
    (steane_encoding, False, True),
]


file_to_delete = DEFAULT_DETECTOR_DATABASE_PATH

for graph_func, do_not_use_database, reuse_db in test_cases:
    if not reuse_db:
        if os.path.exists(file_to_delete):
            os.remove(file_to_delete)

    print(f"Testing {graph_func.__name__} with do_not_use_database={do_not_use_database}")

    # database = DetectorDatabase()
    database = None
    block_graph = graph_func(Basis.X)
    start = time.perf_counter()
        k,
        manhattan_radius=2,
        detector_database=database,
        do_not_use_database=do_not_use_database,
    )
    elapsed = time.perf_counter() - start
    print(f"Time taken: {elapsed:.2f} seconds\n")

    if os.path.exists(file_to_delete):
        file_size = os.path.getsize(file_to_delete)
        print(f"File '{file_to_delete}' size: {file_size} bytes")
    else:
        print(f"File '{file_to_delete}' does not exist")
```

The results are 

```
Testing cnot with do_not_use_database=True
Time taken: 8.51 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' does not exist
Testing cnot with do_not_use_database=False
Time taken: 8.69 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 877292 bytes
Testing cnot with do_not_use_database=False
Time taken: 1.84 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 877292 bytes
Testing steane_encoding with do_not_use_database=True
Time taken: 17.32 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' does not exist
Testing steane_encoding with do_not_use_database=False
Time taken: 17.65 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 2593607 bytes
Testing steane_encoding with do_not_use_database=False
Time taken: 4.30 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 2593607 bytes

```

As a comparison, the main branch gives

```
Testing cnot with do_not_use_database=True
Time taken: 59.61 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' does not exist
Testing cnot with do_not_use_database=False
Time taken: 27.31 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 877292 bytes
Testing cnot with do_not_use_database=False
Time taken: 1.67 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 877292 bytes
Testing steane_encoding with do_not_use_database=True
Time taken: 134.26 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' does not exist
Testing steane_encoding with do_not_use_database=False
Time taken: 92.48 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 2593607 bytes
Testing steane_encoding with do_not_use_database=False
Time taken: 4.04 seconds

File '/home/zhaoyilun/.local/share/TQEC/detector_database.pkl' size: 2593607 bytes

```

src/tqec/compile/detectors/compute.py
```diff
@@ -482,13 +503,72 @@ def _compute_superimposed_template_instantiations(
...
+    return [s3d[:, :, i] for i in range(s3d.shape[2])]
+
+
+def _compute_detector_for_subtemplate(
+    args: tuple[
+        tuple[int, ...],  # indices
+        npt.NDArray[numpy.int_],  # s3d
+        Sequence[Plaquettes],  # plaquettes
+        Shift2D,  # increments
+        bool,  # only_use_database
```
nelimee: That parameter is always set to `False` in the only place this function is called. I think it can be removed.
```suggestion
```

src/tqec/compile/detectors/compute.py
```diff
@@ -482,13 +503,72 @@ def _compute_superimposed_template_instantiations(
...
+    """Helper function for parallel processing of detector computation.
+
+    Args:
+        args: A tuple containing:
+            - indices: Tuple of indices identifying the subtemplate
+            - s3d: 3D numpy array representing the subtemplate
+            - plaquettes: Sequence of plaquettes for each time slice
+            - increments: Spatial increments between plaquette origins
+            - only_use_database: Whether to only use the database
+
```
nelimee, line 543: Missing documentation for parallel_process_count, and only_use_database might be removed according to a previous comment.

nelimee:
Tell me if/when I should merge that PR. You can merge it yourself if you have the right to do so (not sure) and if you are done :)

Zhaoyilunnn:
> Tell me if/when I should merge that PR. You can merge it yourself if you have the right to do so (not sure) and if you are done :)

Please merge the PR as I don't have the permission. Thanks!

Issue #589, opened 2025-05-21
Add `ruff` in `pyproject.toml`


purva-thakre:
Previously discussed in #497 

`ruff` is  a [ linter](https://docs.astral.sh/ruff/linter/) and [formatter](https://docs.astral.sh/ruff/formatter/). An example of how to edit the `toml` file for `ruff` is provided in https://github.com/tqec/tqec/issues/497#issuecomment-2704216397. 

To enable this in `tqec`, we must first fix [a slew of different errors](https://github.com/tqec/tqec/issues/497#issuecomment-2885373093). A large number of the raised errors can be fixed through the auto-fix options available in `ruff`. See #590 for more. 

Note that resolving this issue also requires enabling the settings in the `toml` file with `tqec/.pre-commit-config.yaml`.  https://github.com/tqec/tqec/pull/590#issuecomment-2898853190

purva-thakre:
Closing as all the sub-issues have been resolved.

Issue #585, opened 2025-05-14
Try migrating to `ty` from the Astral's toolchain


Zhaoyilunnn:
## Is your feature request related to a problem? Please describe.

Not really a problem, but I think it might be good for developers.

Maybe we should add `mypy` to pre-commit, not it will not run automatically when committing. And cloud CI can report type error even when committing successfully at local machine.

I am not sure if this is a problem to others, because including more checks in pre-commit also increase the time cost before a commit and can frustrate people I guess.

## Describe the solution you'd like

maybe add [this](https://github.com/pre-commit/mirrors-mypy) to pre-commit config

## Describe alternatives you've considered

--- 

As suggested [here](https://github.com/tqec/tqec/issues/585#issuecomment-2882888908), and in light of the effort of migrating to Astral's tool chain #589, as discussed [here](https://github.com/tqec/tqec/issues/497#issuecomment-2885983100). Maybe it is worthy trying to use [`ty`](https://github.com/astral-sh/ty) instead of mypy in the future

nelimee:
I am fine with it, but it can be slow as you noted. Have you tried it locally? If yes, what's the usual wait time when committing?

In a related effort, I am thinking about switching to Astral's toolchain since a few months, and the lack of a mypy-like replacement held me, but with the introduction of https://github.com/astral-sh/ty that might be the good time to try the migration. If `ty` is as fast as it claims to be, that might be fine for commit times.

Zhaoyilunnn:
> I am fine with it, but it can be slow as you noted. Have you tried it locally? If yes, what's the usual wait time when committing?I

I haven't try it locally before creating this issue. I usually run `mypy src/` manually and it takes several seconds to finish.

After trying https://github.com/pre-commit/mirrors-mypy, I observe inconsistent output, i.e., the `pre-commit` runs with many errors (seems to be false-positive) but `mypy src/` returns no error. I have checked that the versions of mypy are consistent (v1.15.0), maybe there're some weird issues when running mypy in pre-commit consider it runs in a isolated environment: https://github.com/pre-commit/pre-commit/issues/1580.  

Anybody interested in this can have a try in my [experiment branch](https://github.com/Zhaoyilunnn/tqec/blob/chore/mypy-precommit-test/.pre-commit-config.yaml)

>  If ty is as fast as it claims to be, that might be fine for commit times.

That seems to be a promising solution, I'll try it later if nobody takes this issue

purva-thakre:
[...]

PR #599, opened 2025-06-04
Disable ruff in pre commit and add ruff settings to `pyproject.toml` 

purva-thakre:
Related to #589 

Rules ignored permanently are in https://github.com/tqec/tqec/pull/599#discussion_r2129536736


pyproject.toml
```diff
@@ -198,3 +198,45 @@ typeCheckingMode = "standard"
...
+# I -- isort
+# PL -- pylint
+
+# Rules considered over other incompatible ones
+exclude = ["docs/conf.py", "setup.py"]
+
+ignore = [
+  "D407", "D203", "D213", "D416", "PLR0912", "PLR0911", "PLR0915", "PLR2004", "PLR0913",
+  "D103", "D100", "D101", "D102", "D105", "D104", "D107", "D205", "D301","D400", "D401", "D404", "D415", "D417", "E501",
+  "PLR1714", "PLC0414", "PLC0105", "PLW2901",
```
purva-thakre: Note that I plan to create subissues in #589 to disable these ignored rules one by one.

.pre-commit-config.yaml
```diff
@@ -9,17 +9,6 @@ repos:
...
-    # Ruff version.
-    rev: v0.7.0
-    hooks:
-      # Run the linter.
-      - id: ruff
-        types_or: [python, pyi, jupyter]
-        args: [--fix]
-      # Run the formatter.
-      - id: ruff-format
-        types_or: [python, pyi, jupyter]
```
purva-thakre, line 22: Note that I plan to create a subissue in #589 to revert the commit for this change.

purva-thakre:
This PR adds 

- ruff configuration settings to the `pyproject.toml` file
- deletes `ruff.toml` in favor of `pyproject.toml`
- adds files changed by `ruff format` and `ruff check --fix`
- A couple of rules flagged by `ruff check` are added to the ignore line in the `toml` file. The aim is to disable these one by one and create one PR for each. 

I will take a look at the changes made over the next few days, as I committed them directly after ruff changed them. Feel free to review in the meantime.

pyproject.toml
```diff
@@ -198,3 +198,45 @@ typeCheckingMode = "standard"
 # that is not meant to be exported from that module. The default value for this
 # setting is "error".
 reportPrivateImportUsage = false
+
+[tool.ruff]
+# exclude = ["__init__.py"]
+line-length = 120
```
nelimee: As an initial general comment, and because that setting seems to be responsible for most of the line change. What I say below is **subjective** and is at viewing distance of ~70cm (desk distance) on 27" screens.

At `line-length = 120` I cannot have all of:

1. a side-by-side view (e.g., to compare PR changes with 2 tabs open at the same time with a vertical split),
2. in a big enough font to be readable,
3. with all the lines complete on the screen.

I can mostly have the three points above with `line-length = 100` (I am missing the 5 last characters of each line, but I can see them by reducing the horizontal size of the file tree on the left).

So my opinion is that 120 is too large, 100 is OK.

pyproject.toml
purva-thakre: Created the subissues. I think it would be fine to ignore `"D400", "D401", "D404", "D415"` permanently. 

https://docs.astral.sh/ruff/rules/missing-trailing-period/
https://docs.astral.sh/ruff/rules/non-imperative-mood/
https://docs.astral.sh/ruff/rules/docstring-starts-with-this/
https://docs.astral.sh/ruff/rules/missing-terminal-punctuation/

purva-thakre:
This PR adds pycodestyle, pydocstyle, isort, and pylint. Should I add anything else? The full set of rules is available [here](https://docs.astral.sh/ruff/rules/).

src/tqec/utils/noise_model.py
```diff
@@ -127,15 +127,15 @@ class NoiseRule:
...
-                Only valid when applied to operations that produce measurement results.
+        """Args:
+        after: A dictionary mapping noise rule names to their probability argument.
+            For example, {"DEPOLARIZE2": 0.01, "X_ERROR": 0.02} will add two qubit
+            depolarization with parameter 0.01 and also add 2% bit flip noise. These
+            noise channels occur after all other operations in the moment and are applied
+            to the same targets as the relevant operation.
+        flip_result: The probability that a measurement result should be reported incorrectly.
+            Only valid when applied to operations that produce measurement results.
+
```
nelimee:
```suggestion
        """Describes how to add noise to an operation.
        
        Args:
            after: A dictionary mapping noise rule names to their probability argument.
                For example, {"DEPOLARIZE2": 0.01, "X_ERROR": 0.02} will add two qubit
                depolarization with parameter 0.01 and also add 2% bit flip noise. These
                noise channels occur after all other operations in the moment and are applied
                to the same targets as the relevant operation.
            flip_result: The probability that a measurement result should be reported incorrectly.
                Only valid when applied to operations that produce measurement results.
```

nelimee (commented):
I checked all the files except `pyproject.toml`, I still need to do a detailed read of it.

nelimee:
> This PR adds pycodestyle, pydocstyle, isort, and pylint. Should I add anything else? The full set of rules is available [here](https://docs.astral.sh/ruff/rules/).

Maybe:

- [NumPy-specific rules (NPY)](https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy)
- [pep8-naming (N)](https://docs.astral.sh/ruff/rules/#pep8-naming-n)
- [Perflint (PERF)](https://docs.astral.sh/ruff/rules/#perflint-perf)
- [pydoclint (DOC)](https://docs.astral.sh/ruff/rules/#pydoclint-doc)
- [pyupgrade (UP)](https://docs.astral.sh/ruff/rules/#pyupgrade-up)
- [Ruff-specific rules (RUF)](https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf)

purva-thakre:
> I checked all the files except `pyproject.toml`, I still need to do a detailed read of it.

No worries, take your time. e66aa32865634d2bc2dc771daf38707e13e35c56 added more files to the diff. I will take some time reviewing these changes too. I committed them directly after `ruff check --fix`.

purva-thakre:
**related to pyupgrade**

I think the [current CI failures](https://github.com/tqec/tqec/actions/runs/15496782172/job/43635332077?pr=599#step:7:31) are only for Python 3.10 (+ 3.11 too). I tested it locally in a Python 3.12 environment, and there were no failures. 

Edit: Created a local 3.11 & 3.10 environments, but there were no local unit test failures. 

Edit2: I temporarily disabled 3.10 from the CI, and indeed, the failures are specific to this python version. 
image

Edit3: Added make ruff targets in [75f1218](https://github.com/tqec/tqec/pull/599/commits/75f1218977d38b8f066e9ac47f8545e9ee63c95d)

purva-thakre:
Adding `pyupgrade` might take some time. Changes made by `ruff check --fix` with https://github.com/tqec/tqec/pull/599/commits/b5ffdfa8f18d8bb166e8ca9dac82fe691dc608ee are leading to CI failures in Python 3.10. I cannot replicate these locally. See https://github.com/tqec/tqec/pull/599#issuecomment-2950048192 as these failures are only for 3.10. 

https://github.com/tqec/tqec/actions/runs/15497565731/job/43637834350?pr=599#step:7:1633

Edit: We will have to rely on `ruff check --fix --target-version py310` for the pyupgrade rules. There is some discrepancy in how ruff makes fixes for `typing` and `typing_extensions` in <Python3.11. See https://github.com/astral-sh/ruff/issues/7285

Edit2: Added make ruff targets in [75f1218](https://github.com/tqec/tqec/pull/599/commits/75f1218977d38b8f066e9ac47f8545e9ee63c95d)

src/tqec/_cli/subcommands/viz.py
```diff
@@ -37,7 +37,7 @@ def generate_animation(
         "libx264 -vf format=yuv420p -vf pad=ceil(iw/2)*2:ceil(ih/2)*2 -filter_complex "
         "[0]split=2[bg][fg];[bg]drawbox=c=white@1:t=fill[bg];[bg][fg]overlay=format=auto"
     )
-    result = subprocess.run(command.split(), capture_output=True)
+    result = subprocess.run(command.split(), capture_output=True, check=False)
```
purva-thakre, line 40: Interesting to see ruff add `check = False` here..

nelimee: Yup, but I checked and that is the expected behaviour. I forgot to check in which python version that works, because I remember having that part work without that argument.

Makefile
```diff
@@ -0,0 +1,12 @@
...
+	ruff check --fix --target-version py310
+
+.PHONY: check-format
+check-format:
+	ruff format
+
+.PHONY: check-ruff-all
+check-ruff-all:
+	ruff check --fix --target-version py310
+	ruff format
```
purva-thakre: This addition is related to https://github.com/tqec/tqec/pull/599#issuecomment-2950185959

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -304,8 +289,7 @@ def _check_correlation_surface_validity(
         v_basis = Basis.Z if is_z_no_phase(g, v) else Basis.X
         if counts[v_basis.flipped()] not in [0, len(edges)]:
             raise TQECException(
-                "X (Z) type vertex should have Pauli Z (X) Pauli supported on all"
-                f"or no edges, {v} violates the rule."
+                f"X (Z) type vertex should have Pauli Z (X) Pauli supported on allor no edges, {v} violates the rule."
```
purva-thakre:
```suggestion
                f"X (Z) type vertex should have Pauli Z (X) Pauli supported on all or no edges, {v} violates the rule."
```

src/tqec/computation/block_graph_test.py
```diff
@@ -229,8 +230,8 @@ def test_graph_rotation() -> None:
 
 @pytest.mark.parametrize("obs_basis", [Basis.Z, Basis.X, None])
 def test_cnot_graph_rotation(obs_basis: Basis | None) -> None:
-    from tqec.interop.collada.read_write_test import rotated_cnot
     from tqec.gallery.cnot import cnot
+    from tqec.interop.collada.read_write_test import rotated_cnot
```
purva-thakre: Is there a particular reason to use [deferred imports](https://peps.python.org/pep-0690/) for this unit test?

nelimee: I don't think. These can likely go with the other imports, except if we have a circular dependency on test files (which would be surprising)

src/tqec/circuit/measurement_map.py
```diff
@@ -8,8 +8,8 @@ class used to represent and query measurement offsets in a circuit.
 
 from __future__ import annotations
 
+from collections.abc import Sequence
```
nelimee, line 11: Is there an explanation for not using `typing` and using `collections.abc`?

purva-thakre: Because `ruff` prefers `collections.abc` over `typing`.

src/tqec/utils/noise_model.py
```diff
@@ -519,14 +522,11 @@ def _iter_split_op_moments(
...
-                cur_moment = []
-            else:
-                cur_moment.extend(
-                    _split_targets_if_needed(op, immune_qubits=immune_qubits)
-                )
+        elif op.name == "TICK":
+            yield cur_moment
+            cur_moment = []
+        else:
+            cur_moment.extend(_split_targets_if_needed(op, immune_qubits=immune_qubits))
```
purva-thakre, line 529: To do: Check the logic of this change made by ruff.

src/tqec/computation/block_graph_test.py
purva-thakre: Well, that's the thing. I checked these locally, and there were no circular import errors after I moved the imports. I could not tell if there was another reason for it....

src/tqec/circuit/measurement_map.py
nelimee: Seems like a lot of things in typing are deprecated since python 3.9. OK for that change then :)

src/tqec/utils/noise_model.py
nelimee: That seems correct.

src/tqec/computation/block_graph_test.py
nelimee: Let's do the change, and see if anything breaks. I expect this is just something that slipped through PR and I do not see any way this could impact the code outside of circular imports (that should raise directly)

.github/workflows/ci.yml
```diff
@@ -52,10 +52,6 @@ jobs:
         run: |
           python -m pip install pytest pytest-cov
           python -m pytest --cov=src/tqec $(git ls-files '*_test.py')
-      - name: Mypy type checking
-        run: |
-          python -m pip install mypy
-          mypy src/tqec/
```
nelimee: Why deleting without replacing with something else (e.g., `ty`)?

nelimee:
> **related to pyupgrade**
> 
> I think the [current CI failures](https://github.com/tqec/tqec/actions/runs/15496782172/job/43635332077?pr=599#step:7:31) are only for Python 3.10 (+ 3.11 too). I tested it locally in a Python 3.12 environment, and there were no failures.
> 
> Edit: Created a local 3.11 & 3.10 environments, but there were no local unit test failures.
> 
> Edit2: I temporarily disabled 3.10 from the CI, and indeed, the failures are specific to this python version. image
> 
> Edit3: Added make ruff targets in [75f1218](https://github.com/tqec/tqec/pull/599/commits/75f1218977d38b8f066e9ac47f8545e9ee63c95d)

After a more in-depth look, the errors are valid. `Self` was added to the `typing` module in Python 3.11 (see the end of its [docstring](https://docs.python.org/3/library/typing.html#typing.Self)).
To keep the compatibility with Python 3.10 [that will only drop support next year](https://devguide.python.org/versions/), `Self` should be imported from the `typing_extensions` module that provides a compatibility layer with older Python versions. 

The above is already what we do in TQEC for `override`, and in theory for `Self`:
https://github.com/tqec/tqec/blob/d6f90192269b96651a62f5957b193308ff1bc2c5/src/tqec/utils/scale.py#L21

Some imports were likely missed in PR review and did not raise an explicit error until now.

That _should_ fix the issue with Python 3.10, and then you should be able to remove most of the workarounds around that issue.

Makefile
nelimee: See my [last comment](https://github.com/tqec/tqec/pull/599#issuecomment-2955009016), this will likely not be needed anymore.

pyproject.toml
```diff
@@ -198,3 +198,54 @@ typeCheckingMode = "standard"
 # that is not meant to be exported from that module. The default value for this
 # setting is "error".
 reportPrivateImportUsage = false
+
+[tool.ruff]
+# exclude = ["__init__.py"]
+line-length = 100
+target-version = "py311"
```
nelimee:
```suggestion
target-version = "py310"
```

Once https://github.com/tqec/tqec/pull/599#issuecomment-2955009016 will be addressed and we validated that it fixes the problem, we can go back to Python 3.10 that should be the lowest common denominator of the Python versions we support.

.github/workflows/ci.yml
purva-thakre: `ruff` changes led to `mypy` failures. I thought it would be better to tackle these as a separate issue i.e. plan to add this back as a part of fixing the `mypy` failures. Sorry, I forgot to make a comment for this. 

https://github.com/tqec/tqec/actions/runs/15510589278/job/43671051599#step:8:13

Edit: You can see the failures in the latest run, too. https://github.com/tqec/tqec/actions/runs/15537586891/job/43740345509?pr=599#step:8:17

Interestingly, I do not see the same failures in my local py312 environment. I think it's better to add a subissue to #585 to fix these failures. WDYT?

src/tqec/computation/block_graph_test.py
purva-thakre: Yup, made the change in https://github.com/tqec/tqec/pull/599/commits/728bcee26e8bed9060a6ae58c64b35c851ac7a23. There were no failures. :)

purva-thakre:
> Self should be imported from the typing_extensions module that provides a compatibility layer with older Python versions.

> The above is already what we do in TQEC for override, and in theory for Self:

Yes, if you look through the changes in this PR, you will notice only `Self` and `override` are imported from `typing_extensions`. 

image
image

src/tqec/utils/noise_model.py
purva-thakre: Thanks, I made this comment late last night for today. :)

purva-thakre:
Marking this PR `ready for review` because I have quickly reviewed the changes made by ruff. They seem fine to me. The only item that's left is in https://github.com/tqec/tqec/pull/599#discussion_r2135860758

.github/workflows/ci.yml
nelimee: I would say that the goal is to have the smallest impact possible on other PRs. Removing mypy is not desirable at that point, and considering that there are less than 10 errors, I would prefer them to be solved in this PR and never land on `main`.

purva-thakre: Understood. Although the issue is not less than 10 errors, I cannot understand why these errors only show up for py311. I tested them locally for py310, py311, and py312 virtual environments. There is no explicit target version for py311 in the `pyproject.toml` configuration settings for mypy.

nelimee: From what I see, that comes from improvements (at least, changes) in numpy typing. I'll try to fix that in another branch and see if it is easy.

nelimee: My internet connection is trash-tier at the moment, even downloading another version of `numpy` is out of possibility... I think that forcing `numpy` to be `<2.3` in `requirements.txt` should be sufficient at the moment (and then we can open an issue to fix the specific errors introduced by 2.3, but `mypy` is never disabled on `main`). Not sure though, will be able to test on Wednesday.

purva-thakre: No worries. Thanks for pointing to numpy. You are indeed correct that forcing <2.3 for numpy did not lead to failures. See https://github.com/tqec/tqec/pull/599/commits/b27ccff7b9a7054a6e9b864461e52927e90c8ce2

> open an issue to fix the specific errors introduced by 2.3

I'll create an issue for it. 

FWIW I think it might be better to use dependency syntax to bind version dependencies and avoid the CI from installing an unstable version. The [ release notes for numpy 2.3](https://numpy.org/devdocs/release/2.3.0-notes.html#deprecations) state that we are reading the documentation for an unstable development version.

Edit: We can then use dependabot to keep the dependencies up to date.

pyproject.toml
```diff
@@ -80,8 +80,6 @@ dependencies = { file = "requirements.txt" }
 
 
 [tool.mypy]
-# See https://numpy.org/devdocs/reference/typing.html
-plugins = "numpy.typing.mypy_plugin"
```
purva-thakre: I think it might be better to undo this change as we need this for <2.3

nelimee: I agree.

.github/workflows/ci.yml
nelimee: > FWIW I think it might be better to use dependency syntax to bind version dependencies and avoid the CI from installing an unstable version. The [ release notes for numpy 2.3](https://numpy.org/devdocs/release/2.3.0-notes.html#deprecations) state that we are reading the documentation for an unstable development version.
> 
> Edit: We can then use dependabot to keep the dependencies up to date.

I agree. Can you also open an issue for that?

PR #617, opened 2025-06-12
Fix correlation surface plotting

inmzhang:
The correlation edges were not correctly filtered in the current `_compute_cube_transformations` function when drawing correlation surface in 3D. Only those surfaces with the same basis as the cube's normal basis should be checked and drew in this `if` branch.

I discovered this bug when trying to draw a 3D block graph with correlation surface for logical S gate (implemented by gate teleportation with Y measurement):

```py
from tqec import BlockGraph

g = BlockGraph()
n1 = g.add_cube(Position3D(0, 0, 0), "P", "P1")
n2 = g.add_cube(Position3D(0, 0, 1), "XZX")
n3 = g.add_cube(Position3D(0, 0, 2), "P", "P2")
n4 = g.add_cube(Position3D(1, 0, 1), "XZX")
n5 = g.add_cube(Position3D(1, 0, 2), "Y")
g.add_pipe(n1, n2)
g.add_pipe(n2, n3)
g.add_pipe(n4, n5)
g.add_pipe(n2, n4)
correlation = g.find_correlation_surfaces()
g.view_as_html("diagram.html", show_correlation_surface=correlation[0], pop_faces_at_direction="-Y")
```

P.S. I'm preparing a talk to introduce surface code computation and topological quantum error correction (TQEC) to the authors of [Quon](https://arxiv.org/abs/2505.07804). They believe it might offer a better diagrammatic tool than ZX for high-level representations of quantum computations. We'll see how it goes.

### DIFF
--- a/src/tqec/interop/collada/_correlation.py
+++ b/src/tqec/interop/collada/_correlation.py
@@ -152,9 +152,14 @@ class CorrelationSurfaceTransformationHelper:
 
         # Surfaces with even parity constraint
         if kind.normal_basis in surface_bases:
-            assert len(correlation_edges) in {2, 4}, "Even parity constraint violated"
-            if len(correlation_edges) == 2:
-                e1, e2 = sorted(correlation_edges)
+            normal_basis_edges: set[ZXEdge] = set()
+            for edge in correlation_edges:
+                this_node = edge.u if edge.u.id == v else edge.v
+                if this_node.basis == kind.normal_basis:
+                    normal_basis_edges.add(edge)
+            assert len(normal_basis_edges) in {2, 4}, "Even parity constraint violated"
+            if len(normal_basis_edges) == 2:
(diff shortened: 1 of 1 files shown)


Zhaoyilunnn:
Quon looks promising. It would be great if tqec could support different high level representation of quantum computation. 

BTW: I *guess* things like ZX-graph, Quon, or the dae structures are not that intuitive for general users. Will circuit descriptions like openqasm be considered as input to tqec in the future?

inmzhang:
> Will circuit descriptions like openqasm be considered as input to tqec in the future?

This is definitely possible and should be supported. Once we have a mature compilation pipeline from an intermediate representation, e.g. ZX, to tqec circuits, it is straightforward to compile the qasm circuit to that representation then into tqec.

nelimee (approved):

PR #621, opened 2025-06-19
Build docs only when `docs/` is changed

purva-thakre:


nelimee:
That means that changes in the docstrings (`src/tqec/*`) will not be updated live, but only when something in the documentation changes, right?

purva-thakre:
> That means that changes in the docstrings (`src/tqec/*`) will not be updated live, but only when something in the documentation changes, right?

Good point. Let me check. If not, I will try to find a way around it.

purva-thakre:
What if we allow the additional option of `workflow_dispatch`? This should allow us to manually trigger the docs workflow when we know there is no change in `docs/` but there is a change in a public function's docstring?

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually

.github/workflows/gh-pages.yml
```diff
@@ -1,10 +1,11 @@
...
 on:
-  push:
-    branches:
-      - main
-  pull_request:
+  workflow_dispatch:
+    push:
+      branches:
+        - main
+    pull_request:
```
purva-thakre, line 8: https://github.com/tqec/tqec/pull/621#issuecomment-2988660355 I do not see the `run workflow` button in `Actions` yet. I guess it will appear after the PR has been merged?

What this change does is add the option to only manually trigger the docs build on the previously conditioned branches.

nelimee: I do not see it either and was about to ask you if this was normal behaviour. Thinking about it, I *think* I remember that `workflow_dispatch` actions are only valid when on the `main` branch. I'll review this as if this was the case, and we will see.

nelimee: Yes, that's literally the second sentence of the documentation: 
> To trigger the workflow_dispatch event, your workflow must be in the default branch.

nelimee:
> What if we allow the additional option of `workflow_dispatch`? This should allow us to manually trigger the docs workflow when we know there is no change in `docs/` but there is a change in a public function's docstring?

I need to think about it. The fact that this is hidden in a sub-menu and not directly accessible from the PR is a major inconvenience for everyone in my opinion. Also, if we can speed-up `tqec` circuit generation, we might be able to keep the documentation CI running continuously. 

Let's wait and see how much time is needed by #592 when the cache is fully re-used before taking a decision here.

purva-thakre:
Closing due to https://github.com/tqec/tqec/pull/592#issuecomment-3000999955

PR #626, opened 2025-06-25
Raise better `SomeException` and `SomeError`

### LINKED ISSUE #262: Improve exception expressiveness
For the moment, the code base only include one exception, `TQECException`, that is raised with a descriptive error message whenever an exception need to be raised. It might be beneficial to split that unique exception into several more expressive ones.

What is you opinion on that? If you agree, what should be, in your opinion, the rule to decide whether a new exception inheriting from `TQECException` should be included or an existing generic exception (such as `TQECException`) should be used?

purva-thakre:
Fixes #262 

Resources:
- https://docs.python.org/3/tutorial/errors.html
- https://docs.python.org/3/library/exceptions.html#exception-hierarchy
- https://stackoverflow.com/questions/60708789/exceptions-vs-errors-in-python
- https://realpython.com/python-exceptions/
- https://realpython.com/python-raise-exception/
- https://medium.com/@pbvillaflores/raising-exceptions-is-bad-practice-82503242fc27


purva-thakre:
@nelimee What's the motivation behind having a separate file for exceptions? Wouldn't it be easier to instead define them at the top of a file? Just wondering if I need to create an `exception.py` for each folder?

nelimee:
> @nelimee What's the motivation behind having a separate file for exceptions? Wouldn't it be easier to instead define them at the top of a file? Just wondering if I need to create an `exception.py` for each folder?

Currently there is no real preference. I would say that a separate file is easier for users to find the exception if they want to catch it, but that's only valid for exception and not errors.

Also, if you use the exceptions in different modules, it makes sense to share them by creating a specific file.

purva-thakre:
> I would say that a separate file is easier for users to find the exception if they want to catch it, but that's only valid for exception and not errors.

When I was looking through Python's documentation on this, I noticed that they define errors after the base exception (TQECException) has been defined. I haven't been able to find a definitive resource on the conditions when we should define errors and when we should define errors.

nelimee:
> > I would say that a separate file is easier for users to find the exception if they want to catch it, but that's only valid for exception and not errors.
> 
> When I was looking through Python's documentation on this, I noticed that they define errors after the base exception (TQECException) has been defined. I haven't been able to find a definitive resource on the conditions when we should define errors and when we should define errors. 

I agreed with one of the links you provided: exceptions for non-fatal errors that users may want to catch, errors for fatal errors that cannot be solved. Was that your question?

purva-thakre:
Sorry, it was pretty late when I made my previous comment. What I meant to say was, I can get a good idea based on the message being raised, whether it is supposed to be fatal or non-fatal. For example, with the provided inputs, if I expect the computation will not succeed, then it is fatal. But I haven't found a definitive resource on whether this is the way to go about it.

nelimee:
> Sorry, it was pretty late when I made my previous comment. What I meant to say was, I can get a good idea based on the message being raised, whether it is supposed to be fatal or non-fatal. For example, with the provided inputs, if I expect the computation will not succeed, then it is fatal. But I haven't found a definitive resource on whether this is the way to go about it.

I never used that convention, but I find it reasonable and quite intuitive. My opinion is that we should try to use it, except if someone gives a justified hard-no for such a convention.

purva-thakre:
closing this to create a new PR.

PR #639, opened 2025-07-14
Correctly import from `collections.abc` instead of `typing`

### LINKED ISSUE #611: pyupgrade and ruff
`ruff` was added in https://github.com/tqec/tqec/pull/599

Some of the rules were intentionally ignored in `pyproject.toml` to allow us to enable them one by one and make the required changes. The goal of this issue is to remove `"UP038", "UP035"` from the ignore section and make the changes as required. 

`ruff` should provide information when you run `ruff check` after the rule is removed from the `toml` file. Note that instead of using `ruff check`, it might be preferable to use `ruff check --fix --target-version py310`. See https://github.com/tqec/tqec/pull/599#issuecomment-2950185959 for more. 

https://docs.astral.sh/ruff/rules/non-pep604-isinstance/
https://docs.astral.sh/ruff/rules/deprecated-import/
https://docs.astral.sh/ruff/linter/#the-ruff-linter

nelimee:
Fix #611 by importing typing annotations from `collections.abc`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -248,6 +248,7 @@ ignore = [
   "PLC0414", "PLC0105",
   "PLW2901",
   "RUF017", "RUF009", "RUF005", "RUF015", "RUF002", "RUF003",
+  # UP038 - Soon-to-be-deprecated rule, we prefer using isinstance(obj, (A, B)).
   "UP038"
 ]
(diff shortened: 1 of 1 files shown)


nelimee:
Incorrectly based on `main`.

Issue #647, opened 2025-07-15
Imports should be at the top-level of a file


purva-thakre:
Rule "PLC0415" was ignored in #646 

https://docs.astral.sh/ruff/rules/import-outside-top-level/

Some functions rely on deferred imports probably due to circular import issues. It might be worth diagnosing the source of these failures. 

If we would prefer to ignore them instead, feel free to close this issue.

nelimee:
Note that this is fixed in #643.

EDIT: my bad, I did not realise that the changes were not pushed, will do a separate PR.

Issue #645, opened 2025-07-15
Missing docstrings


purva-thakre:
Related to 646

This issue is a continuation of #600.

I can't add a commit to a local PR because pre-commit is flagging some ruff failures locally. The following are missing docstrings:

- [ ] `generate_stim_circuit` in benchmarks/cnot_all_observables.py
- [ ] `generate_cnot_circuits` in benchmarks/cnot_all_observables.py
- [ ] `main` in benchmarks/cnot_all_observables.py
- [ ] And examples in the gallery

https://docs.astral.sh/ruff/rules/undocumented-public-function/

PR #649, opened 2025-07-16
Fix remaining ruff errors

### LINKED ISSUE #648: Fix PLW1641 errors
This rule is temporarily ignored in #646 

https://docs.astral.sh/ruff/rules/eq-without-hash/

nelimee:
This PR should fix all the remaining ruff errors not covered by past PRs.

In particular, it should fix #648, #647 and #645.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,9 @@ dev = [
   "jupyterlab<=4.4.3",
   "tqec[test, doc, bench]",
   "pre-commit<=4.2.0",
+  # Ruff version fixed to ensure replicability of results. Ruff being part of pre-commit, we
+  # cannot afford someone installing a different version of ruff and being forbidden to commit
+  # anything because the different version sees some errors in the code.
   "ruff==0.12.5",
 ]
 doc = [


purva-thakre:
My bad, I pushed to the wrong branch!! We can merge this one. I will close the other PR. Sorry for messing up the rebase!

nelimee:
> My bad, I pushed to the wrong branch!! We can merge this one. I will close the other PR. Sorry for messing up the rebase!

No problem at all, thanks for the rebase :)

nelimee:
I do not get why CI fails. Let's wait for this PR to point to `main` before investigating further.

purva-thakre:
> I do not get why CI fails. Let's wait for this PR to point to `main` before investigating further.

Maybe because you are not trying to merge this into main?

nelimee:
> > I do not get why CI fails. Let's wait for this PR to point to `main` before investigating further.
> 
> Maybe because you are not trying to merge this into main?

Might be it, but I do not know why. Let's worry if it still fails when trying to merge on `main`.

purva-thakre (approved):

nelimee:
https://github.com/tqec/tqec/actions/runs/16519590664/job/46718181294 shows that the ruff pre-commit hook seems to work as expected as it fails due to some lines being too long (introduced in previous PRs without fixing them).

.github/workflows/ci.yml
```diff
@@ -79,6 +79,12 @@ jobs:
           python -m pip install -e '.[test]'
           python -m pip install pytest pytest-cov
       # Only perform quick tests when pushing on a PR
+      - name: Code health - format
+        uses: astral-sh/ruff-action@v3
+        with:
+          args: "format --check --diff"
+      - name: Code health - lint
+        run: ruff check --fix
```
nelimee, line 87: I do not understand why the second part fails, but I think that is not even an issue because this is already tested in `pre-commit`, that is run unconditionally before that action. 

So I think that code can be removed, because we already check in `pre-commit`. 

Any opinion? Did I miss something and in fact this test is important?

purva-thakre: I think I forgot to remove the second part. Before I was thinking of adding ruff without the ruff action. 

```
      - name: Code health - lint
        run: ruff check --fix
```
As for the failure, I am not too sure why it was not flagged by the first part.

purva-thakre: Oh wait. I think these are making the ruff linter and formatter actions explicit. 

https://github.com/astral-sh/ruff-action?tab=readme-ov-file#use-ruff-format
https://github.com/astral-sh/ruff-action?tab=readme-ov-file#use-to-install-ruff

maybe the better question is why is pre-commit not flagging these errors?

Edit: we are not using ruff in precommit. https://github.com/tqec/tqec/blob/21423afa8cd6790b458c6ab11d7fdb47d062d8ef/.pre-commit-config.yaml

nelimee: > maybe the better question is why is pre-commit not flagging these errors?

Yes I agree, and I cannot find an answer either :/

purva-thakre: I edited my previous comment. I found the possible cause :)

nelimee: I do not understand why considering https://github.com/tqec/tqec/pull/649#issuecomment-3117205139, but you are right. I probably made a mistake somewhere.

nelimee: And ruff is in the diff: 
image

nelimee: Also, `python -m ruff check src/tqec` started locally on my computer does not return any error. Let's try to trigger CI again and see...

nelimee: Ok got it, I had the last version of `ruff` (`0.12.5`) but we require `ruff<=0.12.0`. Will bump ruff.

nelimee: So now the question remains, should we remove those lines (once we triple-checked that `ruff` is executed in `pre-commit`), or should we also add them to the regular tests?

purva-thakre: Ah, ok. That makes sense. I forgot your previous comment about adding ruff to pre-commit (I am currently multi-tasking). I also did not check the diff to see if the pre-commit file was modified. 

if it still fails, we could temporarily ignore the rule?

purva-thakre: > So now the question remains, should we remove those lines (once we triple-checked that ruff is executed in pre-commit), or should we also add them to the regular tests?

Might be fine to remove them...

The recommit file only works locally, right? Suppose I have staged changes for a commit and I don't run ruff on it. It will not let me add them to the commit and keeps raising an error that the ruff check failed. This is what I have noticed in other projects.

purva-thakre: Also, should we allow dependabot to make minor patch updates too? The errors you saw could have been resolved on their own through dependabot.

nelimee: > The recommit file only works locally, right? Suppose I have staged changes for a commit and I don't run ruff on it. It will not let me add them to the commit and keeps raising an error that the ruff check failed. This is what I have noticed in other projects.

It works locally, and because we call it in CI, it is also checked in CI. In theory, if every developer follows the developer guide and have a correctly configured pre-commit checker, the checks in CI are completely redundant. But the above condition is often not checked, so keeping the pre-commit checks also in CI makes, in my opinion, a lot of sense.

And yes, that is the expected behaviour.

> Also, should we allow dependabot to make minor patch updates too? The errors you saw could have been resolved on their own through dependabot.

Was thinking about that too. It might be too spammy, and in theory minor releases should not change user experience. I do not know why `ruff` does not follow that, I will check (and strengthen the version requirement of `ruff`).

nelimee: That was introduced in [`0.12.1`](https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#0121). I'll just fix ruff version to the one tested locally, and we will check that each bump works as expected.

nelimee:
Merging with a constant ruff version requirement and a comment to explain that:
```py
  # Ruff version fixed to ensure replicability of results. Ruff being part of pre-commit, we
  # cannot afford someone installing a different version of ruff and being forbidden to commit
  # anything because the different version sees some errors in the code.
  "ruff==0.12.5",
```

Issue #631, opened 2025-07-04
Spatial Hadamard pipe


inmzhang:

nelimee:
@BSchelpe same as in #384, could you just comment here so that I can assign you? :)

BSchelpe:
Thanks! Yes, please assign me.

PR #651, opened 2025-07-19
Use RPNG to build extended stabilizer

nelimee:
This PR is a first step toward fixing #631. 

It changes slightly how extended plaquettes are built, allowing more freedom in the types of stabilizers that can be measured.

At the moment, debug information and in particular drawers are **NOT** included. That means that extended plaquettes will not be drawn. That is a regression I plan to fix later.


inmzhang:
Sorry for the delay, I'll review at tomorrow.

src/tqec/compile/specs/library/generators/extended_stabilizers.py
```diff
@@ -225,81 +246,14 @@ class ExtendedPlaquetteCollection:
...
-            )
-            for polygons, ptype in zip(up_plaquette_polygons, up_drawers)
-        }
-        down_plaquettes = {
-            ptype: down.with_debug_information(
-                PlaquetteDebugInformation(drawer=down_drawers[ptype])
-            )
-            for ptype in down_drawers
-        }
+        """Build an instance from the provided arguments."""
```
inmzhang:
```suggestion
        """Build an instance from the provided ``RPNGDescription``."""
```

inmzhang (approved):
LGTM. I want to note here that extended plaquettes with left/right plaquettes should be supported in the future for implementing the spatial Hadamard with fixed-boundary convention.

nelimee:
> LGTM. I want to note here that extended plaquettes with left/right plaquettes should be supported in the future for implementing the spatial Hadamard with fixed-boundary convention.

Yes, not sure how to implement them correctly yet, and I wanted to start with something simple to then eventually complexify. Thanks for the review! :)

Issue #594, opened 2025-05-30
[Question]: Do we need to worry about unit test coverage?


purva-thakre:
While working on a PR, I noticed two modules lack proper unit test coverage. 

![Image](https://github.com/user-attachments/assets/37600e1b-065f-4c83-877a-9a41caf733f0)

I added the option to show which lines are not covered by unit tests in #593. The coverage report showed that other modules in `tqec` also have lines missing test coverage. The number is quite high across the entire project. 

https://github.com/tqec/tqec/actions/runs/15336433395/job/43154562414?pr=593#step:6:98

Should we add unit tests for these? It will be quite the project on its own.

nelimee:
For `tqec.simulation` in general, I think that adding tests is less important: everything in this package is a simple convenience wrapper around existing code that is already tested.

Also, I do not see meaningful ways to unit-test plotting (comparing to known plotting image will break each time we make a change or matplotlib changes something).

I am more in favour of ignoring specific submodules from the coverage report (`tqec.simulation.plotting` in particular) and still monitor the other modules. This can be done using the [`omit`](https://coverage.readthedocs.io/en/latest/config.html#run-omit) keyword.

nelimee:
I think this issue has been solved in previous PRs. The modules `tqec.simulation` and `tqec.visualisation` have been ignored from coverage reports. 

Do you think we can close this issue @purva-thakre ? Or do you see anything to improve?

purva-thakre:
> Or do you see anything to improve?

I think we should keep this open, as there are a large number of functions that lack proper unit test coverage. See the `Missing` column in https://github.com/tqec/tqec/actions/runs/16550153712/job/46803839285?pr=658#step:8:102

Should I go ahead and create subissues?

nelimee:
> I think we should keep this open, as there are a large number of functions that lack proper unit test coverage. See the `Missing` column in https://github.com/tqec/tqec/actions/runs/16550153712/job/46803839285?pr=658#step:8:102

I suspect at least half of these lines to be in methods that are not really "interesting" to test (e.g., getters). Nevertheless, we should aim at a 100% test coverage, knowing that this will likely never be reached and setting some guidelines. 

> Should I go ahead and create subissues?

The amount of work needed does not seem huge, maybe renaming this issue is enough? If you think sub-issues make sense then yes go ahead :)

purva-thakre:
> I suspect at least half of these lines to be in methods that are not really "interesting" to test (e.g., getters).

[...]

PR #660, opened 2025-07-29
Ignore uninteresting lines from code coverage

purva-thakre:
Related to https://github.com/tqec/tqec/issues/594#issuecomment-3126060849

### DIFF
--- a/src/tqec/circuit/measurement.py
+++ b/src/tqec/circuit/measurement.py
@@ -117,7 +117,7 @@ class Measurement(AbstractMeasurement):
         return f"{self.__class__.__name__}({self.qubit}, {self.offset})"
 
     def __str__(self) -> str:
-        return f"M[{self.qubit},{self.offset}]"
+        return f"M[{self.qubit},{self.offset}]"  # pragma: no cover
 
     @override
     def map_qubit(self, qubit_map: Mapping[GridQubit, GridQubit]) -> Measurement:


src/tqec/compile/detectors/database.py
```diff
@@ -422,14 +422,14 @@ def to_file(self, filepath: Path, format: Literal["pickle", "json"] = "pickle")
...
         if format == "pickle":
             with open(filepath, "wb") as f:
                 pickle.dump(self, f)
         else:
-            filepath = filepath.with_suffix(".json")
-            with open(filepath, "w") as f:
-                json.dump(self.to_dict(), f)
+            filepath = filepath.with_suffix(".json")  # pragma: no cover
+            with open(filepath, "w") as f:  # pragma: no cover
+                json.dump(self.to_dict(), f)  # pragma: no cover
```
purva-thakre: to do: check if we need to write a unit test for this. This is technically relying on an external function i.e. no need to add a unit test for this.

src/tqec/compile/detectors/database.py
```diff
@@ -447,9 +447,9 @@ def from_file(filepath: Path, format: Literal["pickle", "json"] = "pickle") -> D
             with open(filepath, "rb") as f:
                 database = pickle.load(f)
         else:
-            with open(filepath) as f:
-                data = json.load(f)
-                database = DetectorDatabase.from_dict(data)
+            with open(filepath) as f:  # pragma: no cover
+                data = json.load(f)  # pragma: no cover
+                database = DetectorDatabase.from_dict(data)  # pragma: no cover
```
purva-thakre: Same as https://github.com/tqec/tqec/pull/660/files#r2241121022

purva-thakre:
Just a heads up, I will wrap this PR up by next week. I am traveling over this weekend.

purva-thakre:
A large number of the uncovered lines are because the code is raising an Exception or TQECError, and we don't have a corresponding unit test. I think it might be better for me to add these unit tests as a part of #262

purva-thakre:
@nelimee This PR is ready for a review. 

There will be 2 more PRs related to code coverage. One is discussed in https://github.com/tqec/tqec/pull/660/#issuecomment-3166362918 and the other will be for those functions where we need to add unit tests.

pyproject.toml
```diff
@@ -268,3 +268,4 @@ ignore = [
 
 [tool.pytest.ini_options]
 markers = ["slow: mark a test that takes a long time to run."]
+addopts = ["src/tqec/interop/collada/html_viewer.py"]
```
nelimee: What is the rationale behind this change? I do not understand it.

src/tqec/circuit/moment.py
```diff
@@ -197,7 +197,7 @@ def __add__(self, other: Moment) -> Moment:
             raise TQECError("Trying to add an overlapping quantum circuit to a Moment instance.")
         cpy = deepcopy(self)
         cpy += other
-        return cpy
+        return cpy  # pragma: no cover
```
nelimee: Might be worth a real test instead of ignoring.

src/tqec/compile/blocks/positioning.py
```diff
@@ -50,7 +50,7 @@ def __add__(self, other: Self | tuple[int, int]) -> Self:
         if isinstance(other, tuple):
             x, y = other
             return self.__class__(self._x + x, self._y + y)
-        return self.__class__(self._x + other._x, self._y + other._y)
+        return self.__class__(self._x + other._x, self._y + other._y)  # pragma: no cover
```
nelimee: Might be worth a real test instead of ignoring.

src/tqec/compile/specs/library/generators/fixed_boundary.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, translator: RPNGTranslator, compiler: PlaquetteCompiler):
         self._mapper = PlaquetteMapper(translator, compiler)
 
     def _not_implemented_exception(self) -> NotImplementedError:
-        calling_method_name = inspect.stack(context=0)[1].function
+        calling_method_name = inspect.stack(context=0)[1].function  # pragma: no cover
```
nelimee: If this method is not used, we should remove it instead.

src/tqec/compile/tree/annotations.py
```diff
@@ -56,9 +56,9 @@ class Polygon:
 
     def _sorted_qubits(self) -> list[GridQubit]:
         """Return the qubits in a sorted order that can be used to draw the polygon."""
-        cx = sum(q.x for q in self.qubits) / len(self.qubits)
-        cy = sum(q.y for q in self.qubits) / len(self.qubits)
-        return sorted(self.qubits, key=lambda q: math.atan2(q.y - cy, q.x - cx))
+        cx = sum(q.x for q in self.qubits) / len(self.qubits)  # pragma: no cover
+        cy = sum(q.y for q in self.qubits) / len(self.qubits)  # pragma: no cover
+        return sorted(self.qubits, key=lambda q: math.atan2(q.y - cy, q.x - cx))  # pragma: no cover
```
nelimee: You should rather add the `# pragma: no cover` line to the method definition, and that should exclude the whole method. 

From https://coverage.readthedocs.io/en/latest/excluding.html#excluding-code-from-coverage-py:

> By default, any line with a comment of pragma: no cover is excluded. If that line introduces a clause, for example, an if clause, or a function or class definition, then the entire clause is also excluded.

nelimee (commented):
There are several places where you ignore all the lines of a method. In that case, I think it is better to ignore the method itself with a `# pragma: no cover` line on the method.

src/tqec/compile/tree/annotations.py
purva-thakre: Good point. I used a script to quickly make the changes in this PR based on the coverage report.

src/tqec/circuit/moment.py
purva-thakre:
```suggestion
        return cpy
```

src/tqec/compile/blocks/positioning.py
purva-thakre: but isn't this a simple getter method with an operation?

src/tqec/compile/specs/library/generators/fixed_boundary.py
purva-thakre: yeah, it looks like this method is not used in the fixed boundary convention.

pyproject.toml
purva-thakre: going to go ahead and remove this. I think I made this change last week, but forgot why I added it.

src/tqec/compile/blocks/positioning.py
nelimee: Yes it is, but if it performs an operation, it might be worth having at least one test. Should not be too hard to test that, and only a rudimentary unit-test will be enough in my opinion: the operation is simple and well understood. Basically, the test is simply here as a safeguard for potential future changes.

purva-thakre:
This PR is ready to be reviewed/merged. 

To resolve #594, I will have two PRs. One for the functions that need unit tests. The other one will be for those lines that raise `TQECError` but do not have unit tests. This will be resolved as part of #262 .

PR #670, opened 2025-08-17
JOSS Paper

BSchelpe:
Things I am not sure about:

1. There was a title in the sample paper the JOSS submission guidelines link to, so I added one. 
2. The sample paper has a few sentences explaining the references, so I added some bland ones. If there's anything more accurate or detailed to say about the references please do make suggestions for changes. 
3. I wasn't sure how to cite the three posters which use tqec given they haven't been published in proceedings yet (and perhaps won't be?). I cited them using the inproceedings template, because it seemed the best fit of the ones I know, but there may be a better one.
4. The citations in the text aren't displaying as I think they are meant, but I can't see what is wrong. Does anyone with more experience of markdown/bibtex see what I've done wrong?
5. I am not sure how to preview the first section (title, authors, affiliation), nor the references. My md previewer in pycharm 
doesn't display either of them. Could someone with a better previewer check they look ok?

PR #672, opened 2025-08-19
Remove JOSS pdf compiler workflow

purva-thakre:
After #670 is merged, we want to get rid of the workflow that compiles `.md` into a pdf. Otherwise, this workflow will keep running on main and other subsequent PRs.


inmzhang (approved):

purva-thakre:
Closing in favor of https://github.com/tqec/tqec/pull/670#issuecomment-3215924690

I figured out a way to push directly to Bella's PR.

PR #682, opened 2025-09-01
Migrate to uv

### LINKED ISSUE #606: Add ruff to the precommit configuration
`ruff` was added in #599 while it was removed from `.pre-commit-config.yaml` in a0b60ca44d8aa768c12e98e11c0c53b068301d48

Some of the rules were intentionally ignored in `pyproject.toml` to allow us to enable them one by one and make the required changes. After #600, #601, #602, #603, #604, #605, #608, #609, #610, and #611 have been resolved, we want to add `ruff` back to  `.pre-commit-config.yaml`. 

We also want to ensure the `ruff` configuration in the `toml` file is utilized by `.pre-commit-config.yaml` too. 

https://docs.astral.sh/ruff/linter/#the-ruff-linter 
https://github.com/pre-commit/pre-commit

purva-thakre:
Fixes #606

~~the first commit in this PR uses https://mkniewallner.github.io/migrate-to-uv/~~


requirements-dev.txt
```diff
@@ -1,14 +0,0 @@
...
-pydata-sphinx-theme<=0.16.1
-nbsphinx<=0.9.7
-sphinxcontrib-mermaid<=1.0.0
-pytest<=8.4.1
-mypy<=1.17.0
-types-networkx<=3.5.0.20250610
-pytest-cov<=6.2.1
-pandoc<=2.4
-pre-commit<=4.3.0
-scipy-stubs<=1.16.0.0
```
purva-thakre: To do: the CI failures are related to this file and `requirements.txt` being deleted by https://mkniewallner.github.io/migrate-to-uv/

setup.py
```diff
@@ -1,3 +0,0 @@
-from setuptools import setup
-
-setup()
```
purva-thakre, line 3: ~~@inmzhang Do you happen to know why we have a `setup.py` file in the repo? Since there is a `toml` file, I think we no longer need `setup.py`.~~ 

https://pip.pypa.io/en/latest/reference/build-system/setup-py/#editable-installation

Edit: I just realized it is to build the wheels for installation purposes. 
https://github.com/tqec/tqec/blob/bee930ce3c02abd33a9f7d375e497666da9e58e9/pyproject.toml#L1-L3

purva-thakre:
@inmzhang What's the motivation behind adding the following line? Git blame shows that you added it. 

https://github.com/tqec/tqec/blob/bee930ce3c02abd33a9f7d375e497666da9e58e9/pyproject.toml#L37

I am having trouble migrating to `uv` with this line due to the following error. Wanted to double check with you before I remove it. 

```
  × Failed to build `tqec @ file:///home/purva/tqec`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_editable` failed (exit status: 1)

      [stdout]
      configuration error: You cannot provide a value for
      `project.dependencies` and list it under `project.dynamic` at the same
      time
      DESCRIPTION:
          According to PEP 621:  Build back-ends MUST raise an error if the
      metadata
          specifies a field statically as well as being listed in dynamic.
```

inmzhang:
> What's the motivation behind adding the following line? Git blame shows that you added it.

I remember that it's required if you specify the dependencies with `requirements.txt` file, or it's the standard way to set that up at that time. 

Do you think we still need the `requirement(-dev).txt` as separate files to specify dependencies? I intend to include them all in `pyproject.toml` file and make them static.

purva-thakre:
> Do you think we still need the requirement(-dev).txt as separate files to specify dependencies? I intend to include them all in pyproject.toml file and make them static.

No, uv requires this change anyway. In my local commit,  dependencies are now specified in the toml file, requirements and requirements-dev will be deleteed.

.github/workflows/ci.yml
```diff
@@ -19,11 +19,10 @@ jobs:
...
       - name: Set up Python
         uses: actions/setup-python@v5
-        with:
-          python-version: "3.13"
-          cache: "pip"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
       - name: Install dependencies
-        run: python -m pip install --upgrade pip pre-commit
+        run: uv pip install --upgrade pip pre-commit
```
purva-thakre: to do: edit this entire file to be compatible with uv.. https://docs.astral.sh/uv/guides/integration/github/

.github/workflows/gh-pages.yml
```diff
@@ -42,7 +41,8 @@ jobs:
           restore-keys: |
             docs_database_and_stim_results_cache
       - name: Generate the documentation
-        run: make html
+        run: |
+          uv run --extra all make html
         working-directory: docs/
```
purva-thakre: To do: Find the source of this error: 

Image

I believe this is because the `Generate the documentation` step relies on using a working directory for the caching action to reduce the docs build time.

.github/workflows/ci.yml
```diff
@@ -38,27 +39,25 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up Python
         uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
-          cache: "pip"
-      - name: Install dependencies
```
purva-thakre: this caching action was removed because it is not compatible with uv. A separate action can be added for uv caching. 

https://docs.astral.sh/uv/concepts/cache/

.github/workflows/gh-pages.yml
```diff
@@ -20,18 +20,17 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}
     steps:
       - uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
       - name: Set up Python
         uses: actions/setup-python@v5
-        with:
-          python-version: "3.13"
-          cache: "pip"
```
purva-thakre: To do: 

Image

purva-thakre:
@inmzhang Apologies for the excessive notifications from this PR! I used `uv` for the first time and my preferred method of learning is by trial and error. I couldn't test anything related to Github Actions locally. 

I think this PR is ready for a review.

I only need to figure out if we need a dependency caching job because the uv docs make it seem like this is done by default. I will work on this later in the afternoon today. https://github.com/tqec/tqec/pull/682#discussion_r2331910573

~~Edit: I forgot `exotic-oses.yml`!!~~

.github/workflows/ci.yml
```diff
@@ -38,27 +39,25 @@ jobs:
...
+      - name: Install dependencies after creating a virtual environment
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install -r requirements.txt -r requirements-dev.txt
-          python -m pip install pytest mypy
+          uv venv
+          uv pip install ".[all]"
       - name: Build
-        run: python -m pip install -e .
+        run: uv pip install -e .
```
purva-thakre: To do: double check if we need this. As L47 seems to do the same thing. These lines were added before I started editing the workflow.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build >= 0.7.19, <0.9.0"]
+build-backend = "uv_build"
```
purva-thakre, line 3: I added this based on https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#declaring-the-build-backend

Click on the `uv-build` code block. 

For this reason, I have deleted the `setup.py` file.

.github/workflows/ci.yml
```diff
@@ -19,13 +19,14 @@ jobs:
...
-          python-version: "3.13"
-          cache: "pip"
-      - name: Install dependencies
-        run: python -m pip install --upgrade pip pre-commit
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+      - name: Create virtual environment and install dependencies
+        run: |
+          uv venv
+          uv pip install --upgrade pip pre-commit
```
inmzhang: why do we need to install `pip` here?

README.md
```diff
@@ -35,10 +35,10 @@ Documentation is available at <https://tqec.github.io/tqec/index.html>
 
 ## Installation
 
-Currently, you need to install `tqec` from source:
+Currently, you need to install `tqec` from source with [`uv`](https://docs.astral.sh/uv/):
```
inmzhang: With the current setup, it's still installable by `pip` without `uv`. It should be a recommendation way to install `tqec` with `uv` but not necessary for users who do not have `uv` available. What do you think?

.github/workflows/ci.yml
purva-thakre: Mostly because I need to read up on what the differences are between `uv pip install` and `uv add`. I haven't found a definitive resource for it except a couple of issue discussions in the uv repo.

README.md
purva-thakre: Good point. I have added code tabs for both pip and uv. 

image

I just realized there is a separate installation guide in the user guide. The readme links to it. Will work towards updating it tomorrow.

.github/workflows/ci.yml
```diff
@@ -38,27 +39,23 @@ jobs:
...
+        uses: astral-sh/setup-uv@v6
+      - name: Install dependencies after creating a virtual environment
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install -r requirements.txt -r requirements-dev.txt
-          python -m pip install pytest mypy
-      - name: Build
-        run: python -m pip install -e .
+          uv venv
+          uv sync --group all
```
purva-thakre, line 55: I have removed the 'Build' step because `uv sync` installs the editable version of tqec by default.

docs/user_guide/installation.rst
```diff
@@ -42,9 +56,20 @@ The ``tqec`` package is a regular Python package that can be installed using ``p
...
+
+        .. code-block:: bash
+
+                python -m pip install git+https://github.com/tqec/tqec.git
+
+    .. tab-item:: uv
+
+        .. code-block:: bash
+
+                uv pip install git+https://github.com/tqec/tqec.git
```
purva-thakre, line 71: I have kept `uv pip install` in two places (here and readme) because this is the only option to install a non-editable version of tqec with uv. The default `uv sync` option to install a package is the same as `pip install -e .`. 

https://docs.astral.sh/uv/concepts/projects/sync/#syncing-the-environment

There is a non-editable flag with `uv sync` but I need to experiment with it to be sure....

.github/workflows/ci.yml
```diff
@@ -72,23 +71,23 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.13"
-          cache: "pip"
-      - name: Install dependencies and tqec package
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+      - name: Install dependencies and tqec package after creating virtual environment
```
purva-thakre:
```suggestion
      - name: Install dependencies and tqec package after creating a virtual environment
```

docs/contributor_guide.rst
```diff
@@ -18,34 +18,63 @@ If you want to help maintaining and improving the ``tqec`` package, you will nee
...
+
+            # Clone the repository to have local files to work on
+            git clone https://github.com/tqec/tqec.git
+            # Go in the tqec directory
+            cd tqec
+            # Install the library with developer dependencies
+            # Note the "-editable" option, that's important.
+            uv sync --group all
+            # enable pre-commit
+            pre-commit install
```
inmzhang:
```suggestion
            # enable pre-commit
            uv run pre-commit install
```
Need to use `uv run` otherwise we need to activate the venv.

inmzhang (approved):

docs/contributor_guide.rst
purva-thakre: I forgot to edit it. Thanks for catching this!

PR #685, opened 2025-09-04
Address initial JOSS review requests

BSchelpe:
Address two requests from the JOSS review: add countries to people's affiliations and add dois to the references (I think this is just a case of changing the label of the field from url to doi..)

### DIFF
--- a/paper/paper.bib
+++ b/paper/paper.bib
@@ -1,5 +1,5 @@
 @article{polian:2015,
-  	url = {https://doi.org/10.1145/2744769.2747921},
+  	doi = {https://doi.org/10.1145/2744769.2747921},
   	Author = {{Polian}, Ilia and {Fowler}, Austin G.},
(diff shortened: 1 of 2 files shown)


paper/paper.bib
```diff
@@ -1,5 +1,5 @@
 @article{polian:2015,
-  	url = {https://doi.org/10.1145/2744769.2747921},
+  	doi = {https://doi.org/10.1145/2744769.2747921},
```
inmzhang:
```suggestion
  	doi = {10.1145/2744769.2747921},
```

From https://www.bibtex.com/f/doi-field/:

> It's recommended to simply use the DOI, and not a DOI link like https://dx.doi.org/10.... or https://doi.org/10... .

inmzhang (changes requested):

paper/paper.bib
BSchelpe: Ah I didn't realise. I'll change the others too.

PR #689, opened 2025-09-10
fix description

BSchelpe:
A quick fix to the description of one of the functions for the spatial horizontal hadamard. The function was behaving correctly, but the description confusingly didn't match. 

I thought it was clearer to push this separately, rather than wrap it into my more complicated push implementing the other spatial hadamard situations.


inmzhang (approved):

BSchelpe:
thank you!

PR #692, opened 2025-09-11
Add Astral's type checker & remove mypy

### LINKED ISSUE #585: Try migrating to `ty` from the Astral's toolchain

purva-thakre:
Fix #585


src/tqec/compile/blocks/enums.py
```diff
@@ -23,6 +23,10 @@ def to_template_border(self) -> TemplateBorder:
                 return TemplateBorder.TOP
             case SpatialBlockBorder.Y_POSITIVE:
                 return TemplateBorder.BOTTOM
+            # add a wildcard pattern when function returns `None`
+            # flagged by ty
+            case _:
+                raise ValueError(f"Cannot find the template border corresponding to {self} .")
```
purva-thakre:
```suggestion
                raise ValueError(f"Cannot return the template border corresponding to {self} .")
```

purva-thakre:
`ty` currently checks for a limited set of rules. https://docs.astral.sh/ty/reference/rules/

I think the added workflow will fail in the future when new rules are added. Plus, given that `ty` is still in development, we will have to manually upgrade it to the latest version. `dependabot` is currently set to ignore all minor versions to avoid excessive dependabot PRs. 

https://github.com/tqec/tqec/blob/e632f6c3486d2bebd2570206f0609c7c2349fc90/.github/dependabot.yml#L12-L16

src/tqec/compile/tree/node.py
```diff
@@ -45,7 +45,7 @@ def exit_node(self, node: LayerNode) -> None:
 class LayerNode:
     def __init__(
         self,
-        layer: LayoutLayer | BaseComposedLayer,
+        layer: BaseLayer | LayoutLayer | BaseComposedLayer,
```
inmzhang: The type of layer should always be `LayoutLayer | BaseComposedLayer` and it's checked at line 69 and line 77. The typed should be cast there because `TypeGuard` does not work in `ty` yet.

purva-thakre: > The typed should be cast there because TypeGuard does not work in ty ye

Would it be preferable to use type casting or ignore these lines via `# type: ignore` until `TypeGuard` works in `ty`? I had tried casting, but `ty` still kept flagging the same lines. Which is why I tried the quick & dirty way out. I am not that experienced with type checking. 

I also tried type assertions as an alternative to type casting. This led to multiple unit test failures.

inmzhang: I agree that we can `# type: ignore` it for now.

src/tqec/compile/tree/node.py
```diff
@@ -62,7 +62,7 @@ def __init__(
         self._annotations = dict(annotations) if annotations is not None else {}
 
     @staticmethod
-    def _get_children(layer: LayoutLayer | BaseComposedLayer) -> list[LayerNode]:
+    def _get_children(layer: BaseLayer | LayoutLayer | BaseComposedLayer) -> list[LayerNode]:
```
inmzhang:
```suggestion
    def _get_children(layer: LayoutLayer | BaseComposedLayer) -> list[LayerNode]:
```
This also needs to be fixed.

inmzhang (approved):

PR #695, opened 2025-09-12
Fix two issues in `NoiseModel`

HaoTy:
Two fixes in `NoiseModel`:
1. `_noise_rule_for_split_operation()` may incorrectly try to get a measure basis for a non-measurement operation, which triggers the assertion error.
2. The `si1000` model does not have rules for X basis reset and measurement, which for some circuits causes stim to complain.

### DIFF
--- a/docs/contributor_guide.rst
+++ b/docs/contributor_guide.rst
@@ -29,11 +29,15 @@ of ``tqec`` through ``pip`` or ``uv``.
 
             # Clone the repository to have local files to work on
               git clone https://github.com/tqec/tqec.git
-              # Install the library with developer dependencies
-              # Note the "-e" option, that's important.
-              python -m pip install -e 'tqec[all]'
-              # Go in the tqec directory and enable pre-commit
+              # Update pip to at least v25.1
+              python -m pip install --upgrade pip>=25.1
+              # Go in the tqec directory
               cd tqec
+              # Install developer dependencies
+              python -m pip install --group all
+              # Install tqec in editable mode (the "-e" option)
+              python -m pip install -e .
+              # enable pre-commit
               pre-commit install
 
         .. attention::
--- a/src/tqec/compile/observables/abstract_observable.py
+++ b/src/tqec/compile/observables/abstract_observable.py
@@ -358,7 +358,7 @@ def _check_correlation_surface_validity(correlation_surface: CorrelationSurface,
                 )
             continue
         v_basis = Basis.Z if is_z_no_phase(g, v) else Basis.X
-        if counts[v_basis.flipped()] not in [0, len(edges)]:
+        if counts[v_basis.flipped()] not in [0, len(g.incident_edges(v))]:  # type: ignore
             raise TQECError(
                 "X (Z) type vertex should have Pauli Z (X) Pauli supported on "
                 f"all or no edges, {v} violates the rule."
--- a/src/tqec/compile/observables/abstract_observable_test.py
+++ b/src/tqec/compile/observables/abstract_observable_test.py
@@ -5,10 +5,11 @@ from tqec.compile.observables.abstract_observable import (
     compile_correlation_surface_to_abstract_observable,
 )
 from tqec.computation.block_graph import BlockGraph
-from tqec.computation.cube import Cube, ZXCube
+from tqec.computation.cube import Cube, YHalfCube, ZXCube
 from tqec.computation.pipe import Pipe
 from tqec.gallery.cnot import cnot
 from tqec.gallery.memory import memory
+from tqec.gallery.move_rotation import move_rotation
 from tqec.gallery.stability import stability
 from tqec.gallery.three_cnots import three_cnots
 from tqec.utils.enums import Basis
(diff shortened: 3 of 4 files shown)


purva-thakre:
> The si1000 model does not have rules for X basis reset and measurement, which for some circuits causes stim to complain.

@HaoTy Can you provide some examples of these circuits?

Can you also add comments on the modifications you made? See below:
https://github.com/tqec/tqec/blob/a23f1d0927da7e92def05ff98816bd69abf18c8c/src/tqec/utils/noise_model.py#L14-L27

HaoTy:
> Two fixes in `NoiseModel`:
> 
> 1. `_noise_rule_for_split_operation()` may incorrectly try to get a measure basis for a non-measurement operation, which triggers the assertion error.
> 2. The `si1000` model does not have rules for X basis reset and measurement, which for some circuits causes stim to complain.

I was inaccurate - the second issue is from tqec instead of stim, resulting from the first fix. (The stim error was another bug, likely related to the Hadamard implementation, but I haven't been able to work out a reproducible example.)

```
ValueError: No noise (or lack of noise) specified for split_op=stim.CircuitInstruction('RX', [stim.GateTarget(93), stim.GateTarget(104), stim.GateTarget(105), stim.GateTarget(106), stim.GateTarget(115), stim.GateTarget(116), stim.GateTarget(117), stim.GateTarget(126), stim.GateTarget(127), stim.GateTarget(128), stim.GateTarget(136), stim.GateTarget(137), stim.GateTarget(138), stim.GateTarget(148), stim.GateTarget(149), stim.GateTarget(150), stim.GateTarget(160), stim.GateTarget(177), stim.GateTarget(185), stim.GateTarget(188), stim.GateTarget(189), stim.GateTarget(190), stim.GateTarget(196), stim.GateTarget(197), stim.GateTarget(198), stim.GateTarget(200), stim.GateTarget(201), stim.GateTarget(202), stim.GateTarget(208), stim.GateTarget(209), stim.GateTarget(210), stim.GateTarget(211), stim.GateTarget(212), stim.GateTarget(213), stim.GateTarget(219), stim.GateTarget(220), stim.GateTarget(221), stim.GateTarget(222), stim.GateTarget(223), stim.GateTarget(224), stim.GateTarget(230), stim.GateTarget(231), stim.GateTarget(232), stim.GateTarget(234), stim.GateTarget(235), stim.GateTarget(236), stim.GateTarget(242), stim.GateTarget(243), stim.GateTarget(244), stim.GateTarget(247), stim.GateTarget(255)], []).
```

> @HaoTy Can you provide some examples of these circuits?

Simulating the `steane_encoding` block graph with `si1000` (with the first fix) gives me the above error.

> Can you also add comments on the modifications you made? See below:
> 
> https://github.com/tqec/tqec/blob/a23f1d0927da7e92def05ff98816bd69abf18c8c/src/tqec/utils/noise_model.py#L14-L27

Done.

inmzhang:
Note that it's not a bug that SI1000 missed the noise rules for `MX`/`MY` measurements. SI1000 is the noise mode inspired by superconducting qubits, where only Z-basis measurements are available. The circuits should be compiled into Z basis interactions and measurements before using SI1000 noise model.

HaoTy:
> Note that it's not a bug that SI1000 missed the noise rules for `MX`/`MY` measurements. SI1000 is the noise mode inspired by superconducting qubits, where only Z-basis measurements are available. The circuits should be compiled into Z basis interactions and measurements before using SI1000 noise model.

I see. Thank you for the explanation! I assume that means the block graph compilation should consider the noise model and produce compatible physical circuits?

inmzhang:
>  I assume that means the block graph compilation should consider the noise model and produce compatible physical circuits?

This is a missing feature that transpile the circuit into Z-basis interactions. You can use [Craig's `gen` package](https://github.com/Strilanc/quantum-factor-2025-code/blob/main/src/gen/_layers/_transpile.py) to do that. Or you can define your own noise model that can work with MX, RX, etc. gates with different noise strength.

HaoTy:
> This is a missing feature that transpile the circuit into Z-basis interactions.

As a temporary workaround, shall we merge this PR for now and revert the changes to si1000 once this missing feature is implemented? Otherwise, the si1000 model is out-of-the-box unusable, and the errors can be confusing to users.

inmzhang:
> As a temporary workaround, shall we merge this PR for now and revert the changes to si1000 once this missing feature is implemented?

I think you should define your own `NoiseModel` for the workaround instead of modifying si1000 model itself.

> Otherwise, the si1000 model is out-of-the-box unusable, and the errors can be confusing to users.

SI1000 cannot be used for the circuit generated by TQEC for now but can be used for any other circuit that only includes Z-basis interactions. I agree the error message is not good and maybe we should add more details to the si1000 model docstring.

HaoTy:
> I think you should define your own `NoiseModel` for the workaround instead of modifying si1000 model itself.
> 
> SI1000 cannot be used for the circuit generated by TQEC for now but can be used for any other circuit that only includes Z-basis interactions. I agree the error message is not good and maybe we should add more details to the si1000 model docstring.

Understandable. I reverted the changes to si1000 and slightly modified the docstring and the error message to improve clarity. How about now? 
In any case, the first issue should be fixed.

inmzhang (approved):
LGTM.

Issue #384, opened 2024-11-01
Implement logical hadamard transition


inmzhang:
# Logical Hadamard Transition

The logical Hadamard transition allows X(Z) type correlation surface to be changed to Z(X) type either in space or time. The necessary ingredients to implement the feature is ready in the codebase and it should  be a good starting point for the new contributors to get themselves familiar with the project.

The following code assumes that you check out the branch from #379. You can install `tqec` package from that branch with:

```py
python -m pip install git+https://github.com/QCHackers/tqec.git@refactor/computation
```

### 1.1 Construction

The temporal Hadamard transition is simply a transversal layer of H gates applied to all data qubits. Note that the boundary conditions of the code patches before and after the transition are flipped, meaning the X(Z) boundary changes to Z(X), and the stabilizer types of the patch are also inverted.

![图片](https://github.com/user-attachments/assets/bc51a86a-1920-41d5-a23c-1da5cd0ce2cc)

### 1.2 Example

You can create a minimal example of a computation that includes the temporal Hadamard transition:

```py
from tqec import BlockGraph, Cube, Position3D, ZXCube

g = BlockGraph()
g.add_edge(
    Cube(Position3D(0, 0, 0), ZXCube.from_str("ZXZ")),
    Cube(Position3D(0, 0, 1), ZXCube.from_str("XZX")),
)
correlation_surfaces = g.to_zx_graph().find_correration_surfaces()
g.view_as_html(
    pop_faces_at_direction=SignedDirection3D(Direction3D.Y, False),
    show_correlation_surface=correlation_surfaces[0],
)
```

![Temporal Logical Hadamard](https://github.com/user-attachments/assets/68a2ac7a-b064-4da5-babb-a245734b4246)

### 2.1 Construction

The spatial Hadamard transition is more complex. It involves replacing plaquettes on the boundaries of connected cubes. Specifically, new plaquettes measuring `ZXZX` and `XZXZ` stabilizers must be introduced at the boundary.

![图片](https://github.com/user-attachments/assets/d80ba83f-6d1d-4283-8707-00946ddcd5b0)

### 2.2 Example

You can create a minimal example of a computation that includes the spatial Hadamard transition with the following code:

```py
from tqec import BlockGraph, Cube, Position3D, ZXCube

g = BlockGraph()
g.add_edge(
    Cube(Position3D(0, 0, 0), ZXCube.from_str("ZXZ")),
[...]

PR #697, opened 2025-09-12
Implement remaining spatial hadamard code paths in the fixed boundary convention

BSchelpe:
This will close issue #631 (and also issue #384, I believe - I have not checked the temporal hadamard implementations, as far as I know they do not have issues and so the only remaining work for this issue was the spatial hadamards).

I've made this a draft PR because it definitely isn't ready to be merged yet. I know I have some issues I need to check 
about whether I've got the right basis (spatial boundary vs top left,...) and whether the resets and measurements are 
correct in the hadamard versions, for a start. I also haven't done any testing yet - I am planning to follow @inmzhang's suggestions in issue #384.  

I wanted to make a draft PR to get initial feedback on the rough shape of the code - does it look ok, before I start polishing it?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,8 +54,8 @@ jobs:
           uv venv
           uv sync --group all
       # Type check the code base.
-      - name: Mypy type checking
-        run: uv run mypy src/tqec/
+      - name: Type checking
+        run: uv run ty check
       # Perform the full tests (including slow ones) when pushing on main.
       - name: Complete tests
         if: github.ref == 'refs/heads/main'
--- a/.github/workflows/exotic-oses.yml
+++ b/.github/workflows/exotic-oses.yml
@@ -33,8 +33,8 @@ jobs:
           uv venv
           uv sync --group all
       # Perform type checking before testing because that is quicker.
-      - name: Mypy type checking
-        run: uv run mypy src/tqec/
+      - name: Type checking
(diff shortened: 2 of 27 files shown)


inmzhang:
Hi Bella, thanks for the draft PR! I will take time to look into this PR in the next few days.

BSchelpe:
> Hi Bella, thanks for the draft PR! I will take time to look into this PR in the next few days.

Thank you!

inmzhang:
Hi Bella, the code structure looks good to me. Would you like me to review the implementation details, such as the template and plaquette definitions, at the current stage of this PR? The logic is kind of complex, and I’m not fully confident that I can check every detail thoroughly. It would be helpful if you could share those details in a some visual format (even a simple hand-drawn draft would do), as that would make it easier for me to review :-)

BSchelpe:
Thank you very much for looking at it! That's all the feedback I wanted at this stage. Let me make all the small changes I know are wrong, and run the test you suggested in #384, and then it will be ready for a detailed review (at that point I will also write and draw something to make it easier to follow - it took me ages to get my head round it).

smburdick:
@BSchelpe does your PR #774 supersede this?

BSchelpe:
No, I'm afraid not. This is for the fixed boundary convention. #774 is for the fixed bulk. I will revisit this once I've finished the fixed bulk PR.

Issue #696, opened 2025-09-12
Improve the performance of finding correlation surfaces


HaoTy:
## Is your feature request related to a problem? Please describe.

Finding correlation surfaces is extremely slow (as in, it may take months) for moderately large block graphs (~30 ports, ~500 volume), effectively unusable beyond trivial scale.

## Describe the solution you'd like

A simple improvement is that, in `tqec.interop.pyzx.correlation.find_correlation_surfaces()`, replace 
```python
  correlation_surfaces: set[CorrelationSurface] = set()
  for leaf in leaves:
      correlation_surfaces.update(_find_correlation_surfaces_from_leaf(g, leaf))
```
with
```python
from functools import partial
from multiprocessing import Pool

with Pool() as pool:
    correlation_surfaces = set(
        sum(
            pool.imap_unordered(
                  partial(_find_correlation_surfaces_from_leaf, g), leaves
              ),
            start=[],
        )
    )
```
to get a `min(num_cpus, num_ports)` speedup, since `_find_correlation_surfaces_from_leaf` is embarrassingly parallel.

However, it is not adequate at all, as the poor scalability easily dominates this constant speedup. A potentially more effective solution I'm considering is to partition a large block graph, find the correlation surfaces for the partitioned subgraphs, and generate combinations that satisfy boundary conditions.

## Describe alternatives you've considered

- Caching the intermediate results to avoid recomputation, which turned out slower due to the hashing overhead.
- PyZX's Pauli web finding. Ran into the same gFlow issue (see #528).

Zhaoyilunnn:
Hi @HaoTy I am interested in investigating this issue, do you have a readily available test case?

purva-thakre:
> - Caching the intermediate results to avoid recomputation, which turned out slower due to the hashing overhead.
> - PyZX's Pauli web finding. Ran into the same gFlow issue (see
https://github.com/tqec/tqec/issues/528).

Making a comment to come back to this issue later: I have a couple of handwritten notes on this in addition to some refactor proposals. Will get to typing these up after I have resolved the other issues that I have started to work on. Sorry for the delay.

- One main issue I have found over the past week is that the project is based on python when we ought to be thinking of using Rust[^2]. Austin also pointed this out over an email conversation, but his suggestion was to rewrite parts of tqec in C++ for very big Clifford circuits[^1]. 
- The other one is that PyZX's Pauli web finding is insufficient. I have a couple of ideas on how to improve/refactor this in my handwritten notes. 

[...]

PR #704, opened 2025-09-22
feat: reduce branches for passthrough node when finding correlation surfaces

Zhaoyilunnn:
As reported in #696, finding correlation surfaces has poor performance scaling. 

Before we come up with more advanced algorithm, there's one clear "redundancy" in current implementation that can be fixed with simple change.

When visiting a passthrough node (same color observable), current algorithm exploits all possble combinations of remained edges that satisfy the even parity constraint. 

image

However, the highlighted branch is unnecessary (and should not be included in this case), because when dragging the blue line from left to right, you cannot pass all edges.

But of course all 4 edges can be included, this happens when visiting the same node the next time. So the right situation should be,

1. When first visiting the node, we have three possible branches

image
 
 2. The next time visiting the node, e.g., in the blow case, we have one branch with all edges included. 
 
image

We don't need to include the 4-edge case when first visiting a node, the recursive algorithm will eventually visit the node again as long as there are remaining edges. 

See also: https://github.com/tqec/tqec/issues/696#issuecomment-3316633727


inmzhang:
> But of course all 4 edges can be included, this happens when visiting the same node the next time. 

I realized that there might be cases where a node will not be visited for a second time, then some correlation surfaces will not be constructed with the current method. For example, in this structure:

图片

The current method will not find a single valid correlation surface. Reverting the changes made in this PR will make it work.

Zhaoyilunnn:
> there might be cases where a node will not be visited for a second time

I think you are correct...  Sorry I am attending a conference and have little time at this moment, could you revert this change, or I'll do it later.

Issue #548, opened 2025-03-27
Y-Basis Initialization and Measurement


KabirDubey:
We want to support circuits with Y basis initialization and measurement. These are represented by the green rectangular prism in SketchUp. Fig 7 of [1] goes through the layer by layer construction, which involves sqrt{X} gates and some scheduling details. 

In my opinion, the first step is a design document that outlines how this can be implemented in the layers framework and other modules. We can make sub-issues once the expected outcomes are better understood. 

Anyone that is interested can be assigned to this issue. 

[1] [Inplace Access to the Surface Code Y Basis](https://quantum-journal.org/papers/q-2024-04-08-1310/pdf/)
[2] [Slides explaining the usage of Y Init and Meas](https://docs.google.com/presentation/d/184GHX9jffq9dcwWzbku0K90V9xdA8_25lD8pfeEcgRg/edit?slide=id.g32ef451badc_0_588#slide=id.g32ef451badc_0_588)

KabirDubey:
A Y basis memory experiment is a good goal. Here's how it compares to a Z and X basis memory experiments:

<img width="890" alt="Image" src="https://github.com/user-attachments/assets/a8c5aeb7-5ba5-4394-b1ce-ee14b259d27d" />

The shortest Y-basis memory experiment is shown on right (note the top face is green, not blue):

![Image](https://github.com/user-attachments/assets/3ec05557-e235-4280-aebc-525735a2c4d7)

You could run in principle run a memory experiement for longer, but`tqec` natively supports d rounds of "hold". Note the vertical connector has zero space-time volume, so all steps of the memory experiment are encoded in the Y prisms.

afowler:
The X, Y and Z memory experiments should be exactly the same length. In the
SketchUp figure, they are all d rounds long.

angelaelisa:
How does `rotate` (block_graph.py) affect —or shouldn't affect— to a blockgraph previously built in SketchUp that contains a `YHalfCube`?

nelimee:
> How does `rotate` (block_graph.py) affect —or shouldn't affect— to a blockgraph previously built in SketchUp that contains a `YHalfCube`?

In theory, just like any BlockGraph. In practice, the YHalfCube currently only has 4 valid orientations (the half dimension should always be in the Z axis). Because it makes sense in theory, I am in favour of rotating the YHalfCube just like any other cube, and raise an error when generating the circuit (such as NotImplementedError)

angelaelisa:
[...]

PR #707, opened 2025-09-26
A polynomial-time algorithm for finding correlation surfaces

HaoTy:
A new algorithm for finding correlation surfaces, exponentially faster than the original algorithm, doesn't require gflow, and doesn't need info about input and output leaf nodes. See also #696.

It's currently in `src/tqec/interop/pyzx/correlation_experimental.py` as it's still in a draft form. I plan to add mid-exploration pruning based on leaf nodes, change the data structure to be X-and-Z-based, and polish it more. 

Happy to discuss how it should be modified to best integrate into the existing codebase. For example, the `reduce_to_minimal_generators` argument doesn't make sense anymore since the algorithm doesn't return all possible correlation surfaces. And how the tests should be adapted accordingly (it should pass all tests up to equivalence, except for those requiring all possible correlation surfaces).

Basic benchmarking using stacked Steane encoding block graphs:
```
Finding 7 correlation surfaces for block graph with 19 cubes, 20 pipes, and 7 leaf cubes took 0.01 s
Finding 8 correlation surfaces for block graph with 35 cubes, 40 pipes, and 8 leaf cubes took 0.03 s
Finding 9 correlation surfaces for block graph with 51 cubes, 60 pipes, and 9 leaf cubes took 0.07 s
Finding 10 correlation surfaces for block graph with 67 cubes, 80 pipes, and 10 leaf cubes took 0.13 s
Finding 11 correlation surfaces for block graph with 83 cubes, 100 pipes, and 11 leaf cubes took 0.29 s
Finding 12 correlation surfaces for block graph with 99 cubes, 120 pipes, and 12 leaf cubes took 0.31 s
Finding 13 correlation surfaces for block graph with 115 cubes, 140 pipes, and 13 leaf cubes took 0.81 s
Finding 14 correlation surfaces for block graph with 131 cubes, 160 pipes, and 14 leaf cubes took 0.84 s
Finding 15 correlation surfaces for block graph with 147 cubes, 180 pipes, and 15 leaf cubes took 1.63 s
Finding 16 correlation surfaces for block graph with 163 cubes, 200 pipes, and 16 leaf cubes took 1.62 s
Finding 17 correlation surfaces for block graph with 179 cubes, 220 pipes, and 17 leaf cubes took 2.22 s
Finding 18 correlation surfaces for block graph with 195 cubes, 240 pipes, and 18 leaf cubes took 2.85 s
Finding 19 correlation surfaces for block graph with 211 cubes, 260 pipes, and 19 leaf cubes took 4.63 s
Finding 20 correlation surfaces for block graph with 227 cubes, 280 pipes, and 20 leaf cubes took 5.85 s
Finding 21 correlation surfaces for block graph with 243 cubes, 300 pipes, and 21 leaf cubes took 7.03 s
Finding 22 correlation surfaces for block graph with 259 cubes, 320 pipes, and 22 leaf cubes took 7.10 s
Finding 23 correlation surfaces for block graph with 275 cubes, 340 pipes, and 23 leaf cubes took 10.39 s
Finding 24 correlation surfaces for block graph with 291 cubes, 360 pipes, and 24 leaf cubes took 12.19 s
Finding 25 correlation surfaces for block graph with 307 cubes, 380 pipes, and 25 leaf cubes took 14.27 s
Finding 26 correlation surfaces for block graph with 323 cubes, 400 pipes, and 26 leaf cubes took 16.79 s
Finding 27 correlation surfaces for block graph with 339 cubes, 420 pipes, and 27 leaf cubes took 19.30 s
Finding 28 correlation surfaces for block graph with 355 cubes, 440 pipes, and 28 leaf cubes took 18.09 s
Finding 29 correlation surfaces for block graph with 371 cubes, 460 pipes, and 29 leaf cubes took 20.49 s
Finding 30 correlation surfaces for block graph with 387 cubes, 480 pipes, and 30 leaf cubes took 25.62 s
```

### DIFF
--- a/src/tqec/interop/pyzx/correlation_experimental.py
+++ b/src/tqec/interop/pyzx/correlation_experimental.py
@@ -7,6 +7,7 @@ from functools import partial, reduce
 from itertools import chain, combinations, product
 
 import networkx as nx
+import numpy as np
 import stim
 from pyzx.graph.graph_s import GraphS
 from pyzx.pauliweb import PauliWeb, multiply_paulis
@@ -19,7 +20,6 @@ from tqec.interop.pyzx.utils import (
     is_s,
     is_x_no_phase,
     is_z_no_phase,
-    is_zx_no_phase,
 )
 from tqec.utils.enums import Basis, Pauli
 from tqec.utils.exceptions import TQECError
@@ -140,10 +140,8 @@ def _find_correlation_surfaces_from_leaf(
             bases_at_leaves[closed_leaf] = Pauli.Y
         elif is_z_no_phase(g, closed_leaf):
             bases_at_leaves[closed_leaf] = Pauli.X
-        elif is_x_no_phase(g, closed_leaf):
-            bases_at_leaves[closed_leaf] = Pauli.Z
         else:
-            raise TQECError(f"Unsupported leaf node type: {g.type(closed_leaf)}.")
+            bases_at_leaves[closed_leaf] = Pauli.Z
     neighbor = next(iter(g.neighbors(leaf)))
     pauli_graphs = []
     for basis in (
@@ -159,48 +157,115 @@ def _find_correlation_surfaces_from_leaf(
     if g.vertex_degree(neighbor) == 1:  # make sure no leaf node will be in the frontier
         frontier = []
         if neighbor in bases_at_leaves:
-            pauli_graphs = filter(
-                lambda pg: pg.nodes[neighbor][leaf] in (bases_at_leaves[neighbor], Pauli.I),
-                pauli_graphs,
+            pauli_graphs = list(
+                filter(
+                    lambda pg: pg.nodes[neighbor][leaf] in (bases_at_leaves[neighbor], Pauli.I),
+                    pauli_graphs,
+                )
             )
     else:
         frontier = [neighbor]
         explored_leaves = [leaf]
 
     while frontier:
+        stabilizer_nodes = explored_leaves + frontier
         cur = frontier.pop()
         unconnected_neighbors = sorted(
             filter(lambda n: n not in pauli_graphs[0][cur], g.neighbors(cur))
         )
-        neighbor_leaves_bases = {
-            n: _flip_pauli_based_on_edge_type(g, (cur, n), bases_at_leaves[n])
-            for n in unconnected_neighbors
-            if n in bases_at_leaves
-        }
+        # neighbor_leaves_bases = {
+        #     n: _flip_pauli_based_on_edge_type(g, (cur, n), bases_at_leaves[n])
+        #     for n in unconnected_neighbors
+        #     if n in bases_at_leaves
+        # }
         unexplored_neighbors = [n for n in unconnected_neighbors if n not in pauli_graphs[0]]
+        broadcast_basis = "X" if g.type(cur) == VertexType.Z else "Z"
+        passthrough_basis = Basis[broadcast_basis].flipped()
 
-        new_pauli_graphs = []
+        valid_graphs, invalid_graphs, syndromes = [], [], []
         stabilizer_list = []
         for pauli_graph in pauli_graphs:
-            incident_paulis = list(pauli_graph.nodes[cur].values())
-            if not is_zx_no_phase(g, cur):
-                raise TQECError(
-                    f"Unsupported non-leaf node type: {g.type(cur)} with phase {g.phase(cur)}."
-                )
-            broadcast_basis = "X" if g.type(cur) == VertexType.Z else "Z"
+            incident_paulis = pauli_graph.nodes[cur].values()
+            passthrough_parity = sum(p.has_basis(passthrough_basis) for p in incident_paulis) % 2
             broadcast_supports = [p.has_basis(broadcast_basis) for p in incident_paulis]
+            valid = True
             if all(broadcast_supports):
                 broadcast_pauli = Pauli[broadcast_basis]
             elif not any(broadcast_supports):
                 broadcast_pauli = Pauli.I
             else:  # invalid broadcast
-                continue
-
-            passthrough_basis = Basis[broadcast_basis].flipped()
-            passthrough_parity = sum(p.has_basis(passthrough_basis) for p in incident_paulis) % 2
+                valid = False
             if not unconnected_neighbors and passthrough_parity:  # invalid passthrough
-                continue
+                valid = False
+                broadcast_supports += [True]
 
+            stabilizer = stim.PauliString(
+                "".join(
+                    pauli.value
+                    for pauli in chain.from_iterable(
+                        pauli_graph.nodes[n].values() for n in stabilizer_nodes
+                    )
+                )
+            )
+            if _can_be_generated_by(stabilizer, stabilizer_list):
+                continue
+            if valid:
+                if stabilizer != stim.PauliString("I" * len(stabilizer)):
+                    stabilizer_list.append(stabilizer)
+                valid_graphs.append((pauli_graph, broadcast_pauli, passthrough_parity))
+                if len(stabilizer_list) == 2 * len(stabilizer_nodes):
+                    break
+            else:
+                invalid_graphs.append(pauli_graph)
+                syndromes.append(
+                    np.array(broadcast_supports).dot(1 << np.arange(len(broadcast_supports)))
+                )
+
+        for i, (pauli_graph, syndrome) in enumerate(zip(invalid_graphs, syndromes)):
+            if len(stabilizer_list) == 2 * len(stabilizer_nodes):
+                break
+            for indices in chain.from_iterable(
+                combinations(range(len(invalid_graphs)), num_flips)
+                for num_flips in range(1, len(invalid_graphs) + 1)
+            ):
+                if i in indices:
+                    continue
+                new_pauli_graph = reduce(
+                    _multiply_pauli_graphs, [invalid_graphs[j] for j in indices]
+                )
+
+                incident_paulis = new_pauli_graph.nodes[cur].values()
+                passthrough_parity = (
+                    sum(p.has_basis(passthrough_basis) for p in incident_paulis) % 2
+                )
+                broadcast_supports = [p.has_basis(broadcast_basis) for p in incident_paulis]
+                valid = True
+                if all(broadcast_supports):
+                    broadcast_pauli = Pauli[broadcast_basis]
+                elif not any(broadcast_supports):
+                    broadcast_pauli = Pauli.I
+                else:  # invalid broadcast
+                    continue
+                if not unconnected_neighbors and passthrough_parity:  # invalid passthrough
+                    continue
+
+                stabilizer = stim.PauliString(
+                    "".join(
+                        pauli.value
+                        for pauli in chain.from_iterable(
+                            new_pauli_graph.nodes[n].values() for n in stabilizer_nodes
+                        )
+                    )
+                )
+                if _can_be_generated_by(stabilizer, stabilizer_list):
+                    continue
+                if stabilizer != stim.PauliString("I" * len(stabilizer)):
+                    stabilizer_list.append(stabilizer)
+                valid_graphs.append((new_pauli_graph, broadcast_pauli, passthrough_parity))
+                break
+
+        new_pauli_graphs = []
+        for pauli_graph, broadcast_pauli, passthrough_parity in valid_graphs:
             passthrough_nodes_list = list(
                 chain.from_iterable(
                     combinations(unconnected_neighbors, num_passthrough)
@@ -224,26 +289,11 @@ def _find_correlation_surfaces_from_leaf(
             ]
 
             for out_paulis in out_paulis_list:
-                if any(
-                    out_paulis[n] not in (pauli, Pauli.I)
-                    for n, pauli in neighbor_leaves_bases.items()
-                ):
-                    continue
-                stabilizer = stim.PauliString(
-                    "".join(
-                        pauli.value
-                        for pauli in list(
-                            chain.from_iterable(
-                                pauli_graph.nodes[n].values() for n in explored_leaves + frontier
-                            )
-                        )
-                        + list(out_paulis.values())
-                    )
-                )
-                if _can_be_generated_by(stabilizer, stabilizer_list):
-                    continue
-                if stabilizer != stim.PauliString("I" * len(stabilizer)):
-                    stabilizer_list.append(stabilizer)
+                # if any(
+                #     out_paulis[n] not in (pauli, Pauli.I)
+                #     for n, pauli in neighbor_leaves_bases.items()
+                # ):
+                #     continue
                 new_pauli_graph = pauli_graph.copy()
                 for n, pauli in out_paulis.items():
                     if n not in new_pauli_graph:
@@ -267,7 +317,19 @@ def _find_correlation_surfaces_from_leaf(
         explored_leaves.extend(filter(lambda n: g.vertex_degree(n) == 1, unexplored_neighbors))
 
     correlation_surfaces = list(map(partial(_pauli_graph_to_correlation_surface, g), pauli_graphs))
-    return correlation_surfaces
+    return list(filter(lambda s: _leaf_nodes_can_support_span(g, s.span), correlation_surfaces))
+
+
+def _multiply_pauli_graphs(
+    pauli_graph_a: nx.Graph,
+    pauli_graph_b: nx.Graph,
+) -> nx.Graph:
+    """Multiply two Pauli graphs of the same scope to form a new Pauli graph."""
+    new_pauli_graph = pauli_graph_a.copy()
+    for v in new_pauli_graph:
+        for n in new_pauli_graph.neighbors(v):
+            new_pauli_graph.nodes[v][n] *= pauli_graph_b.nodes[v][n]
+    return new_pauli_graph
 
 
 def _pauli_graph_to_correlation_surface(
--- a/src/tqec/interop/pyzx/correlation_experimental.py
+++ b/src/tqec/interop/pyzx/correlation_experimental.py
@@ -193,6 +193,12 @@ def _find_correlation_surfaces_from_leaf(
                 if passthrough_parity:
                     valid = False
 
+            # if any(
+            #     out_paulis[n] not in (pauli, Pauli.I)
+            #     for n, pauli in neighbor_leaves_bases.items()
+            # ):
+            #     continue
+
             stabilizer = stim.PauliString(
                 "".join(
                     pauli.value
@@ -207,8 +213,8 @@ def _find_correlation_surfaces_from_leaf(
                 if stabilizer != stim.PauliString("I" * len(stabilizer)):
                     stabilizer_list.append(stabilizer)
                 valid_graphs.append((pauli_graph, broadcast_pauli, passthrough_parity))
-                if len(stabilizer_list) == 2 * len(stabilizer_nodes):
-                    break
+                # if len(stabilizer_list) == 2 * len(stabilizer_nodes):
+                #     break
             else:
                 invalid_graphs.append(pauli_graph)
                 syndromes.append(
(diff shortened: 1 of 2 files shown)


afowler:
Excellent, would you be willing and able to talk about your new algorithm
at an upcoming group meeting? Perhaps in two weeks?

Best,
Austin.

HaoTy:
> Excellent, would you be willing and able to talk about your new algorithm at an upcoming group meeting? Perhaps in two weeks?

Of course, glad to!

Zhaoyilunnn:
Awesome! I'll try to understand and test this implementation.

HaoTy:
@Zhaoyilunnn The "wrong" correlation surfaces in the failed tests may just be products of what the tests want, which should be totally valid (e.g., the three cnots example seems valid to me; I don't see why a correlation surface can't have both X and Z bases), but I will take a closer look tomorrow.
I've seen the "non-deterministic observables" Stim error with the original algorithm before, so potentially there could be a more foundamental cause.

Zhaoyilunnn:
> @Zhaoyilunnn The "wrong" correlation surfaces in the failed tests may just be products of what the tests want, which should be totally valid (e.g., the three cnots example seems valid to me; I don't see why a correlation surface can't have both X and Z bases), but I will take a closer look tomorrow.
> I've seen the "non-deterministic observables" Stim error with the original algorithm before, so potentially there could be a more foundamental cause.

ok I got it. So your algorithm does not produce surfaces with minimal area. I agree the above correlation surface itself is correct. However, in the original implementation, it will not generate a surface that has both X and Z basis. Because starting from each leaf node, it will propogate either a X operator or a Z operator, and a three-cnot block graph will not generate a surface like yours. 
Thus I guess this kind of correlation surface leads to the failure of tests.

I am not sure if the current observable annotation algorithm made any assumption on this. Maybe @inmzhang could comment on this.

inmzhang:
> I am not sure if the current observable annotation algorithm made any assumption on this. Maybe @inmzhang could comment on this.

A correlation surface that is a product of a Z-basis and X-basis correlation surface is valid but may lead to Y basis observable terminating at the ports, which requires Y-basis measurement/initialization to support the observable. The current observable annotation implementation does not take the Y-basis observable into account yet.

Nice work! I don't have much time these days and will try to review and understand the code in the next week.

HaoTy:
> So your algorithm does not produce surfaces with minimal area.

Ah, yes, I should've noted this in my original post. Because it doesn't generate all the possible correlation surfaces, we need to solve a combinatorial optimization problem if we want to minimize the area. However, I don't think minimizing the area has a huge benefit. Please correct me if my understanding is limited.

> However, in the original implementation, it will not generate a surface that has both X and Z basis. Because starting from each leaf node, it will propogate either a X operator or a Z operator, and a three-cnot block graph will not generate a surface like yours. Thus I guess this kind of correlation surface leads to the failure of tests.
> 
> A correlation surface that is a product of a Z-basis and X-basis correlation surface is valid but may lead to Y basis observable terminating at the ports, which requires Y-basis measurement/initialization to support the observable. The current observable annotation implementation does not take the Y-basis observable into account yet.

Thank you for the clarifications! After the algorithm returns a generator set, we can potentially find another generator set that doesn't have Y-basis observables at the ports. But it may be easier and desirable to implement the Y-basis observable annotation.

inmzhang:
> But it may be easier and desirable to implement the Y-basis observable annotation.

Having Y-basis observable at ports means that we need to use Y-basis initialization/measurement block there to terminate the observable, which not only changes observable annotation but the circuit as well. It should always be helpful to not mix the X/Z basis when possible.

HaoTy:
> Having Y-basis observable at ports means that we need to use Y-basis initialization/measurement block there to terminate the observable, which not only changes observable annotation but the circuit as well. It should always be helpful to not mix the X/Z basis when possible.

I have some confusion. There are two cases when a port may get a Y from a correlation surface:
1. The port is filled by YHalfCube, indicating a Y-basis initialization or measurement. Correlation surfaces containing Ys are perhaps unavoidable. How was this handled previously? Is it TBD with #548?
2. The port is open. In this case, don't the X/Z-basis correlation surfaces also "change the circuit", as in filling the port, or am I interpreting this wrong?

inmzhang:
> The port is filled by YHalfCube, indicating a Y-basis initialization or measurement. Correlation surfaces containing Ys are perhaps unavoidable. How was this handled previously?

We do not support **simulate** a block graph with `YHalfCube`s yet, it's only possible after #548 is fixed. Yes, the Y-basis correlation basis is unavoidable  and expected then, we'll handle it during building the circuit for `YHalfCube`.

> The port is open. In this case, don't the X/Z-basis correlation surfaces also "change the circuit", as in filling the port, or am I interpreting this wrong?

Block graph with open ports can not be used for simulation or building a circuit. The concern is always at that we are required to fill a YHalfCube to terminate a Y-basis correlation surface which is more expensive than transversal X/Z basis initialization/measurement.

If your algorithm can find a generating set of correlation surface, I believe we can always reproduce a set of CSS-style correlation surface that does not mix the X/Z bases when the structure of block graph allows that.

Zhaoyilunnn (commented):
--

Zhaoyilunnn (approved):
Just want to remove the "Request changes"

HaoTy:
> We do not support **simulate** a block graph with `YHalfCube`s yet, it's only possible after #548 is fixed. Yes, the Y-basis correlation basis is unavoidable and expected then, we'll handle it during building the circuit for `YHalfCube`.

> Block graph with open ports can not be used for simulation or building a circuit. The concern is always at that we are required to fill a YHalfCube to terminate a Y-basis correlation surface which is more expensive than transversal X/Z basis initialization/measurement.

I see. 

> If your algorithm can find a generating set of correlation surface, I believe we can always reproduce a set of CSS-style correlation surface that does not mix the X/Z bases when the structure of block graph allows that.

Although I haven't implemented it, I do have a simple solution in my mind, which can reuse part of the code in the algorithm. Will update soon.

HaoTy:
@afowler Hi Austin, I'm ready to present the algorithm. Would next Wednesday be a good time?

afowler:
Yes, that would work well. Would you like the whole meeting?

HaoTy:
> Would you like the whole meeting?

The presentation itself should be ~20 minutes, but could use more time for Q&A and discussing how it can be best integrated to align with the rest of tqec.

afowler:
Do you have a couple of slides on your algorithm that we could feature at
Munich? Please see this deck
<https://docs.google.com/presentation/d/1RcMXB3kWtWWyM_LOQLkxo19Zaq4atLaV1fCsWqd78PQ/edit?usp=sharing>
in preparation. Discussing finding correlation surfaces would be next. Do
you have a deck for Wednesday?

Best,
Austin.

HaoTy:
> Do you have a couple of slides on your algorithm that we could feature at Munich? 

Yes! Please see [https://1drv.ms/p/c/6cb205327248e7b5/EeHjUTUmXUFBqu9RmlAuPcQBxKYEghPdpFXGoLfB6q1xAw?e=hXLfm2](https://1drv.ms/p/c/6cb205327248e7b5/EeHjUTUmXUFBqu9RmlAuPcQBxKYEghPdpFXGoLfB6q1xAw?e=hXLfm2). 
For the last slide, I plan to add benchmarking on ZX diagrams generated from Clifford+T circuits if I can get it done by Wednesday, or the above rudimentary stacked Steane encoding example otherwise.

afowler:
Thanks for the slides, look forward to seeing the benchmarking data.
Honestly, that would be the best thing to include in this short talk. Brief
description of the algorithm, and benchmarking data. I need to submit the
slides to Munich on Wednesday, but I'm sure there can be some flexibility
if you still don't have the data on Wednesday.

Best,
Austin.

HaoTy:
@afowler Preliminary results updated in my slides. Some of the previous slides were also slightly modified

HaoTy:
Through a series of algorithmic and engineering efforts, the performance of the algorithm has been drastically improved (notice the significantly expanded x-axis):
image
The scaling in practice is more like quadratic, due to the bottleneck being the multiplication of multiple correlation surfaces, which, in the current implementation, would need to read and perform XORs on all values in several dicts of dicts. This may not be a bottleneck when we implement it in, for example, rust, so I don't intend to attempt more involved improvements at this moment. 
 
 The problem of avoiding Y cubes is trickier than I thought. The solution I had in mind, constructing and solving a linear system, doesn't seem applicable now. I will see what else can be utilized to solve this problem.

maximilianruesch:
Hey, just wanted to throw another suggestion into this: In https://github.com/maximilianruesch/fault-gadgets/blob/develop/src/faultgadget/web/compute.py (and accompanying files) I implemented a Gaussian solver for Pauli webs based on the definitions from https://arxiv.org/abs/2410.17240, and in particular based on the algorithm from https://www.cs.ox.ac.uk/people/aleks.kissinger/papers/borghans-thesis.pdf. The context in which it is implemented is a more theoretical approach in https://arxiv.org/abs/2510.08477.

This implementation should currently be roughly qubic (I have not tested it integrated with tqec since I do not have anything set up), although during development I had a similar idea to this PR: Find cuts in the graph such that combination of the different stabiliser basis of the subgraphs will be cheap, then compute the web basis of the subgraphs using the Gaussian solver (or further divide / conquer). This would also allow the computation to become parallelised (unsure if this PR does this already), and could allow finding Pauli webs on machines with limited RAM (as the linear systems you would need to store are limited in size by the maximum of the subgraph size and the most edges cut in one cut). I am currently working on implementing this, although its not quite ready yet.

Although I do not fully understand the problem with Y-cubes, perhaps it might be easier to place linear constraints on the system to be solved with the above approach?

afowler:
If we're talking 3D placement of Y half cubes, a time duration must be
chosen and the short dimension of these must be parallel to this time
direction.

If we are talking finding correlation surfaces in a ZX graph, then only
Y correlation surfaces can terminate on a Y node, or alternatively if an X
correlation surface goes in, a Z correlation surface must come out, and
vice versa.

maximilianruesch:
@HaoTy is the problem that Y-cubes (and thus Y correlation surfaces terminating at an open node) should generally be avoided because they require some choice of time? Or is it something else?

HaoTy:
> Hey, just wanted to throw another suggestion into this: In https://github.com/maximilianruesch/fault-gadgets/blob/develop/src/faultgadget/web/compute.py (and accompanying files) I implemented a Gaussian solver for Pauli webs based on the definitions from https://arxiv.org/abs/2410.17240, and in particular based on the algorithm from https://www.cs.ox.ac.uk/people/aleks.kissinger/papers/borghans-thesis.pdf. The context in which it is implemented is a more theoretical approach in https://arxiv.org/abs/2510.08477.
> 
> This implementation should currently be roughly qubic (I have not tested it integrated with tqec since I do not have anything set up), although during development I had a similar idea to this PR: Find cuts in the graph such that combination of the different stabiliser basis of the subgraphs will be cheap, then compute the web basis of the subgraphs using the Gaussian solver (or further divide / conquer). This would also allow the computation to become parallelised (unsure if this PR does this already), and could allow finding Pauli webs on machines with limited RAM (as the linear systems you would need to store are limited in size by the maximum of the subgraph size and the most edges cut in one cut). I am currently working on implementing this, although its not quite ready yet.
> 
> Although I do not fully understand the problem with Y-cubes, perhaps it might be easier to place linear constraints on the system to be solved with the above approach?

Thank you for the information! 
I had a similar idea of constructing and solving a giant linear system covering all nodes, but thought it would be less favorable compared to a native graph algorithm. Although it would be naturally suitable for accelerated computing, it requires solving an O(n) by O(n) system (larger if m is not bounded by n), which, as you said, takes about O(n^3), where n is the number of vertices. When the graph is sparse, such as the bounded-degree graphs we have, the matrix is also sparse (because the rules are local), and the speed and memory footprint can likely be improved by sparse numerical methods. 
Still, I think a graph algorithm is a more natural solution to this problem. Eventually, with the help of time ordering from the circuit, we'd want a linear algorithm, essentially a stabilizer propagation in the circuit (but the graph doesn't necessarily have a good correspondance with the circuit after optimization/converting to lattice surgery). The algorithm in this PR (along with some changes not yet pushed) is quite close to this goal. Nothing is parallelized at this moment, but surely it can be achieved.

The Y cube problem has multiple complications. It is not very easily solvable by Gaussian elimination, if possible at all. We want to minimize the use of Y cubes, because they need an additional d/2-depth physical circuit to implement compared to X/Z initialization/measurements. But in some cases, Y cubes are needed in some of the ports for the correlation surfaces to be valid. Additionally, they can only be in the temporal direction (see #719). I think we should impose some constraints on the port-filling of the input to properly handle Y cubes.

afowler:
Interesting point about Y termination of outputs. I personally think it is
ok to flag a structure as not possible to simulate if the only valid
correlation surfaces involve one or more Y terminations on a port that is
not time aligned.

HaoTy:
@inmzhang @Zhaoyilunnn This PR is ready for review. There is a linear algorithm in development but may take another while, so I thought it would be of value to perfect and merge the polynomial algorithm (which anyway functions as a subroutine in the linear algorithm). Otherwise, we won't be able to compile any practical-size circuits.

Now the algorithm avoids Y-terminating correlation surfaces whenever possible. Note that it cannot minimize the number of Y leaves when they are unavoidable, nor does it avoid temporal `YHalfCube`s (there is no positional information passed in). I think the best solution is to instruct the user to specify all necessary `YHalfCube`s.

To be backward compatible, now a full set of correlation surfaces will be constructed from the generators if `reduce_to_minimal_generators=False`. Nonetheless, I suggest we remove it or rename it to indicate the "construct from generators" logic.

I believe all failed tests are because the returned correlation surfaces do not exactly match the hardcoded values, which can be addressed by modifying the tests to check that the generating sets are equivalent instead of identical. However, if the tests are to ensure the correct behavior of the notebooks, then we should modify the hardcoded values and the notebooks. 

In general, `FilledGraph`s and `fill_ports_for_minimal_simulation()` may behave differently, and the number of filled graphs may be different from the old algorithm, unless `search_small_area_observables=True` (even so, the order can be different), which is not exposed in the `BlockGraph` interface and not scalable. Thus, I modified the Steane encoding construction to not rely on `fill_ports_for_minimal_simulation()`. The filled graph tests are untouched and failing.

Let me know what you think and how we should modify the tests! I'm also available for meetings to better explain things.

inmzhang:
Hi Tianyi, good job! I'll review the code at Thursday.

I noticed that you add a placeholder argument for vertex order. Do you have any idea of how to calculate the vertex order without using the gflow-based approach that can share here? I'm really curious about that.

HaoTy:
> I noticed that you add a placeholder argument for vertex order. Do you have any idea of how to calculate the vertex order without using the gflow-based approach that can share here? I'm really curious about that.

Rather than flow-like orderings, I envision this argument to be more like an ordered partitioning of the graph, which is derived with extra information from the ZX graph construction, e.g., from the original circuit for circuit-induced ZX graphs, from positional information for block-graph-induced ZX graphs, etc. My current understanding is that these partitions must be circuit-like (can only go through optimizations preserving certain flow properties) for a chance to be linear in everything; otherwise, the algorithm can only be linear in the number of vertices but still polynomial in the number of leaves, which is proportional to the number of T gates due to T injections. 

It is also related to the execution-time computation of correlation surfaces that is required for conditional operations. The interface may change, though, to work with the conditional block graph I just started implementing.

src/tqec/interop/pyzx/correlation.py
```diff
@@ -88,218 +124,690 @@ def find_correlation_surfaces(
...
+def _multiply_pauli_graphs(pauli_graphs: list[PauliGraphBase]) -> PauliGraph:
+    """Multiply multiple correlation surfaces together."""
+    result = PauliGraph()
+    others = pauli_graphs[1:]
+    for v, neighbors in pauli_graphs[0].items():
+        result_neighbors = result.setdefault(v, {})
+        other_neighbor_rows = [pg[v] for pg in others]
+        for n, pauli in neighbors.items():
+            for neighbor_row in other_neighbor_rows:
+                pauli ^= neighbor_row[n]  # noqa: PLW2901
```
inmzhang, line 300:
```suggestion
            pauli = reduce(operator.xor, (nr[n] for nr in other_neighbor_rows), pauli)
```

inmzhang (commented):
Hi Tianyi, the algorithm and implementation look solid. Please update the failing tests however works best for your use case. Just make sure to adjust the documentation notebooks as well.

HaoTy:
> Hi Tianyi, the algorithm and implementation look solid. Please update the failing tests however works best for your use case. Just make sure to adjust the documentation notebooks as well.

Thanks! What's your opinion on the `reduce_to_minimal_generators` argument? (See quote below)

> To be backward compatible, now a full set of correlation surfaces will be constructed from the generators if `reduce_to_minimal_generators=False`. Nonetheless, I suggest we remove it or rename it to indicate the "construct from generators" logic.

inmzhang:
> What's your opinion on the `reduce_to_minimal_generators` argument?

I intend to remove it as it is the default behavior now.

src/tqec/interop/pyzx/correlation.py
HaoTy: My initial implementation of this function was even more concise (https://github.com/tqec/tqec/pull/707/changes/1c9afb58ba85bfd77c6fccdd5af76b8509dfb98a). After profiling on graphs with thousands of vertices, it turns out that this function is the bottleneck of the algorithm, and the slight overhead of writing everything concisely becomes noticeable at that scale. I don't exactly remember if `reduce` is part of the overhead; let me verify, and if yes, I will add a comment to document the intentionally verbose implementation.

HaoTy:
The current Y-elimination method cannot get rid of the Y in generators that need to contain both X and Z (but do not need any Y). For example, it cannot turn $\\{IZIZ, XYYX, XZXI, ZIZI\\}$ into $\\{IXZX, ZIZI, IZIZ, XIXZ\\}$ in the case of the CZ gate. I have a still efficient solution in mind that should work, but needs some time to implement and test.

inmzhang (approved):
LGTM! Let me know if you're ready for me to merge this.

HaoTy:
Thanks! Polished it a little bit more; ready to merge now!

A few further questions:
1. Up next, I want to reimplement `CorrelationSurface`'s interface with `PauliGraph`, because the former is in many ways inefficient and cannot leverage the internal methods designed for the latter, which can come in handy, say, when I want to reform the generating set in dynamic compilation. Is there any place in the source code that relies on `CorrelationSurface` being immutable or hashable?
2. Should `Pauli` and `PauliGraph` be moved to somewhere else (like enums.py) and/or renamed (e.g., with a leading underscore)? 
3. On a related note, why does the correlation surface computation reside in interop/pyzx instead of computation/correlation.py?

inmzhang:
>  Is there any place in the source code that relies on `CorrelationSurface` being immutable or hashable?

The hashable property was used to deduplicate correlation surface which is not required now. I can't think of anywhere else that requires it to be immutable or hashable.

> 2. Should `Pauli` and `PauliGraph` be moved to somewhere else (like enums.py) and/or renamed (e.g., with a leading underscore)?

There is a `PauliBasis` enum defined in https://github.com/tqec/tqec/blob/020266133acf5e0ebbd444b0c365ee14a7873eb0/src/tqec/plaquette/rpng/rpng.py#L12

 I remember that we use the string value of this enum in several places. I do think the two can be unified as the int enum you defined which is much more powerful in my opinion, and refactor into some enum module,. If we finally replace `CorrelationSurface` with `PauliGraph`, it may be better to rename it directly as correlation surface, which gained its familirity in the community.

> 3. On a related note, why does the correlation surface computation reside in interop/pyzx instead of computation/correlation.py?

At that time, I depicted that we would finally compute the correlation surface with the aid of some builtin features from PyZX. Therefore, the whole computation of correlation surface resides in the `interop` module. It can definitely be refactored outside of it now, like in `computation` module.

HaoTy:
>  If we finally replace `CorrelationSurface` with `PauliGraph`, it may be better to rename it directly as correlation surface, which gained its familirity in the community.


Yes, that's the plan. I think we can merge this PR for now, and I can address these additional potential changes in a future PR.

PR #709, opened 2025-09-29
Add ipykernel

purva-thakre:
This PR adds ipykernel as a dependency because I was having trouble working with a jupyter notebook in a uv venv. `import tqec` would fail with `ModuleNotFoundError: no module named tqec`. 

With the dependency, I was able to create a separate kernel for tqec. Now, I can import tqec without any issues.


purva-thakre:
Going to go ahead and close this PR. I think this issue was specific to my local virtual environment.

PR #713, opened 2025-09-30
Minor details

### LINKED ISSUE #712: Couple minor bugs
## Describe the bug

While carrying out integration tests, I ran into a couple of very minor fixes that seem nonetheless important to correct.

- The [contributors guidelines](https://github.com/tqec/tqec/blob/main/docs/contributor_guide.rst) do not make explicit mention of `venv` creation. While anyone contributing is in theory aware of `venv` and its benefits, it's one of those things that have become so mechanical that one can easily forget. Best to add it. 
- There is a typo in the[ logical_cnot example](https://github.com/tqec/tqec/blob/main/examples/logical_cnot.py), probably left after some debugging. The num_workers is set to 20, whereas it probably should be set to `cpu_count` or not included at all being that it's an optional parameters that defaults to `cpu_count` 

## Steps to reproduce the behavior

Check the files linked above. 

## Expected behavior

Include `venv` creation in contributors guidelines.
Remove `num_workers` from logical_cnot example. 

## Python dependency versions

They're minor details. I will push a branch with fixes soon. I just didn't want to push a branch and then ask a merge without an issue, for clear historical documentation.

jbolns:
A couple very minor details:
- add explicit mention of venv creation to contributors guide
- remove typo from logical_cnot example

Closes #712


docs/contributor_guide.rst
```diff
@@ -28,17 +28,27 @@ of ``tqec`` through ``pip`` or ``uv``.
...
+
+            # Go in the tqec directory
+            cd tqec
+
+            # Create an environment (optional but highly recommended)
+            python -m venv .venv
+            # Activate the environment
+            source .venv/bin/activate  # GNU/Linux and MacOS
+            .venv\Scripts\activate.bat # Windows (CMD)
+            .venv\Scripts\Activate.ps1 # Windows (PowerShell)
```
purva-thakre: we shouldn't specify a particular way to activate a virtual environment. There are multiple ways to create a venv for the same platform. what you have is just one such option.  

Make a comment before this code block stating we suggest use of a virtual environment and add links for some options for different platforms.

examples/logical_cnot.py
```diff
@@ -37,7 +37,6 @@ def generate_graphs(convention: Convention, observable_basis: Basis) -> None:
         manhattan_radius=2,
         convention=convention,
         observables=correlation_surfaces,
-        num_workers=20,
```
purva-thakre, line 40: what is the motivation behind removing this? 

You might want to take a look at the examples in the gallery. They rely on `cpu_count()`. My suggestion would be to change this line with the function `from multiprocessing import cpu_count`

docs/contributor_guide.rst
jbolns: Done. Linked most common approach. Anyone who prefers a different approach, doesn't need a link.

examples/logical_cnot.py
jbolns: No need. `cpu_count()` is the default when no parameter is given.

jbolns:
Just added a tiny additional commit to .gitignore. 
- examples_database is user generated, upon simulation 
- it can end up placed basically anywhere in the repo
- Best to ignore regardless of location to avoid anyone accidentally pushing it (I almost did)

examples/logical_cnot.py
purva-thakre: Yes, I am aware. It is better to make this explicit by adding the line.

docs/contributor_guide.rst
```diff
@@ -21,24 +21,31 @@ recommended to use an editable installation.
 Currently, ``tqec`` is compatible with Python 3.10, 3.11, 3.12 and 3.13. You can install the editable version
 of ``tqec`` through ``pip`` or ``uv``.
 
+.. NB!::
+    Creating an environment before running `pip install` is optional but recommended to avoid everything installing globally.
+    Click [here](https://docs.python.org/3/library/venv.html) for a common approach.
+
```
purva-thakre: this does not appear in the docs build because the `NB!` block is not a valid format. If you scroll down, you can see two options for an attention and a warning block. 

https://tqec.github.io/tqec/pull/713/contributor_guide.html

Either way, I think this should not be a highlighted block. The installation guide already has a similar line right above the installation code block. 

https://tqec.github.io/tqec/user_guide/installation.html#optional-but-highly-recommended-create-a-new-environment

examples/logical_cnot.py
purva-thakre: After #716 is merged, we don't need to worry about this file. Disregard my previous suggestion.

jbolns: Reverted to full line with cpu_count. Pushed already.

docs/contributor_guide.rst
jbolns: I thought you could write anything between the dots. Tiny bit dissapointed with this Sphinx language after finding out that's not the case. 
But anyway, found a favourite amongst their comment-like classes: "hint". I like the emoticon. Let's use that. :)

docs/contributor_guide.rst
```diff
@@ -21,24 +21,31 @@ recommended to use an editable installation.
 Currently, ``tqec`` is compatible with Python 3.10, 3.11, 3.12 and 3.13. You can install the editable version
 of ``tqec`` through ``pip`` or ``uv``.
 
+.. hint::
+    Creating an environment before running `pip install` is optional but recommended to avoid everything installing globally.
```
purva-thakre:
```suggestion
    Creating an environment before running ``pip install`` is optional but recommended to avoid everything installing globally.
```

docs/contributor_guide.rst
```diff
@@ -21,24 +21,31 @@ recommended to use an editable installation.
...
+    Click [here](https://docs.python.org/3/library/venv.html) for a common approach.
```
purva-thakre:
```suggestion
    Click `here <https://docs.python.org/3/library/venv.html>`_ for a common approach.
```

purva-thakre (approved):

PR #714, opened 2025-10-01
Fix code block syntax of `uv`/`pip` install in readme

inmzhang:


README.md
```diff
@@ -37,19 +37,16 @@ Documentation is available at <https://tqec.github.io/tqec/index.html>
 
 Currently, you need to install `tqec` from source with `pip` or [`uv`](https://docs.astral.sh/uv/):
 
-````{tab-set-code}
```
purva-thakre, line 40: why remove the tab code option?

inmzhang: Does it render correctly at the Github markdown render environment? And It seems clear enough to list the two options without tabs.

purva-thakre: > Does it render correctly at the Github markdown render environment?

Yes, I defined the following for it:

https://github.com/tqec/tqec/blob/4650643d0926f4e99043128b975cb8b0fd32d7fc/docs/conf.py#L152-L158

You can also see it in the [readme](https://tqec.github.io/tqec/#installation):

image

inmzhang: It only works for docs environment but not for Github default render environment: https://github.com/tqec/tqec?tab=readme-ov-file#installation.

inmzhang:
Wrongly pushed some commits into this branch... force push to remove them.

README.md
purva-thakre: Ok, well can you also remove the linked lines in conf.py? The lines were only added for the readme.

purva-thakre (approved):

PR #716, opened 2025-10-02
Remove examples

### LINKED ISSUE #715: Why is there an `examples` folder outside of `docs`?
I am referring to `tqec/examples`, which contains a logical CNOT example as well as a jupyter notebook. Would it not be better to move these to the docs or remove them if they already exist in the docs?

@inmzhang

purva-thakre:
Fixes #715


examples/notebooks/create_plaquette.ipynb
```diff
@@ -1,259 +0,0 @@
-{
```
purva-thakre, line 1: this notebook is terribly out of date. It is a good idea I pushed for executable docs in #592, but unfortunately missed this one.

inmzhang: Then maybe we can simply remove it.

purva-thakre: Let me try to make changes to the notebook. If I am not successful, I am leaning towards removing it.

purva-thakre: @inmzhang Do you happen to know why this was added as a jupyter notebook and not unit tests for the `RPNG` class?  I spent some time making changes to the notebook, but realized that there is a significant overlap with this notebook and `src/tqec/plaquette/rpng/rpng_test.py`.

I see that Adrien and Gian were responsible for this notebook in #411 .

inmzhang (approved):
LGTM.

examples/notebooks/create_plaquette.ipynb
inmzhang: Sorry I don't remember.

PR #720, opened 2025-10-04
Framework integration

jbolns:
Not ready to merge. Simply draft PR-ing for discussion.


docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "27f0f36b",
+   "metadata": {},
+   "source": [
+    "This notebook illustrates the process of importing into TQEC circuits originally created with external NISQ circuit designe software/frameworks, such as, for instance, Qiskit and Qrisp.\n",
+    "\n",
+    "**NB! prerequisite.** The notebook assumes you have installed ***topologiq***: `pip install git+https://github.com/tqec/topologiq.git`.\n",
+    "\n",
+    "> **Additional temporary prerequisite.** While this notebook does not use them directly, for the time being, you also need to install the library of the NISQ circuit design framework you are looking to use. Currently supported frameworks: Qiskit (`pip install qiskit`), Qrisp (`pip install qrisp`)."
```
purva-thakre: you will have to add toplogiq, qrisp, and qiskit as docs dependencies for this notebook in the `pyproject.toml` file. The docs build will fail due to this.

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "5986b007",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import random\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "%matplotlib widget\n",
```
purva-thakre: right now the docs build is failing because it does not like this line. 

https://github.com/tqec/tqec/actions/runs/18241486912/job/51943885803?pr=720#step:7:97

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8914774a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from topologiq.scripts.runner import runner\n",
+    "from topologiq.utils.interop_pyzx import pyzx_g_to_simple_g\n",
```
purva-thakre: `build-and-test-python` workflows are failing because `uv/ty` cannot find the `topologiq` installation. Looks like topologiq should be a tests dependency as well as docs dependency in the toml file. 

https://github.com/tqec/tqec/blob/df953e6e1c28ab7db8be3a87dd3a3c2405dd1bd0/pyproject.toml#L72

https://github.com/tqec/tqec/blob/df953e6e1c28ab7db8be3a87dd3a3c2405dd1bd0/pyproject.toml#L65

purva-thakre: Or a better option is to add topologiq to the main set of dependencies. 

https://github.com/tqec/tqec/blob/df953e6e1c28ab7db8be3a87dd3a3c2405dd1bd0/pyproject.toml#L38

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "\n",
+    "ASSETS_FOLDER = Path(\"../../assets/\").resolve()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "27f0f36b",
+   "metadata": {},
+   "source": [
+    "This notebook illustrates the process of importing into TQEC circuits originally created with external NISQ circuit designe software/frameworks, such as, for instance, Qiskit and Qrisp.\n",
```
purva-thakre:
```suggestion
    "This notebook illustrates the process of importing into TQEC circuits originally created with external NISQ circuit design software/frameworks, such as, for instance, Qiskit and Qrisp.\n",
```

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "# Draw graph of incoming circuit\n",
+    "zx.draw(zx_graph, labels = True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "de2c9b1d",
+   "metadata": {},
+   "source": [
+    "The incoming circuit is not optimised. Topologiq can handle an unoptimised circuit, but the space-time volume of the diagram produced by topologiq is increasingly proportional to the size of the input graph. It is therefore a good idea to optimise the incorming circuit. \n",
```
purva-thakre:
```suggestion
    "The incoming circuit is not optimised. Topologiq can handle an unoptimised circuit, but the space-time volume of the diagram produced by topologiq is increasingly proportional to the size of the input graph. It is therefore a good idea to optimise the incoming circuit. \n",
```

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ad835257",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Apply states\n",
+    "# Credits. C, Purva Thakre\n",
```
purva-thakre:
```suggestion

```
no need for this. Thanks for being thoughtful!

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "num_apply_state = zx_graph.num_inputs()\n",
+    "zx_graph.apply_state('0' * num_apply_state)\n",
+    "\n",
+    "# Apply post-select only to the outputs of the ancilla qubits\n",
+    "# Credits. C, Purva Thakre\n",
```
purva-thakre:
```suggestion

```
no need for this. Thanks for being thoughtful!

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "            list(np.logspace(-4, -1, 10)),\n",
+    "            NoiseModel.uniform_depolarizing,\n",
+    "            manhattan_radius=2,\n",
+    "            observables=correlation_surfaces_for_simulation,\n",
+    "            max_shots=1_000_000,\n",
+    "            max_errors=5_000,\n",
+    "            decoders=[\"pymatching\"],\n",
+    "            print_progress=False,\n",
+    "            save_resume_filepath=Path(\n",
+    "                f\"../_examples_database/steane_stats_{support_observable_basis.value}.csv\"\n",
```
purva-thakre: what are your thoughts on renaming this? There already exists a database with this name due to `docs/gallery/steane_encoding.ipynb`. It shouldn't matter that much but this is the demo for topologiq.

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "            ax.grid(axis=\"both\")\n",
+    "            ax.legend()\n",
+    "            ax.loglog()\n",
+    "            ax.set_title(\"Logical CNOT Error Rate\")\n",
+    "            ax.set_xlabel(\"Physical Error Rate\")\n",
+    "            ax.set_ylabel(\"Logical Error Rate\")\n",
+    "            fig.savefig(\n",
+    "                SAVE_DIR\n",
+    "                / f\"logical_cnot_result_{support_observable_basis}_observable_{i}.png\"\n",
+    "            )\n",
```
purva-thakre: I think you forgot to rename these.

docs/gallery/using_nisq_circuits.ipynb
```diff
@@ -0,0 +1,530 @@
...
+    "\n",
+    "# Function to cal simulation cycle for all relevant bases\n",
```
purva-thakre:
```suggestion
    "# Function to call the simulation cycle for all relevant bases\n",
```

src/tqec/interop/nisq/load.py
```diff
@@ -0,0 +1,134 @@
...
+    qc.cx(1, 7)
+    qc.cx(1, 8)
+    qc.h(1)
+
+    qc.h(2)
+    qc.cx(2, 3)
+    qc.cx(2, 5)
+    qc.cx(2, 7)
+    qc.cx(2, 9)
+    qc.h(2)
```
Zhaoyilunnn: Maybe we could move qiskit/qrisp circuits into assets folder? As we are not sure whether we should add these external frameworks as optional dependencies, it seems that we should avoid `import qiskit/qrisp` statements in `src` folder.

Use native format like `qpy` or just python file are both fine to me.

jbolns: Yes, we should add it as .qpy, `steane.qpy` should do it. Do you want to do it or should I do it? I'm fine either way. 

In a conversation with @afowler we were surprised that qiskit is not a default dependency already. It used to be, no? Since it's the major such framework, it makes sense to ship it default. At the very least, it keeps our own versioning in check.

As the notebook runs qiskit by default, this suffices to avoid runtime errors.

Other frameworks... Let's pin that until after the next push in this draft PR. I have a suggestion, but best not to mix topics.

afowler:
My surprise was actually about cirq, not qiskit, since I thought cirq was
required for stim. It's possible this tool simply gets installed when you
installed stim.

src/tqec/interop/nisq/load.py
Zhaoyilunnn: Maybe it would be easier for you to do so. I thinkd I don't have write access to this branch. Just created the `steane.qpy` file: https://drive.google.com/file/d/12984bNVUZWTwr_KOosTaVPkIyY_dHrol/view?usp=drive_link here to facilitate the process.

It can be loaded by 

```python
from qiskit import qpy

with open("steane.qpy", "rb") as fd:
    qc = qpy.load(fd)[0]
```

jbolns: Done. Just pushed. I'll add some general comments as comment below.

docs/gallery/using_nisq_circuits.ipynb
jbolns: Just made a new push considering this.

docs/gallery/using_nisq_circuits.ipynb
jbolns: Just made a new push considering this.

docs/gallery/using_nisq_circuits.ipynb
jbolns: Just made a new push considering this. Will add some relevant notes as a general comment at the end of current conversation, as it's applicable to a few suggestions.

docs/gallery/using_nisq_circuits.ipynb
jbolns: Commented it out. If running locally, this is needed for interactive 3D visualisations, so don't want to completely delete. But it makes sense that it won't run in the docs, as, even locally, it requires authorising the widget download.

docs/gallery/using_nisq_circuits.ipynb
jbolns: Just made a new push considering this. Will add some relevant notes as a general comment at the end of current conversation, as it's applicable to a few suggestions.

jbolns:
> cirq, not qiskit

My bad, I mixed them up. 

Cirq is not installing at all, even after my last push containing additional dependencies. 

I was going to add it explicitly but:
- we don't yet have a cirq steane (in that conversation I mixed up acronyms, we have qiskit and qrisp)
- there's a versioning problem with cirq that would require loosening TQEC's main Python requirements (to include 3.10) and I can't immediately tell if this could have collateral effects.

jbolns:
So, the latest push has a few things worth noting explicitly. 

**Dependencies**
I added a new dependency group to `pyproject.toml`. 
- A foundational install wouldn't install topologiq, nor qiskit, nor qrisp, not any other framework we might add in the future. 
- If you add the `docs` dependencies, it would install all pre-existing doc dependecies plus topologiq and qiskit because that's all the notebook needs to run in its default configuration
- There is an additional `integration` group with topologiq and all supported frameworks, which is what you would install if running comprehensive integration tests locally (or, more practically, you would `uv add <applicable_framework>` only the framework you're interested in). 

The notebook contains instructions to clarify all this. It's easier to explain for an user than to give the full summary above, as they just need to know they need to run the group or add their chosen framework library.  

**Notebook & circuits**
Additionally:
- Uploaded the .qpy Yilun linked above to `./assets`
- Updated the nisq importer and notebook to reflect this
- Updated notebook to consider all feedback in this thread
- Updated notebook to consider above-noted changes to dependencies.

src/tqec/interop/nisq/load.py
purva-thakre: I might be wrong, but you can't load Qrisp circuits from a file yet. You will have to define the circuit in a collapsible code block.

purva-thakre:
> My surprise was actually about cirq, not qiskit, since I thought cirq was
required for stim.

@afowler There is now a separate package called [stimcirq](https://pypi.org/project/stimcirq/). We only install Stim without the additional package.

purva-thakre:
@jbolns Now, the docs build is failing because your notebook needs to be added to `docs/gallery/index.rst`.

pyproject.toml
```diff
@@ -81,6 +81,8 @@ docs = [
   "sphinx-copybutton<=0.5.2",
   "sphinxcontrib-bibtex<=2.64",
   "jupyter-sphinx<=0.5.3",
+  "qiskit>=2.1.1",
+  "topologiq@git+https://github.com/tqec/topologiq.git"
```
purva-thakre: since you added the integration group, you can directly use it here https://packaging.python.org/en/latest/specifications/dependency-groups/

Edit: Also, just an FYI, I am a little bit busy over the next couple of days. Will respond to things after Sunday. Sorry for any delays.

jbolns: I'm not sure I understand. Repository management at this level ain't my forte. 
You mean changing those two lines for "integration" and that triggers the install of the group?

src/tqec/interop/nisq/load.py
jbolns: Latest push solves this, I think. It's now defined in Python, as .py, and simply imported into the notebook. See comment below about potentially needing separate doc files if we add more examples (in which case, perhaps qrisp would be a good candidate for a notebook of its own).

jbolns:
With the exception of the comment I noted explicitly not understanding, the latest push should address all comments thus far. 

**Changelog summary**
The most significant change is that, having externalised the circuits to a corresponding file in `./assets/`, it is now possible to import them directly into the notebook. 

As a consequence, I went ahead and completely eliminated the `./src/tqec/interops/nisq` I had introduced earlier, as it was no longer being used by the notebook. 

**A few additional notes**
While I was very sold on the idea of a single notebook for many frameworks, I tried adding a cirq version using the circuit Angela uploaded to the algorithm's folder. At that point, the notebook did become a bit too long. I think two frameworks fit perfectly into a single notebook, but if we are going to add examples of more frameworks, we might want to consider several framework-specific notebooks. Problem is, this would significantly add simulations to the deployment workflow, one full set of sims per notebook.

Zhaoyilunnn:
Now the demo and code structure look good to me. Nice work J!

afowler:
I think it is okay for a user to focus on a single framework and do a
single set of sims per framework. It is not obvious that two different
frameworks even with in principle the same quantum circuit will lead to
exactly the same lattice surgery and therefore exactly the same simulations.

jbolns:
Latest push splits the notebook into separate Qiskit and Qrisp notebooks. 

Additional notebooks possible by repeating the general recipe in these two notebooks.
- The top of a new notebook would be 100% specific to framework of choice
- The QASM-->PyZX bit might differ for some frameworks, given parsing difficulties
- The PyZX optimisation might vary a little according to the previous step
- The topologiq and TQEC bits are probably almost a copy+paste irrespective of base framework

KabirDubey:
Hi J, 

I am getting the following error at the point of the `qiskit_integration.ipynb` notebook where we use `get_stim_circuit` to get the Stim circuit for different bases. 

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[39], line 39
     37     # Call Stim circuit generation for all relevant bases
     38     if filled_block_graphs:
---> 39         get_stim_circuit(filled_block_graphs)
     41 else:
     42     print("ERROR. Block didn't run because there is no block_graph or it was not possible to create filled block_graphs from it.")

Cell In[39], line 22, in get_stim_circuit(pre_filled_block_graphs)
     20 if block_graph_for_computation:
     21     compiled_graph = compile_block_graph(block_graph_for_computation)
---> 22     stim_circuit = compiled_graph.generate_stim_circuit(
     23         k=1, noise_model=NoiseModel.uniform_depolarizing(p=0.001)
     24     )
     25     print("\nFirst 10 lines of Stim circuit (Basis X):\n")
     26     print(stim_circuit[:10])

File ~/Projects/dev/ytqec/tqec_pr720/src/tqec/compile/graph.py:501, in TopologicalComputationGraph.generate_stim_circuit(self, k, noise_model, manhattan_radius, detector_database, database_path, do_not_use_database, only_use_database)
    465 def generate_stim_circuit(
    466     self,
    467     k: int,
   (...)    473     only_use_database: bool = False,
    474 ) -> stim.Circuit:
    475     """Generate the ``stim.Circuit`` from the compiled graph.
    476 
    477     Args:
   (...)    499 
    500     """
--> 501     circuit = self.to_layer_tree().generate_circuit(
    502         k,
    503         manhattan_radius=manhattan_radius,
    504         detector_database=detector_database,
    505         database_path=database_path,
    506         do_not_use_database=do_not_use_database,
    507         only_use_database=only_use_database,
    508     )
    509     # If provided, apply the noise model.
    510     if noise_model is not None:

File ~/Projects/dev/ytqec/tqec_pr720/src/tqec/compile/tree/tree.py:311, in LayerTree.generate_circuit(self, k, include_qubit_coords, manhattan_radius, detector_database, database_path, do_not_use_database, only_use_database, lookback)
    309 if detector_database is None:  # Nothing passed in,
    310     if database_path.exists():  # look for an existing database at the path.
--> 311         detector_database = DetectorDatabase.from_file(database_path)
    312     else:  # if there is no existing database, create one.
    313         detector_database = DetectorDatabase()

File ~/Projects/dev/ytqec/tqec_pr720/src/tqec/compile/detectors/database.py:507, in DetectorDatabase.from_file(filepath, format)
    501 if format not in DetectorDatabase._READERS:
    502     raise TQECError(
    503         f"Could not read the database: the provided format {format} is not supported. "
    504         + "Supported formats are:\n  -"
    505         + "\n  -".join(DetectorDatabase._READERS.keys())
    506     )
--> 507 return DetectorDatabase._READERS[format](filepath)

File ~/Projects/dev/ytqec/tqec_pr720/src/tqec/compile/detectors/database.py:222, in _DetectorDatabaseIO.from_pickle_file(filepath)
    220 try:
    221     with open(filepath, "rb") as f:
--> 222         database = pickle.load(f)
    223 except pickle.PickleError as e:
    224     moving_location = filepath.parent / f"faulty_database_{hash(e)}.pkl"

AttributeError: Can't get attribute 'ExtendedBasis' on <module 'tqec.utils.enums' from '/Users/kabirdubey/Projects/dev/ytqec/tqec_pr720/src/tqec/utils/enums.py'>
```

I think you need to add these lines to `src/tqec/utils/enums,py`, as in `main`:

```
class PauliBasis(Enum):
    X = "X"
    Y = "Y"
    Z = "Z"

    def __str__(self) -> str:
        return self.value  # pragma: no cover

    def to_extended_basis(self) -> ExtendedBasis:
        """Return ``self`` as an extended basis."""
        return ExtendedBasis(self.value)


class ExtendedBasis(Enum):
    X = PauliBasis.X.value
    Y = PauliBasis.Y.value
    Z = PauliBasis.Z.value
    H = "H"

    def __str__(self) -> str:
        return self.value  # pragma: no cover
```

Also, this is probably because this PR is still in progress, but I wasn't able to use the `docs/_examples_database` pre-computed in the branch to generate the plots. I needed to delete it and then wait it out. I'll make a discussion post trying to find standard operating proceduring for using the database in demos. 

Thanks,
Kabir

docs/gallery/qiskit.ipynb
inmzhang, line 1: It seems better to me that if we can put these notebooks under a new section like `Framework Integration` in the user guide page. The widgets in the gallery are more like well-defined primitive gates or circuits whose structure will not vary between different runs.

jbolns:
Latest push has some minor updates AND updates branch against `main`. 

@KabirDubey Want to re-pull the branch and give it a go again to see if it was something related to applying recent updates on `main` to this branch?

I'm not sure that'll do the trick, for reasons below. But worth trying

Could this be a Mac thing?
- The contents of `src/tqec/utils/enums.py` in the latest version of `main` I've pulled (just now) are different than what you suggest. In particular, there is no `PauliBasis` or `ExtendedBasis` whatsover. 
- The database note is odd. I'm using it as given, and it's been working from the first try in both Linux and Windows machines.
- Can you run other gallery notebooks like, for instance, the [manually-built Steane](https://tqec.github.io/tqec/gallery/steane_encoding.html)?

docs/gallery/qiskit.ipynb
jbolns: This seems appropriate. Additionally, there will be several more if we get more foundational circuits, and there's still a need to add demos for PyZX graphs with Pauli Webs.

Would I add it to the gallery index as `<folder_name>/<notebook_filename>.ipynb`?

inmzhang: Oh, I mean we can create a new section called `Framework Integration` in https://github.com/tqec/tqec/blob/main/docs/user_guide/index.rst and list the notebooks there. However, I’m not sure if the automatic notebook execution will work correctly with this—maybe @purva-thakre can confirm.

jbolns:
> from the first try
* from the first successful try, after I figured out Linux versioning problems (but that wasn't related to the notebooks or the database, it was a flawed numpy re-versioning)

SMC17:
@jbolns - Quick heads up: found and fixed a coordinate transformation bug in `topologiq.py` while testing the integration.

https://github.com/tqec/tqec/pull/726

jbolns:
Merged a fork-to-branch PR just now. #726, by @SMC17. 

It's failed some automatic checks, but:
- the import function is a significant improvement to existing one.
- it's adds a comprehensive set of tests. 

Will pull and see about fixing whatever caused the failed checks.

jbolns:
I have to say I do not currently understand why the type checks are failing.

There were two types of error.
- Something related to variable naming. That is no longer appearing. 
- Something related to how I call topologiq's main runner. 

It shouldn't be failing. 
- It's unlikely related to the merge by @SMC17 because they failing files were not touched by that merge. 
- The runner function hasn't changed from the last working version.
- I did remove one line, but then I introduced the line back in (with a different `fig_data` payload but nonetheless the same kind of object as was being passed in last working version). 
- It works on my computer. =D
- The GH Pages deployment succeeds, which I think means it does build.

Will sleep on it, but if anyone's got a clue what is going on, by all means. 

@purva-thakre @Zhaoyilunnn ?

Azuya334:
Hi, I would like to work on some tasks here.

jbolns:
@Azuya334 , as you have a physics background, would you happen to know how to design a circuit in any of the following?
- [https://quantumai.google/qsim](https://quantumai.google/qsim)
- [https://github.com/ProjectQ-Framework/ProjectQ](https://github.com/ProjectQ-Framework/ProjectQ)
- [https://github.com/CQCL/hugr](https://github.com/CQCL/hugr)
- [https://github.com/CQCL/guppylang](https://github.com/CQCL/guppylang)
- [https://github.com/CQCL/selene](https://github.com/CQCL/selene)

The main ask underneath this PR is to create a Steane encoding using one framework, export it to QASM, and see if it loads to PyZX (qsim might load directly to PyZX without a need for QASM, though).

These frameworks above have not yet been taken. 

Example implementations in `docs/gallery/qiskit_integration.ipynb` and `docs/gallery/qrisp_integration.ipynb`, with the corresponding circuits `assets/steane_qiskit.qpy` and `assets/steane_qrisp.py`.

Azuya334:
@jbolns I learned to use qsim a while ago for a class. I can go pick it up. I am not too familiar with implementing Steane code, I will go read the examples.

Zhaoyilunnn:
> I have to say I do not currently understand why the type checks are failing.
> 
> There were two types of error.
> 
> * Something related to variable naming. That is no longer appearing.
> * Something related to how I call topologiq's main runner.
> 
> It shouldn't be failing.
> 
> * It's unlikely related to the merge by @SMC17 because they failing files were not touched by that merge.
> * The runner function hasn't changed from the last working version.
> * I did remove one line, but then I introduced the line back in (with a different `fig_data` payload but nonetheless the same kind of object as was being passed in last working version).
> * It works on my computer. =D
> * The GH Pages deployment succeeds, which I think means it does build.
> 
> Will sleep on it, but if anyone's got a clue what is going on, by all means.
> 
> @purva-thakre @Zhaoyilunnn ?

It also works on my computer (linux ubuntu 22.04) with the same python version, ty version, and other package versions as the github ci env. I think this may be a bug of ty? I also don't understand why. 

Since the error occurs in the notebook of docs directory, I would suggest to skip ty checking on these notebooks. Because, in any way, they are not the source of tqec framework itself. I don't know if others agree on this point. 

https://github.com/tqec/tqec/pull/728

SMC17:
This topologiq PR (tqec/topologiq#17) directly supports the Qiskit/Qrisp integration work:

**Pipeline Strengthening:**
The Qiskit/Qrisp integration relies on the PyZX → topologiq → TQEC pipeline. This work strengthens the topologiq component:

1. **Performance:** 1.78x faster processing reduces pipeline latency
2. **Reliability:** Comprehensive testing validates topologiq behavior
3. **Debugging:** Visualization tools help debug graph conversions
4. **Reproducibility:** Forced parameters enable pipeline debugging

**Testing Patterns:**
The testing patterns in this PR can inform integration testing:
- Parametrized tests for variant validation
- Reproducible test execution with forced parameters
- Statistical validation of performance improvements
- Zero-warning test output standards

**Integration Debugging:**
If the Qiskit/Qrisp demos encounter topologiq issues:
```python
# Reproduce specific pipeline behavior
result = graph_manager_bfs(
    circuit,
    force_src_id=specific_id,
    force_src_kind=specific_kind,
    ...
)

# Visualize topologiq's graph processing
visualize_exploration_3d(explored, path, src, tgt)
```

Happy to help debug any topologiq-related integration issues!

jbolns:
I took weekend of, @SMC17, so I didn't answer your comment from yesterday asking what to prioritise. Apologies for that. 

Very excited to start looking at [topologiq's 17](https://github.com/tqec/topologiq/pull/17) this coming week. It looks very cool!

For clarity in expectations. I was able to merge #726 to this PR without extensive review because it was a merge-to-branch. We will still review all code in this PR extensively prior merging to main. [Topologiq's 17](https://github.com/tqec/topologiq/pull/17) is a merge-to-main, so I *will* need to go over everything manually and with a lot of patience.

jbolns:
@inmzhang Is the new folder structure in last push along the lines of what you'd suggested. 
Also, I still don't get why `uv run ty check` is failing here. When I test locally, the checks clear. @purva-thakre ?

purva-thakre:
@jbolns Sorry for the late reply. I was traveling and ended up sick as well. I will go through the comments in this PR today.

assets/steane_qrisp.py
```diff
@@ -0,0 +1,48 @@
+# The following is how you would generate a Steane code using Qrisp.
```
purva-thakre: Please add some comments along the lines of this file is used by so and so notebook in the documentation. Same for any other files you add to `assets`. 

it might also be easier if you instead create an `assets` folder in the docs directory.

purva-thakre:
@jbolns The docs build is failing due to the messages provided here: https://github.com/tqec/tqec/actions/runs/18498220732/job/52707752197?pr=720#step:7:102

docs/integration/index.rst
```diff
@@ -0,0 +1,13 @@
+.. _integration-reference-label:
+
+Integration Gallery
+===================
+
+This section contains examples of how to use (NISQ) circuits designed with external frameworks including, but not limited to, Qiskit and Cirq.
+
+.. nbintegration::
```
purva-thakre: I do not think it likes this line. It is explicitly marked as a reason for failure in https://github.com/tqec/tqec/actions/runs/18498220732/job/52707752197?pr=720#step:7:102. 

`docs/gallery/index.rst` uses `nbgallery`.

https://github.com/tqec/tqec/blob/2a196b8cab2a2527bacb9971c427945245645a4f/docs/gallery/index.rst?plain=1#L8

purva-thakre: You also need to add the new `index.rst` file to `docs/gallery/index.rst` if you want to list the integration examples on the gallery page.

pyproject.toml
purva-thakre: See the example at the provided link:

Image

purva-thakre:
> @inmzhang Is the new folder structure in last push along the lines of what you'd suggested. Also, I still don't get why `uv run ty check` is failing here. When I test locally, the checks clear. @purva-thakre ?

@jbolns You might want to check the version of `ty` you have locally. I have it setup such that the latest version of uv is installed in the CI. What is installed by the CI for this PR is screenshotted below:

image

Looks like a new version tag was added 5 days ago. https://github.com/astral-sh/ty/tags

image

src/tqec/interop/nisq/load.py
purva-thakre: Ah, yes, you can do that. I was referring to how you can load a qiskit qasm circuit. But what you have done is ok too.

purva-thakre:
> I think it is okay for a user to focus on a single framework and do a single set of sims per framework. It is not obvious that two different frameworks even with in principle the same quantum circuit will lead to exactly the same lattice surgery and therefore exactly the same simulations.

I agree with @afowler comment mostly from the POV of how long it will take for the docs build to run if we demo the Steane code circuit for all frameworks. 

What are everyone's thoughts on having one notebook only for framework integration, where the code randomly selects which framework's circuit to use, prints a couple of lines about this, and implements the integration pipeline?

docs/gallery/qiskit.ipynb
purva-thakre: > Oh, I mean we can create a new section called `Framework Integration` in https://github.com/tqec/tqec/blob/main/docs/user_guide/index.rst and list the notebooks there. However, I’m not sure if the automatic notebook execution will work correctly with this—maybe @purva-thakre can confirm.

It should. I added some comments which will hopefully fix the docs failures. https://github.com/tqec/tqec/pull/720#discussion_r2430971401

assets/steane_qrisp.py
jbolns: Done. Pushes soon. Let's pin the assets folder discussion for once we decide placement and number of notebooks. Putting circuits in the assets folder, wherever place, made sense under the assumption most frameworks have a native file format. Since some but not all do, I'm wondering if we need a third approach we haven't yet considered.

docs/integration/index.rst
jbolns: Oh, so that's a pre-established tag. I thought it was the section tile. I really need to learn some Sphinx syntax. 🤣
- Changed line. 
- Added integration gallery to docs index, to see if it displays as full section. I think that's what @inmzhang meant by separate gallery (which makes sense if we only use a single notebook, else we put back to main gallery if there's only one integration notebook).

src/tqec/interop/nisq/load.py
jbolns: I think this one is resolved.

pyproject.toml
jbolns: Last push has this, I think.

docs/gallery/qiskit.ipynb
jbolns: Last push follows this structure. It appears to have built, but I can't seem to find where the build can be previewed.

jbolns:
The ty issue is resolved. I reckon it deserves a explicit summary. Resolved by including all optional parameters in function call. It was failing, seemingly, because some optional parameters were deemed necessary, which then messed up the position where it expected other parameters.

jbolns:
**Summary of last push.**
- Improves dependency management
- Improves readability of notebooks
- Eliminates ty failures
- Realises the multi-gallery approach: standard gallery + integration gallery. 

**Pending**
Per my count, the pending issues to discuss are:
- Optimising tests introduced in #726 by using parametrize to condense multiple unit tests like this into a single unit test
- Reducing docstrings of `.../interops/pyzx/topologiq.py` bug-fix introduced in #726 to the standard TQEC length and style
- Deciding between ONE or MANY notebooks approach, alongside their placement. 

**One or many?**

> I think it is okay for a user to focus on a single framework and do a single set of sims per framework. It is not obvious that two different frameworks even with in principle the same quantum circuit will lead to exactly the same lattice surgery and therefore exactly the same simulations.

I think @afowler meant it is preferable to use several notebooks focusing each on a single framework. This avoids a Qiskit user having to worry about Qrisp/pytket/Cirq code and vice-versa. 

> One notebook only for framework integration, where the code randomly selects which framework's circuit to use

This is similar to my initial idea of letting users select a framework, with the improvement that this would automatically run different frameworks every time there's a merge. So, I can't say I fundamentally disagree. It would be fine. 

I'm siding with different notebook for each framework tested successfully because they're easier to read by users. 

Having said that, I would prefer if the designers of the circuits for this or that framework prepare and push their respective notebooks. Not too worried about the Qiskit and Qrisp notebooks since both @purva-thakre and @Zhaoyilunnn are established contributors. But newer contributors who go over the pain of designing a circuit and testing it should get the lines/credit.

docs/integration/index.rst
purva-thakre: > I think that's what @inmzhang meant by separate gallery (which makes sense if we only use a single notebook

I am confused. I though we were adding a new section to the examples gallery (similar to below):

image

docs/integration/qiskit.ipynb
```diff
@@ -0,0 +1,574 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "27f0f36b",
+   "metadata": {},
+   "source": [
+    "This notebook illustrates the process of importing into TQEC circuits originally created with Qiskit.\n",
+    "\n",
+    "The notebook assumes you have already designed a Qiskit circuit and saved it using Qiskit's `.qpy` format. \n",
+    "\n",
+    "> **NB!** Run `uv add --only-group integration` to install the libraries used in this notebook (ignore if you installed all dependency groups when setting up TQEC)."
```
purva-thakre: You cannot use this in sphinx. 

```suggestion
    .. note::
        Run `uv add --only-group integration` to install the libraries used in this notebook (ignore if you installed all dependency groups when setting up TQEC)."
```

purva-thakre:
> Having said that, I would prefer if the designers of the circuits for this or that framework prepare and push their respective notebooks. 

If you have access to their notebooks in the shared folder, you can add them as co-authors in github. https://gist.github.com/lisawolderiksen/f9747a3ae1e58e9daa7d176ab98f1bad

reach out to each person for their github handle.

purva-thakre:
> This is similar to my initial idea of letting users select a framework, with the improvement that this would automatically run different frameworks every time there's a merge. So, I can't say I fundamentally disagree. It would be fine.

> I'm siding with different notebook for each framework tested successfully because they're easier to read by users.

My primary motivation for this is that the docs build time is going to increase per framework notebook. Locally, if I build the docs for the first time (without the cache), each Steane circuit simulation is going to take at least 1 hour per notebook (the Steane blockgraph notebook takes half this time on my thinkpad). If I look at the frameworks considered in Austin's presentation, this PR will add at least 6 hours for the first local build (if the Steane simulation per framework takes just an hour). 

The time in the CI is lower due to caching, for now, because the cache exists[^1]. But for the cache to exist locally, you have to at least build it once i.e., wait 6+ hours the first time. Plus, you would be against using `make clean` locally because the docs build time has increased so much. 

Edit: I just pulled the notebooks in this PR locally. Each notebook takes approximately 25-30 minutes on my Thinkpad with `make clean html` and 15 minutes with `make html`. The latter is using the locally saved cache. Multiply these times by 6-7 for each framework. Assume double the time for the CI (without a proper cache). 


[^1]: Plus, something is wrong with the CI cache. I won't be able to look it over until after next week.

docs/integration/index.rst
```diff
@@ -0,0 +1,11 @@
+.. _integration-reference-label:
+
+Integration Gallery
+===================
+
+Examples of how to use NISQ circuits designed with external frameworks (e.g., Qiskit, Qrisp, etc.).
```
KabirDubey: What about these circuits are NISQ? I thought their motivation was that they would be core primitives in a fault tolerant qc system.

purva-thakre: lol. I was about to type the same comment.

docs/integration/qrisp.ipynb
```diff
@@ -0,0 +1,581 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "27f0f36b",
+   "metadata": {},
+   "source": [
+    "This notebook illustrates the process of importing into TQEC circuits originally created with Qrisp.\n",
+    "\n",
+    "The notebook assumes you have already designed a Qrisp circuit. \n",
+    "\n",
+    "> **NB!** Run `uv add --only-group integration` to install the libraries used in this notebook (ignore if you installed all dependency groups when setting up TQEC)."
```
purva-thakre: Apply https://github.com/tqec/tqec/pull/720#discussion_r2434133702 here.

purva-thakre:
I am not sure if this is a bug related to `tqec` or `topologiq` but the HTML viewer is really zooming out for the pyzx graphs. Is there a way to apply auto_layout for the inset?

image

Same for below:

image

docs/integration/qiskit.ipynb
purva-thakre: Also, this is assuming the person reading this has `uv` when we provide installation instructions for `pip` and `uv`.

docs/integration/qiskit.ipynb
```diff
@@ -0,0 +1,574 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3bdc28c4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(f\"\\n=> Imported Qiskit circuit of name {circuit_name.capitalize()}:\\n\")\n",
+    "print(qc) if qc else print(\"ERROR. Failed to load a circuit\")"
```
purva-thakre:
```suggestion

```
is there a particular motivation to print an error message if the circuit is not printed correctly? If it can't load the circuit, I imagine there will be an error printed to the console anyway.

docs/integration/qiskit.ipynb
```diff
@@ -0,0 +1,574 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bee986d0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(f\"\\n=> Converted Qiskit circuit of name {circuit_name.capitalize()} into the following QASM string:\\n\")\n",
+    "print(qasm_str) if qasm_str else print(\"ERROR. Failed to convert circuit to QASM\")"
```
purva-thakre:
```suggestion

```
Same question as https://github.com/tqec/tqec/pull/720/files#r2434193440

docs/integration/qiskit.ipynb
```diff
@@ -0,0 +1,574 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "7d784a93",
+   "metadata": {},
+   "source": [
+    "The structure of `lattice_nodes` and `lattice_edges` is essentially the same as TQEC's `block_graph`, with one caveat.\n",
+    "\n",
+    "There is a need to perform some re-scaling and re-indexing of edges. In TQEC's `block_graph`, pipes occupy no space, which leads to adjacent cubes being closer to one another than in a space-time diagram where pipes do occupy space for the purposes of visualising things clearly.\n",
+    "\n",
+    "Fortunately, TQEC already has a way to import 3D objects like `lattice_nodes` and `lattice_edges`. TQEC can import 3D models encoded in COLLADA format (.dae), which use the same kind of coordinates and artifacts as `lattice_nodes` and `lattice_edges`. Accordingly, the following block of code relies on TQEC's COLLADA interoperability to realise similar transformations for `lattice_nodes` and `lattice_edges`, leading into a `block_graph`."
```
purva-thakre:
```suggestion
    "Fortunately, TQEC already has a way to import 3D objects like `lattice_nodes` and `lattice_edges`. TQEC can import 3D models encoded in COLLADA format (`.dae`), which use the same kind of coordinates and artifacts as `lattice_nodes` and `lattice_edges`. Accordingly, the following block of code relies on TQEC's COLLADA interoperability to realise similar transformations for `lattice_nodes` and `lattice_edges`, leading into a `block_graph`."
```

purva-thakre:
Out of curiosity, what further work is needed to reduce the circuit volume in topologiq? I am simply comparing Austin's manually built structure in his slides with the [one from topologiq](https://docs.google.com/presentation/d/1RcMXB3kWtWWyM_LOQLkxo19Zaq4atLaV1fCsWqd78PQ/edit?slide=id.g39455a4c89c_0_581#slide=id.g39455a4c89c_0_581).

afowler:
>
> I think @afowler <https://github.com/afowler> meant it is preferable to
> use several notebooks focusing each on a single framework. This avoids a
> Qiskit user having to worry about Qrisp/pytket/Cirq code and vice-versa.
>
When it comes to precise implementation, I'm always flexible, but in my
mind it would be nice if a user of any algorithm specification tool can
create an output file, and then easily have tqec consume it. Whether that
is achieved with a command line argument and a switch to indicate file
format, or a notebook, or some other method, comes down to
user-friendliness and developer friendliness.

purva-thakre:
Something is not working as expected in the build. Will dig through the code later today. The cache hit is finding older caches when I don't want it to use a cache (to check the initial build time). I guess the fallback option is to clear all caches?

It might be straightforward to think of the one without the cache as `make clean html` and the one with the cache as `make html` locally. 

I tried triggering the build without the cache, and it took shorter than it should. My experience with the Steane simulation is that it should take the longest without the cache locally or in the CI. In this scenario, we have 3 notebooks for the Steane code that should have dominated the build time (but they don't).

afowler:
All ideas are welcome. If anyone has some machine learning
experience, curious to know if that can be adapted to the problem of
mapping a ZX graph to an explicit 3D lattice surgery structure.

Re: more algorithmic approaches, a systematic method of exploring genuinely
different sequences of adding nodes using the current algorithm would be of
great interest. It would also be great to try to systematically generate
genuinely different variants of the ZX graph.

docs/integration/qrisp.ipynb
jbolns: Right, fun story. The sphinx syntax created some issues with how the diff of the other notebook was read, rendering the file unreadable. But I've re-written in both notebooks as normal paragraph.

docs/integration/index.rst
jbolns: I see your point point. The idea was that some of these frameworks are currently communicated as NISQ platforms. But I see how the label doesn't travel well when you use the circuits in a fault-tolerant pipeline. Removed.

docs/integration/index.rst
jbolns: That's exactly what I tried to achieve. 🤣
Not sure if it's been evident, but I'm not particularly getting the whole Sphinx docs rationale. 

Will push commits with all the other fixes but, to avoid trying recipes blind, could you clarify the structure for me. 

Considering the two structural rationales I've pushed are not the intended outcome, the next approach I would try would be: 
- move `qiskit.ipynb` and `qrisp.ipynb` notebooks inside `./docs/gallery`
- edit `docs/gallery/index.rst` as follows

```
This section contains a collection of constructed logical computations and their simulation results.

.. nbgallery::

   memory.ipynb
   cnot.ipynb
   move_rotation.ipynb
   three_cnots.ipynb
   steane_encoding.ipynb

.. nbgallery::

   qiskit.ipynb
   qrisp.ipynb

```

Would this be correct, i.e., lead to what this picture just above shows?

jbolns: Nevermind, in line with recent discussions about AI, I asked AI. =D
Let's see if it got the structure right. 
Not auto-generated. Asked canonical form and implemented it myself against our needs.

jbolns:
> Whether that is achieved with a command line argument and a switch to indicate file format, or a notebook, or some other method, comes down to user-friendliness and developer friendliness.

The switch command would become larger and larger as we add frameworks, to the point the reader might get lost in the implementation of the switch rather than be able to focus in the specifics of the circuit. 

Alternative. How about we do different circuits? Steane for some notebooks, a CNOT for another, a few CNOTs for another.
- Should lower the runtimes for building docs.
- Should lead to less redundant notebooks (and more varied thumbnails).
- Allows taking part of the initial suggestions by @purva-thakre , i.e., keep (some) circuits inside the notebook (for instance, it makes sense to have a separate circuit file for Qiskit but truth be said, the Qrisp circuit would be better inside the Qrisp notebook).
- Sets the foundational mechanism for additional contributions: design a circuit and make a notebook showing how to take it all the way into a simulation.

> Out of curiosity, what further work is needed to reduce the circuit volume in topologiq? I am simply comparing Austin's manually built structure in his slides with the [one from topologiq](https://docs.google.com/presentation/d/1RcMXB3kWtWWyM_LOQLkxo19Zaq4atLaV1fCsWqd78PQ/edit?slide=id.g39455a4c89c_0_581#slide=id.g39455a4c89c_0_581).

The sky is the limit. If you want suggested ideas for improvement, I have at least 5. But I'm pretty sure there's many more than 5 ideas, so if you have ideas, don't be shy to open an issue. 

Just a note. As soon as we finish this integration demos I will prioritise getting the (100, 95%) rates up to (100%, 100%). It seems reckless to opening the repository up for broader experimentation before doing this.

Also, refactoring before letting the codebase increase significantly in size seems important. It was a very iterative process. The code is an absolute mess. Truth be said, while I moved it as soon as viability became evident, I do feel a little embarrassed of not having taken a few weeks to literally just clean code up.

jbolns:
> I am not sure if this is a bug related to tqec or topologiq but the HTML viewer is really zooming out for the pyzx graphs. Is there a way to apply auto_layout for the inset?

Not exactly a bug, but indeed a little annoying. 

I'll remove the overlays entirely from the notebooks, as they're not really necessary given all the explanation before and after.

For the full explanation. Feature needs to be improved to consider the following. 
- You need a matplotlibe figure.
  - When PyZX defaults to Matplotlib (in VSCode), you typically get a nice image that scales appropriately. 
  - When PyZX defaults to D3, the `zx.draw()` cannot be used so you need to create an additional matplotlib fig.
- PyZX's `zx.draw_matplotlib()` does not have auto-layout, so the result is very ugly for many graphs
  - The backup for when a pretty PyZX figure is not available is to visualise the `simple_graph` used internally by topologiq
- The topologiq `simple_graph` is truly a simple grah.
  - Visualisation loses all sense of PyZX'ness
  - Visualisation scales weirdly, which is what you're seeing. 

So, in sum, you're seeing bad scaling of the backup of the backup.

jbolns:
@SMC17 , based on feedback by @purva-thakre with which I agree. 

Could you?

(1) join the tests currently in `src/tqec/interop/pyzx/topologiq_test.py` to a single unified test?

For this, you can use parametrize to condense multiple unit tests like this into a single unit test. 

There is also no need for all tests currently in the file. The `test_steane_like_structure()` covers many of the simpler situations you test earlier in the file, including simpler structures and the ability to verify multiple ports and port uniqueness. 

Suggested guidance, if needed. [https://docs.pytest.org/en/7.1.x/how-to/parametrize.html](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html)

And (2)

Reduce the docstrings in `read_from_lattice_dicts()` to the standard we use across TQEC?
- One line description
- A small itemised list of any relevant constraints
- Args: ..., Raises: ..., Returns: ...
- No "-" at start of any items in args, raises, or return.

Thanks in advance.

purva-thakre:
> The switch command would become larger and larger as we add frameworks, to the point the reader might get lost in the implementation of the switch rather than be able to focus in the specifics of the circuit.
> Alternative. How about we do different circuits? Steane for some notebooks, a CNOT for another, a few CNOTs for another.

@jbolns I will let you decide what you want to do. I do not think it will be that bad. Indeed, logical CNOT does not take that long, but it is not that interesting compared to the Steane code. 

Have some statements along the lines of: This notebook is compatible with multiple frameworks: Qiskit, Qrisp, etc. However, to reduce the simulation time, we demo the full pipeline in a randomly selected framework.

```py
import random

circuits = {'Qrisp': qrisp_circuit, 'Qiskit': qiskit_circuit}
selected_name = random.choice(list(circuits.keys()))
selected_circuit = circuits[selected_name]
print(f"{selected_name} circuit was selected")
```
Then, based on the value for `selected_name`, call the method that converts the circuit to QASM. Here, again, clarify why we are doing this. If this code block becomes too large, you can hide the code block in the jupyter notebook and allude to which method is used by which framework, etc. The rest of the pipeline is the same for all frameworks.

purva-thakre:
> The sky is the limit. If you want suggested ideas for improvement, I have at least 5. But I'm pretty sure there's many more than 5 ideas, so if you have ideas, don't be shy to open an issue.

thanks. I was simply curious as to why topologiq did not try to further reduce the circuit volume. I guess it isn't trying to create structures with the lowest possible circuit volume? If the tool does this, it will probably increase the run/execution time.

jbolns:
> I will let you decide what you want to do

I'll stick with the current approach two-section-gallery and maybe just add more specifics to each notebook about how the Qiskit file imported or the Qrisp circuit imported is used and generated and how you could do the same for other circuits. 

I think this invites others to submit their notebooks using other frameworks, i.e., Cirq, pytket, ... 

A single notebook would limit room for further contributions. 

Concerning build times, which is important. Does it not have something like a database, too? Right now, since we've been changing folders and whatnot, the build needs to do everything. Once this settles, wouldn't it take shorter? The initial run of each notebook takes 14min on my computer, but subsequent ones are very quick. 

> I guess it isn't trying to create structures with the lowest possible circuit volume?

Well, it is. 18 is a very decent volume. Most of us got something like that when putting the Steane manually together. While I am obviously biased, I'd frame it differently. Topologiq did such a good job at delivering low volumes that one can now reasonably think of an automated-12-cube Steane. 

Having said that, the initial implementation is purposely conservative in that it favours as-high-as-possible completion rates over potentially lower volumes.

purva-thakre:
> Once this settles, wouldn't it take shorter? The initial run of each notebook takes 14min on my computer, but subsequent ones are very quick.

Not really, no. The way CI uses the cache and how `make html` uses the cache are a bit different, i.e., two different things to consider. The CI with the cache will take 5-10 minutes per notebook and 30-45 minutes per notebook without the cache. 
`make html` after you have used `make clean html` takes just a few seconds locally if you have not made any changes to the docs. 

I need to dig into issues with the CI cache. Probably will get to it after next week. I am still a bit sick. 

But building the docs locally with multiple notebooks becomes a headache because I often use `make clean html` when I am making changes to the docs. Which is why the build time without the cache is one of my main concerns....

purva-thakre:
> Well, it is. 18 is a very decent volume. Most of us got something like that when putting the Steane manually together. While I am obviously biased, I'd frame it differently. Topologiq did such a good job at delivering low volumes that one can now reasonably think of an automated-12-cube Steane.

I missed this earlier. Austin's slides for MQSF show topologiq constructed a 3D structure of volume 18, whereas the structure that Austin built by hand in SketchUp is of volume 12. Which is why I asked about it...

jbolns:
> Austin's slides for MQSF show topologiq constructed a 3D structure of volume 18, whereas the structure that Austin built by hand in SketchUp is of volume 12. 

Yes. I'm just interpreting the difference positively: Topologiq is doing such a good job reducing volumes that we're literally *only* 6 cubes from Austin's manual build.

Nonetheless, it remains true that it would be fantastic to get even lower volumes.

So, I'm glad you pressed on.

I just tried something and the **new best volume is now 14 cubes**.

Topologiq's 18-cube result corresponds to the graph in @afowler 's MQSF slide 16, i.e., exactly the graph PyZX outputs.

In slide 17, Austin adds a red spider to the ZX-graph before building manually.

If you feed the exact graph in slide 17 to Topologiq you get, amongst all variants, the following 14-cube output.

steane_opt_output

I suppose this means a worthy avenue for further research is to find a generalisable way to decide when and how to perform the kind of operation in slide 17. 

A good starting point for this would be to find the original algorithm proposal. It contained a similar operation, but we've lost the document and I can't recall the specifics.

Azuya334:
@jbolns I made a integration notebook for qsim with the circuit defined in cirq. Should I make a branch and PR for it?

jbolns:
Go for it @Azuya334 . Can't promise you it will merge to main for sure or at least in it's current form. The cache thing discussed above seems a significant consideration. 

Of course, make sure you PR to this branch rather than to main.

afowler:
I'll make revisiting this algorithm a priority when I get back. I can
re-create the lost document. Plus add a bunch of new ideas. Have added a
slide on the below to the presentation.

jbolns:
Super! 
Ps. Really like slides 27 & 28 (Y, S), too. #epic @inmzhang @KabirDubey !

purva-thakre:
I have had some time to look into the cache issue. My 2 cents on this is that it would be preferable to have just one notebook for all frameworks to demo one specific example. The only thing that is different across all frameworks is how the external frontend's circuit is lowered to QASM, the rest is the same for all (pyzx, tqec simulation). If we go down the path of one notebook per frontend, working on the docs locally will become quite the hassle. 

One reason for this is that in the future, we will have more complicated examples than the Steane encoding example. Sticking to the practice of one notebook per framework will increase the docs build time such that it might even time out in the future, i.e., the docs won't be built.

purva-thakre:
> I'll make revisiting this algorithm a priority when I get back. I can
re-create the lost document.

@afowler If you are referring to the google doc on how to transform zx graphs to 3D structures, I made a copy of it before it was lost. See https://docs.google.com/document/d/1P_ZSvlILxpPBYPshq7rY5ahLSTZ024WoBn8Iu_r90rI/edit?usp=sharing

@jbolns My questions about topoloqig's output and Austin's MQSF slides are due to trying to understand if there is a plug-and-chug capability in topologiq. If I want to change the algorithm from the random greedy to something else that guarantees optimal performance, can I do so? As we have discussed so far, the current approach might not provide an optimal solution.

afowler:
>
> *purva-thakre* left a comment (tqec/tqec#720)
> <https://github.com/tqec/tqec/pull/720#issuecomment-3445467893>
>
> I'll make revisiting this algorithm a priority when I get back. I can
> re-create the lost document.
>
> @afowler <https://github.com/afowler> If you are referring to the google
> doc on how to transform zx graphs to 3D structures, I made a copy of it
> before it was lost. See
> https://docs.google.com/document/d/1P_ZSvlILxpPBYPshq7rY5ahLSTZ024WoBn8Iu_r90rI/edit?usp=sharing
>
Excellent, many thanks for finding that and sharing your copy.

> @jbolns <https://github.com/jbolns> My questions about topoloqig's output
> and Austin's MQSF slides are due to trying to understand if there is a
> plug-and-chug capability in topologiq. If I want to change the algorithm
> from the random greedy to something else that guarantees optimal
> performance, can I do so? As we have discussed so far, the current approach
> might not provide an optimal solution.
>
It should definitely be the case that it is possible to change the
algorithm that maps the zx graph to a block graph. There will indeed be the
need to try different ways.

Best,
Austin.

jbolns:
> If I want to change the algorithm from the random greedy to something else that guarantees optimal performance, can I do so? 

Yes. I want you to do so, in fact. 

Topologiq is structured in a way that makes it possible to variant or even entirely different algorithms with relative ease.

It's very onion-like. A `runner` files calls a `graph-manager` algorithm which quite literally manages the entire process including order of operations, potential placements, choice between several valid placements. In turn, for each edge, the `graph_manager` calls a `pathfinder` algorithm to get a topologically-sound 3D for the path/edge being processed. 

The `runner` could be easily adapted to call a different `graph_manager`. The `graph_manager` could be easily adapted to call a different `pathfinder`. If you want your algorithm to handle both graph management and pathfinding, your `new_graph_manager` could simply not call the pathfinder and instead use it's own operations for defining each edge. 

Additionally, your implementation could be fully self-contained or it could aid itself on several "utils" files containing  ZX, graph, and symbolic micro-features that I reckon would be needed in any algorithm. Your choice entirely. 

There is truly a menu of possibilities. 

Having said that, I do suggest you start from getting to know the codebase by helping with some of the issues currently open.

The issues currently open are absolutely necessary for:
1. getting to know the codebase,
2. enabling sound management of contributions that add a significant amount of new code.

It would be irresponsible of me to allow the codebase to explode in complexity before currently open issues get resolved.

jbolns:
> The only thing that is different across all frameworks is how the external frontend's circuit is lowered to QASM

After seeing examples of usage using other frameworks, having notebooks for each does seem helpful. Users generally prefer to see examples using their specific framework. Furthermore, not all frameworks are as similar to one another as Qiskit and Qrisp. In particular, some frameworks will altogether skip QASM.

> If we go down the path of one notebook per frontend, working on the docs locally will become quite the hassle.

For now, a compromise solution is to comment out the lines that trigger simulations in some of the notebooks. We can specify in the notebook that since the procedure to run a simulation is the same, to economise resources, we commented the call out. 

On the long-run, it would be ideal to get something like the examples_database work nicely for GH and local builds.

At the end of the day, our docs should scale alongside what we can do.

If we're having build-time issues with but-a-couple-notebooks, we need a different approach to docs.

purva-thakre:
> After seeing examples of usage using other frameworks, having notebooks for each does seem helpful.

Because their build times are not as bad as ours. 

> In particular, some frameworks will altogether skip QASM.

I am not sure what you mean by this. The current pipeline is literally external frontend -> qasm -> pyzx -> tqec. The only thing that is different across all is that the first step differs across all.  

> If we're having build-time issues with but-a-couple-notebooks, we need a different approach to docs.

It's not a couple notebooks. It is the entire gallery. As the examples become more complicated, simulation time takes over.

> On the long-run, it would be ideal to get something like the examples_database work nicely for GH and local builds.

The issue is also not just with this. There are things we can't control or further optimize, like Stim/sinter, which can take longer as the examples scale up. 

I think I am going to pull away from maintaining the docs after this. It is quite the headache making changes to the docs as the build time increases. I have tried to make my case, and the conversation is going nowhere.

afowler:
We should definitely try to avoid excessive simulation. Updating docs
should not involve simulating again.

jbolns:
I made the following change, which I believe is a good compromise that satisfies 100% of all input by everyone involved in this conversation. 

The Qiskit notebook now uses a 16-qubit GHZ circuit and does not include simulation. Instead of generating the Stim circuit and calling a simulation, it explicitly directs readers to other notebooks with simulation examples. 

The Qrisp notebook remains unchanged. I can remove the Stim and simulation sections to match. 

Other notebooks can follow the same principle.

The emphasis is then on the circuit design for each framework and how it can be imported to PyZX and then TQEC. Some frameworks, pytket and cirq/qsim, for instance, will altogether skip QASM, and the initialisation and post-selection specifics can vary depending on how the person designed the initial circuit.

Ps. The build-test errors are for the same reason as previous ones. I introduced an additional optional parameter in Topologiq's main runner function, used solely for debugging purposes. It's demanding I include this parameter in the function call. I really do think this rule should be removed/skipped. Optional parameters are supposed to be optional.

jbolns:
This PR is dated. Conversation on #862 is progressing under the radar. This PR can be closed
Will give a couple days in case of last thoughts by anyone who has commented in it, then close.

smburdick:
> This PR is dated. Conversation on #862 is progressing under the radar. This PR can be closed Will give a couple days in case of last thoughts by anyone who has commented in it, then close.

@jbolns Are there any components you want to cherry pick from this, and submit as a PR? My suggestion would be to do the smaller parts individually, which makes them easier to review

jbolns:
@smburdick 
Hi man, thanks for asking. I'd forgotten to close issue.

Nothing here is both important and salvageable. Too old. Even the example notebooks would need to be rewritten. 

The replacement is very minimalist and we have reached an agreement on how to proceed with it, so will tag you there in case you want to be involved. Tomorrow, though. Want to do it from computer (now on mobile) to copy some code over.

Issue #564, opened 2025-04-09
Implement controllable temporal height


KabirDubey:
Enable a user to be able to adjust the temporal height of a block. This height represents the number of rounds of stabilizer measurement. 

A memory experiment typically needs d rounds. Subtracting one layer for initialization and another for measurement, we get d-2 rounds of memory. In `tqec` we parameterize circuits by k (defined in section VIII of [the review paper](https://drive.google.com/file/d/1gCNpSdvE7Xtnf6Onh05vz9PX_WkjNv2r/view?pli=1)) such that d=2k+1. The `start_simulation_using_sinter` method allows users to specify a range of values for k to generate circuits that scale with the code distance. In the case of this Issue, we want to keep the spatial code distance fixed and only increase along the time axis. 

Additional details and some ways this could be useful are discussed [here](https://github.com/tqec/tqec/issues/533#issuecomment-2787759680). We don't expect programs which control k to be an integral part of the tool's usage, but the flexibility could be useful for certain users, and refactoring the code to support it could be a good exercise for those interested.

afowler:
For maximum usability, since I suspect there are real situations where you
would want to do this everywhere in a computation, I would suggest a factor
alpha that scales the temporal height. By default this should be 1, but the
user could set it to any desired factor. Ideally, this should be specified
in just one place and affect the construction of the entire computation.

nelimee:
> Hi! I'm still relatively new to the tool but, if there is still interest for this feature, id like to work on it as i wanted to check out the internals of how the mapping to circuits is done anyways.

Yes please go ahead :) Note that #562 will soon be merged and it is likely that the changes needed to fix this issue are at least partially based on that PR (I think that amounts at having the possibility to replace the default value for `_DEFAULT_BLOCK_REPETITIONS` to a runtime value):

https://github.com/tqec/tqec/blob/6cce91342dd7a6fdcb15b2f196c20f02aeb95f40/src/tqec/compile/specs/library/fixed_parity.py#L30

purva-thakre:
@SchmidtMoritz Do you need any help with this issue?

SchmidtMoritz:
ah im so sorry! i was waiting on the mentioned PR but then got busy and lost track of the issue. If it is not too urgent i would take care of it in the next days. Im not up to date on the latest changes. If you have any helpful suggestions or pointers please let me know.

purva-thakre:
> If it is not too urgent i would take care of it in the next days

[...]

Issue #533, opened 2025-03-22
[Question]: Why "Pipe must have different basis walls"


Zhaoyilunnn:
**Question**: 

https://github.com/tqec/tqec/blob/f8ebe11d68eda66345568cf48a8e14f8a59686fb/src/tqec/computation/pipe.py#L44-L45

Hi Team, I have a question.. why a pipe must have different basis walls?
Condider two `XZZ` cube, we cannot connect them in X axies? (i.e., have a `OZZ` pipe)

afowler:
In principle, such a same color wall pipe could be used to build an
experiment that tests the strength of error correction along a single
access. This is typically done to test the strength of error correction
along the time access, and is called a stability experiment. Such
experiments are not part of a computation, but can be used to characterize
the hardware. At the moment, I don't think we need to support such pipes,
though they do have real meaning.

KabirDubey:
> Such experiments are not part of a computation, but can be used to characterize
> the hardware. At the moment, I don't think we need to support such pipes,
> though they do have real meaning.
> […](#)

I thought the fact that we can simulate a logical CNOT implied that we can simulate a stability experiment. If we choose an appropriate space time axis and look at the [logical CNOT](https://tqec.github.io/tqec/gallery/cnot.html), we seem to have a structure that is the same as the [“moving a qubit” structure in figure 2](https://quantum-journal.org/papers/q-2022-08-24-786/pdf/) (a stability experiment). What am I missing?

![Image](https://github.com/user-attachments/assets/fc485085-a97f-4068-9ea9-6ca639c66b56)

afowler:
A single XXZ block or ZZX block is a stability experiment, but ideally you
want to be varying the number of rounds of stabilizer measurement in time
so that you see the rate of decay of logical error as the number of rounds
increases. This enables you to work out the necessary size of the temporal
code distance to achieve the desired logical error rate.

If there is strong interest to support stability experiments, my vote would
be for giving the user the power to define a different function of k to set
the temporal thickness of each block and connector, or indeed simply set
the thickness to a chosen number. This would enable one to create a plot of
the probability of getting the parity of a horizontal sheet of stabilizers
correct at different temporal heights, and extract the constant
characterizing the exponential decay, and extrapolate to the desired height.

Does this make sense?

Regards,
Austin.

KabirDubey:
> Does this make sense?

It took me a while, but I get it now. I think it would be interesting to support stability experiments.

afowler:
[...]

PR #721, opened 2025-10-07
controllable temportal height by adjusting compile_block_graph and Cu…

SchmidtMoritz:
This PR is addressing https://github.com/tqec/tqec/issues/564.

I added the block_temporal_height parameter and adjusted the CubeBuilderand PipeBuilder base classes as well as the implementations for both fixed_boundaryand fixed_bulk. For now i only adjusted the callings of compile_block_graph in simulation/simulation.py start_simulation_using_sinter so i could easily check if everything is working as it should. I also ran all compiling tests locally with everything passing.

For now there are no additional or adjust existing tests for the new teporal height feature. Is this needed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["uv_build >= 0.7.19, <0.9.0"]
+requires = ["uv_build >= 0.7.19, < 0.10.0"]
 build-backend = "uv_build"
 
 [project]


src/tqec/compile/compile.py
```diff
@@ -72,6 +74,7 @@ def compile_block_graph(
     block_graph: BlockGraph,
     convention: Convention = FIXED_BULK_CONVENTION,
     observables: list[CorrelationSurface] | Literal["auto"] | None = "auto",
+    block_temporal_height: LinearFunction = LinearFunction(2, -1),
```
inmzhang:
```suggestion
    block_temporal_height: LinearFunction = _DEFAULT_BLOCK_REPETITIONS,
```

inmzhang (commented):
Thanks for the contribution! This looks good to me. I suggest to add a simple test in `compile_test.py` for adjusting temporal height.

inmzhang:
Sorry for no feedback in #564, I missed that message somehow...

KabirDubey:
Nice, thanks! I would suggest updating the documentation to explain that one can change the temporal height from the default. A script that produces this plot:

> A single XXZ block or ZZX block is a stability experiment, but ideally you
> want to be varying the number of rounds of stabilizer measurement in time
> so that you see the rate of decay of logical error as the number of rounds
> increases. This enables you to work out the necessary size of the temporal
> code distance to achieve the desired logical error rate.

could be interesting. 

This is def not a priority, I think it's just worth documenting so users [know why this functionality exists](https://github.com/tqec/tqec/issues/533#issuecomment-2787759680). No worries if you don't want to do it, we can always make a issue for this flagged not a priority.

inmzhang:
Please run `uv run ty check` and fix the type errors.

SchmidtMoritz:
I fixed the tests in `graph_test.py` and now `uv run ty check` runs without issues. I also added a new test in `compile_test.py` that checks a few linear functions for different block configurations.

PR #725, opened 2025-10-09
Fix topologiq coordinate transformation and add comprehensive testsPr 720

SMC17:
## Summary

This PR fixes a coordinate transformation bug in the topologiq integration and adds comprehensive test coverage.

## Problem

The original implementation used midpoint calculation for pipe positioning, which caused incorrect cube connections in the BlockGraph. The code calculated a midpoint between source and target cubes, then attempted to shift from that midpoint to find the actual cube positions. This approach is conceptually incorrect because:

1. Pipes in TQEC don't have positions - they connect adjacent cubes
2. The midpoint calculation led to non-adjacent cube connections in some cases
3. This caused test failures for complex lattice structures

## Solution

Changed the pipe positioning logic to directly transform source and target cube coordinates from topologiq space to TQEC space, matching the approach used in the existing JSON importer (`collada/read_write.py` lines 384-396).

**Before:**
```python
# Calculate midpoint and shift from it
shift_coords_from_src = tuple([(u - v) / 3 for u, v in zip(tgt_pos, src_pos)])
directional_multiplier = int(sum(shift_coords_from_src))
coords = [u + v for u, v in zip(src_pos, shift_coords_from_src)]
translation = FloatPosition3D(*coords)
head_pos = int_position_before_scale(
    translation.shift_in_direction(pipe_kind.direction, -1 * directional_multiplier),
    pipe_length,
)
tail_pos = head_pos.shift_in_direction(pipe_kind.direction, 1 * directional_multiplier)

### DIFF
--- a/src/tqec/interop/pyzx/topologiq.py
+++ b/src/tqec/interop/pyzx/topologiq.py
@@ -1,4 +1,27 @@
-"""Read and write block graphs to and from Collada DAE files."""
+"""Read and write block graphs to and from topologiq space-time diagrams.
+
+This module provides interoperability between topologiq's lattice surgery
+representation and TQEC's BlockGraph representation. The primary function
+`read_from_lattice_dicts` converts topologiq's output (lattice nodes and edges)
+into a TQEC BlockGraph that can be compiled and simulated.
+
+Typical usage:
+    ```python
+    from topologiq.scripts.runner import runner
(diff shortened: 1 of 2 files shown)


SMC17:
Closing this - I accidentally branched from the wrong base and included unrelated 
commits. I've commented on #720 with the relevant bugfix instead.

Issue #724, opened 2025-10-09
[Question]: How to use detector database in demos?


KabirDubey:
**Question**: Suppose I want to present a demo of `tqec` generating and running a `Stim` circuit. I create a Jupyter notebook with all the code needed for this, just like our gallery examples. Running the notebook puts a `database.pkl` file locally. As I understand, this database has pre-computed the detectors which Stim samples from. We want an external user to clone my program and the database so that they can simply perform the sampling rather than waiting to re-compute the detectors.

What is the standard operating procedure for how the notebook should be written so that the user does not encounter bugs? Bugs are often a `pickle.load()` failure, where either a plaquette, circuit, or operation was missed, or a database versioning issue. 

Apologies if this has already been explained. I want to make sure I understand what is supposed to work everytime so I can nail down any bugs with the caching if there are any. 

Related: #681

KabirDubey:
the JSON format with a [to_file](https://github.com/tqec/tqec/blob/064cd70f87d238f6446704c4b7a8d54b16a515c8/src/tqec/compile/detectors/database.py#L465) line seems promising

@BSchelpe @Zhaoyilunnn let me know if you have any thoughts

Zhaoyilunnn:
> the JSON format with a [to_file](https://github.com/tqec/tqec/blob/064cd70f87d238f6446704c4b7a8d54b16a515c8/src/tqec/compile/detectors/database.py#L465) line seems promising
> 
> @BSchelpe @Zhaoyilunnn let me know if you have any thoughts

If a pickle database built in one environment cannot be loaded correctly in another environment, I believe using json is ok. 

The expected user experience should be like (if I understand correctly):

1. Build a logical operation
2. Simulate it, and the database will be built/updated in local env
3. If the user simulate the same operation again, it will reuse the existing database

In this process, the user should not need to manually specify the database path (unless he really want a customized path), and should not need to delete/updated the database file manually.

I haven't looked into the issues related to #681 but I plan to investigate it. There must be some way to make the above process smoother

BSchelpe:
I agree with everything @Zhaoyilunnn says. 

The only thing I thought was worth mentioning from @KabirDubey's description of what bugs show up: the database versioning issue superficially sounds like a different sort of issue from "plaquette, circuit, operation" being missed, but I think it is likely to have the same cause. If the version number is missing, the code assumes it is 0.0.0. This would then be incompatible with the current version number in the code and so lead to a versioning issue. 

[...]

Issue #681, opened 2025-09-01
Mac/Linux database incompatibility


purva-thakre:
@KabirDubey came across an error/bug while running the notebook created for the QCE25 tutorial. This notebook had a saved database, which was created on my Linux device. Kabir had to delete the database, run the notebook on his device, and then generate a new database file. 

I do not have access to a Mac device. But it might be worth looking into what might be causing this. @KabirDubey Can you provide the full error message?

I also noticed the docs workflow only tests the build on Linux devices. Should we add additional non-linux runners to `.github/workflows/gh-pages.yml`?

purva-thakre:
I think this is related to #576

Zhaoyilunnn:
Do we have the notebook source code? I am not sure whether pickle ensures cross-platform compatibility. And it *seems* that [different python versions may also cause compatibility issue.](https://stackoverflow.com/a/68987412/14092312) (it is _backward_ compatible with python versions). Do you have the same python version?

KabirDubey:
Thanks, busy this week but will follow up with a description of this issue and others.

purva-thakre:
Thanks, Kabir!

> I am not sure whether pickle ensures cross-platform compatibility. 

@Zhaoyilunnn See https://docs.python.org/3/library/pickle.html#data-stream-format and https://stackoverflow.com/a/1850806

I will let Kabir add a link to the final notebook as the repo currently has multiple versions. I am not too sure which one raised the errors.

PR #727, opened 2025-10-10
chore: automatic check format when reading/writing detector database

Zhaoyilunnn:
## Motivation

In a previous PR https://github.com/tqec/tqec/pull/584 we've added `json` as an alternative to the default `pickle` format, but it seems that it has not been used. 

In this issue: https://github.com/tqec/tqec/issues/724  @KabirDubey would like to try json version of database because there are some issues with pickle load/store crossing platforms: https://github.com/tqec/tqec/issues/681

I found that currently, users cannot specify the format of database smoothly.

The current user experience is like

**Program 1**
```python
from pathlib import Path

from tqec.compile import compile_block_graph
from tqec.gallery import cnot

db = DetectorDatabase()
g = cnot(Basis.X)
g = compile_block_graph(g)
g.generate_stim_circuit(1, detector_database=db)
db.to_file(Path("cnot_database.json"), format="json")
```

There are some problems with the above experience.
1. The user must specify the format manually, and even if the code is `db.to_file(Path("cnot_database.pkl"), format="json")`, it will also write the json file to the `.pkl` file. If in an extreme scenario the user write a json format `.pkl` to the default database path, future re-use of the cached database will cause problems.
2. Running the above still tries to read and write to the system default database path. And the above code will still save a `.pkl` file to the path. 
3. Even if the user specify the `database_path` to be `path/to/database.json`, it will still save pickle file to it..., because we don't have an argument specifying the format in `generate_stim_circuit()` 

For point 3, one can test the below program and will see that the saved `cnot_database.json` is a pickle file

**Program 2**
```python
from pathlib import Path

from tqec.compile import compile_block_graph
from tqec.gallery import cnot

db = DetectorDatabase()
g = cnot(Basis.X)
g = compile_block_graph(g)
g.generate_stim_circuit(1, detector_database=db, database_path=Path("cnot_database.json"))
```

## Solutions

1. Of course we can add an option `format` to the `generate_stim_circuit()`. But this will require the user to set both `database_path` and `format`, and if the path suffix and `format` does not match, it can still run without error, I think this will cause confusion.
2. Implicitly check format based on file path (the current implementation)

I prefer the second solution, please let me know if you prefer the first or you have other proposals.

With current implementation, the user experience is that if one want to use json format database, they can run program 2 and the json format database will be saved in the specified location. Also, if they want to reuse a json format database, they just need to set the `database_path`.

### DIFF
--- a/src/tqec/compile/compile.py
+++ b/src/tqec/compile/compile.py
@@ -26,6 +26,8 @@ _DEFAULT_SCALABLE_QUBIT_SHAPE: Final = PhysicalQubitScalable2D(
     LinearFunction(4, 5), LinearFunction(4, 5)
 )
 
+_DEFAULT_BLOCK_REPETITIONS: LinearFunction = LinearFunction(2, -1)
+
 
 def _get_template_from_layer(
     root: BaseLayer | BaseComposedLayer,
@@ -72,6 +74,7 @@ def compile_block_graph(
     block_graph: BlockGraph,
     convention: Convention = FIXED_BULK_CONVENTION,
     observables: list[CorrelationSurface] | Literal["auto"] | None = "auto",
+    block_temporal_height: LinearFunction = _DEFAULT_BLOCK_REPETITIONS,
 ) -> TopologicalComputationGraph:
     """Compile a block graph.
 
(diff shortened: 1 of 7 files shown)


inmzhang:
LGTM. @BSchelpe would you like to review this PR?

BSchelpe:
> LGTM. @BSchelpe would you like to review this PR?

Thanks! Will do.

PR #728, opened 2025-10-11
chore: only include src for ty checking

Zhaoyilunnn:


inmzhang:
I still think it’s worth doing type checking on the notebooks, as it can catch some unexpected typos.

Zhaoyilunnn:
> I still think it’s worth doing type checking on the notebooks, as it can catch some unexpected typos.

OK, that makes sense.

PR #729, opened 2025-10-11
Add opacity controls for BlockGraph visualizations

### LINKED ISSUE #690: Implement an option to control the opacity of blockgraph structures
If one wants to port the 3D structures to an image, it is difficult to show the structures beyond what the image shows because the boundaries are 100% opaque. A person will have to take a screenshot from all sides of the blockgraph in tqec. 

We want to add an opacity option such that a single image can convey all the information about boundaries. For example, compare the following:

| from our docs      | Fig 2d from https://arxiv.org/abs/2410.05202 |
| ----------- | ----------- |
| Image| Image|

The image on the right also relies on something like `pop_faces_at_directions`(tqec function) as it is not showing all the corresponding boundaries.

SMC17:
Fixes #690

## Changes
- Add `node_opacity`, `edge_opacity`, `hadamard_opacity` to `DrawerConfiguration`
- Update `draw_positioned_zx_graph_on()` and `plot_positioned_zx_graph()` with opacity parameters
- Apply alpha transparency in matplotlib calls

## Testing
All existing tests pass
Backward compatible (defaults to 1.0)
Pre-commit hooks pass

### DIFF
--- a/CONTRIBUTION_VALIDATION.md
+++ b/CONTRIBUTION_VALIDATION.md
@@ -1,5 +1,5 @@
 # Contribution Validation & Self-Review
-**Date**: October 10, 2025  
+**Date**: October 10, 2025
 **Purpose**: Ensure all contributions meet TQEC standards and actually solve problems
 
 ---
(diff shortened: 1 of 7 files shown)


SMC17:
Closing to clean up commits. Will reopen with cleaner history.

Issue #673, opened 2025-08-20
Add text option for interactive visualization in dark mode


purva-thakre:
If I use `view_as_html` in dark mode, the options to interact with the 3D model are hard to see in VS Code.

<img width="541" height="83" alt="Image" src="https://github.com/user-attachments/assets/f01b7347-c89d-468a-85f8-9eae9da2dab2" />

This is supposed to look like below:

<img width="638" height="64" alt="Image" src="https://github.com/user-attachments/assets/0df782a1-e030-4316-9c50-375aac19adc3" />

**Option 1:** Add a background to the `Mouse Wheel = Zoom. Left Drag = Orbit. Right Drag = Strafe.` section of `view_as_html`

**Option 2:** When it is detected dark mode is in use, change the text color.

ItachyBird:
I plan to fix this by updating the styles for dark mode (using prefers-color-scheme / background color). Please let me know if that’s okay.

inmzhang:
> I plan to fix this by updating the styles for dark mode (using prefers-color-scheme / background color). Please let me know if that’s okay.

It sounds good. The related code is at https://github.com/tqec/tqec/blob/main/src/tqec/interop/collada/read_write.py.

purva-thakre:
@ItachyBird Just checking, do you still plan to work on this issue?

MarkAppprogrammer:
@purva-thakre Can I work on this issue?

inmzhang:
Closed by #733.

purva-thakre:
@MarkAppprogrammer In the future, use one of the allowed keywords to directly close an issue after a PR is merged. 

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

PR #733, opened 2025-10-12
Set background color to white in HTML viewer

MarkAppprogrammer:
Changed the background color to white in the HTML Viewer for readability (based on #673). The original issue sought to change only the options text to have a white background, but when testing it with a couple different themes, it didn't look right. I originally tried to get the background color using the following code:
```
getComputedStyle(document.documentElement)
    .getPropertyValue('--vscode-editor-background')
    .trim()
    .toLowerCase();
```
I got the color, but choosing when to change the color of text (since VS Code has such a wide assortment of color themes is quite difficult, so I stuck to keeping the default background color to white.


purva-thakre:
@MarkAppprogrammer Thanks for the PR. There are some pre-commit hook failures in your branch. Do you mind fixing these? The contributor guide in the docs provides additional information. 

In VS Code, these failures are automatically caught by the hook such that it does not let you add a commit until the failures are fixed.

purva-thakre (approved):

MarkAppprogrammer:
@purva-thakre Just fixed it.

PR #734, opened 2025-10-14
fix(detector-database): some pickle load errors not handled

Zhaoyilunnn:
As [described](https://github.com/tqec/tqec/issues/724#issuecomment-3397118181) by @KabirDubey, there's an issue in loading faulty/old-version detector database. 

Previously Adrien has added some exception handling, when failing to load a db, we'll raise warining, move the db to a `faulty_xx` file, and return an empty database.

https://github.com/tqec/tqec/blob/99e0b66d1b8359038604867c8a02833f2eadf350/src/tqec/compile/detectors/database.py#L223-L238

However, it only handles `PickleError`, according to [the documentation](https://docs.python.org/3/library/pickle.html#pickle.UnpicklingError), other exceptions may also happen, so I modified the code to handle all exceptions.

While I prefer not to use general `Exception`, I _think_ it is ok here for user experience


inmzhang (approved):
LGTM.

@Zhaoyilunnn Yilun, I've invited you to be a member of TQEC org. Now you should have the permission to merge the PR by yourself. Could you check on this?

Zhaoyilunnn:
> LGTM.
> 
> @Zhaoyilunnn Yilun, I've invited you to be a member of TQEC org. Now you should have the permission to merge the PR by yourself. Could you check on this?

Thanks, it works perfectly

PR #736, opened 2025-10-15
Y half cube updates

KabirDubey:
- Documented explanations of `InjectedBlock`
- Replaced  `additional_metadata` with a `z` coordinate for linking
- Added a class data structure for organizing the Y basis transition round circuit generation
- Eliminated unnecessary flattening by keeping REPEAT loops intact


src/tqec/compile/tree/tree.py
```diff
@@ -376,37 +376,72 @@ def generate_circuit(
...
+            assert layer_z is not None, "z_coordinate must be set for temporal injection coordination"
             circuits_by_z[layer_z] = node.generate_circuit(k, qubit_map)
 
         # included observable indices at each injection position
         injection_obs_indices: dict[BlockPosition3D, list[int]] = {}
         for obs_idx, obs in enumerate(self._abstract_observables):
-            for cube in obs.y_half_cubes:
-                pos = BlockPosition3D(*cube.position.as_tuple())
-                injection_obs_indices.setdefault(pos, []).append(obs_idx)
+            if hasattr(obs, 'y_half_cubes'):
```
inmzhang: Why do we need this change?

src/tqec/compile/tree/injection.py
```diff
@@ -115,11 +116,14 @@ def append_tree_circuit(self, circuit: stim.Circuit) -> None:
...
         remap their lookbacks against the updated global indices.
+
+        Performance note: Flattening only occurs in _weave_into_cur_circuit
+        when injection requires merging circuits.
         """
         # commit the previous circuit
         self._commit_prev_circuit()
         self._mtracker.start_new_slice(circuit.num_measurements)
-        self._cur_circuit = circuit.flattened()
+        self._cur_circuit = circuit
```
inmzhang, line 136: Note that for some circuit slice, flatten is still required because there may be in flows from the previous slice and the first chunk accepting the in flows happen to be a repeated circuit. This seems will not happen for Y basis block but is a corner case need to be cautious about.

src/tqec/compile/specs/library/generators/y_basis.py
```diff
@@ -18,6 +20,186 @@
...
+    MAIN: Top-left to bottom-right diagonal where real + imag = constant
+    ANTI: Bottom-left to top-right diagonal where real - imag = constant
+    """
+    MAIN = "main"
+    ANTI = "anti"
+
+
+@dataclass(frozen=True)
+class TwistLine:
+    """Represents a domain wall in the Y-basis transition.
```
inmzhang:
```suggestion
    """Represents a twist line in the Y-basis transition.
```

src/tqec/compile/specs/library/generators/y_basis.py
```diff
@@ -18,6 +20,186 @@
...
+
+    MAIN: Top-left to bottom-right diagonal where real + imag = constant
+    ANTI: Bottom-left to top-right diagonal where real - imag = constant
+    """
+    MAIN = "main"
+    ANTI = "anti"
+
+
+@dataclass(frozen=True)
+class TwistLine:
```
inmzhang, line 37: Is it possible to remove the all the calls to `_split_ul_md_dr` and use this class instead?

inmzhang:
And please make sure pre-commit hooks and `ty` checks are fixed.

src/tqec/compile/tree/tree.py
KabirDubey: It's not needed, I misunderstood the abstract observable object, will delete this line

src/tqec/compile/tree/injection.py
KabirDubey: Oh right okay I will make a commit that checks for this corner case

src/tqec/compile/specs/library/generators/y_basis.py
KabirDubey: Yeah I think so I'll make a commit with `_split_ul_md_dr` as a class function

PR #740, opened 2025-10-25
Workflow to verify broken links in docs

### LINKED ISSUE #738: Typo in the example
There's a typo ("does not exists") in the error message of the main example from the documentation.

### LINKED ISSUE #739: Bad link
The link in section 8 of https://tqec.github.io/tqec/user_guide/quick_start.html doesn't work.

purva-thakre:
This PR adds a workflow that verifies if there are broken links in the docs. 

Fixes #739 
Fixes #738

### DIFF
--- a/src/tqec/compile/blocks/positioning.py
+++ b/src/tqec/compile/blocks/positioning.py
@@ -10,9 +10,6 @@ from tqec.utils.position import BlockPosition2D, BlockPosition3D, SignedDirectio
 
 
 class LayoutPosition2D(ABC):
-    _x: int
-    _y: int
-
     def __init__(self, x: int, y: int) -> None:
         """Represent the local indexing used to represent both cubes and pipes.
 


docs/conf.py
```diff
@@ -194,3 +194,20 @@ def setup(app):
...
+
+# options for the linkcheck workflow
+# we expect these links to be externally valid all the time
+# the workflow leads to 403 if we check a link via an automated workflow
+linkcheck_ignore = [
+    r"https://arxiv\.org/.*",
+    r"https://doi\.org/.*",
+    r"https://link\.aps\.org/doi/.*",
+    r"http://dx\.doi\.org/.*",
+    r"https://www\.quantiki\.org/.*",
```
purva-thakre:
```suggestion

```

src/tqec/compile/blocks/positioning.py
```diff
@@ -10,6 +10,9 @@
 
 
 class LayoutPosition2D(ABC):
+    _x: int
+    _y: int
```
inmzhang: I don't think we need this?

inmzhang (dismissed):

src/tqec/compile/blocks/positioning.py
purva-thakre: Good catch! I think this was a leftover from a strategy I tried initially of using explicit type declarations.

purva-thakre:
Did something change with the Github UI or a new repo setting was enabled? I have only recently started seeing the stale review thing...

image

Looks like if we are suggesting any changes to a PR, it is better not to pre-approve them?

inmzhang:
> Did something change with the Github UI or a new repo setting was enabled? I have only recently started seeing the stale review thing...
> image
> 
> Looks like if we are suggesting any changes to a PR, it is better not to pre-approve them?

Yes, Adrien made the setting change. You’re absolutely right, I’ll avoid pre-approving until all suggested changes have been addressed.

inmzhang (approved):

PR #745, opened 2025-10-29
close #743 move pymatching dependency from dev to normal dependency

### LINKED ISSUE #743: pymatching dependency
Does it maybe make sense to add pymatching to the dependencies list? Because when I'm running

```
!python -m pip install git+https://github.com/tqec/tqec.git
!tqec run-example --out-dir ./results
```

in Colab to reproduce the main example, it says I need to install pymatching, which is used for decoding. I understand that it is maybe not a dependency per se, but maybe it is worth handling this in the documentation of the example to avoid confusion. Thanks!

Zhaoyilunnn:


pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
     "pycollada<=0.9.1",
     "pycryptosat<=5.11.21",
     "python-sat[cryptosat]<=1.8.dev24",
+    "pymatching<=2.3.0",
```
inmzhang:
```suggestion
    "pymatching<=2.3.1",
```

The latest version is v2.3.1 which fixed some bugs related to correlated matching.

PR #750, opened 2025-11-07
JOSS review fixes

burgholzer:
Hey everyone 👋🏼 

While doing my review as part of the ongoing JOSS submission https://github.com/openjournals/joss-reviews/issues/9142 I accumulated some of the smaller things to fix in small, digestible commits.

They mostly declutter existing configuration files and adopt one or two best practices from https://learn.scientific-python.org/development/guides/repo-review/

I'll post a couple of follow-up issues separately.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,8 @@ authors = [
 ]
 description = "A design automation framework for Topological Quantum Error Correction."
 readme = "README.md"
-license = { file = "LICENSE" }
+license = "Apache-2.0"
+license-files = ["LICENSE"]
 keywords = ["topological quantum error correction", "qec"]
 requires-python = ">=3.10"
 classifiers = [


Zhaoyilunnn (approved):

pyproject.toml
```diff
@@ -10,9 +10,10 @@ authors = [
 ]
 description = "A design automation framework for Topological Quantum Error Correction."
 readme = "README.md"
-license = { file = "LICENSE" }
+license = "Apache-2.0"
+license-files = ["LICENSE"]
 keywords = ["topological quantum error correction", "qec"]
-requires-python = ">= 3.10, <3.14"
+requires-python = ">=3.10"
```
purva-thakre, line 16: Hi @burgholzer, my knowledge might be outdated, but I thought it was considered best practice to upper-bound python versions for a project? 

I get why you are suggesting we skip this for the project dependencies in #751. I also get your point about having a lower bound. My personal opinion here is that bumping up to a major python version should be done manually[^1]. 

[^1]: FWIW, I haven't had the chance to read https://iscinumpy.dev/post/bound-version-constraints/#tldr completely

burgholzer: One of the main reasons for not upper capping python versions is that it simply does not work as you would expect and it puts unnecessary pressure on maintainers.
Upper capping the Python version means that your package will definitely require an update and a new release even though it might "just work" with the new Python version without any changes.
Additionally, if a previous version has been published without an upper bound, then resolvers will typically back solve and pick an older version of your package for a newer Python version, which is totally not wants you want. The last point is not as critical here, because there has not been a release of this package yet.

Overall, the right "tool" to use for indicating Python version compatibility is the classifier system in pyproject.toml
Tools like nox use that information for determining the versions of Python that are supported.

purva-thakre: > Tools like nox use that information for determining the versions of Python that are supported.

@burgholzer Sorry for the very late reply. 

Dumb question: What is your motivation behind using `nox`?

Perhaps I am not understanding how the tool works exactly. But I don't really see a difference between adding a new version of python to the workflow below and using something along the lines of [`@nox.session(python=["3.10", "3.11", "3.12, 3.13"])`](https://nox.thea.codes/en/stable/tutorial.html#testing-against-different-and-multiple-pythons)

https://github.com/tqec/tqec/blob/2cdea6ff0506b850dc13cf38c9ee7f3265c3ee32/.github/workflows/ci.yml#L35

burgholzer: nox is an extremely convenient task runner. It helps to abstract away many different tasks. In our MQT projects, no matter in which, you can run
- `nox -s lint` to run all kinds of linter tasks
- `nox -s tests` to run tests on all supported Python versions, or `nox -s tests-3.12` to run on a specific version
- `nox -s minimums` to run tests with minimal dependency versions (similar to above)
- `nox -s docs` to build and locally host the project docs.

We only set `PYTHON_ALL_VERSIONS = ["3.10", "3.11", "3.12", "3.13", "3.14"]` (https://github.com/munich-quantum-toolkit/core/blob/e9e7a8ee88fe03b6876dd5b9de963a208b7fa51a/noxfile.py#L35) in the nox file once and every command is set up properly based on that.
The main advantage over setting this in a CI workflow is that developers can use the very setup that is running in CI also locally, which is great for reproducibility and debugging.

Issue #752, opened 2025-11-07
CD workflow and release on PyPI


burgholzer:
## Is your feature request related to a problem? Please describe.

The repository is currently missing a CD workflow for automatic releases and has no official release available from PyPI.
The fact that no official release is available and the version number is still at `v0.0.1` makes it hard to communicate changes. Even though the project might not be ready for a 1.0 release, it would certainly deserve a `0.1.0` release with subsequent updates when new developments roll in.

## Describe the solution you'd like

A GitHub workflow triggered on GitHub releases that builds the source distribution as well as the wheel and uploads them to PyPI (preferably using trusted publishing).
Some information on trusted publishing https://docs.pypi.org/trusted-publishers/

You could take https://github.com/munich-quantum-toolkit/bench/blob/main/.github/workflows/cd.yml as an inspiration, which internally uses
- https://github.com/munich-quantum-toolkit/workflows/blob/main/.github/workflows/reusable-python-packaging-sdist.yml
- https://github.com/munich-quantum-toolkit/workflows/blob/main/.github/workflows/reusable-python-packaging-wheel-build.yml

Afterwards, an official release should be published.

## Describe alternatives you've considered

Obviously, manual releases may be an option here, but I'd personally recommend going with the automatic workflow which also provides automatic build attestations assuring the integrity of the distributed assets.

## Additional context

This is part of https://github.com/openjournals/joss-reviews/issues/9142

purva-thakre:
> The repository is currently missing a CD workflow for automatic releases and has no official release available from PyPI.

If I recall older discussions correctly, we plan on releasing an official version sometime next year. We were procrastinating adding this workflow until then. 😅

burgholzer:
> > The repository is currently missing a CD workflow for automatic releases and has no official release available from PyPI.
> 
> If I recall older discussions correctly, we plan on releasing an official version sometime next year. We were procrastinating adding this workflow until then. 😅 

Is there any compelling reason for not starting to create releases. I totally get procrastinating a 1.0 release, but something like development releases would certainly rather help than hurt in my opinion.
Especially since no available releases make it unnecessarily hard for other projects to build on the TQEC library. From experience, people have a hard time adding git dependencies to their Python projects.

purva-thakre:
> something like development releases would certainly rather help than hurt 

To be honest, we didn't think about a dev release.  But I agree with you. This workflow should be added.

inmzhang:
[...]

Issue #751, opened 2025-11-07
Proper Python dependency configuration


burgholzer:
## Is your feature request related to a problem? Please describe.

The current dependency setup in the project seems to be less than ideal as it places upper bounds on all dependencies.
Please consider reading https://iscinumpy.dev/post/bound-version-constraints/ in that regard. In almost all instances, upper caps are considered a bad practice.

To make matters a little more extreme, none of the packages have lower bounds specified. As such, people might be running fully outdated versions of dependencies that might, by any chance, install and yield a completely broken configuration.

## Describe the solution you'd like

I assume the main goal here is to get reproducible environments.
However, in that case, the proper tool for that would be a lock file.
As you are using `uv` already, adopting that is as straight-forward as it gets.
Simply run `uv lock` and commit the respective lock file.
Commands like `uv sync` or `uv pip install` will then use the known-good versions from the lockfile.
Services like dependabot or renovate can regularly update the lockfile for you.

As for the lower bound problem: `uv` also offers great tooling support for this through its `--resolution-strategy=lowest` flag, which automatically sets up an environment with the minimum versions it can resolve according to the project dependencies.

If you want some inspiration for how this could look like (also in CI), have a look at https://github.com/munich-quantum-toolkit/core/blob/98c31eb50cc6e84b676b0545237aee4ce21a6339/noxfile.py#L120-L130 and https://github.com/munich-quantum-toolkit/workflows/blob/main/.github/workflows/reusable-python-tests.yml

## Describe alternatives you've considered

N/A

## Additional context

This is part of https://github.com/openjournals/joss-reviews/issues/9142

purva-thakre:
> As you are using uv already, adopting that is as straight-forward as it gets.
Simply run uv lock and commit the respective lock file.

To fix one of Lukas' suggestions, whoever works on this should remove the following line from `gitignore`. I can't recall why we decided to ignore this file. 

https://github.com/tqec/tqec/blob/91e05ace570a203b6ff065297c303e1ac067f5d4/.gitignore#L193

jbolns:
I think most of the changes needed to address this can be included in #720:

#720 already does some lower-bounding to avoid broken configurations when you install TQEC including Topologiq. To be fair, I had "eye-balled" the lower bounds, but I can do some homework and add *existing* lower bounds for all packages. 

I can also include the uv.lock, as @purva-thakre suggests. Keeping it up to date ensures the latest version of Topologiq is installed when installing all dependency groups. 

[...]

PR #755, opened 2025-11-10
Add CD workflow publishing to PYPI

inmzhang:
Related to #752. Once #751 is resolved, we can hopefully release a v0.1.0 version of tqec.


Zhaoyilunnn:
I realized that I forgot to revert the change in https://github.com/tqec/tqec/pull/704 which should be a bug at this moment. Let me fix this before publishing?

Zhaoyilunnn (approved):

PR #757, opened 2025-11-10
Python dependency configuration

### LINKED ISSUE #751: Proper Python dependency configuration

inmzhang:
Closes #751.

This PR removes most upper bounds on dependency versions and adds reasonable lower bounds. It also introduces a CI workflow to verify that the minimum dependency versions are satisfied when `pyproject.toml` changes. I don’t have a systematic way to determine the exact minimum versions yet, but I confirmed that the current configuration works across all supported Python versions (3.10–3.13).

@burgholzer Hi Lukas, does this change make sense to you?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,18 +17,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.13"
       - name: Install uv
         uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
-      - name: Create virtual environment and install dependencies
+      - name: Setup Python development environment
         run: |
-          uv venv
-          uv sync --group dev
+          uv sync --group dev --python "3.13"
       - name: Run pre-commit hooks
         run: uv run pre-commit run --all-files
 
@@ -41,18 +36,13 @@ jobs:
     steps:
       # Checkout and install TQEC
       - uses: actions/checkout@v4
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
       - name: Install uv
         uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
-      - name: Install dependencies after creating a virtual environment
+      - name: Setup Python development environment
         run: |
-          uv venv
-          uv sync --group all
+          uv sync --group all --python ${{ matrix.python-version }}
       # Type check the code base.
       - name: Type checking
         run: uv run ty check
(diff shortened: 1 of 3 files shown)


uv.lock
```diff
@@ -0,0 +1,4012 @@
...
+]
+
+[[package]]
+name = "accessible-pygments"
+version = "0.0.5"
+source = { registry = "https://mirrors.bfsu.edu.cn/pypi/web/simple/" }
+dependencies = [
+    { name = "pygments" },
+]
+sdist = { url = "https://mirrors.bfsu.edu.cn/pypi/web/packages/bc/c1/bbac6a50d02774f91572938964c582fff4270eee73ab822a4aeea4d8b11b/accessible_pygments-0.0.5.tar.gz", hash = "sha256:40918d3e6a2b619ad424cb91e556bd3bd8865443d9f22f1dcdf79e33c8046872", size = 1377899, upload-time = "2024-05-10T11:23:10.216Z" }
```
Zhaoyilunnn: Personally I believe it's fine to set the pypi registry as the mirror site. But I am not sure if this will make sync slower for people outside China 😄

inmzhang: Did not notice that. Good catch! I'll set it to the default index :-)

Zhaoyilunnn (dismissed):

inmzhang:
Yilun, thanks for the review! I’d like to wait a little longer to see if @burgholzer has any comments before merging.

.github/workflows/test-min-versions.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+
+      - name: Install with minimum dependency versions
+        run: |
+          uv venv
+          uv sync --group test --resolution=lowest-direct
```
burgholzer:
```suggestion
      - name: Install with minimum dependency versions
        run: |
          uv sync --group test --resolution=lowest-direct
```
There is no need to explicitly create a virtual environment when using `uv sync` (neither for `uv pip install <...>`).

I am also pretty sure that you could skip the `actions/setup-python@v5` step above, and simply call
```
uv sync --group test --resolution=lowest-direct --python ${{ matrix.python-version }}
```

Issue #748, opened 2025-10-30
Implement spatial hadamard in the fixed_bulk convention


BSchelpe:
This is the remaining subissue in #561. It is expected to require very similar functions to #631, and so the intention is to work on this after #631 is finished.

BSchelpe:
I am planning to work on this now, to get a draft PR ready, before finishing the one for issue #631, so that Gilad can start to look at how his new schedule proposal will interact with the hadamard in the fixed bulk convention.

nelimee:
Double-checking on this issue: https://github.com/tqec/tqec/pull/774 says that it should solve this issue. Should this issue be closed now? Or is there still work to do?

BSchelpe:
Unfortunately it isn't finished. I made that comment early on in my work on #774, when I thought I could do all types of spatial hadamard in one PR. #774 has only tackled one type of spatial hadamard in the fixed bulk convention. The other three remain to be implemented, so this issue needs to stay open for the moment.

PR #774, opened 2025-11-12
Implement spatial hadamards in the fixed bulk

BSchelpe:
This will close issue #748. It is currently a draft PR, because it still needs more work doing (most obviously the left-right hadamard implementing) and testing. I am pushing it now, so that Gilad can do testing on top of it.


BSchelpe:
This PR is now ready for review. 

### Schematic code flow:

get_spatial_pipe_block (one cube spatial) -> get_spatial_cube_pipe_block -> get_spatial_cube_arm_plaquettes
-> get_left_right_arm_spatial_cube_arm or get_up_down_spatial_cube_arm. 
[starting in library/fixed_bulk.py, moving to generators/fixed_bulk.py at get_spatial_cube_arm_plaquettes]

get_spatial_cube_arm_plaquettes now has branches for if is_hadamard and there are two new functions: get_left_right_hadamard_cube_arm_plaquettes and get_up_down_hadamard_cube_arm_plaquettes. 

The left_right_hadamard is not yet implemented in this code push. 

Each configuration then has its own function, for example get_spatial_z_above_rght_and_2_arm_extended_stabiliser_hadamard_plqts

Below are diagrams of the junctions that this now implements. The dotted cubes can either be there or not - it makes no difference to the hadamard. sbb is the boundary basis of the top cube in these pictures. 

IMG_3138
IMG_3139
IMG_3140
IMG_3141

### The issues that will remain after this PR: 
1. This implements spatial hadamards on an up-down spatial cube pipe (ie a pipe connecting at least one spatial cube). Spatial hadamards on a left-right pipe remain to be implemented. This will involve building new differently-shaped extended stabilisers.
2. It doesn't yet have the temporal alternation of schedules for the extended stabilisers. So, at the moment structures containing the extended stabilisers will currently reduce the code distance. I plan to work on this next, but I thought it was less complicated to split it off from this PR and will be easier to test the two changes separately.
3. Spatial hadamards on regular spatial pipes (ie pipes which connect temporal cubes) have not yet been implemented in the fixed bulk convention. 

### Other points to note about the code:

1. I have changed the left-right arm code. It now deletes 'corners'from its pipe, so that it doesn't interfer with those put in by the up-down arms. This is needed for the up-down hadamards. However, this code will need making more sophisticated once left-right hadamards are implemented, eg by also keeping tracking of which arms are hadamards as well as which directions they point in.

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -862,14 +879,15 @@ def get_spatial_cube_arm_raw_template(self, arms: SpatialArms) -> RectangularTem
         else:
             raise TQECError(f"Unrecognized spatial arm(s): {arms}.")
 
-    def get_spatial_cube_arm_rpng_descriptions(
+    def get_spatial_cube_arm_plaquettes(
```
smburdick, line 881: Does renaming the function break any tests or external use cases

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -923,72 +943,29 @@ def get_spatial_cube_arm_rpng_descriptions(
             SpatialArms.RIGHT,
             SpatialArms.LEFT | SpatialArms.RIGHT,
         ]:
-            return self._get_left_right_spatial_cube_arm_rpng_descriptions(
-                spatial_boundary_basis, arms, linked_cubes, reset, measurement
-            )
+            if is_hadamard:
```
smburdick, line 945: Logic feels a little redundant, maybe you could generalize this sequence, which will make it easier to extend in the future

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -923,72 +943,29 @@ def get_spatial_cube_arm_rpng_descriptions(
...
+                return self._get_up_down_spatial_hadamard_cube_arm_plaquettes(
+                    spatial_boundary_basis, arms, linked_cubes, reset, measurement
+                )
+            else:
+                return self._get_up_down_spatial_cube_arm_plaquettes(
+                    spatial_boundary_basis, arms, linked_cubes, reset, measurement
+                )
         raise TQECError(f"Got an invalid arm: {arms}.")
 
-    def get_spatial_cube_arm_plaquettes(
```
smburdick, line 939: Maybe this function is still needed in some places, unless we agreed to delete it at some point. Just being careful

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1031,23 +1008,97 @@ def _get_left_right_spatial_cube_arm_rpng_descriptions(
...
         if SpatialArms.LEFT in arms and _sbb == Basis.Z and SpatialArms.UP in v.spatial_arms:
-            mapping[up] = _corners[0]
+            del mapping[up]  # mapping[up] = _corners[0]
         if SpatialArms.RIGHT in arms and _sbb == Basis.X and SpatialArms.UP in u.spatial_arms:
-            mapping[up] = _corners[1]
-        # Replaces the bottom plaquette if it should be a 3-body stabilizer.
+            del mapping[up]  # mapping[up] = _corners[1]
+        # Replaces the bottom plaquette if there is an arm, which might modify the plaquette to a
+        # corner, or extended triangle.
         if SpatialArms.LEFT in arms and _sbb == Basis.X and SpatialArms.DOWN in v.spatial_arms:
```
smburdick, line 1030: Maybe combine these `if`s into a single one with an `or`?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1031,23 +1008,97 @@ def _get_left_right_spatial_cube_arm_rpng_descriptions(
...
+
+    def _get_left_right_spatial_hadamard_cube_arm_plaquettes(
+        self,
+        spatial_boundary_basis: Basis,
+        arms: SpatialArms,
+        linked_cubes: tuple[CubeSpec, CubeSpec],
+        reset: Basis | None = None,
+        measurement: Basis | None = None,
+    ) -> Plaquettes:
+        raise NotImplementedError()
```
smburdick, line 1098: Why not implemented yet, or do we need the signature around

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1106,6 +1158,108 @@ def _get_up_down_spatial_cube_arm_rpng_descriptions(
...
+        Note:
+        A spatial cube is defined as a cube with all its spatial boundaries
+        in the same basis.
+        Such a cube might appear in stability experiments (e.g.,
+        http://arxiv.org/abs/2204.13834), in spatial junctions (i.e., a cube
+        with more than one pipe in the spatial plane) or in other QEC gadgets
+        such as the lattice surgery implementation of a ``CZ`` gate.
+
+        Warning:
+        This method is tightly coupled with
```
smburdick, line 1178: If they are tightly coupled, should they really be different methods?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1106,6 +1158,108 @@ def _get_up_down_spatial_cube_arm_rpng_descriptions(
...
+            if _sbb == Basis.X and SpatialArms.LEFT not in u.spatial_arms:
+                plqts = self.get_spatial_above_right_arm_extended_stabiliser_hadamard_plqts(
+                    _sbb, reset, measurement
+                )  # confusing name, but same plaquettes as this situation
+            if _sbb == Basis.Z and SpatialArms.RIGHT not in u.spatial_arms:
+                plqts = self.get_spatial_below_left_arm_extended_stabiliser_hadamard_plqts(
+                    _sbb, reset, measurement
+                )
+
+        return plqts
```
smburdick, line 1290: Consider `return`ing in the actual `if` statements

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1384,3 +1538,445 @@ def get_spatial_horizontal_hadamard_plaquettes(
         return self._mapper(self.get_spatial_horizontal_hadamard_rpng_descriptions)(
             top_left_is_z_stabilizer, reset, measurement
         )
+
+    ###############################################################
+    #                Extended stabiliser Hadamards                #
+    ###############################################################
+
+    def get_spatial_z_above_rght_and_2_arm_extended_stabiliser_hadamard_raw_template(
```
smburdick: Would it be better to have one `get_spacial_z_above_right` and provide additional parameters (like the template)?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1384,3 +1538,445 @@ def get_spatial_horizontal_hadamard_plaquettes(
...
+            a description of the plaquettes needed to implement a
+            spatial Hadamard in the extended stabiliser row above a spatial junction which is either
+            double-armed, or has a left arm.
+
+        """
+        # tlb: top-left basis, otb: other basis.
+        tlb, otb = top_left_basis, top_left_basis.flipped()
+        # Generating plaquette descriptions we will need later.
+        extended_plaquette_collection = self.get_extended_plaquettes(reset, measurement)
+        bulk = dict()
```
smburdick: consider rewriting this as
```
bulk = {
   tlb: extended_plaquette_collection[tlb].bulk
   <etc>
}
<etc>
``` 
(Or, do some kind of unpacking)

smburdick (commented):
* Overall, many of the new functions  contain a lot of the same code. I think they can be generalized, to make the code easier to maintain and understand.
* Please update the testing modules to include your code changes, and provide some use cases.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: This function still exists, it is just in a different place in the file now (line 882).

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: The function was only called in the old function get_spatial_cube_arm_plaquettes. It is no longer needed in the new get_spatial_cube_arm_plaquettes. [The old get_spatial_cube_arm_plaquettes function wrapped the get_spatial_cube_arm_rpng_plaquettes. I have simply removed the wrapping.]

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: I find this a useful way to keep track of the fact we need to implement this soon, and where the functions for the left-right hadamard should sit.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: I'm not sure why this would be better?

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: I'm not quite sure what you're proposing as a change? Do you perhaps mean you would rather have is_hadamard as a parameter into a function which contains both the _get_left_right_spatial_hadamard_cube_arm_plaquettes and _get_left_right_spatial_cube_arm_plaquettes code? I agree the code in general feels a bit over-explicit, but I am trying to keep it as simple as possible to make the code flow easy to follow for someone else stepping through how different configurations will pass through tqec.  In this particular place, I think it would be confusing to merge those two functions.

src/tqec/compile/specs/library/generators/fixed_bulk.py
smburdick: I think it's better to raise a specific issue or have a checklist item to implement this, in order to raise more visibility, rather than making the code as-is more complex.

src/tqec/compile/specs/library/generators/fixed_bulk.py
smburdick: It's just a little cleaner than declaring an empty dict and assigning each member imperatively.

src/tqec/compile/specs/library/generators/fixed_bulk.py
smburdick: Okay, just as long as it doesn't break anything.

src/tqec/compile/specs/library/generators/fixed_bulk.py
smburdick: I was just noticing that the new functions look similar and could possibly be combined, for the sake of easier maintainability, like if we decide to have more than 2 arms in the future.

src/tqec/compile/specs/library/generators/fixed_bulk.py
smburdick: ok.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: All existing tests run fine, but thank you for being cautious!

smburdick: Ok great, can you also add tests for your new/updated functions under `tests/compile/specs/library/generators`?

BSchelpe: Lourens has very kindly offered to do this.

BSchelpe:
Thanks for the review!

> * Overall, many of the new functions  contain a lot of the same code. I think they can be generalized, to make the code easier to maintain and understand.

When first looking at it, I agree the code in general feels over-explicit. This was a deliberate general design principle for this part of the code that @nelimee started. Because the code flow through this is very complicated and hard to follow, we didn't want to make it any more complicated to read. That's why the functions are all separate.  It is possible I may have taken @nelimee's principle too far. My preference for how to approach this would be to push the current working very simple code, and then do a second pass to look through and work out whether any shared functions can be pulled out without losing readability. That said if we find specific cases in this code review where a shared function can easily be pulled out I would not be against it :)

> * Please update the testing modules to include your code changes, and provide some use cases.

Lourens has kindly offered to do this.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: Thanks for the explanation. Happy to change it.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: The logic here splits into arms in the y-direction and arms in the x-direction. This is only assuming arms coming out of spatial cubes will always be either in the x or y directions. I think this assumption is hardcoded all over the tqec codebase, and so is ok to bake in here. Unless you were referring to something else by '2 arms'?

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: Sorry if I am being slow, but I don't understand. Could you demonstrate with a code snippet?

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: I have also raised an issue. There will be some error thrown when a graph with a left-right hadamard is sent in. I think it is easier for the user to understand if the code throws a specific notimplemented error. I disagree that it makes the code more complex. I think having this stub here makes the code easier to read, otherwise if I was coming to it for the first time I would not understand why left-right hadamards were not catered for.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: This pattern of 'template, rpng, plaquettes' or 'template, plaquettes' functions is used throughout at least this folder of scripts, and predates my changes. I would be nervous about deviating from it without more understanding of the code outside this folder.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: Happy to change to that.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: See comment on earlier suggestion to combine the template and plaquettes function.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: [but your comment does apply to _get_up_down_spatial_hadamard_cube_arm_plaquettes! Will change there now.]

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: Actually, I can't because it causes the lint error of too many return statements.

BSchelpe:
After running more test configurations, I realised that the above diagrams do not fully capture all possible configurations of the hadamard pipe, in particular situations where both u and v have arms beyond their connecting arm. Below are more comprehensive (also more schematic) diagrams of the possible junctions. The PR logic has also been updated to reflect this. 
Note that in the 'horseshoe' configurations (ie the first in each case) it is a full-sized rectangle added to the boundary of the connecting pipe, whereas in the 'single pipe' (4th) configurations they are half-width rectangles.
IMG_3200
IMG_3199

BSchelpe:
@smburdick, the test is now added

smburdick:
> @smburdick, the test is now added

OK, I'll take a look tomorrow morning.

tests/compile/compile_test.py
```diff
@@ -447,7 +447,6 @@ def test_compile_spatial_hadamard_vertical_correlation_surface(
         )
 
 
-@pytest.mark.skip(reason="Hadamard around spatial junction is not implemented yet.")
 @pytest.mark.parametrize(
     ("k", "convention", "direction", "obs_basis"),
     generate_inputs(CONVENTIONS, (Direction3D.X, Direction3D.Y), (Basis.X, Basis.Z)),
```
smburdick, line 452: I assume we're not testing `Direction3D.Z` because we're not adding Hadamards in the temporal direction?

tests/compile/compile_test.py
```diff
@@ -459,15 +458,27 @@ def test_compile_spatial_hadamard_horizontal_correlation_surface(
...
     kind_before_hadamard = "ZZX" if obs_basis == Basis.Z else "XXZ"
     n1 = g.add_cube(Position3D(0, 0, 0), kind_before_hadamard)
     kind_after_hadamard = "XXZ" if obs_basis == Basis.Z else "ZZX"
     n2 = g.add_cube(Position3D(0, 0, 0).shift_in_direction(direction, 1), kind_after_hadamard)
     g.add_pipe(n1, n2)
 
     d = 2 * k + 1
-    if convention.name == "fixed_bulk":
+    if convention.name == "fixed_boundary":
+        # This is not currently correctly implemented and throws an error complaining about
```
smburdick: Can you make sure this case is documented where it should be thrown?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1106,6 +1158,114 @@ def _get_up_down_spatial_cube_arm_rpng_descriptions(
...
+        )
+
+    def _get_up_down_spatial_hadamard_cube_arm_plaquettes(
+        self,
+        spatial_boundary_basis: Basis,
+        arms: SpatialArms,
+        linked_cubes: tuple[CubeSpec, CubeSpec],
+        reset: Basis | None = None,
+        measurement: Basis | None = None,
+    ) -> Plaquettes:
```
smburdick, line 1219: I think this needs a docstring but I don't know the exact policy on private methods.

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -165,7 +166,9 @@ def get_3_body_rpng_descriptions(
 
         Returns:
             the four 3-body stabilizer measurement plaquettes. Their order follow the usual
-            convention: ``(top_left, top_right, bottom_left, bottom_right)``.
+            convention: ``(top_left, top_right, bottom_left, bottom_right)``. NB the colour of the
+            "corners" is pre-determined in the fixed bulk convention, so we do not need to send in a
+            basis to be chosen by the user.
```
smburdick: "... so we don't need a user-defined basis"?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1031,23 +1008,97 @@ def _get_left_right_spatial_cube_arm_rpng_descriptions(
...
-        # Replaces the top plaquette if it should be a 3-body stabilizer.
+        # Replaces the top plaquette if there is an arm, which might modify the plaquette to a
+        # corner, or extended triangle.
         if SpatialArms.LEFT in arms and _sbb == Basis.Z and SpatialArms.UP in v.spatial_arms:
-            mapping[up] = _corners[0]
+            del mapping[up]  # mapping[up] = _corners[0]
         if SpatialArms.RIGHT in arms and _sbb == Basis.X and SpatialArms.UP in u.spatial_arms:
-            mapping[up] = _corners[1]
-        # Replaces the bottom plaquette if it should be a 3-body stabilizer.
+            del mapping[up]  # mapping[up] = _corners[1]
```
smburdick: Please leave a comment explaining the commented out code, specific to each one.


https://stackoverflow.com/questions/3664288/is-it-bad-practice-to-leave-commented-out-code-in-production-releases

smburdick (commented):
minor comments.

tests/compile/compile_test.py
BSchelpe: Yes, these are only the spatial hadamard tests.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: I left these commented out, because I wasn't sure deleting these corners was the best solution and I wanted to easily be able to reinstate them. I make a comment about this at the beginning of this code section on line 1018.

BSchelpe: To expand on this: this part of the code needs to change relatively soon again after this PR to accommodate left-right hadamards. It is possible that the solution to implementing the left-right hadamards involves reinstating the commented out code, and so I wanted to leave it here to make it easy to reinstate.

tests/compile/compile_test.py
BSchelpe: I've made an explicit NotImplementedError at the point in the code where it should be implemented. And changed this to catch NotImplemented.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: Actually thinking some more. I don't think it is likely the solution will be to reinstate the commented out code. It would be cleaner to remove the comments, and they could always be found in the commit history if needed.

smburdick:
@BSchelpe is this ready for re-review?

BSchelpe:
> @BSchelpe is this ready for re-review?

Yes please. I believe I've addressed all the comments now. From my point of view I think this is finally ready to merge. @KabirDubey is there anything else you would like to see tested/changed?

src/tqec/compile/specs/library/fixed_boundary.py
```diff
@@ -295,7 +295,8 @@ def _get_spatial_cube_pipe_block(
         x, y, z = spec.pipe_kind.x, spec.pipe_kind.y, spec.pipe_kind.z
         assert x is not None or y is not None
         spatial_boundary_basis: Basis = x if x is not None else y  # type: ignore
-        # Get the plaquette indices mappings
+        if spec.pipe_kind.has_hadamard:
```
smburdick, line 298: Maybe leave a comment about why this isn't implemented yet.

src/tqec/visualisation/computation/plaquette/extended.py
```diff
@@ -35,10 +35,10 @@ def flip(self) -> ExtendedPlaquettePosition:
...
 class ExtendedPlaquetteType(Enum):
     BULK = auto()
-    LEFT_WITH_ARM = auto()
-    LEFT_WITHOUT_ARM = auto()
-    RIGHT_WITH_ARM = auto()
-    RIGHT_WITHOUT_ARM = auto()
+    BOTTOM_RIGHT_TRIANGLE = auto()
+    RIGHT_HALF_RECTANGLE = auto()
+    TOP_LEFT_TRIANGLE = auto()
+    LEFT_HALF_RECTANGLE = auto()
```
smburdick, line 41: Can you explain why these are the same, or are the old values being deprecated? Will there be other potential values added here?

Also, I would add a line to the commit when this merges that this enum is being swapped out, since it presumably affects other parts of the system.

tests/compile/compile_test.py
smburdick, line 454: There's no code coverage report, but, I might add just another few tests, especially if it's around 80%.

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1384,3 +1568,614 @@ def get_spatial_horizontal_hadamard_plaquettes(
         return self._mapper(self.get_spatial_horizontal_hadamard_rpng_descriptions)(
             top_left_is_z_stabilizer, reset, measurement
         )
+
+    ###############################################################
```
smburdick, line 1572: Can you add more tests specifically under `library/` and `library/generators`?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1106,6 +1156,140 @@ def _get_up_down_spatial_cube_arm_rpng_descriptions(
...
+        NotImplementedError: if spatial_boundary_basis is not X or Z.
+
+        Returns:
+        the plaquettes needed to implement a hadamard pipe connecting to a
+        spatial cube.
+
+        """
+        u, v = linked_cubes
+        _sbb = spatial_boundary_basis
+        _sbb_f = _sbb.flipped()
```
smburdick, line 1246: variable `_sbb_f` is not used.

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1106,6 +1156,140 @@ def _get_up_down_spatial_cube_arm_rpng_descriptions(
...
+        # Note that in the below names, 'above' refers to above in the flat plane when the
+        # y-axis is pointing downwards. Ie 'above' refers to lower values of y.
+        if _sbb == Basis.Z:
```
smburdick, line 1249: I can't help but feel there's a more elegant (and less error prone) way to implement these if-else statements using a `dict`, but I could be wrong.

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1106,6 +1156,140 @@ def _get_up_down_spatial_cube_arm_rpng_descriptions(
...
+            if (SpatialArms.RIGHT in v.spatial_arms) and (SpatialArms.RIGHT in u.spatial_arms):
+                plqts = self.get_spatial_z_right_horseshoe_extended_stabiliser_hadamard_plqts(
+                    _sbb, reset, measurement
+                )
+            elif (SpatialArms.RIGHT in v.spatial_arms) and (
+                SpatialArms.RIGHT not in u.spatial_arms
+            ):
+                plqts = self.get_spatial_z_above_rght_arm_extended_stabiliser_hadamard_plqts(
```
smburdick, line 1257: Why not just return `plqts` right after the assignment?

Ah, I made a comment about this 2 months ago. I think it should fit on the line if you return right after the assignment. Just makes it less error-prone.

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1384,3 +1568,614 @@ def get_spatial_horizontal_hadamard_plaquettes(
...
+        """
+        return QubitHorizontalBorders()
+
+    def get_spatial_z_above_rght_arm_extended_stabiliser_hadamard_plqts(
+        self,
+        top_left_basis: Basis,
+        reset: Basis | None = None,
+        measurement: Basis | None = None,
+    ) -> Plaquettes:
+        """Return a description of the plaquettes needed to implement a
```
smburdick, line 1593: set of the plaquettes?

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1384,3 +1568,614 @@ def get_spatial_horizontal_hadamard_plaquettes(
...
+            otb: extended_plaquette_collection[otb].bulk,
+        }
+        triangle = {
+            tlb: extended_plaquette_collection[tlb].bottom_left_triangle,
+            otb: extended_plaquette_collection[otb].bottom_left_triangle,
+        }
+        return Plaquettes(
+            FrozenDefaultDict(
+                {
+                    2: triangle[tlb].top,
```
smburdick, line 1644: I would add a comment or use some kind of data structure to illustrate what the range of indices means

src/tqec/compile/specs/library/generators/fixed_bulk.py
```diff
@@ -1384,3 +1568,614 @@ def get_spatial_horizontal_hadamard_plaquettes(
...
+    ) -> RectangularTemplate:
+        """Return the :class:`~tqec.templates.base.RectangularTemplate` instance needed to
+        implement a spatial Hadamard in the extended stabiliser row below a spatial junction
+        which has a right arm, with sbb of the top cube = Z. Note 'below' refers to below in
+        the flat plane when the y-axis is pointing downwards. Ie 'below' refers to higher
+        values of y.
+        """
+        return QubitHorizontalBorders()
+
+    def get_spatial_z_below_rght_arm_extended_stabiliser_hadamard_plqts(
```
smburdick, line 1821: Again, a lot of these methods feel a little redundant. For example:

`get_spatial_z_below_rght_arm_extended_stabiliser_hadamard_plqts`:
```
        tlb, otb = top_left_basis, top_left_basis.flipped()
        # Generating plaquette descriptions we will need later.
        extended_plaquette_collection = self.get_extended_plaquettes(reset, measurement)
        bulk = {
            tlb: extended_plaquette_collection[tlb].bulk,
            otb: extended_plaquette_collection[otb].bulk,
        }
        top_triangle = {
            tlb: extended_plaquette_collection[tlb].top_left_triangle,
            otb: extended_plaquette_collection[otb].top_left_triangle,
        }

        return Plaquettes(
            FrozenDefaultDict(
                {
                    2: top_triangle[tlb].top,
                    4: top_triangle[otb].bottom,
                    5: bulk[tlb].top,
                    6: bulk[otb].top,
                    7: bulk[otb].bottom,
                    8: bulk[tlb].bottom,
                }
            )
        )
```
vs
`get_spatial_x_no_lft_arms_extended_stabiliser_hadamard_plqts`
```
        # tlb: top-left basis, otb: other basis.
        tlb, otb = top_left_basis, top_left_basis.flipped()
        # Generating plaquette descriptions we will need later.
        extended_plaquette_collection = self.get_extended_plaquettes(reset, measurement)
        bulk = {
            tlb: extended_plaquette_collection[tlb].bulk,
            otb: extended_plaquette_collection[otb].bulk,
        }
        right_rectangle = {  # ie it is the right half of a 'normal' plaquette
            tlb: extended_plaquette_collection[tlb].right_half_rectangle,
            otb: extended_plaquette_collection[otb].right_half_rectangle,
        }
        return Plaquettes(
            FrozenDefaultDict(
                {
                    1: right_rectangle[tlb].top,
                    3: right_rectangle[otb].bottom,
                    5: bulk[otb].top,
                    6: bulk[tlb].top,
                    7: bulk[tlb].bottom,
                    8: bulk[otb].bottom,
                }
            )
        )
```

I recommend keeping them separate methods, but using something like a configuration dict to grab the specific triangle and indices that's specific to that method. That will save a lot of code, and make it a lot clearer and easier to extend.

smburdick (changes requested):
* I do think the methods in `fixed_bulk` can be generalized quite a bit, using a configuration, which would save a lot of code.
* Double check that the tests provide good coverage.

BSchelpe:
Below are the configurations I've tested this code for. They are configurations connecting two spatial cubes, and they have been tested for both the sbb of the top cube being X and Z. 

The code that they were tested with checked for the code distance. I give an example of it below. 

IMG_3206

    g = BlockGraph()
    n0 = g.add_cube(Position3D(0, 0, 0), "ZZX", "")
    n1 = g.add_cube(Position3D(0, 1, 0), "XXZ", "")
    g.add_pipe(n0, n1)

    correlation_surfaces = g.find_correlation_surfaces()

    compiled_g = compile_block_graph(
        block_graph=g,
        convention=FIXED_BULK_CONVENTION,  
    )
    k = 1
    noise_model = NoiseModel.uniform_depolarizing(1e-3)
    circuit = compiled_g.generate_stim_circuit(k, noise_model=noise_model)
    shortest_logical_error = circuit.shortest_graphlike_error(
        ignore_ungraphlike_errors=False, canonicalize_circuit_errors=True
    )
    assert len(shortest_logical_error) == 2 * k + 1, (
        "Circuit distance does not match expected value!"
    )

src/tqec/compile/specs/library/fixed_boundary.py
BSchelpe: There isn't a reason, beyond no one has done it yet.

src/tqec/visualisation/computation/plaquette/extended.py
BSchelpe: The reason I've changed the names are they were tightly coupled to one use of the extended stabilisers, and the names were not applicable for other uses. These new names are purely based on the shape of the stabilisers and so are more general. I wouldn't have changed them as part of this PR, but I needed to add 2 new ones and the old naming convention could not be extended to them. I can't think of any other values that would want to be added in the future. I have searched for uses of this enum in the codebase and changed all instances.

tests/compile/compile_test.py
BSchelpe: I think it's best to add more tests separately from this PR, as the PR is already pretty big.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: I think it's best to add more tests separately from this PR, as the PR is already pretty big.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: oops. Thanks. I've removed it.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: It wasn't the length of line that was problematic in your suggested change. It was the lint error of too many return statements.

src/tqec/compile/specs/library/generators/fixed_bulk.py
BSchelpe: Which indices are you referring to? Do you mean 2,4,5,6,7,8? Those are defined in the raw template, which the function's docstring refers to.

smburdick (dismissed):
Looks good. Thank you for addressing my comments
* We can add an action item for more tests if needed (maybe check code coverage again. I think it only shows up on the PR summary if you have write access due to the permissions settings)
* I think stylistically it is fine. We could probably do more as a team to specifically define what our coding standards should be, but it's consistent with the rest of the code.

BSchelpe:
> Looks good. Thank you for addressing my comments

Thank you! I've been thinking about your comments about changing to a configuration dict to reduce the amount of repeated code. I think I see how to do it, but it would be nice to do it in its own PR since the rearrangement is a bit complicated, and the code is currently functional so it would be nice to push this PR now. Thank you for approving! (Would you be able to reapprove? Sorry, I've just pushed the two small changes you suggested in your last review.)

smburdick (approved):
LGTM.

smburdick:
> > Looks good. Thank you for addressing my comments
> 
> Thank you! I've been thinking about your comments about changing to a configuration dict to reduce the amount of repeated code. I think I see how to do it, but it would be nice to do it in its own PR since the rearrangement is a bit complicated, and the code is currently functional so it would be nice to push this PR now. Thank you for approving! (Would you be able to reapprove? Sorry, I've just pushed the two small changes you suggested in your last review.)

Sure, I will raise a couple issues on these points.

PR #777, opened 2025-11-15
feat: support re-schedule measurements

### LINKED ISSUE #388: `Schedule`s are not really flexible
As noted in #384, `Plaquette` scheduling is quite inflexible right now. We basically need to have an implicit global schedule for the library of plaquettes used, and each plaquette need to respect that implicit global schedule.

As an example, for the CSS set of plaquettes, the global schedule is:

1. Resets,
2. Potential Hadamard gate to change the reset basis,
3. First CNOT,
4. Second CNOT,
5. Third CNOT,
6. Fourth CNOT,
7. Potential Hadamard gate to change the measurement basis,
8. Measurements

#384 needs to add a transversal Hadamard layer on all the data-qubits. For that very specific issue, and for the CSS plaquettes, it turns out that we can likely push the Hadamard gate to either `2` or `7` using the fact that an Hadamard transition cannot happen before the data qubits have been initialized (meaning that Hadamard gates inserted in an hypothetical schedule `0` could be moved forward to `2` without any issue) or after they have been measured (meaning that Hadamard gates inserted in an hypothetical schedule `9` could be moved backward to `7` without any issue), canceling out duplicated Hadamard gates. But this approach might not be valid for all the sets of plaquettes.

Zhaoyilunnn:
close #388 

Work in progress, there are bugs for now.

### DIFF
--- a/tests/circuit/schedule/circuit_test.py
+++ b/tests/circuit/schedule/circuit_test.py
@@ -229,3 +229,33 @@ def test_scheduled_circuit_filter_by_qubit() -> None:
     filtered_circuit = circuit.filter_by_qubits([GridQubit(0, 0), GridQubit(0, 1)])
     assert filtered_circuit.get_circuit() == stim.Circuit("QUBIT_COORDS(0, 0) 0\nH 0")
     assert filtered_circuit.schedule == Schedule([0])
+
+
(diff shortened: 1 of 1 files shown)


Zhaoyilunnn:
needs more effort to debug, close for not bothering other people

Issue #804, opened 2025-12-30
Improve circuit generation runtimes by reducing redundancies and unnecessary checks


jbolns:
## Feature request description.

While discussing Pandora with @alexandrupaler, we found a couple things that need to be done TQEC-side to ensure TQEC is able to produce circuits as quickly as possible. 

The improvements seem relevant irrespective of any Pandora-related conversations. 

In essence, TQEC is leaving a lot of performance on the table. 

## Preliminary ideas for solution

Three relatively small adjustments have stood out thus far, and additional ideas are welcome. 

(*Do not be shy to point errors in the below. I'm only recently starting to engage in depth with this side of TQEC, so caution is advisable.*)

### Cache for cube specs
It seems like we are underutilising caches. For instance, it is possible to **very significantly** cut the runtimes needed to generate a very long memory experiment by adding caches to methods in [fixed_bulk.py](https://github.com/tqec/tqec/blob/3c0db694339614f537772a7bf6a183c8ffcb78a7/src/tqec/compile/specs/library/fixed_bulk.py) and a couple other in [plaquettes.py](https://github.com/tqec/tqec/blob/1414538334a378fcf5c7c35451bd07f5987e22aa/src/tqec/compile/blocks/layers/atomic/plaquettes.py).

I am not suggesting the above as *the* solution. I might be over- or under-utilising them. Alternatively, it might be possible to achieve the same using less caches placed elsewhere.

But it does seem it wouldn't hurt to add a cache here and there, especially where it related to operations that repeat in plenty during circuit creation.

### Not considering observables.
There are some checks related to observables that can probably be avoided if observables are not considered.

It is possible to not consider observables by sending `None` to [compile_block_graph](https://github.com/tqec/tqec/blob/8acd186cbdcbea05fc33acdd73503b984f3d44f2/src/tqec/compile/compile.py#L71). Alternatively, it might be worth considering changing the default value for that optional parameter from "auto" to `None`. 

```
def compile_block_graph(
    block_graph: BlockGraph,
    convention: Convention = FIXED_BULK_CONVENTION,
    observables: list[CorrelationSurface] | Literal["auto"] | None = "auto", 
)
```

Additional gains seem possible by ensuring that when observables are not considered, the `observable_builder` does not trigger unnecessary operations down the line, specifically, in `TopologicalComputationGraph` --> `to_layer_tree` --> `_annotate_observables`.

### Yielding rather than returning circuits
[...]

PR #805, opened 2026-01-02
Add streaming from LayerTree to Stim circuit

BluSteve:
This rather large PR addresses some points in #804. It is currently incomplete, but I am drafting this to gather feedback.

### What was done
The annotation generation for circuits, detectors, and observables have been moved to the main compilation recursion. Annotation generation used to be preprocessed, which took up 90% of the runtime of `layertree.generate_circuit`.

By streaming annotation generation together with "annotation -> Stim code", TQEC is now able to output granular slices of a circuit before annotating the full circuit, reducing compilation response time massively. There is however no change to compilation turnaround time.

Qubit map posed a challenge as the original implementation parsed the entire circuit to determine what qubits were involved. As such, a qubit map could only be generated upon the full compilation of the circuit. This runs contrary to the philosophy of a streaming compiler, where instructions must be executable on a quantum computer before the compilation has completed.

Therefore, a qubit map that is a superset of the qubits needed must be passed into the `generate_circuit_stream` method. I believe this would not be a difficult parameter for the end user to determine on their side, because one would know how many qubits they have before starting a circuit to run on those qubits.

It is also expected that the memory footprint of `generate_circuit_stream` is much lower than its unstreamed counterpart, because one can pipe each circuit slice to disk instead of storing it entirely in memory.

### Demo output from `streaming.py (nx = 3, ny = 3, t = 10, k = 2)`
```
Benchmarking streaming with nx=3, ny=3, t=10, k=2

Layer tree generation time (s): 1.3971402645111084

Unstreamed circuit generation time (s): 3.438138484954834

Initial stream generation time (s): 2.1457672119140625e-06

Index, Time Taken (s), Circuit Snippet
0, 0.5272216796875, 
1, 0.003019094467163086, QUBIT_COORDS(0, 0) 0
2, 0.08830690383911133, RX 2 4 8 10 14 16 18
3, 0.08120489120483398, REPEAT 3 {     TICK 
4, 8.916854858398438e-05, TICK
5, 0.08701610565185547, RX 2 4 8 10 14 16 35
6, 0.00010919570922851562, TICK
7, 0.08097243309020996, RX 2 4 8 10 14 16 35
8, 0.07716012001037598, REPEAT 3 {     TICK 
9, 9.441375732421875e-05, TICK
10, 0.08183670043945312, RX 2 4 8 10 14 16 35
11, 0.00012612342834472656, TICK
12, 0.09161663055419922, RX 2 4 6 8 10 12 14 
13, 0.08293986320495605, REPEAT 3 {     TICK 
14, 8.392333984375e-05, TICK
15, 0.09255671501159668, RX 2 4 6 8 10 12 14 
16, 0.00011968612670898438, TICK
17, 0.08626627922058105, RX 2 4 8 10 14 16 35
18, 0.07943367958068848, REPEAT 3 {     TICK 
19, 9.131431579589844e-05, TICK
20, 0.08783793449401855, RX 2 4 8 10 14 16 35
21, 0.000118255615234375, TICK
22, 0.09179830551147461, RX 2 4 6 8 10 12 14 
23, 0.08345770835876465, REPEAT 3 {     TICK 
24, 8.7738037109375e-05, TICK
25, 0.09314703941345215, RX 2 4 6 8 10 12 14 
26, 0.00010776519775390625, TICK
27, 0.09747004508972168, RX 2 4 8 10 14 16 35
28, 0.08734583854675293, REPEAT 3 {     TICK 
29, 0.00011706352233886719, TICK
30, 0.0986020565032959, RX 2 4 8 10 14 16 35
31, 0.0001227855682373047, TICK
32, 0.0910801887512207, RX 2 4 8 10 14 16 35
33, 0.08283138275146484, REPEAT 3 {     TICK 
34, 8.559226989746094e-05, TICK
35, 0.09275221824645996, RX 2 4 8 10 14 16 35
36, 0.00010037422180175781, TICK
37, 0.08691787719726562, RX 2 4 6 8 10 14 16 
38, 0.0796966552734375, REPEAT 3 {     TICK 
39, 9.059906005859375e-05, TICK
40, 0.08720612525939941, RX 2 4 6 8 10 14 16 
41, 0.00011301040649414062, TICK
42, 0.08677363395690918, RX 2 4 8 10 14 16 35
43, 0.07991170883178711, REPEAT 3 {     TICK 
44, 9.012222290039062e-05, TICK
45, 0.08720779418945312, RX 2 4 8 10 14 16 35
46, 9.751319885253906e-05, TICK
47, 0.0709226131439209, RX 2 4 8 10 14 16 35
48, 0.07046747207641602, REPEAT 3 {     TICK 
49, 8.845329284667969e-05, TICK
50, 0.07999014854431152, RX 2 4 8 10 14 16 35
Total streamed circuit generation time (s): 3.1571872234344482

Comparing streamed vs unstreamed circuits...
Circuits are the same!
```

In this example, the response time was lowered from 3.5s to 0.63s (if we look index 2, which is the first Stim gate instruction. A longer circuit would make the response time improvement even more significant. 

### To do
- Proper testing (moving `streaming.py` to somewhere more appropriate than the root directory)
- Streaming from BlockGraph to LayerTree. I believe this would lead to an even shorter response time. For a 10x10 logical qubit circuit with depth 200, compiling the BlockGraph takes upwards of 40s.


alexandrupaler:
@HaoTy would you be willing to review this PR?

HaoTy:
> @HaoTy would you be willing to review this PR?

I'd be happy to give comments and suggestions, but I don't have write access so can't approve anything

nelimee:
> I'd be happy to give comments and suggestions, but I don't have write access so can't approve anything

You should have write access now (or an invitation to get it in your mailbox) :+1:

HaoTy:
> You should have write access now (or an invitation to get it in your mailbox) :+1: 

Thanks!

src/tqec/compile/graph.py
```diff
@@ -517,6 +518,61 @@ def generate_stim_circuit(
...
+            detector_database=detector_database,
+            database_path=database_path,
+            do_not_use_database=do_not_use_database,
+            only_use_database=only_use_database,
+            reschedule_measurements=reschedule_measurements,
+        )
+        # If provided, apply the noise model.
+        # if noise_model is not None: # TODO noise_model not supported with streaming
+        #     circuit = noise_model.noisy_circuit(circuit)
+        # return circuit
```
HaoTy: What are the difficulties of applying the noise model to the circuit being streamed, the qubit indices? We can adapt the `NoiseModel` class if necessary.

src/tqec/compile/tree/node.py
```diff
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
 import itertools
-from collections.abc import Mapping, Sequence
-from typing import Any, TypeGuard
+from collections.abc import Iterator, Mapping, Sequence
+from typing import Any, TypeGuard, override, Dict
```
HaoTy:
```suggestion
from typing import Any, TypeGuard
from typing_extensions import override
```

`override` from `typing` is only available in 3.12+

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+        self,
+        k: int,
+        global_qubit_map: QubitMap,
+        reschedule_measurements: bool,
+        detectors_walker: AnnotateDetectorsOnLayerNode,
+        subtree_to_z: Dict[LayerNode, int], # Maybe this doesn't have to be passed down so many layers
+        abstract_observables,
+        observable_builder,
+        add_polygons: bool = False,
+        leaf_dict: Dict | None = None,
```
HaoTy:
```suggestion
        leaf_dict: dict[] | None = None,
```

Please add key and value typings as well.

src/tqec/compile/tree/node.py
```diff
@@ -244,3 +552,39 @@ def generate_circuit(self, k: int, global_qubit_map: QubitMap) -> stim.Circuit:
             assert isinstance(circuit, stim.Circuit)
             ret += circuit
         return ret
+
+    def generate_circuit_stream(self, k: int, global_qubit_map: QubitMap,
+                                reschedule_measurements: bool,
+                                detectors_walker: AnnotateDetectorsOnLayerNode,
+                                subtree_to_z: Dict[LayerNode, int],
```
HaoTy:
```suggestion
                                subtree_to_z: dict[LayerNode, int],
```

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
             return ret
         raise TQECError(f"Unknown layer type found: {type(self._layer).__name__}.")
 
+    def generate_circuits_with_potential_polygons_stream(
+        self,
+        k: int,
+        global_qubit_map: QubitMap,
+        reschedule_measurements: bool,
+        detectors_walker: AnnotateDetectorsOnLayerNode,
+        subtree_to_z: Dict[LayerNode, int], # Maybe this doesn't have to be passed down so many layers
```
HaoTy:
```suggestion
        subtree_to_z: dict[LayerNode, int], # Maybe this doesn't have to be passed down so many layers
```

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+        abstract_observables,
+        observable_builder,
```
HaoTy: Add typings

src/tqec/compile/tree/tree.py
```diff
@@ -372,6 +371,134 @@ def generate_circuit(
...
+
+        if include_qubit_coords:
+            yield annotations.qubit_map.to_circuit()
+
+        subtree_to_z = {subtree_root : z for (z, subtree_root) in enumerate(self._root.children)}
+
+        yield from self._root.generate_circuit_stream(k, annotations.qubit_map,
+                                                      reschedule_measurements, detectors_walker,
+                                                      subtree_to_z, self._abstract_observables,self._observable_builder)
+
```
HaoTy, line 473: This has a lot of duplicate code with non-stream version of the method. We should reuse code when possible. Option 1, create a new method for common code. Option 2, implement the non-stream version using the stream version. Of course, if, at the end, we successfully make streaming not lead to any downsides, we don't have to maintain two versions and can just have a single stream version.

src/tqec/compile/graph.py
```diff
@@ -517,6 +518,61 @@ def generate_stim_circuit(
...
+        """
+        yield from self.to_layer_tree().generate_circuit_stream(
+            k,
+            manhattan_radius=manhattan_radius,
+            detector_database=detector_database,
+            database_path=database_path,
+            do_not_use_database=do_not_use_database,
+            only_use_database=only_use_database,
+            reschedule_measurements=reschedule_measurements,
+        )
```
HaoTy: Required argument `qubit_map` is missing. Judging from the PR description, it should be passed down from `compile_block_graph()`. However, I don't think the user has to specify it; because the block graph is always known before the compilation, we should be able to figure it out at the latest point where we still have the complete picture of the computation.

HaoTy: I think one solution is to decouple `_annotate_qubit_map()` from `_generate_annotations()` and call the former before streaming, because `to_layer_tree()` is not streamed and we already have every node in the layer tree. In any case, it won't degrade the performance since the annotation walks were separate in the old implementation.

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
             return ret
         raise TQECError(f"Unknown layer type found: {type(self._layer).__name__}.")
 
+    def generate_circuits_with_potential_polygons_stream(
```
HaoTy: The code reuse suggestion also applies here. But let's address the qubit_map limitation and only have the stream version.

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+
+        """
+        if isinstance(self._layer, LayoutLayer):
+            annotations = self.get_annotations(k)
+
+            # circuit
+            base_circuit = self._layer.to_circuit(
+                k, reschedule_measurements=reschedule_measurements
+            )
+            if base_circuit is None:
```
HaoTy: This check seems unnecessary with the current implementation.

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+            mapped_circuit.append_annotation(
+                stim.CircuitInstruction(
+                    "SHIFT_COORDS", [], StimCoordinates(0, 0, 1).to_stim_coordinates()
+                )
+            )
+
+            if add_polygons:
+                yield annotations.polygons
+
+            yield mapped_circuit.get_circuit(include_qubit_coords=False)
```
HaoTy: Does it matter that the circuit and polygon order is reversed? I'm not particularly familiar with Crumble.

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+                leaves = _get_ordered_leaves(self)
+
+                for obs_idx, observable in enumerate(abstract_observables):
+                    obs_slice = observable.slice_at_z(z)
+
+                    # Python is an awful language
+                    def ao0(leaf: LayerNode, obs_slice=obs_slice, k=k,
+                            obs_idx=obs_idx, observable_builder=observable_builder):
+                        _annotate_observable_at_node(leaf, obs_slice, k, obs_idx, observable_builder,
+                                                     ObservableComponent.BOTTOM_STABILIZERS)
```
HaoTy:
```suggestion
                    def ao(leaf: LayerNode, component: ObservableComponent):
                        _annotate_observable_at_node(leaf, obs_slice, k, obs_idx, observable_builder, component)
```

The three annotate observable functions can be merged into one (with `partial` on `component`). In general, the implementation here looks a bit clunky. I'm sure we can find a way to make `leaf_dict`, etc. more elegant.

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+                (i for i in body if isinstance(i, stim.Circuit)),
+                start=stim.Circuit(),
+            )
+            body_circuit.insert(0, stim.CircuitInstruction("TICK"))
+
+            if add_polygons:
+                yield from body  # only keep the first set of polygons
+
+            yield body_circuit * self._layer.repetitions.integer_eval(k)
+
```
HaoTy, line 370:
```suggestion
        raise TQECError(f"Unknown layer type found: {type(self._layer).__name__}.")

```

src/tqec/compile/tree/node.py
```diff
@@ -222,6 +357,179 @@ def generate_circuits_with_potential_polygons(
...
+        detectors_walker.exit_node(self)
```
HaoTy: Did `detectors_walker` enter node if it's a `SequencedLayers`?

streaming.py
```diff
@@ -0,0 +1,176 @@
...
+                continue
+            qubits[i] = GridQubit(b, c)
+            i += 1
+
+    qm = QubitMap(qubits)
+
+    return qm
+
+
+def _random_block_graph(nx: int, ny: int, t: int) -> BlockGraph:
```
HaoTy: This can be very useful for testing and benchmarking other things if we improve it with controllable connectivity and port specifications. We can worry about it in a future PR.

HaoTy (changes requested):
Let's find a way to address the qubit map problem. The user shouldn't be burdened to learn how to specify it. I've proposed a solution in the review.

If qubit map is the only downside of the streaming implementation, we can have it as the default and get rid of the non-streaming version (after addressing qubit map).

I assume things were moved to avoid circular imports. I don't have a strong opinion on where things should be. Perhaps @inmzhang can check whether the moves are appropriate, and also suggest where `streaming.py` should be placed.

Please run `pre-commit install` (see https://tqec.github.io/tqec/contributor_guide.html) so that ruff checks (and fixes some) formatting issues before each commit. Run `ty check` to check for typing errors.

BluSteve:
Thanks for the feedback Tianyi! Yep the moves were haphazardly done to resolve circular dependencies. 

I will take a closer look at the suggested changes within a few days.

src/tqec/compile/tree/node.py
BluSteve: I will modify it such that it enters and exits node regardless of type. But currently detectors_walker just returns if it is a SequencedLayer.

src/tqec/compile/graph.py
BluSteve: We would definitely need to adapt the `NoiseModel` class. I don't think I'm familiar enough with Stim as of yet to do that. I suggest that we keep it as a work in progress for now as the circuit streaming may be useful even without noise models.

src/tqec/compile/tree/node.py
BluSteve: I'm also not familiar enough with Crumble.

src/tqec/compile/tree/node.py
BluSteve: I've cleaned this up and used `partial`s.

src/tqec/compile/tree/tree.py
BluSteve: I've refactored out the verbose database generation code into a shared method.

BluSteve:
@HaoTy Hi Tianyi, my apologies for the delay. All the suggestions have been made except those bottlenecked by the `qubit_map` problem. Type issues have been fixed and pre-commit checks pass.

At present, I think the streaming version is still too unstable to fully replace the non-streaming version. I've thus written the code such that the functionality of the non-streaming version is not affected whatsoever.

I think the streaming version could be merged and used by interested users first while the last TODOs (mainly the noise model) are being ironed out.


I believe only the following issues prevent this PR from being merged:
1. The `qubit_map` problem
2. I'm unsure what testing standards are necessary
3. (related to point 1) I'm not sure where to put `streaming.py` and how to convert it to an automated test

HaoTy:
Your ruff and ty configs need to be tweaked so that they follow pyproject.toml. Currently, the imports are not sorted, some lines are too long, and the type check also fails. Be sure to also run the slow tests with `pytest -m "slow"`. The current code passes all non-slow tests but fails for almost all tests in compile_test.py. 

Tests can be added to the corresponding test files wherever the source code has the streaming version added. The best way in my mind is to add a parameter for streaming vs non-streaming with some small modifications to the existing testing code.

streaming.py basically does three things: 1. generate random block graphs, 2. address the qubit mapping of the streaming version, and 3. check that streaming version circuits match the non-streaming version. Random block graph generation can be polished and added to the gallery module in a future PR. Qubit mapping should be addressed in the implementation of the streaming version. Validation of correctness is addressed by tests. No need to place streaming.py somewhere; its content can be devided and absorbed into different components.

src/tqec/compile/graph.py
HaoTy: I still don't understand what's the obstacle here. What's the issue if we just apply the noise model to every piece of the generated circuit in the stream?

src/tqec/compile/tree/node.py
HaoTy: Please add key and value typings as well.

src/tqec/compile/tree/node.py
HaoTy: Add typings

BluSteve:
Hi Tianyi, ruff and ty now pass, and all slow tests pass except for one. 

I have added a very complete but perhaps excessive way of testing that streaming produces the same circuits as nonstreaming, where I added a few lines comparing the unstreamed with streamed in the main `generate_circuit_and_assert` method in `compile_test.py`.

All of the tests that use this method pass except for one. As this method is widely used in downstream test cases, I believe that this is a good way of verifying the streaming version performs equivalently to the nonstreaming version.

The test that doesn't pass is below:

`FAILED tests/compile/compile_test.py::test_compile_move_rotation[1-convention2-X] - AssertionError: assert stim.Circuit(...(0, 0, 1)\n''') == stim.Circuit(...(0, 0, 1)\n''')`

image

Here, for some reason unclear to me after investigation, the streaming version adds three extra ticks. Note that the presence of two consecutive ticks is not itself a bug, as the unstreamed version also outputs consecutive ticks.

I am not sure if this behavior is fundamentally wrong.

BluSteve:
Regarding the qubit map issue, I personally believe that the user should have this information available before they have the full circuit, so they can pass it in.

Any algorithm that returns the qubit map dynamically from the circuit-- needs to have the full circuit first, which doesn't work with streaming.

The workflow that streaming allows is to
1) have a quantum computer with some fixed number of qubits
2) stream instructions to the quantum computer
3) qc executes streamed instructions while the rest of the circuit is compiled.

HaoTy:
Sorry for the late response, and thanks for fixing the format and type hinting.

For the workflow you mentioned, we indeed need to perform run-time compilation and stream instructions to the hardware, but it's because of conditional operations needed for QEC, not that we can't wait for the compilation. To be clear, in practice we most definitely want every static part compiled at compile time, so the chance of circuit execution being delayed by compilation is minimized.

For the qubit map, it's something that the compiler can figure out and doesn't matter to the user, so we shouldn't rely on the user to provide it. We want to keep the current user-friendliness, where the user only needs to construct the `BlockGraph`, nothing lower-level than that. In the future, the goal is to let the user input the logical circuit only, without worrying about compilation details unless they explicitly want to.
I see that the added tests get the qubit map by `layer_tree._get_global_qubit_map(k)`. Is it not something we can do in the compilation since the streaming circuit is compiled from the layer tree?
Also note that I have previously proposed a solution:
> I think one solution is to decouple `_annotate_qubit_map()` from `_generate_annotations()` and call the former before streaming, because `to_layer_tree()` is not streamed and we already have every node in the layer tree. In any case, it won't degrade the performance since the annotation walks were separate in the old implementation.

For the repeated tick issue, indeed I don't immediately see it breaking any functionality (afaik, TICKs are checked by tqecd to generate detectors, which look correct), but it's a bit annoying that the two versions sometimes don't match. Maybe we can do some postprocessing to remove the duplicate TICK annotations if we can't fixed the root cause.

Lastly, I still have my question about the noise model issue. What will be wrong if we just apply the noise model to every piece of the generated circuit in the stream?

afowler:
I would say that in the future we will want to constrain the compilation to
a specific rectangular grid of logical qubits. So long term the user might
have to specify two numbers that give the dimensions of this array, or
alternatively specify a compilation target such as minimum space or minimum
time and the compiler will output these numbers.

HaoTy:
> I would say that in the future we will want to constrain the compilation to a specific rectangular grid of logical qubits. So long term the user might have to specify two numbers that give the dimensions of this array, or alternatively specify a compilation target such as minimum space or minimum time and the compiler will output these numbers.

Yes, I forgot to mention that. But in my mind, it can be figured out by the compiler from the target hardware specs if the user doesn't want to explicitly specify it. The compiler should be able to decide a target code distance from the physical error rate and the logical circuit. Then with the physical qubit array information, the logical qubit grid size can be decided.

To add a bit of clarity for this PR, this discussion is separate from the qubit map issue, which concerns physical qubits.

BluSteve:
> I think one solution is to decouple `_annotate_qubit_map()` from `_generate_annotations() `and call the former before streaming, because `to_layer_tree()` is not streamed and we already have every node in the layer tree. In any case, it won't degrade the performance since the annotation walks were separate in the old implementation.

I tried to implement the above solution, but the issue is that the QubitLister (which is the main method in `_annotate_qubit_map()`) needs every node to have a circuit annotation.

``` py
    def visit_node(self, node: LayerNode) -> None:
        if not node.is_leaf:
            return
        annotations = node.get_annotations(self._k)
        if annotations.circuit is None:
            raise TQECError("Cannot list qubits without the circuit annotation.")
        self._seen_qubits |= annotations.circuit.qubits
```

Qubit map annotation is fast, but circuit annotation is slow and needs to be streamed. 

In the tests, I had to use the annotated LayerTree from the non-streaming version to "jump start" the streaming version.

HaoTy:
We should be able to address the qubit map problem after carrying out https://github.com/tqec/tqec/issues/847#issuecomment-3961217651.

BluSteve:
Is this PR blocked until #847 clears? I believe I can fix the noise model issue, after which qubit map would be the sole remaining obstacle.

smburdick:
> Is this PR blocked until #847 clears? I believe I can fix the noise model issue, after which qubit map would be the sole remaining obstacle.

I think it will be much faster once #857 is merged

smburdick:
@BluSteve please resolve the merge conflicts and bring this branch up-to-date, and request a new review, if you still want us to review it. Otherwise, please close it.

BluSteve:
I see that #847 has been resolved. That should be all the bottlenecks fixed for this PR, so I'll patch it up soon and merge.

HaoTy:
To clarify, #847 was closed because the performance issue was addressed in another way, not because #847 itself was addressed. I will try to take another look at the qubit map blocker this weekend.

HaoTy:
@BluSteve see https://github.com/BluSteve/tqec/pull/1 for the qubit map issue.

HaoTy:
With #924, the extra TICK issue is resolved, and we can implement `generate_circuit` with the streaming version.

BluSteve:
I've started to resolve the merge conflict with tqec:main. The only todo after that is to figure out applying a noise model.

Both should be done by the end of this week.

BluSteve:
@HaoTy I believe I've resolved all merge conflicts with tqec:main. I've also made applying noise models work. The key was to partition the streamed slices into moment-aligned pieces, then pass in the `system-qubits` parameter into noise model so it's aware of the global set of qubits.

`ty check` fails with errors in parts of the code I never touched, so I'm not sure it was the result of this PR.

I ran all tests, fast and slow, and the unstreamed and streamed versions produce the same circuit. I've ran this for the case where there's no noise model applied, and I'm running the noise model version right now. It would take about 1.5 hours. 

I think this PR is very close to merging.

BluSteve:
Update: all tests pass, which means unstreamed and streamed are congruent. Should the streamed version fully replace the unstreamed one?

HaoTy:
> @HaoTy I believe I've resolved all merge conflicts with tqec:main. I've also made applying noise models work. The key was to partition the streamed slices into moment-aligned pieces, then pass in the `system-qubits` parameter into noise model so it's aware of the global set of qubits.

Nice!

> `ty check` fails with errors in parts of the code I never touched, so I'm not sure it was the result of this PR.

It could be that you used a different `ty` version. It's fine if you can pass the GitHub CI. Currently, I think there are some formatting issues causing the pre-commit-ci to fail.

> Update: all tests pass, which means unstreamed and streamed are congruent. Should the streamed version fully replace the unstreamed one?

Is there a non-trivial performance difference? If not, we can implement the unstreamed method using the streamed one to avoid duplicate code.

BluSteve:
There's seems to be no test for add_polygons, even before this PR. I think the current code works by inspection, but I'm not familiar enough with add polygons to create a meaningful test case for it.

BluSteve:
The performance between unstreamed and streamed is comparable. You can use this benchmark file. I'm not adding to the repository because it requires commenting out these lines in `block_graph`.

image


[streaming.py](https://github.com/user-attachments/files/28021492/streaming.py)

BluSteve:
I've thought about it some more. Replacing the unstreamed version with the streamed one naively wouldn't work as they do not return the same types. Unstreamed is a circuit whereas streamed is an Iterator.

Maybe we could use the streamed one in the implementation of the unstreamed one and keep both to uphold backwards compatibility?

HaoTy:
> I've thought about it some more. Replacing the unstreamed version with the streamed one naively wouldn't work as they do not return the same types. Unstreamed is a circuit whereas streamed is an Iterator.
> 
> Maybe we could use the streamed one in the implementation of the unstreamed one and keep both to uphold backwards compatibility?

This is what I was proposing.

BluSteve:
Do you want write a polygons test? I could also just make unstreamed use streamed already.

BluSteve:
Hi Tianyi, following up on this.

HaoTy:
> Hi Tianyi, following up on this.

Sorry, missed your previous message. (@ me next time if I don't respond timely). I will look into whether and how we should add tests for polygons. As far as I know, it's currently only used by Crumble?

BluSteve:
Yes, I believe it's only used for visualizations.

KabirDubey:
A polygon is just a background color to decorate the Crumble visualization of the stabilizer of the circuit's data. These are static in the support of the logical qubit(s). If the patch is just doing memory, then I think they can just be added once; they aren't like the flow annotations on measure qubits (detection regions) which change shape in time. If the circuit has some lattice surgery in space and if we wanted to commit to decorating all stabilizers it would look something like [this](https://algassert.com/crumble#circuit=Q(0,2)0;Q(0.5,0.5)1;Q(0.5,1.5)2;Q(0.5,2.5)3;Q(1,0)4;Q(1,1)5;Q(1,2)6;Q(1.5,0.5)7;Q(1.5,1.5)8;Q(1.5,2.5)9;Q(2,1)10;Q(2,2)11;Q(2,3)12;Q(2.5,0.5)13;Q(2.5,1.5)14;Q(2.5,2.5)15;Q(3,0)16;Q(3,1)17;Q(3,2)18;Q(3.5,0.5)19;Q(3.5,1.5)20;Q(3.5,2.5)21;Q(4,1)22;Q(4,2)23;Q(4,3)24;Q(4.5,0.5)25;Q(4.5,1.5)26;Q(4.5,2.5)27;Q(5,0)28;Q(5,1)29;Q(5,2)30;Q(5.5,0.5)31;Q(5.5,1.5)32;Q(5.5,2.5)33;Q(6,1)34;Q(6,2)35;Q(6,3)36;Q(6.5,0.5)37;Q(6.5,1.5)38;Q(6.5,2.5)39;Q(7,1)40;POLYGON(0,0,1,0.25)7_13_14_8;POLYGON(0,0,1,0.25)2_8_9_3;POLYGON(0,0,1,0.25)7_1;POLYGON(0,0,1,0.25)15_9;POLYGON(0,0,1,0.25)31_37_38_32;POLYGON(0,0,1,0.25)26_32_33_27;POLYGON(0,0,1,0.25)31_25;POLYGON(0,0,1,0.25)39_33;POLYGON(1,0,0,0.25)1_7_8_2;POLYGON(1,0,0,0.25)8_14_15_9;POLYGON(1,0,0,0.25)2_3;POLYGON(1,0,0,0.25)13_14;POLYGON(1,0,0,0.25)25_31_32_26;POLYGON(1,0,0,0.25)32_38_39_33;POLYGON(1,0,0,0.25)26_27;POLYGON(1,0,0,0.25)37_38;TICK;R_1_2_3_7_8_9_13_14_15_25_26_27_31_32_33_37_38_39;MARKZ(1)25_26_27_13_14_15;TICK;R_6_10_4_12_30_34_28_36;RX_5_11_0_17_29_35_23_40;MARKX(0)1_7_13_25_31_37;TICK;CX_5_1_2_6_7_10_11_8_9_12_17_13_29_25_26_30_31_34_35_32_33_36_40_37;TICK;CX_8_6_13_10_5_2_11_9_15_12_17_14_32_30_37_34_29_26_35_33_39_36_40_38;TICK;CX_3_6_8_10_1_4_5_7_0_2_11_14_27_30_32_34_25_28_29_31_23_26_35_38;TICK;CX_5_8_9_6_14_10_11_15_0_3_7_4_29_32_33_30_38_34_35_39_23_27_31_28;TICK;M_6_10_4_12_30_34_28_36;MX_5_11_0_17_29_35_23_40;DT(1,2,0)rec[-16];DT(2,1,0)rec[-15];DT(1,0,0)rec[-14];DT(2,3,0)rec[-13];DT(5,2,0)rec[-12];DT(6,1,0)rec[-11];DT(5,0,0)rec[-10];DT(6,3,0)rec[-9];TICK;POLYGON(0,0,1,0.25)2_8_9_3;POLYGON(0,0,1,0.25)7_1;POLYGON(0,0,1,0.25)15_9;POLYGON(0,0,1,0.25)31_37_38_32;POLYGON(0,0,1,0.25)31_25;POLYGON(0,0,1,0.25)39_33;POLYGON(0,0,1,0.25)7_13_14_8;POLYGON(0,0,1,0.25)26_32_33_27;POLYGON(1,0,0,0.25)1_7_8_2;POLYGON(1,0,0,0.25)8_14_15_9;POLYGON(1,0,0,0.25)2_3;POLYGON(1,0,0,0.25)25_31_32_26;POLYGON(1,0,0,0.25)32_38_39_33;POLYGON(1,0,0,0.25)37_38;POLYGON(1,0,0,0.25)20;POLYGON(1,0,0,0.25)19;POLYGON(1,0,0,0.25)21;POLYGON(1,0,0,0.25)13_19_20_14;POLYGON(1,0,0,0.25)20_26_27_21;TICK;R_6_10_4_12_30_34_28_36_18_22_24_16;RX_5_11_0_17_29_35_23_40_19_20_21;MARKX(0)19;MARKZ(1)18_16_24_22;TICK;CX_5_1_2_6_7_10_11_8_9_12_17_13_29_25_26_30_31_34_35_32_33_36_40_37_14_18_19_22_23_20_21_24;TICK;CX_8_6_13_10_5_2_11_9_15_12_17_14_32_30_37_34_29_26_35_33_39_36_40_38_20_18_23_21_27_24_25_22;TICK;CX_3_6_8_10_1_4_5_7_0_2_11_14_27_30_32_34_25_28_29_31_23_26_35_38_15_18_20_22_17_19_13_16;TICK;CX_5_8_9_6_14_10_11_15_0_3_7_4_29_32_33_30_38_34_35_39_23_27_31_28_17_20_26_22_19_16_21_18;TICK;M_6_10_4_12_30_34_28_36_24_16_22_18;MX_5_11_0_17_29_35_23_40;MARKZ(1)24_16_22_18;DT(1,2,1)rec[-20]_rec[-36];DT(2,1,1)rec[-19]_rec[-35];DT(1,0,1)rec[-18]_rec[-34];DT(2,3,1)rec[-17]_rec[-33];DT(5,2,1)rec[-16]_rec[-32];DT(6,1,1)rec[-15]_rec[-31];DT(5,0,1)rec[-14]_rec[-30];DT(6,3,1)rec[-13]_rec[-29];DT(1,1,1)rec[-8]_rec[-28];DT(2,2,1)rec[-7]_rec[-27];DT(0,2,1)rec[-6]_rec[-26];DT(3,1,1)rec[-5]_rec[-25];DT(5,1,1)rec[-4]_rec[-24];DT(6,2,1)rec[-3]_rec[-23];DT(4,2,1)rec[-2]_rec[-22];DT(7,1,1)rec[-1]_rec[-21];OI(0)rec[-9]_rec[-10]_rec[-11]_rec[-12]_rec[-17]_rec[-18]_rec[-19]_rec[-20];TICK;POLYGON(0,0,1,0.25)2_8_9_3;POLYGON(0,0,1,0.25)7_1;POLYGON(0,0,1,0.25)15_9;POLYGON(0,0,1,0.25)31_37_38_32;POLYGON(0,0,1,0.25)31_25;POLYGON(0,0,1,0.25)39_33;POLYGON(0,0,1,0.25)7_13_14_8;POLYGON(0,0,1,0.25)26_32_33_27;POLYGON(0,0,1,0.25)19_25_26_20;POLYGON(0,0,1,0.25)19_13;POLYGON(0,0,1,0.25)27_21;POLYGON(0,0,1,0.25)14_20_21_15;POLYGON(1,0,0,0.25)1_7_8_2;POLYGON(1,0,0,0.25)8_14_15_9;POLYGON(1,0,0,0.25)2_3;POLYGON(1,0,0,0.25)25_31_32_26;POLYGON(1,0,0,0.25)32_38_39_33;POLYGON(1,0,0,0.25)37_38;POLYGON(1,0,0,0.25)13_19_20_14;POLYGON(1,0,0,0.25)20_26_27_21;TICK;R_6_10_4_12_30_34_28_36_18_22_24_16;RX_5_11_0_17_29_35_23_40;TICK;CX_5_1_2_6_7_10_11_8_9_12_17_13_29_25_26_30_31_34_35_32_33_36_40_37_14_18_19_22_23_20_21_24;TICK;CX_8_6_13_10_5_2_11_9_15_12_17_14_32_30_37_34_29_26_35_33_39_36_40_38_20_18_23_21_27_24_25_22;TICK;CX_3_6_8_10_1_4_5_7_0_2_11_14_27_30_32_34_25_28_29_31_23_26_35_38_15_18_20_22_17_19_13_16;TICK;CX_5_8_9_6_14_10_11_15_0_3_7_4_29_32_33_30_38_34_35_39_23_27_31_28_17_20_26_22_19_16_21_18;TICK;M_6_10_4_12_30_34_28_36_24_16_22_18;MX_5_11_0_17_29_35_23_40;DT(1,2,2)rec[-20]_rec[-40];DT(2,1,2)rec[-19]_rec[-39];DT(1,0,2)rec[-18]_rec[-38];DT(2,3,2)rec[-17]_rec[-37];DT(5,2,2)rec[-16]_rec[-36];DT(6,1,2)rec[-15]_rec[-35];DT(5,0,2)rec[-14]_rec[-34];DT(6,3,2)rec[-13]_rec[-33];DT(4,3,2)rec[-12]_rec[-32];DT(3,0,2)rec[-11]_rec[-31];DT(4,1,2)rec[-10]_rec[-30];DT(3,2,2)rec[-9]_rec[-29];DT(1,1,2)rec[-8]_rec[-28];DT(2,2,2)rec[-7]_rec[-27];DT(0,2,2)rec[-6]_rec[-26];DT(3,1,2)rec[-5]_rec[-25];DT(5,1,2)rec[-4]_rec[-24];DT(6,2,2)rec[-3]_rec[-23];DT(4,2,2)rec[-2]_rec[-22];DT(7,1,2)rec[-1]_rec[-21];TICK;R_6_10_4_12_30_34_28_36_18_22_24_16;RX_5_11_0_17_29_35_23_40;TICK;CX_5_1_2_6_7_10_11_8_9_12_17_13_29_25_26_30_31_34_35_32_33_36_40_37_14_18_19_22_23_20_21_24;TICK;CX_8_6_13_10_5_2_11_9_15_12_17_14_32_30_37_34_29_26_35_33_39_36_40_38_20_18_23_21_27_24_25_22;TICK;CX_3_6_8_10_1_4_5_7_0_2_11_14_27_30_32_34_25_28_29_31_23_26_35_38_15_18_20_22_17_19_13_16;TICK;CX_5_8_9_6_14_10_11_15_0_3_7_4_29_32_33_30_38_34_35_39_23_27_31_28_17_20_26_22_19_16_21_18;TICK;M_6_10_4_12_30_34_28_36_24_16_22_18;MX_5_11_0_17_29_35_23_40_19_20_21;MARKX(0)19;DT(1,2,3)rec[-23]_rec[-43];DT(2,1,3)rec[-22]_rec[-42];DT(1,0,3)rec[-21]_rec[-41];DT(2,3,3)rec[-20]_rec[-40];DT(5,2,3)rec[-19]_rec[-39];DT(6,1,3)rec[-18]_rec[-38];DT(5,0,3)rec[-17]_rec[-37];DT(6,3,3)rec[-16]_rec[-36];DT(4,3,3)rec[-15]_rec[-35];DT(3,0,3)rec[-14]_rec[-34];DT(4,1,3)rec[-13]_rec[-33];DT(3,2,3)rec[-12]_rec[-32];DT(1,1,3)rec[-11]_rec[-31];DT(2,2,3)rec[-10]_rec[-30];DT(0,2,3)rec[-9]_rec[-29];DT(3,1,3)rec[-8]_rec[-28];DT(5,1,3)rec[-7]_rec[-27];DT(6,2,3)rec[-6]_rec[-26];DT(4,2,3)rec[-5]_rec[-25];DT(7,1,3)rec[-4]_rec[-24];TICK;POLYGON(0,0,1,0.25)7_13_14_8;POLYGON(0,0,1,0.25)2_8_9_3;POLYGON(0,0,1,0.25)7_1;POLYGON(0,0,1,0.25)15_9;POLYGON(0,0,1,0.25)31_37_38_32;POLYGON(0,0,1,0.25)26_32_33_27;POLYGON(0,0,1,0.25)31_25;POLYGON(0,0,1,0.25)39_33;POLYGON(1,0,0,0.25)1_7_8_2;POLYGON(1,0,0,0.25)8_14_15_9;POLYGON(1,0,0,0.25)2_3;POLYGON(1,0,0,0.25)13_14;POLYGON(1,0,0,0.25)25_31_32_26;POLYGON(1,0,0,0.25)32_38_39_33;POLYGON(1,0,0,0.25)26_27;POLYGON(1,0,0,0.25)37_38;TICK;R_6_10_4_12_30_34_28_36;RX_5_11_0_17_29_35_23_40;TICK;CX_5_1_2_6_7_10_11_8_9_12_17_13_29_25_26_30_31_34_35_32_33_36_40_37;TICK;CX_8_6_13_10_5_2_11_9_15_12_17_14_32_30_37_34_29_26_35_33_39_36_40_38;TICK;CX_3_6_8_10_1_4_5_7_0_2_11_14_27_30_32_34_25_28_29_31_23_26_35_38;TICK;CX_5_8_9_6_14_10_11_15_0_3_7_4_29_32_33_30_38_34_35_39_23_27_31_28;TICK;M_6_10_4_12_30_34_28_36;MX_5_11_0_17_29_35_23_40;MARKX(0)1_7_13_25_31_37;DT(1,2,4)rec[-16]_rec[-39];DT(2,1,4)rec[-15]_rec[-38];DT(1,0,4)rec[-14]_rec[-37];DT(2,3,4)rec[-13]_rec[-36];DT(5,2,4)rec[-12]_rec[-35];DT(6,1,4)rec[-11]_rec[-34];DT(5,0,4)rec[-10]_rec[-33];DT(6,3,4)rec[-9]_rec[-32];DT(1,1,4)rec[-8]_rec[-27];DT(2,2,4)rec[-7]_rec[-26];DT(0,2,4)rec[-6]_rec[-25];DT(3.5,1.5,4)rec[-5]_rec[-18]_rec[-19]_rec[-24];DT(5,1,4)rec[-4]_rec[-23];DT(6,2,4)rec[-3]_rec[-22];DT(3.5,2.5,4)rec[-2]_rec[-17]_rec[-18]_rec[-21];DT(7,1,4)rec[-1]_rec[-20];TICK;M_1_2_3_7_8_9_13_14_15_25_26_27_31_32_33_37_38_39;DT(1,0,5)rec[-15]_rec[-18]_rec[-32];DT(1,2,5)rec[-13]_rec[-14]_rec[-16]_rec[-17]_rec[-34];DT(2,1,5)rec[-11]_rec[-12]_rec[-14]_rec[-15]_rec[-33];DT(2,3,5)rec[-10]_rec[-13]_rec[-31];DT(5,0,5)rec[-6]_rec[-9]_rec[-28];DT(5,2,5)rec[-4]_rec[-5]_rec[-7]_rec[-8]_rec[-30];DT(6,1,5)rec[-2]_rec[-3]_rec[-5]_rec[-6]_rec[-29];DT(6,3,5)rec[-1]_rec[-4]_rec[-27];OI(1)rec[-16]_rec[-17]_rec[-18];OI(2)rec[-7]_rec[-8]_rec[-9]).  

Anyways if you've (1) checked that the streamed output stim file is the same as the unstreamed output stim file and (2) kept the same polygon adding, I think all should be fine. Even if you didn't, a polygon is extra-visualization on top of Crumble that a human debugger that is relying on viz probably does not strictly need. There def should be a way to test and automate the polygon annotation with correlation surface and distance data about the circuit but it seem appropriate to make a separate issue about it and move on.

HaoTy:
It seems there is no formal format specification for polygon in Stim/Crumble. @BluSteve I agree with Kabir that if the polygon-decorated circuits produced by the streaming version display correctly on Crumble, we're good. I don't think additional tests are needed.

BluSteve:
@HaoTy I have made all the changes discussed. All the non-streaming versions of get_circuit now point to the streaming version. I have also verified that the polygons work as expected.

I think this PR is ready for final review.

HaoTy:
> @HaoTy I have made all the changes discussed. All the non-streaming versions of get_circuit now point to the streaming version. I have also verified that the polygons work as expected.
> 
> I think this PR is ready for final review.

Awesome! Will try to do it over the weekend, but no promise... Meanwhile, can you take a look at the CI errors?

BluSteve:
I've fixed the CI errors!

BluSteve:
@HaoTy Oh forgot to reformat. Try again please

src/tqec/compile/tree/tree.py
```diff
@@ -53,6 +53,60 @@ def seen_qubits(self) -> set[GridQubit]:
...
+    """
+
+    @override
+    def visit_node(self, node: LayerNode) -> None:
+        if not isinstance(node._layer, LayoutLayer):
+            return
+        self._seen_qubits |= node._layer.qubits(self._k)
+
+
+def _generate_detector_database(
```
HaoTy: This is not used anywhere. I remember it to be from the previous de-duplication effort, but now itself is the duplication.

src/tqec/compile/tree/node.py
```diff
@@ -42,6 +53,132 @@ def exit_node(self, node: LayerNode) -> None:
         pass
 
 
+class AnnotateDetectorsOnLayerNode(NodeWalker):
```
HaoTy: Is circular import the reason that this class was moved here? We should consider moving it back, as `annotators/detectors.py` is clearly the dedicated place for detector annotation.

HaoTy (changes requested):
`_generate_annotations` (along with the four `_annotate_xxx` methods it calls) now is only used in `generate_crumble_url`. Because the streaming version essentially uses pretty much the same set of code broken into pieces, we should avoid maintaining the same things in two places. Do you think you can handle it in this PR or need another one?

One problem arises from breaking the code into pieces is that some inconspicuous but important logic may be lost. For example, `_annotate_detectors` is responsible for saving/updating the detector database. The current detector annotating path seems to never save/update the database.

I will need more time to look more closely. Meanwhile, if anyone else wants to provide feedback on this PR, please feel free to do so.

src/tqec/compile/tree/node.py
BluSteve: Yes it was circular import. I've now moved it back.

BluSteve:
@HaoTy I believe I've fixed the detector database issue. I would prefer to keep this PR contained at its current complexity, with a followup PR specifically to reduce code duplication issues. As it's already a plug-and-play replacement for all existing usages.

src/tqec/compile/tree/node.py
```diff
@@ -42,6 +52,34 @@ def exit_node(self, node: LayerNode) -> None:
...
 
 
+def _get_ordered_leaves(root: LayerNode) -> list[LayerNode]:
+    """Return the leaves of the tree in time order."""
+    if root.is_leaf:
+        return [root]
+    return [n for child in root.children for n in _get_ordered_leaves(child)]
+
+
+def _annotate_observable_at_node(
```
HaoTy: Can we move these back too? Basically, everything in their original place.

src/tqec/compile/tree/tree.py
```diff
@@ -294,67 +351,161 @@ def generate_circuit(
...
+            lookback: number of QEC rounds to consider to try to find detectors.
+                Including more rounds increases computation time.
+            reschedule_measurements: whether to reschedule measurements in a ``LayoutLayer``
+                to be in the same moment. Since each plaquette may have its own measurement
+                schedule, setting this may be necessary for hardware that requires
+                measurements to be synchronous.
+
+        Returns:
+            a ``stim.Circuit`` instance implementing the computation described
+            by ``self``.
```
HaoTy: Stale docstring

src/tqec/compile/tree/tree.py
```diff
@@ -294,67 +351,161 @@ def generate_circuit(
...
+                    detectors_walker,
+                    subtree_to_z,
+                    self._abstract_observables,
+                    self._observable_builder,
+                )
+            finally:
+                # The database will have been updated inside the above function
+                # with AnnotateDetectorsOnLayerNode, and here at the end of the
+                # computation we save it to file.
+                if detector_database is not None and database_path is not None:
```
HaoTy, line 471: If `detector_database` is `None` and there's no file at the specified path, the database will be instantiated in `AnnotateDetectorsOnLayerNode` and still not saved.

src/tqec/compile/tree/tree.py
```diff
@@ -53,6 +54,60 @@ def seen_qubits(self) -> set[GridQubit]:
...
+    """
+
+    @override
+    def visit_node(self, node: LayerNode) -> None:
+        if not isinstance(node._layer, LayoutLayer):
+            return
+        self._seen_qubits |= node._layer.qubits(self._k)
+
+
+def _generate_detector_database(
```
HaoTy: Please remove if not used.

src/tqec/compile/tree/node.py
```diff
@@ -162,65 +200,213 @@ def generate_circuits_with_potential_polygons(
...
+                            for fn, component in fns:
+                                fn(self, component=component)
+
+                base_circuit = annotations.circuit
+                if base_circuit is None:
+                    raise TQECError(
+                        "Cannot generate the final quantum circuit before annotating "
+                        "nodes with their individual circuits. Did you call "
+                        "LayerTree.annotate_circuits before?"
+                    )
```
HaoTy, line 262: Stale error message.

src/tqec/compile/tree/node.py
```diff
@@ -236,11 +422,57 @@ def generate_circuit(self, k: int, global_qubit_map: QubitMap) -> stim.Circuit:
...
+            subtree_to_z: mapping from direct children of root to their z-coordinate values,
+                used for determining which abstract observables to apply at each node.
+            abstract_observables: collection of abstract observable definitions to be
+                annotated in the circuit.
+            observable_builder: builder instance used to construct observable annotations
+                from abstract observable definitions.
+
+        Returns:
+            a ``stim.Circuit`` instance representing ``self`` with the provided
+            ``global_qubit_map``.
```
HaoTy: Stale docstring.

src/tqec/compile/tree/node.py
```diff
@@ -236,11 +422,57 @@ def generate_circuit(self, k: int, global_qubit_map: QubitMap) -> stim.Circuit:
...
+
+    def _generate_circuit_stream(
+        self,
+        k: int,
+        global_qubit_map: QubitMap,
+        reschedule_measurements: bool = False,
+        detectors_walker: AnnotateDetectorsOnLayerNode | None = None,
+        subtree_to_z: dict[LayerNode, int] | None = None,
+        abstract_observables: list[AbstractObservable] | None = None,
+        observable_builder: ObservableBuilder | None = None,
```
HaoTy: These arguments are passed through all recursive calls and the last three are always all-provided or all-None. I feel a single object is more suitable, something like

```python
@dataclass
class AnnotationContext:
    detectors_walker: AnnotateDetectorsOnLayerNode | None
    subtree_to_z: dict[LayerNode, int]
    abstract_observables: list[AbstractObservable]
    observable_builder: ObservableBuilder
```

Then the `pre_annotated` in `_generate_circuits_with_potential_polygons_stream` can use the existence of this object as the judgement, instead of doing three is-None.

src/tqec/compile/tree/tree.py
HaoTy: I meant `_generate_detector_database`.

BluSteve:
@HaoTy No rush but just wanted to confirm that I'm not bottlenecking anything anymore.

HaoTy (changes requested):
Sorry for the delay. I was worried that an early termination of the iterator would lead to unexpected behaviors on the second call. But I think the option to partially consume the stream is just a byproduct, not the major objective of this PR, so we don't need to worry about this practically not really meaningful use case now. Considering this PR is already substantial and has been around for quite a while, I lean toward merging first and fixing later if there turn out to be issues.

src/tqec/compile/tree/tree.py
BluSteve:
```    
def _annotate_detectors(
        self,
        k: int,
        manhattan_radius: int = 2,
        database_path: Path | None = DEFAULT_DETECTOR_DATABASE_PATH,
        lookback: int = 2,
        parallel_process_count: int = 1,
    ) -> None:
        if manhattan_radius <= 0:
            return  # pragma: no cover
        self._root.walk(
            AnnotateDetectorsOnLayerNode(
                k,
                manhattan_radius,
                detector_database,
                lookback,
                parallel_process_count,
            )
        )
        # The database will have been updated inside the above function, and here at
        # the end of the computation we save it to file.
        if detector_database is not None and database_path is not None:
```

I believe this follows the behavior of _annotate_detectors from before.

src/tqec/compile/tree/node.py
BluSteve: Agreed and implemented

src/tqec/compile/tree/tree.py
```diff
@@ -294,67 +351,161 @@ def generate_circuit(
...
+
+        """
+        # If already annotated, no need to re-annotate.
+        if k in self._annotations:
+            annotations = self._get_annotation(k)
+            assert annotations.qubit_map is not None
+            yield from self._root._generate_circuit_stream(
+                k,
+                annotations.qubit_map,
+            )
```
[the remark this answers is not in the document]
BluSteve: valid

src/tqec/compile/tree/tree.py
```diff
@@ -294,67 +351,161 @@ def generate_circuit(
...
+                        )
+                    else:  # ie using the default
+                        warnings.warn(
+                            f"The default detector database that you have saved on your system is "
+                            f"out of date (version {loaded_version}). The version in the TQEC code "
+                            f"you are running is newer (version {current_version}). The database "
+                            "will be regenerated.",
+                            TQECWarning,
+                        )
+
```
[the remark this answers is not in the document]
BluSteve: @HaoTy I think this one is valid even though this was the same as in `generate_circuit()`.

src/tqec/compile/tree/node.py
```diff
@@ -162,65 +200,213 @@ def generate_circuits_with_potential_polygons(
...
-                    "SHIFT_COORDS", [], StimCoordinates(0, 0, 1).to_stim_coordinates()
+
+                if add_polygons:
+                    yield annotations.polygons
+
+                yield mapped_circuit.get_circuit(include_qubit_coords=False)
+
+            elif isinstance(self._layer, SequencedLayers):
+                leaf_dict: dict[LayerNode, list[tuple[Callable, ObservableComponent]]] = {}
+
```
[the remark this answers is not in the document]
BluSteve: leaf_dict is always None at this point

src/tqec/compile/tree/node.py
```diff
@@ -162,65 +200,213 @@ def generate_circuits_with_potential_polygons(
...
-            ret = []
-            for child, next_child in itertools.pairwise(self._children):
-                ret += child.generate_circuits_with_potential_polygons(
-                    k, global_qubit_map, add_polygons
+
+            elif isinstance(self._layer, RepeatedLayer):
+                body = list(
+                    self._children[0]._generate_circuits_with_potential_polygons_stream(
+                        k,
+                        global_qubit_map,
```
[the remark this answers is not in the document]
BluSteve: Impossible to stream further as we need to keep this moment-aligned for adding noise model.

src/tqec/compile/tree/tree.py
```diff
@@ -53,6 +54,60 @@ def seen_qubits(self) -> set[GridQubit]:
...
+
+def _generate_detector_database(
+    database_path: Path, detector_database: DetectorDatabase | None
+) -> DetectorDatabase:
+    # We need to know for later if the user explicitly provided a database or
+    # not to decide if we should warn or raise.
+    user_defined = detector_database is not None or database_path != DEFAULT_DETECTOR_DATABASE_PATH
+    # If the user has passed a database in, use that, otherwise:
+    if detector_database is None:  # Nothing passed in,
```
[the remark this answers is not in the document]
BluSteve: done

src/tqec/compile/tree/tree.py
```diff
@@ -294,67 +351,161 @@ def generate_circuit(
...
-                        f"of date (version {loaded_version}). The version in the TQEC code you are "
-                        f"running is newer (version {current_version}). The database will be "
-                        "regenerated.",
+                        f"An exception occurred when loading {database_path}: {e}\n"
+                        f"Database not opened.",
                         TQECWarning,
                     )
+                    detector_database = None
+
+            if detector_database is not None:
```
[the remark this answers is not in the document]
BluSteve: @HaoTy I believe generate_circuit() from the current main branch exhibits the same behavior

BluSteve:
@HaoTy I believe I've fixed everything except the detector database problems which I don't fully understand. AI also identified some issues with it but I'm not sure what the original intended behavior is in the main branch.

src/tqec/compile/tree/tree.py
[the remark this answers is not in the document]
HaoTy: I think Copilot missed the full picture. When the `detector_database` passed to `AnnotateDetectorsOnLayerNode` is `None`, an empty one will be automatically initialized:
https://github.com/tqec/tqec/blob/da44a246e7cde00c1a4c7ddd244cc02f6466d031/src/tqec/compile/tree/annotators/detectors.py#L205

src/tqec/compile/tree/tree.py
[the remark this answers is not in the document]
HaoTy: Confirmed it's a pre-existing bug we introduced in a previous refactor. The diagnosed issue here is not the actual problem. Let's fix it in a separate PR then.

src/tqec/compile/tree/tree.py
HaoTy: Same as above; fix in a separate PR.

HaoTy (approved):
Let's get it merged!

BluSteve:
Thank you for reviewing! :D

PR #815, opened 2026-01-15
Update sketchup template

### LINKED ISSUE #814: Add patch rotation to template asset
We need patch rotations according to #571. 

[Slide deck about patch rotations](https://docs.google.com/presentation/d/1WfJbPpF27uh5KaIkR341Bo4M4ZfeoB7SS0GBe9POkVY/edit?slide=id.p#slide=id.p)

Here's the directory it needs to be added to:

https://github.com/tqec/tqec/tree/main/assets

Here's what the new template looks like:

Image

[Here's the the actual .skp file](https://drive.google.com/file/d/1Qb6ulVrmrahnBb6p2bLRCt8vDXidRzNF/view?usp=sharing)

**Two things I want to confirm before making the PR**
1. Note that the magic state cultivation block will change from this:

Image

to this:

Image

Which one do we want? I'd probably vote for the one with the black base because it's more descriptive, and I don't foresee the added .dae file complexity being a slowdown.  

2. This template rotation is a ZXZ to XZX map. Is it correct to think that this is sufficient because we can read rotated blocks in SketchUp?

Image

KabirDubey:
Closes #814

~~I updated the dae file parser to read patch rotation and cultivation blocks as distinct block types (neither cubes nor pipes) in the `BlockKind` union. It felt wrong to type them as cubes because their position geometry is unique. 
I put `NotImplementedError`s when they're recognized.~~

I attempted to update the parser but reverted because I think that will involve more care than I originally thought. This PR is just updating the template files. The _v8 files correspond to the old version of SketchUp.


jbolns (dismissed):
The `template_v8.skp` opens in Old Sketchup. All blocks seem to be there. 

I need someone else to confirm the new SketchUp version works to be able to approve. @angelaelisa ? 

Additionally, I don't see a need for either of the ***`.dae`*** templates. In fact, I suggest we delete the current `assets/template.dae` altogether. @inmzhang ?
- We currently have  `assets/logical_cnot.dae` to test simple DAE import.
- We should, when everything is fully supported, add a separate `assets/some_circuit_with_all_cubes.dae`.
- BUT it makes no sense to import a DAE of the actual template into TQEC.

Ps. I don't think updating the parser should be part of this issue. We should review I/O generally before. It's becoming a little too add-hoc.

jbolns:
Additionally, but I need to make this as a question for @afowler and @inmzhang , I think we need two magic state cultivation blocks? The naming conventions we use for SketchUp are such that each potential position/rotation has its own cube. If the magic state cultivation block can be rotated along z-axis (meaning red faces can end up where blue faces are currently), to keep in line with the convention, we need two, one facing in one direction and the other in the other.

angelaelisa:
@jbolns it works perfect in SketchUp Pro 2022 :) Anything else you need, please tell me 

Screenshot 2026-01-16 at 10 30 46

jbolns:
@angelaelisa do both of the `.skp` work in 2022?
The screenshot is of `_v8`. It's good that it opens but the one meant for 2022 is the other `.skp`.

angelaelisa:
@jbolns sorry :) here you go the `template.skp`:

**SketchUp Por 2022**
Screenshot 2026-01-16 at 10 55 00

**SketchUp 2026** 
Screenshot 2026-01-16 at 10 54 04

angelaelisa:
> Additionally, but I need to make this as a question for @afowler and @inmzhang , I think we need two magic state cultivation blocks? The naming conventions we use for SketchUp are such that each potential position/rotation has its own cube. If the magic state cultivation block can be rotated along z-axis (meaning red faces can end up where blue faces are currently), to keep in line with the convention, we need two, one facing in one direction and the other in the other.

From my point of view, I agree with @jbolns. I think it would be better to have two different components because it would be easier to build `blockgraphs` (at least at the frontend) without having to rotate, just using one or another component depending on the kind of pipe it's connected to. Just following the way we do with `pipes`, `cubes`, and `Hadamard pipes`

afowler:
We could add a second rotated cultivation structure, but I don't think we
want to add every possible rotation of the patch rotation structure. The
philosophy behind the single cultivation structure was that since it is so
big and used rarely, it would be ok for the user to need to rotate it.
After all, the user could rotate it once and have a copy.

jbolns:
I suppose I have only one question. Will the user of SketchUp know the cultivation block *can* be rotated? If so, then I will revise my position based on the above. 

Because, I meant it only to facilitate usage in SketchUp. There's no difference for TQEC. For TQEC import, rotating the existing cultivation block suffices: it will trigger rotation of the original kind. As long as we have internal definitions for all possible kinds arising from rotating the original kind, a single structure suffices.

@KabirDubey On second inspection, the _v8 is missing the port cube. The port cube is *not* needed for SketchUp -> TQEC, but it is needed to move things TQEC -> SketchUp.

afowler:
I'm ok with a second cultivation since it should only be rotated in one way
and we can put Y and cultivation in a section of "do not rotate" blocks.
Everything else can be rotated arbitrarily, including patch rotation.

angelaelisa:
I think that having a second cultivation will be more intuitive for SketchUp users and makes easier to implement frontend solutions. Having a second component doesn't create problems and improves the user experience in many ways

inmzhang:
Actually I think we can simply represent the cultivation block as a purple cube without specifying blue/red boundaries, and determine the cultivated code boundaries with the connecting pipe instead.

afowler:
I think we will have to have two explicit orientations because a
potentially common use case could be a T factory and it will matter what
orientation is used when it comes to dynamically routing successfully
prepared states to where they will be used.

The t state will determine the pipe coloring, not the other way around in
this case.

inmzhang:
> I think we will have to have two explicit orientations because a
potentially common use case could be a T factory and it will matter what
orientation is used when it comes to dynamically routing successfully
prepared states to where they will be used.

That makes sense for the dynamical routing in the long term. What I have in mind is still static position for cultivation and consumption.

afowler:
Agree that until we tackle dynamic routing a single t block is sufficient.

KabirDubey:
Apologies for the delay in addressing reviewer feedback. I'm reviving this PR @jbolns @angelaelisa @afowler @inmzhang. 

My latest commit updates `template.skp` and `template.dae` with a rotated magic state cultivation block. Here's what the template looks like:

SCR-20260624-ldtx-2

#### why a rotated MSC block?

> Actually I think we can simply represent the cultivation block as a purple cube without specifying blue/red boundaries, and determine the cultivated code boundaries with the connecting pipe instead.

>> I think we will have to have two explicit orientations because a potentially common use case could be a T factory and it will matter what orientation is used when it comes to dynamically routing successfully prepared states to where they will be used.
The t state will determine the pipe coloring, not the other way around in this case.

>>> That makes sense for the dynamical routing in the long term. What I have in mind is still static position for cultivation and consumption.

The benefit to including the rotated MSC block seems to be that now a control system simulator can distinguish between T states that can be connected to blue nodes versus red nodes. The "dynamic routing" scenario I am imagining here is that the control system observes that the most-likely-to-succeed T state in a factory is escaping into a patch with one boundary orientation and therefore allocates nodes which more efficiently (in STV) support the injection of that T state orientation. Or conversely, the control system calculates that red or blue is less expensive to supply than the other and consequently schedules a larger fraction of T states to escape in that boundary orientation rather than the other. 

^ That's my best guess. I do not completely understand, probably because I do not know exactly what dynamical routing refers to. If you're worried about the teleportation basis couldn't you always schedule a temporal Hadamard for pretty-much-zero cost?

Either way, there's no downside to including the rotated MSC block, so the above does not affect this PR.

~#### does someone have sketchup 8?~ EDIT: might have found a workaround, see below

~Before I can lock this PR in for re-review, I need someone with SketchUp 8 to provide me with a SketchUp version of the updated `template.skp` and `template.dae` files (or, if you want, commit it to this branch yourself). If you can't access the file yourself from my branch, email me and I'll send you a drive link or something.~

#### do we even need `.dae` files?

> Additionally, I don't see a need for either of the .dae templates. In fact, I suggest we delete the current assets/template.dae altogether. @inmzhang ?
> * We currently have assets/logical_cnot.dae to test simple DAE import.
> * We should, when everything is fully supported, add a separate assets some_circuit_with_all_cubes.dae.

My understanding is that the purpose of hosting this `assets` directory is to support hand-compilation and visualization. The exact user experience for the former is as follows: (1) user imports the template into their favorite GUI; (2) user manually assembles and optimizes their desired circuit; (3) user exports the block graph as a `.dae` file to `tqec` and records its performance. 

Is the alternative that the `assets` directory only contains `.skp` files? I think that might be worse. All GUIs should be able to read `.dae` files--it's an open-standard file format and efficient XML parsing has been worked on for [at least 17 years](https://stackoverflow.com/questions/606304/the-most-efficient-way-to-parse-xml)--but not all GUIs can read `.skp` files. Quality `.skp` to `.dae` conversion is lowkey paywalled by SketchUp.    

> * BUT it makes no sense to import a DAE of the actual template into TQEC.

One could run [this batch parsing](https://github.com/KabirDubey/tqec/blob/dae-component-reproduction/_reproduction/inputs/parser.py) script over the template file and benchmark their `tqec` system configuration over all computational primitives in a single workflow. Automatic port filling should cover all deterministic fillings of the pipes. I think the port cube would need to get programmatically skipped. 

Thanks for reviewing!!

jbolns:
I can send you the v8 version within this week, @KabirDubey .

But since I don't have Pro so I can't open the other to check, how do I call the blocks, i.e., what's their entity name? 
- In Topologiq, I call the gray one "ooo" and that's how they're showing in BGRAPH.
- I think we had agreed with @HaoTy to call the Y cube "yi" and "ym" depending on whether it's initialisation or measurement, or just "y" if it's a multipurpose. 
- I can't recall if we discussed cultivation. 
- We most definitely did not discuss the patch rotation
 
To stay in theme, my preference is on descriptive characters rather than words.

@nelimee

jbolns:
Ps1. The port cube was added for the import plugin to enable roundtrips, previously needed in v8 and pro and now only in v8. I can't really recall why. I think we thought that if the blockgraph has explicit ports on either end, the data should maintain them explicit across.

KabirDubey:
> I can send you the v8 version within this week, @KabirDubey .

See the `_SU8` files in most recent commit. Used `SuForYou` to downversion the `.skp` file and [this](https://3dencoder.com/SKP-to-dae) to convert to `.dae`--may have been lossy.

> 
> But since I don't have Pro so I can't open the other to check, how do I call the blocks, i.e., what's their entity name?

I go to Entity Info > Definition and see Cultivation is "cultivation" and the patch rotation is "patch rotation". The gray block is "port#1". Let me know you are unable to call the blocks by those addresses. 

> To stay in theme, my preference is on descriptive characters rather than words.

I suggest the following way to wrap the components (referring to the picture from my most recent screenshot):

the leftmost cultivation block with `zxt`, 
the rightmost cultivation block with `xzt`,
the patch rotation block with `pr`, and
the gray block with `ooo` or `port` (no harm in doing both).

>     * I think we had agreed with @HaoTy to call the Y cube "yi" and "ym" depending on whether it's initialisation or measurement, or just "y" if it's a multipurpose.

I saw this, and I think it's fine, but `y` should be sufficient. EDIT: Oh wait I guess it's ~worth distinguishing~ necessary to distinguish if you want to do both y basis memory and twist braiding experiments?: 

SCR-20260624-ozjj-2

HaoTy:
> the leftmost cultivation block with `zxt`, the rightmost cultivation block with `xzt`, the patch rotation block with `pr`, and the gray block with `ooo` or `port` (no harm in doing both).

I suggest we also consider supporting cultivating the Tx (= HTH) state.

> > ```
> > * I think we had agreed with @HaoTy to call the Y cube "yi" and "ym" depending on whether it's initialisation or measurement, or just "y" if it's a multipurpose.
> > ```
> 
> I saw this, and I think it's fine, but `y` should be sufficient. EDIT: Oh wait I guess it's ~worth distinguishing~ necessary to distinguish if you want to do both y basis memory and twist braiding experiments?:

To quote Austin, y half cubes have intrinsic time ordering, so yi and ym have real meaning. yi starts with initialization gates. ym ends with measurement gates.
We should aim to always explicitly specify the direction and avoid using just "y", even though the direction often can be deduced.

jbolns:
Update. I sent the v8 version to @KabirDubey. Exact same as in his screenshot, with named entities.

I see not reason why not to include any other cubes that might become necessary. Saves having to do this process twice. In particular, Topologiq will write give "Yi" or "Ym" to BGRAPH based on relative positioning, never simply "Y". So, I implicitly agree with @HaoTy . Having said that, SketchUp -> DAE doesn't have a way to change entity name depending on relative position of Y, so one would need two cubes, which also implies a need for a way to easily differentiate visually between the two (just changing the bottom/top faces will 100% led into layer 8 situations).

KabirDubey:
Thanks @jbolns and @HaoTy, latest commit should have the updated files with the correction orientations.

> I suggest we also consider supporting cultivating the Tx (= HTH) state.

Sorry I think I'm getting tripped up by states versus gates here. The state |Tx> = H|Tz>, so can't we express these states by connecting the MSC block to a X cube instead of a Z cube? Or connect a Hadamard pipe. Is there some special circuit that prepares |Tx> differently?

I can't think of a Tx gate block graph that you _can_ translate from the essential structure below but _can't_ build from the blocks in this PR: 

image

> To quote Austin, y half cubes have intrinsic time ordering, so yi and ym have real meaning. yi starts with initialization gates. ym ends with measurement gates.
We should aim to always explicitly specify the direction and avoid using just "y", even though the direction often can be deduced.
> I see not reason why not to include any other cubes that might become necessary. Saves having to do this process twice. In particular, Topologiq will write give "Yi" or "Ym" to BGRAPH based on relative positioning, never simply "Y". So, I implicitly agree with @HaoTy 
> Having said that, SketchUp -> DAE doesn't have a way to change entity name depending on relative position of Y, so one would need two cubes, which also implies a need for a way to easily differentiate visually between the two (just changing the bottom/top faces will 100% led into layer 8 situations).

Just to make sure I understand, is the ask here to update the SketchUp template in this PR with different blocks for yi and ym? The easiest way to do this off the top of my head would just be having two half cubes that appear the same visually but actually one has entity name yi and the other has entity name ym. Idk what a layer 8 situation is.

I'm not convinced updating the SketchUp template is productive though. I see the importance of distinguishing between the blocks, but am leaning towards making the distinction entirely with translation functions. Specifically, I am in okay with the `.bgraph` file type accepting yi's and ym's, and the `BlockGraph` class having a specialized yi and ym subclasses (or something like that) for `compile.py`, `LayerTree`, etc. to generate the exact circuit they correspond to. I think this is pretty much the status-quo.

jbolns:
> Idk what a layer 8 situation is

Haha. Yes. Fair enough. User error. If the blocks looks the same or are sufficiently similar as for cognitive dissonance to arise, we will all end up using the "Yi" when we need a "Ym" and vice-versa. 

> I'm not convinced updating the SketchUp template is productive though.

If one can avoid confusion, I see no downside in having both. As SketchUp is also a learning device, this can help communicate the existence of both cubes. 

Now, for fairness and in full honesty, in Topologiq (in branch, still, not yet available in main), I do as @Kabir suggests later in the same paragraph. I have a "Y" cube used in the algorithm and then switch it to "Yi" or "Ym" when writing to BGRAPH by checking if the neighbour's position in Z-axis is above or below. This is done at the same time as checking the Y cube has only one neighbour, so it takes, I think, an additional line or two relative to the integrity checks alone.

afowler:
Regarding having two y half cube blocks, if you chose to go down this path
one would be at the bottom of the cube and the other would be at the top of
the cube, so visually they would be clearly different. I don't think this
is necessary, and would not be in favor of this, in my opinion, this should
be inferred from the position of the block.

KabirDubey:
Okay @jbolns @HaoTy my vote is for only one Y half cube block. If y'all are okay with that and with the `.dae` and `.skp` files in the latest commit, please approve. 

> If one can avoid confusion, I see no downside in having both. If one can avoid confusion, I see no downside in having both. As SketchUp is also a learning device, this can help communicate the existence of both cubes.

Although Austin's suggested new components make sense [^1], I think the honest downside is the effort in making this new component. There's also a case to be made that two half cubes slightly increases rather than decreases hand compiling hardness because you have to scan between two cubes. I think this outweighs the marginal benefit of clarifying things for a hypothetical user. If we're going to silently fix the circuit based on its position then there's no functional benefit. 

[^1]: my vision is one d^{3} cube with its edges colored black; ym fills the bottom d^{3}/2 half with green and yi the top d^{3}/2.

afowler:
To be clear, when it comes to SketchUp, I'm not suggesting new components
:-)

I like the current single half cube.

jbolns:
Oh yeah, sorry @KabirDubey . Forgot to acknowledge explicitly.
I read Austin's earlier message as 100% for having a single Y cube.
I am ready to approve whenever you think you have a final version. It's just I've seen a few commits still coming in, so don't want to accidentally hit approve before the final commit is in to avoid accidents.

jbolns (approved):
All changes have been applied or discussed and agreed upon. 
Good to go now.

jbolns:
Done, approved. I'll leave it to you to hit the button in case there's final adjustments needed for any reason whatsoever.

Issue #813, opened 2026-01-14
Reimplement `CorrelationSurface`


HaoTy:
Following https://github.com/tqec/tqec/pull/707#issuecomment-3695369924 and the discussion in the community meeting, we want to reimplement the current `CorrelationSurface` interface for efficiency and to leverage the internal correlation surface computation methods designed for the internal data structure:

https://github.com/tqec/tqec/blob/f1e0a9d182c4738f8e410a6b0df5f9b815cce041/src/tqec/interop/pyzx/correlation.py#L183 

@nelimee Anything I should pay particular attention to? I already consulted Yiming that nothing should rely on `CorrelationSurface` being immutable or hashable. Two nontrivial discrepancies I noticed between the current `CorrelationSurface` and the internal data structure are 

1. Due to how the new algorithm works, the latter will have "filler" Pauli I operators at empty parts of the ZX graph. This is also related to how the `Pauli` operator used in the latter is defined (inheriting the `Flag` enum).

2. In the latter, the Hadamard information is not automatically implied from the edge when it's I-I or Y-Y, and requires either the original ZX graph or an explicit attribute added to the class to track.

smburdick:
Go ahead and assign me @HaoTy

HaoTy:
> Go ahead and assign me @HaoTy 

Ah, I should've mentioned that there's already a draft implementation at the time of writing. My apologies! This issue is mainly for tracking what's being done and gathering feedback, following Adrien's suggestion in the meeting. Thanks for volunteering to help!

HaoTy:
Addressed by #822.

Issue #818, opened 2026-01-17
Allow node colours in BlockGraph to be conditioned on correlation surfaces


nelimee:
## Is your feature request related to a problem? Please describe.

We need a way to condition a node colour in the `BlockGraph` on the corrected parity of a correlation surface.

## Describe the solution you'd like

I think @HaoTy has something ongoing, I'll let him describe what he thinks would be a good solution.

## Describe alternatives you've considered

*None*

## Additional context

*None*

smburdick:
Related: #817

HaoTy:
What I had in mind was to handle conditional basis together with the more general conditional operations (1.2 and 1.3 in #791), based on the assumption that we want `BlockGraph` to remain completely static (otherwise many properties and functions need to be rediscussed and refactored, e.g., spacetime volume, visualization, etc.). But I do think that allowing `BlockGraph`s to be dynamic is more elegant and worth upgrading the affected functions.

The other notable design choice is what serves as the condition. I originally planned to let users (or upstream compilers) specify the logical measurements using cube positions in a given block graph. This is less natural than using a correlation surface but has the benefit of constructing the block graph in one step. On the contrary, correlation surfaces are computed after the block graph is fixed, and the specific correlation surface that is associated with the target logical measurement may need to be reformed from the generating set after previous conditional operations are fixed (executed). In my mind, the user should not need to fiddle with correlation surface objects.

afowler:
I would advocate for saying there is no such thing as a logical
measurement, there are only correlation surfaces. What people traditionally
think of as a logical measurement is really just one edge of a correlation
surface that extends backward in time. The entire parity of the correlation
surface is the measurement result.

HaoTy:
> I would advocate for saying there is no such thing as a logical
> measurement, there are only correlation surfaces. What people traditionally
> think of as a logical measurement is really just one edge of a correlation
> surface that extends backward in time. The entire parity of the correlation
> surface is the measurement result.

I get that. My point is that the relevant correlation surface is only computed after specifying the block graph, so if we want to have some data structure serving as conditions at the time a dynamic block graph is constructed, we better avoid using the entirety of a correlation surface but rather specify a part of it that's enough for recovering/computing it.

afowler:
That may be how correlation services are calculated now, but it cannot stay
that way. Correlation surfaces really need to be computed in PyZX so they
can be used to inform how to compile the Block graph. It is important that
parts of the graph with correlation surfaces that determine future
[...]

PR #822, opened 2026-01-21
Improve `CorrelationSurface`-related efficiency and usability

HaoTy:
Addresses #813. Reimplements the `CorrelationSurface` class using dict of dicts to interop with internal correlation surface computation methods and for better efficiency in certain scenarios, such as constant instead of linear time read given a vertex id.

### DIFF
--- a/src/tqec/computation/correlation.py
+++ b/src/tqec/computation/correlation.py
@@ -187,18 +187,7 @@ class CorrelationSurface:
             The Pauli operator supported on the given ports.
 
         """
-        # Avoid pulling pyzx when importing that module.
-        from pyzx.pauliweb import multiply_paulis  # noqa: PLC0415
-
-        paulis = []
-        for port in io_ports:
-            basis_set = {b.value for b in self.bases_at(port)}
-            result = "I"
-            for basis in basis_set:
-                result = multiply_paulis(result, basis)
-            paulis.append(result)
-
-        return "".join(paulis)
+        return "".join(str(Pauli.from_basis_set(self.bases_at(port))) for port in io_ports)
 
     def external_stabilizer_on_graph(self, graph: BlockGraph) -> str:
         """Get the external stabilizer of the correlation surface on the graph.
--- a/src/tqec/utils/enums.py
+++ b/src/tqec/utils/enums.py
@@ -85,6 +85,11 @@ class Pauli(Flag):
         bases = list(Basis)
         return {bases[basis.value >> 1] for basis in self.iter_xz() if basis in self}
 
+    @staticmethod
+    def from_basis_set(bases: set[Basis]) -> Pauli:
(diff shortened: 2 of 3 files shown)


HaoTy:
Considering how #818 will be carried out, the `CorrelationSurface` class might need to be immutable and hashable afterall.

HaoTy:
This PR improves the efficiency and usability of `CorrelationSurface`-related code, including
1. Improves the `Flag`-based `Pauli` enum, moves it to enums.py, and adds conversions between `Pauli` and `Basis`;
2. Changes `ZXNode` and `ZXEdge` to `NamedTuple` to reduce memory footprint, speedup attribute access, and benefit from C-level optimizations;
3. Adds cached properties to `CorrelationSurface` so that the previously linear operations become constant after the first call;
4. Avoid creating identical `ZXNode` instances when constructing `CorrelationSurface`;
5. Moves the correlation surface finding algorithm to the more suitable computation module, in a separate _correlation.py file to avoid pulling pyzx when not used;
6. Adds pyzx vertex type to `Pauli` and `Basis` conversions, and simplifies implementations that previously relied on a series of `is_zx_no_phase()`, etc.;
7. Adds a conversion from the public `CorrelationSurface` back to the internal graph representation;
8. Adds more correlation, pyzx utils, and enums tests.

@inmzhang @Zhaoyilunnn Want to take a review?

inmzhang:
I'll do the review before this weekend.

HaoTy:
> I'll do the review before this weekend.

Thanks!

HaoTy:
Since we have `Pauli` now, it is perhaps more rigorous for `external_stabilizer()` to return a tuple of `Pauli`s, or a new `PauliString` class. However, I don't see Pauli strings being utilized very much (besides the correlation surface algorithm, which internally represents them as ints), and switching to a non-built-in type only adds inconvenience for the user (consider the old `add_cube()` interface and the current gallery constructions requiring a `Basis` instance but not a simple "X" or "Z" string). That said, I'm up for discussions.

inmzhang:
> Since we have `Pauli` now, it is perhaps more rigorous for `external_stabilizer()` to return a tuple of `Pauli`s, or a new `PauliString` class. However, I don't see Pauli strings being utilized very much (besides the correlation surface algorithm, which internally represents them as ints), and switching to a non-built-in type only adds inconvenience for the user (consider the old `add_cube()` interface and the current gallery constructions requiring a `Basis` instance but not a simple "X" or "Z" string). That said, I'm up for discussions.

`external_stabilizer()` is now mainly used internally and all the options look fine for me.

HaoTy:
> Looks all good and solid to me!

Thanks!

> `external_stabilizer()` is now mainly used internally and all the options look fine for me.

It can be useful for selecting the desired correlation surface, which may be relevant for implementing the condition for conditional operations. But let's worry about it later.

Issue #823, opened 2026-01-21
Improve runtime of pytest suite


smburdick:
As it stands, running the entire pytest suite takes a very long time (about 30 minutes on my M1 mac). I think we should try to reduce it to 5-10 minutes tops, so that it's easier to validate changes against the entire codebase.

I think it would be worth exploring a few things:

* For tests that take longest and are repeated, can we reduce the startup time (eg, via cacheing)?
* We should ensure that we are separating out unit tests from integration tests. The `compile` tests are numerous. For example, `test_compile_two_same_blocks_connected_in_time` is invoked 64 times in the suite and functions as a large integration test. Perhaps we can run it only a few times with randomized block inputs to save time.
* Notably, parallelizing it with the `-n auto` flag reduced performance by about 15% on my machine

HaoTy:
`compile_test.py` is one of the slowest components. A preliminary profiling reveals that the majority of the time is spent on 
1. pickle loading and dumping, when the detector database is present;

<img width="2467" height="958" alt="Image" src="https://github.com/user-attachments/assets/18d6c751-1203-4d96-8b9b-f957da2705c8" />

2. multiprocessing overhead, when detector database is not used (notably, if `do_not_use_database=True` and the database file is present, a lot of time is wasted on loading the pickle, which is immediately discarded);

<img width="2464" height="958" alt="Image" src="https://github.com/user-attachments/assets/e070053b-8d2d-4362-80e6-81ec72e876e0" />

I was using WSL2, so could be biased.

smburdick:
@HaoTy @jbolns @nelimee @inmzhang  Can one of you kindly reopen this, I marked as completed by mistake. Also, feel free to assign me.

As I mentioned in the group, we can have a round table discussion about how we want to tackle this tomorrow

jbolns:
Re-opened as it seems clearly an accidental click.  
Can't assign because, while I'm looking forward to hearing about this tomorrow, I haven't followed the discussion closely. Having said that, the size and importance of the issue suggests it is still too early to assign to anyone in particular. Perhaps best to wait for conversation tomorrow and more exchanges here?

smburdick:
@jbolns Thank you for reopening. Sure, we should probably break this down into more manageable subtasks.

HaoTy:
> I was using WSL2, so could be biased.

[...]

PR #828, opened 2026-02-03
Add slow decorators to speed up testing

smburdick:
This lowers the "non-slow" test runtime from 9 minutes to about 7 seconds on my machine.

Address #823 . As discussed there, `compile.py` is a huge bottleneck, as is mapping qubit indices inside the `Moment` class. Let's mark these tests as slow so we can run the test suite quickly as we develop, then run the slower integration tests once to check our code before committing.


Zhaoyilunnn:
```
/home/runner/work/tqec/tqec/docs/contributor_guide.rst:172: WARNING: Title underline too short.
```

Hi @smburdick you might need to fix the above warning, which caused the doc build failure

HaoTy:
Shall we add a line `addopts = "-m 'not slow'"` under `[pytest]` in pyproject.toml so that pytest defaults to not running slow tests? Configs in pyproject.toml usually are respected by IDEs, so it will also be helpful if one uses the Python testing extension in VSCode, for example. The slow tests can always be ran explicitly, and we should document this in the contributor guide. Some changes to the workflow configs might be needed.

smburdick:
> Shall we add a line `addopts = "-m 'not slow'"` under `[pytest]` in pyproject.toml so that pytest defaults to not running slow tests? Configs in pyproject.toml usually are respected by IDEs, so it will also be helpful if one uses the Python testing extension in VSCode, for example. The slow tests can always be ran explicitly, and we should document this in the contributor guide. Some changes to the workflow configs might be needed.

That's a good idea. I will do that, and update the docs

Zhaoyilunnn (approved):

Zhaoyilunnn:
LGTM. I think this PR can be merged.

smburdick:
@Zhaoyilunnn can you merge the PR? I don't have write access. Thank you.

PR #821, opened 2026-01-20
Add color to Cube class

smburdick:
Address #817 

ZXCubes are either red or blue, YHalfCubes are green, and Port cubes have undefined color.

Issue #817, opened 2026-01-17
Add node colour information on BlockGraph


nelimee:
## Is your feature request related to a problem? Please describe.

The `BlockGraph` data-structure does not have any easy way of recovering the colour of each of its node. That would be a good thing to add.

## Describe the solution you'd like

This can probably be done by implementing a new `Cube.colour` property on the `Cube` instances that are stored in the `BlockGraph`. 

Ideally, this property should return a field of a new enumeration that will at least contain "blue" and "red" but might also contain other colours if needed (e.g., "green" for the Y-basis reset/measurement, "purple" for magic state cultivation, ...).

## Describe alternatives you've considered

Add a new attribute to the nodes composing the `BlockGraph`. This has the downside of encoding some information in two different places, and so potential bugs w.r.t. only one place being updated without the other.

## Additional context

This is the first issue towards making the colour of some nodes in the `BlockGraph` to be dependent on correlation surfaces.

It might also be beneficial to shift the vision from the triplet description of each node (e.g. `"XZX"`) to just having its colour. These two descriptions are equivalent provided that each node has an easy way of knowing which pipes it connects to.

smburdick:
Hey @nelimee , go ahead and assign me.

smburdick:
For example, if the cube were an `XXZ` type `ZXCube`, would the "color" be `[red, blue]`? My understanding was that cube faces had colors corresponding to the Pauli error/stabilizer. I would assume we are also covering yellow (Hadamard) blocks, but maybe there's an abstract understanding of color here that I am missing.

afowler:
There is a direct identification. Z=blue, X=red, a cube with three letters
has the color of the single letter.

smburdick:
So `ZXCube`s can be red or blue; does this mean `YHalfCube`s are always green; what about `Port` cubes?

Which kinds of cubes can be purple (for T state injection) or yellow (for Hadamards)?

afowler:
The bulk of our nodes are red or blue. For example XXZ is blue, one Z, and
ZXZ is red, one X.

Green nodes are Y nodes. Only one connector allowed time oriented only.
Forward or back in time.

Purple nodes are T nodes. Only one connector forward time oriented.

afowler:
Oh, and Hadamard gates are not associated with nodes, rather edges.

smburdick:
> Oh, and Hadamard gates are not associated with nodes, rather edges.
> […](#)

What does that mean @afowler ?

[...]

Issue #791, opened 2025-12-04
Support conditional/dynamic operations


HaoTy:
This is an expansion of the discussion in https://github.com/tqec/tqec/issues/747#issuecomment-3574142412. I appreciate @Zhaoyilunnn's effort in driving this forward.

Classical feedforward/conditional/dynamic operations are essential for running meaningful circuits fault-tolerantly. They arise from various sources, including the repeat-until-success (RUS) nature of magic state preparation, conditional corrections in magic state injection, dynamic logical circuits such as projective rotation synthesis, etc., and need to be supported sooner or later.

At least three enhancements are needed. 
1. Dynamic lattice surgery operations.
1.1 Auto-retry in magic state cultivation based on post-selection. I assume this will be handled in standalone when cultivation is added.
1.2 Measurement basis change based on the result of a previous measurement. This is needed for auto-correcting magic state injections.
1.3 Dynamic logical circuits with a nontrivial structural change in the block graph. This will be quite common in future input circuits, given that repeat-until-success Rz synthesis can reduce the number of T gates by roughly 2/3 compared to ancilla-free Rz synthesis. We can aim to add dynamic block graphs as a starting point, but eventually, PyZX and topologiq need to support dynamic compilation as well. There is a relevant PR by @Zhaoyilunnn in PyZX (https://github.com/zxcalc/pyzx/pull/348).
2. Correlation surface computation for the dynamic block graph. I will try to address this in #707 or a follow-up PR.
3. Physical-level compilation and simulation. Since Stim `Circuit` and the detector sampler we currently use don't support non-Pauli conditional instructions, this needs to be done in a less efficient way, where block graph slices are compiled individually and simulated by Tableau simulations in concatenation. In this case, the classical control flow is realized using Python conditionals, and detector syndromes for decoding and logical observable calculations need to be handled explicitly. There could be more clever approaches; please do comment if you have ideas.

Please feel free to open new issues and work on some of these points.

beitom:
[...]

Issue #723, opened 2025-10-08
Integration of topologiq & related demos


jbolns:
Since the ZX / PyZX / BFS algorithm issues have expanded enormously, perhaps it is best to track the integration of topologiq and related demos in a fresh issue.

Draft PR #720 contains two notebooks with demos for Qiskit and Qrisp. Each notebook covers:
- Loading the foundational circuit
- Converting circuit to QASM
- Importing QASM to PyZX, as a circuit
- Converting PyZX circuit to a PyZX graph
- Converting the PyZX graph into a space-time diagram, using topologiq
- Using the space-time diagram to produce a TQEC `block_graph` 
- Generating Stim circuit and running simulations with the resulting `block_graph`

Anyone contributing a circuit produced with a different framework can use these notebooks as guidelines of what is needed. 

The further into the process the initial implementation is, the better.

But it's completely fine also if they just produce the initial circuit and we go over the steps together. 

... 

Closing #720 should suffice to close this issue. 

Additionally, closing this issue seems to suffice to close #523 , #528 , #449 . If there is anything in those issues topologiq doesn't achieve, perhaps it is new to make a new up-to-date issue.

KabirDubey:
Here is a pipeline describing the intended usage for the tool at the moment:

<img width="907" height="1539" alt="Image" src="https://github.com/user-attachments/assets/d9352c92-495f-4b84-af4b-f85d2e0b1e3b" />

jbolns:
Summary of events. The references just above culminated in a fork-to-branch PR. They're now part of #720.

It's failed some automatic checks, but:
- the import function is a significant improvement to existing one.
- it's adds a comprehensive set of tests. 

Will pull and see about fixing whatever caused the failed checks.

Adding details here instead of referencing back so it's easier to read details.

Thanks @SMC17 !

SMC17:
Glad the coordinate fix and tests were helpful for #720.

Quick note: @staxxx12 mentioned in the PR was an old git config - SMC17 is my only GitHub account. All future commits will show correctly!

I have a couple more improvements ready that might help with the integration:

1. **Error Message Improvements** ([branch](https://github.com/SMC17/tqec/tree/feature/improve-error-messages))
   - Enhanced PyZX/topologiq error messages with detailed context
   - Helps users debug QASM→PyZX→topologiq pipeline issues
   - All tests passing, pre-commit clean

[...]

PR #829, opened 2026-02-04
Introduce conditional cube

HaoTy:
This PR is part of the effort to address #818 and #791, which
1. introduces conditional cube and cube kind (potentially supercedes #817 and #821, as color is not critically relevant in realizing conditional operations and there's already a `TQECColor` enum class for visualization purposes), 
2. refactors static cube kinds into enums (feels natural to use enums when all same-kind instances are identical, but debatable), and
3. supports conditional cubes for most operations in `BlockGraph` (except for compilation, visualization, and from/to ZX graphs and .dae files, which raise `NotImplementedError`s).

Implementation was partly assisted by Claude code. I have reviewed all AI-generated parts to ensure they adhere to my intended design.


HaoTy:
@KabirDubey this one.

smburdick:
@HaoTy are you still working on this?

HaoTy:
> @HaoTy are you still working on this?

I still plan to complete the entire conditional-operation feature, but I'm a bit limited by my available time at the moment. I have a preliminary implementation of the compilation of conditional cubes, which will be a follow-up on this PR and require some more work. This PR, however, is pretty close to being ready for review. I didn't push forward earlier because the prerequisite correlation surface PR hadn't been merged. Now I think I can revisit this after completing some other TODOs on my list. Thanks for the reminder.

src/tqec/computation/block_graph.py
```diff
@@ -178,14 +197,23 @@ def _check_pipe_exists(self, pos1: Position3D, pos2: Position3D) -> None:
...
+        condition: CorrelationSurface | None = None,
+    ) -> Position3D:
         """Add a cube to the graph.
 
         Args:
             position: The position of the cube.
             kind: The kind of the cube. It can be a :py:class:`~tqec.computation.cube.CubeKind`
                 instance or a string representation of the cube kind.
             label: The label of the cube. Default is None.
+            condition: The condition for when the cube kind is conditional, specified as a partial
```
smburdick, line 215: Does the `CorrelationSurface` class delineate between partial/full CSes?

src/tqec/computation/block_graph.py
```diff
@@ -540,9 +612,107 @@ def find_correlation_surfaces(self) -> list[CorrelationSurface]:
...
+    def resolve_conditional_kinds(
+        self, condition_values: bool | int | Mapping[Position3D, bool | int]
+    ) -> BlockGraph:
+        """Resolve every conditional cube to one of its branch kinds and return a static graph.
+
+        Args:
+            condition_values: the runtime value(s) of the conditions. If a single boolean
+                (or integer) is provided, all the conditional cubes are resolved with that
+                value. Otherwise, a mapping from the positions of the conditional cubes to
+                their respective condition values must be provided.
```
smburdick, line 637: Do you think the code should check for these different use cases, and possibly perform some error handling?

src/tqec/computation/block_graph.py
```diff
@@ -540,9 +612,107 @@ def find_correlation_surfaces(self) -> list[CorrelationSurface]:
...
+                raise TQECError(f"There is no port with label {label}.")
+            pos = self._ports[label]
+        else:
+            raise TQECError(f"Invalid port specification: {port}")
+
+        if isinstance(kind, str):
+            kind = cube_kind_from_string(kind)
+
+        fill_node = Cube(pos, kind, label, condition)
+        self._graph.add_node(pos, **{self._NODE_DATA_KEY: fill_node})
```
smburdick, line 702: Does this propagate the condition information correctly?

src/tqec/computation/block_graph.py
```diff
@@ -697,6 +867,11 @@ def rotate(
             rotate_position_by_matrix,
         )
 
+        if any(cube.condition is not None for cube in self.cubes):
+            raise NotImplementedError(
+                "Rotating a block graph containing conditional cubes is not implemented: "
```
smburdick, line 875: Is this non-trivial to implement?

src/tqec/computation/correlation.py
```diff
@@ -266,6 +266,31 @@ def shift_by(self, dx: int = 0, dy: int = 0, dz: int = 0) -> CorrelationSurface:
     def __xor__(self, other: CorrelationSurface) -> CorrelationSurface:
         return CorrelationSurface(self.span.symmetric_difference(other.span))
 
+    def to_dict(self) -> dict[str, Any]:
+        """Return a JSON-serializable dictionary representation of the correlation surface."""
+        return {
+            "span": [
```
smburdick, line 272: Does this capture the conditionality of it?

src/tqec/computation/cube.py
```diff
@@ -129,59 +138,194 @@ def with_basis_along(self, direction: Direction3D, basis: Basis) -> ZXCube:
...
+            The :py:class:`~tqec.computation.cube.LeafCubeKind` instance constructed
+            from the string representation.
+
+        """
+        match string.upper():
+            case "PORT" | "P":
+                return LeafCubeKind.PORT
+            case "Y" | "Y_HALF_CUBE":
+                return LeafCubeKind.Y_HALF_CUBE
+            # case "T":
```
smburdick, line 197: Can (or will) magic state cubes exist as regular cubes?

src/tqec/computation/cube.py
```diff
@@ -129,59 +138,194 @@ def with_basis_along(self, direction: Direction3D, basis: Basis) -> ZXCube:
...
-            return YHalfCube()
-        case _:
-            return ZXCube.from_str(s)
+    s = s.strip().upper()
+    if s in ZXCube.__members__:
+        return ZXCube.from_str(s)
+    if s in LeafCubeKind.__members__ or s in ["PORT", "P", "Y"]:
+        return LeafCubeKind.from_str(s)
+    if "_" in s:
+        return ConditionalCubeKind.from_str(s)
```
smburdick, line 324: I guess I don't fully understand why ConditionalCubeKind is divorced from being a ZXCube or leaf cube. If I understand, it is a kind of placeholder, but if there's information about what it could be, could it also be represented as a hybrid/optional type (That wouldn't map neatly into this context, so leaving it blank for being a string will probably suffice. Just wanted to clarify it for other contexts as well)

src/tqec/computation/open_graph.py
```diff
@@ -164,18 +164,18 @@ def ports_basis_for_clique(
...
+                fg.fill_ports({port: LeafCubeKind.Y_HALF_CUBE})
                 continue
             assert basis in ["Z", "X"]
             # Match the basis of the pipe at the port
             pipe_at_port = graph.pipes_at(port_pos)[0]
             pipe_kind = pipe_at_port.kind
             at_head = pipe_at_port.u == graph[port_pos]
             fill_kind = ZXCube(
-                *[
+                tuple(
```
smburdick, line 175: I'd double check the type safety around these tuple castings if you haven't already. ty check may not catch them.

src/tqec/interop/bgraph/read_write.py
```diff
@@ -120,6 +120,10 @@ def write_bgraph(
     bgraph_lines.append("\nCUBES: index;x;y;z;kind;label;\n")
     write_ids = {}
     for cube in block_graph.cubes:
+        if cube.is_conditional:
+            raise NotImplementedError(
+                "Exporting conditional cubes to BGRAPH file is not yet supported."
```
smburdick, line 125: would be worth bringing up in our ongoing discussion of the bgraph format :-)

src/tqec/interop/collada/read_write.py
```diff
@@ -207,6 +207,10 @@ def write_block_graph_to_dae_file(
     for cube in block_graph.cubes:
         if cube.is_port:
             continue
+        if cube.is_conditional:
+            raise NotImplementedError(
+                "Exporting conditional cubes to DAE file is not yet supported."
```
smburdick, line 212: Have you given any thought as to how you'd want to represent these in a visual editor?

tests/computation/conditional_test.py
```diff
@@ -0,0 +1,243 @@
...
+        g.validate()
+    # A temporal pipe whose walls do not match one of the branches.
+    g = BlockGraph()
+    g.add_cube(Position3D(0, 0, 0), "XZZ")
+    g.add_cube(Position3D(0, 0, 1), "ZXZ_ZXX", condition=condition)
+    g.add_pipe(Position3D(0, 0, 0), Position3D(0, 0, 1), "XZO")
+    with pytest.raises(TQECError, match=r"mismatched\s+colors"):
+        g.validate()
+
+
```
smburdick, line 249: These unit tests are good, but have you considered also updating the compilaiton or other larger integration test suites?

Of course, we will see a lot more use cases as this sees more support on potential frontends too.

smburdick (commented):
* Can you comment whether you were more or less reliant on Claude Code for this one as opposed to say #995?
* Also, is there any dependency crossover you feel might exist between this and #995, or are they completely independent branches?

src/tqec/computation/block_graph.py
HaoTy: No.

src/tqec/computation/block_graph.py
HaoTy: Do you have a broken example?

src/tqec/computation/block_graph.py
HaoTy: Not too difficult but also not trivial. It just requires the rotation of a correlation surface in consideration of its position in the block graph. I consider block graph rotation as a low-priority feature, so will leave this for someone else and another PR :)

src/tqec/computation/correlation.py
HaoTy: The correlation surface is the condition itself.

src/tqec/computation/cube.py
HaoTy: This is a discussion deserves to be on the meeting agenda.

tests/computation/conditional_test.py
HaoTy: Conditional cubes currently do not compile. It's a fairly complex enhancement
refactor that I plan to carry out through a few future PRs.

src/tqec/interop/collada/read_write.py
HaoTy: Looking forward to learning your discussion results through the meeting recording :)

src/tqec/interop/bgraph/read_write.py
HaoTy: Yes, and I have brought it up during BGRAPH's initial implementation. No intention to implement it in this PR though.

src/tqec/computation/cube.py
HaoTy: Not sure what your suggestion is. Can you give a concrete example?

HaoTy:
> * Can you comment whether you were more or less reliant on Claude Code for this one as opposed to say #995?

I wrote a considerable amount of code in this PR. In contrast, #995 was mostly implemented by Claude. How is this question relevant?

> * Also, is there any dependency crossover you feel might exist between this and #995, or are they completely independent branches?

They should be pretty much independent. But both are prerequisites to the compilation of conditional cubes.

src/tqec/interop/collada/read_write.py
KabirDubey: For conditional leaf nodes, I think we've previously seen sketchup diagrams with omitted initialization and/or measurement faces. This is pretty much an HCI question imo, can be added to the list of spinoff issues.

src/tqec/computation/block_graph.py
smburdick: I don't. Would be worth double checking, but maybe the unpacking is what irks me.

src/tqec/computation/cube.py
smburdick: Asked another way: conceptually, can a cube be both a conditional cube as well as another kind (ZX, etc) simultaneously.

smburdick:
> > * Can you comment whether you were more or less reliant on Claude Code for this one as opposed to say [Enhance the correlation surface algorithm with vertex-ordering-based outer loop, partial surface finding, and performance improvements #995](https://github.com/tqec/tqec/pull/995)?
> 
> I wrote a considerable amount of code in this PR. In contrast, #995 was mostly implemented by Claude. How is this question relevant?

Mostly out of my own curiosity. I haven't used any AI agents for writing large amounts of code, and am wondering how effective they really are for research projects like ours.

KabirDubey:
Thanks!! I'm aiming to go over the code this weekend. My main point of clarification is about `BlockGraph.resolve_conditional_kinds` outputting a `BlockGraph`. Is the purpose of this method solely to isolate static block graphs, e.g. for inspecting a faulty execution pathway?

If this is intended as the access point for compilation, then I'm not sure how that would work efficiently. Consider a single-qubit `(HT)...` program with `D` unconditional nodes and `K` conditional leaf nodes. When the parity determining the `i`-th conditional measurement is known, would you recompile the same block graph, but now with `D + i` unconditional nodes and `K - i` conditional leaf nodes? If so, then generating the layer tree before layer `i` would be wasteful, because everything we need to know about the logical operator can be considered already computed and stored by necessity. Therefore, a `resolve` method should output no more than a list of blocks.

An alternate design would feature conditionality as an attribute to any block[^1] and directly respond with physical-level circuits. Perhaps a sibling attribute to the block kind would work. I can try to go into more detail and/or draft a PR if I'm not making any conceptual errors and if it would be helpful.

The content of each physical circuit possibility depends on the system execution configuration, but I think it amounts to the following at runtime:
-if the block has not already been generated, generate a particular circuit whose macroscopic behavior corresponds to the conditional block's kind;
-once that circuit has been executed or simulated, link the appropriate history of its physical measurements to the logical observable.

Some examples that you're probably already accounting for but may help illustrate my POV:

One could obtain conditional logic at the block graph layer, e.g. to instruct a spatial arm plus a Y-basis measurement over ancilla space to inject an S gate, by programming each relevant block as a conditional block and directing the downstream compilation to *only* emit the circuit, monitor its execution or simulation it, and link it to the logical correlator *if* the S-gate correction is needed. The spacetime volume of this gadget does not have to be wasted because it can be filled with another gadget, perhaps one scheduled later or  MSC, that is itself typed as a conditional block whose spacetime location is conditioned on the appearance of the S-gate structure.

In the case where you have enough space to precompute the S-gate injection structure--i.e. when `topologiq` requires an `|i>` state to be initialized and entangled with the data and T state before the joint input/T-state parity is known--`tqec` would emit the circuit, monitor its execution or simulation, and *only* link it to the logical observable *if* the S-gate correction is needed.

In the case where you have a spatial arm, i.e. a lattice-surgery move, and you're running complementary gap decoding, the spatial arm and junctions would be conditional blocks. Then the compilation would emit the circuit and, during or after[^2] execution/simulation, calculate the gap and annotate the observable with whichever of the `O(d)` spatial-to-spatial-boundaried correlation surfaces obtains the confidence satisfying the desired stability/runtime tradeoff.

[^1]: this is mostly a nit, but I think the typing for this feature should say conditional `Block` or conditional node instead of conditional cube. I think this would eventually interoperate most easily with the `BlockGraph` IR and walking surface codes. The instructions are rarely if ever cubes. The Y half-cube and MSC you enable conditionality for in this PR are not macroscopically cubes. Moreover, a compiler engineer may want to program entire conditional block subgraphs, and they can do so by packaging a block graph composed of conditional blocks. Furthermore, a compiler engineer may want to program a conditional measurement `Layer` or conditional face where `O(d)` stabilizer measurements are scheduled just to obtain some temporal distance before a subsequent operation; they can do so by packaging a conditional block which passes the branching logic off the same condition to the circuit generator.

Since any block circuit that can be generated at compile time can also be generated at execution time, I think `Block` is the correct abstraction layer, not `Cube`. It doesn't seem like too much trouble to work toward giving programmers that flexibility in this PR. Up to you though, mainly want to hear your thoughts.

[^2]: depending on decoder speed.

src/tqec/computation/block_graph.py
HaoTy: I checked the other `BlockGraph` methods with type dispatching, and the current treatment is consistent with the convention.

src/tqec/computation/block_graph.py
HaoTy: You cannot distinguish them without a graph. Sometimes you cannot distinguish them even with a graph.

src/tqec/computation/cube.py
HaoTy: A conditional cube is a choice between two regular cubes (with some restrictions on the kinds), where the choice is resolved at runtime.

src/tqec/computation/open_graph.py
HaoTy: This is incorrect on main but correct here, because ZXCube is now an Enum where the values are tuples of Basis

src/tqec/interop/collada/read_write.py
HaoTy: Although the first implementation of conditional cube compilation likely will only support switching between X and Z, but at some point we may also want to support switching between X/Z and Y.

src/tqec/interop/bgraph/read_write.py
HaoTy: Tagging @jbolns

HaoTy:
> > > * Can you comment whether you were more or less reliant on Claude Code for this one as opposed to say [Enhance the correlation surface algorithm with vertex-ordering-based outer loop, partial surface finding, and performance improvements #995](https://github.com/tqec/tqec/pull/995)?
> > 
> > 
> > I wrote a considerable amount of code in this PR. In contrast, #995 was mostly implemented by Claude. How is this question relevant?
> 
> Mostly out of my own curiosity. I haven't used any AI agents for writing large amounts of code, and am wondering how effective they really are for research projects like ours.

They are super effective, if used properly.

I have resolved most of the comments, but feel free to unresolve them if you have additional concerns.

src/tqec/interop/bgraph/read_write.py
jbolns: Agree w. @HaoTy . No need to do here. 

For reference, in Topologiq I call them `zx*` znc `xz*` depending on orientation (with common denominator `xz` for internal usage). But I can write them to BGRAPH however needed.

That said, can be resolved in #723, which can now include basically all gates.

HaoTy: The tricky part is, you also need to represent the condition.

HaoTy:
> Thanks!! I'm aiming to go over the code this weekend. My main point of clarification is about `BlockGraph.resolve_conditional_kinds` outputting a `BlockGraph`. Is the purpose of this method solely to isolate static block graphs, e.g. for inspecting a faulty execution pathway?
> 
> If this is intended as the access point for compilation, then I'm not sure how that would work efficiently. Consider a single-qubit `(HT)...` program with `D` unconditional nodes and `K` conditional leaf nodes. When the parity determining the `i`-th conditional measurement is known, would you recompile the same block graph, but now with `D + i` unconditional nodes and `K - i` conditional leaf nodes? If so, then generating the layer tree before layer `i` would be wasteful, because everything we need to know about the logical operator can be considered already computed and stored by necessity. Therefore, a `resolve` method should output no more than a list of blocks.

This is intended solely for debugging purposes.

> An alternate design would feature conditionality as an attribute to any block[1](#user-content-fn-1-66c6fb08ab93e75c4ef3b1a802877efa) and directly respond with physical-level circuits. Perhaps a sibling attribute to the block kind would work. I can try to go into more detail and/or draft a PR if I'm not making any conceptual errors and if it would be helpful.

There are at least three types of dynamic behavior (see #791), which, IMO, need different treatments. This PR specifically targets type 1.2, the easiest and most ready-to-tackle one, where a switch between X and Z measurements is sufficient.

I think your proposal could be a solution to 1.3 (and help 1.1), but what I had in mind was to have an additional abstraction on top of `BlockGraph` (see https://github.com/tqec/tqec/issues/791#issuecomment-3701535984) so that you can use one condition to control a large piece of structure, where within this piece, everything still has a fixed relationship.

> The content of each physical circuit possibility depends on the system execution configuration, but I think it amounts to the following at runtime: -if the block has not already been generated, generate a particular circuit whose macroscopic behavior corresponds to the conditional block's kind; -once that circuit has been executed or simulated, link the appropriate history of its physical measurements to the logical observable.
> 
> Some examples that you're probably already accounting for but may help illustrate my POV:
> 
> One could obtain conditional logic at the block graph layer, e.g. to instruct a spatial arm plus a Y-basis measurement over ancilla space to inject an S gate, by programming each relevant block as a conditional block and directing the downstream compilation to _only_ emit the circuit, monitor its execution or simulation it, and link it to the logical correlator _if_ the S-gate correction is needed. The spacetime volume of this gadget does not have to be wasted because it can be filled with another gadget, perhaps one scheduled later or MSC, that is itself typed as a conditional block whose spacetime location is conditioned on the appearance of the S-gate structure.
> 
> In the case where you have enough space to precompute the S-gate injection structure--i.e. when `topologiq` requires an `|i>` state to be initialized and entangled with the data and T state before the joint input/T-state parity is known--`tqec` would emit the circuit, monitor its execution or simulation, and _only_ link it to the logical observable _if_ the S-gate correction is needed.
> 
> In the case where you have a spatial arm, i.e. a lattice-surgery move, and you're running complementary gap decoding, the spatial arm and junctions would be conditional blocks. Then the compilation would emit the circuit and, during or after[2](#user-content-fn-2-66c6fb08ab93e75c4ef3b1a802877efa) execution/simulation, calculate the gap and annotate the observable with whichever of the `O(d)` spatial-to-spatial-boundaried correlation surfaces obtains the confidence satisfying the desired stability/runtime tradeoff.
> 
> Since any block circuit that can be generated at compile time can also be generated at execution time, I think `Block` is the correct abstraction layer, not `Cube`. It doesn't seem like too much trouble to work toward giving programmers that flexibility in this PR. Up to you though, mainly want to hear your thoughts.
> 
> ## Footnotes
> 1. this is mostly a nit, but I think the typing for this feature should say conditional `Block` or conditional node instead of conditional cube. I think this would eventually interoperate most easily with the `BlockGraph` IR and walking surface codes. The instructions are rarely if ever cubes. The Y half-cube and MSC you enable conditionality for in this PR are not macroscopically cubes. Moreover, a compiler engineer may want to program entire conditional block subgraphs, and they can do so by packaging a block graph composed of conditional blocks. Furthermore, a compiler engineer may want to program a conditional measurement `Layer` or conditional face where `O(d)` stabilizer measurements are scheduled just to obtain some temporal distance before a subsequent operation; they can do so by packaging a conditional block which passes the branching logic off the same condition to the circuit generator. [↩](#user-content-fnref-1-66c6fb08ab93e75c4ef3b1a802877efa)
> 2. depending on decoder speed. [↩](#user-content-fnref-2-66c6fb08ab93e75c4ef3b1a802877efa)

I have a toy example compiled locally, but it needs some generalization to the observable part. The strategy I use is to have the conditional version of every intermediate representation involved in the compilation, down to dynamic physical circuits, so that at runtime the resolved observable value can lead directly to compiled low-level instructions, rather than high-level branches that would need to be compiled at runtime.

A `Block` is a `SequencedLayers`, an IR during compilation, so it gets its own conditional version. This is another reason I specify the condition at the `Cube` level, besides it's sufficient for type 1.2: `BlockGraph` is specified with `Cube`s and `Pipe`s, not `Block`s; `Block` is a middleware.

Btw, how do you insert footnotes? Pretty neat.

src/tqec/interop/bgraph/read_write.py
jbolns: On the way in, or on the way out?
Actually, unless you absolutely insist, I'm going to stick to your original note for not wanting to implement here, and do us both a favour and leave that conversation for future versions of ourselves. 😁

HaoTy: Conditions should be preserved round-trip, which means we need a text representation for correlation surfaces. No too difficult, but let's have this conversation in a future PR.

HaoTy:
@KabirDubey I finished watching the recording. I have a very clear picture of what we need to perform a T injection since #791 and have been working on subtasks toward this goal, so my recent PRs are very relevant to what's discussed: #995 has implemented partitioning a `BlockGraph` into local chunks and concatenating correlation surfaces across adjacent chunks. It also allows finding the complete correlation surface from a partial one, which is also needed for compiling conditional cubes.
This PR implements the "color abstraction". 
The "delay abstraction" is realized by the additional abstraction on top of `BlockGraph` in my above comment and https://github.com/tqec/tqec/issues/791#issuecomment-3701535984 (not implemented yet).

KabirDubey:
> > Thanks!! I'm aiming to go over the code this weekend. My main point of clarification is about `BlockGraph.resolve_conditional_kinds` outputting a `BlockGraph`. Is the purpose of this method solely to isolate static block graphs, e.g. for inspecting a faulty execution pathway?
> > If this is intended as the access point for compilation, then I'm not sure how that would work efficiently. Consider a single-qubit `(HT)...` program with `D` unconditional nodes and `K` conditional leaf nodes. When the parity determining the `i`-th conditional measurement is known, would you recompile the same block graph, but now with `D + i` unconditional nodes and `K - i` conditional leaf nodes? If so, then generating the layer tree before layer `i` would be wasteful, because everything we need to know about the logical operator can be considered already computed and stored by necessity. Therefore, a `resolve` method should output no more than a list of blocks.
> 
> This is intended solely for debugging purposes.

I see. If you have the chance to give a high level overview of how you envision the functions in this PR eventually passing to `compile.py`,`SequencedLayers`, etc. it might make my planned code-level review a bit faster. 

> > An alternate design would feature conditionality as an attribute to any block[1](#user-content-fn-1-66c6fb08ab93e75c4ef3b1a802877efa) and directly respond with physical-level circuits. Perhaps a sibling attribute to the block kind would work. I can try to go into more detail and/or draft a PR if I'm not making any conceptual errors and if it would be helpful.
> 
> There are at least three types of dynamic behavior (see #791), which, IMO, need different treatments. This PR specifically targets type 1.2, the easiest and most ready-to-tackle one, where a switch between X and Z measurements is sufficient.
> 

Sorry I forgot about #791, your perspective makes sense now. I'll need to do a closer read but in general different treatments (specifically separating 1.1 from the rest) sound good to me. I'm generally cautious about excessive abstractions and a first glance at a class like `ConditionalLeafNode` doesn't seem clean to me but as long as you're not runtime generating blocks that have already been executed and open things to being refactored for programmability I'm not opposed. 

> I think your proposal could be a solution to 1.3 (and help 1.1), but what I had in mind was to have an additional abstraction on top of `BlockGraph` (see [#791 (comment)](https://github.com/tqec/tqec/issues/791#issuecomment-3701535984)) so that you can use one condition to control a large piece of structure, where within this piece, everything still has a fixed relationship.

My understanding is that all of the features are actions conditioned on some calculation between previously enumerated correlation surface possibilities and real time control system feedback (1.1 at a microscopic level, everything else at a macroscopic level). I think your proposal would support the fallback rotation synthesis gadget, but if I understand correctly it would not cover gadgets like the probabilistic T gate where the measurement and fidelity of blocks within the gadget[^1] inform the structure of the static gadget itself. I think this is why you distinguish between 1.2 and 1.3. It's currently unclear to me how your design would elegantly support the gap decoding example I described below, which can implicate any block. 

> > In the case where you have a spatial arm, i.e. a lattice-surgery move, and you're running complementary gap decoding, the spatial arm and junctions would be conditional blocks. Then the compilation would emit the circuit and, during or after[2](#user-content-fn-2-66c6fb08ab93e75c4ef3b1a802877efa) execution/simulation, calculate the gap and annotate the observable with whichever of the `O(d)` spatial-to-spatial-boundaried correlation surfaces obtains the confidence satisfying the desired stability/runtime tradeoff.

Perhaps these would be covered by,
> The "delay abstraction" is realized by the additional abstraction on top of BlockGraph in my above comment and https://github.com/tqec/tqec/issues/791#issuecomment-3701535984 (not implemented yet).

but this is not intuitive to me because you're taking a signal obtained by a microscopic process (a physical measurement specified at the `stim` (or I guess tentatively `stimdx`) circuit generation level) that implicates a microscopic instruction (the scheduling of layers within a Block), but writing that code at a macroscopic (block graph) level. The modification of the block graph must inevitably happen at the level of blocks, but modifying `BlockGraph` is insufficient because a delay may not shift the block graph a macroscopic amount of distance. You agree here that

> A `Block` is a `SequencedLayers`, an IR during compilation, so it gets its own conditional version.

so I guess all I am saying is that a conditional block graph ought to be no more than a block graph consisting of conditional blocks (including ports) whose type, connectivity, and underlying circuit depend on a control system response. I believe this is the most simple implementation that covers all the cases, including making this strategy:

> The strategy I use is to have the conditional version of every intermediate representation involved in the compilation, down to dynamic physical circuits, so that at runtime the resolved observable value can lead directly to compiled low-level instructions, rather than high-level branches that would need to be compiled at runtime.

not have to serialize and evaluate any IR below the block graph IR except the conditional blocks in a block graph. 

> This is another reason I specify the condition at the `Cube` level, besides it's sufficient for type 1.2: `BlockGraph` is specified with `Cube`s and `Pipe`s, not `Block`s; `Block` is a middleware.

Makes sense. I'd vote to refactor the typing because it's confusing to call blocks that aren't cubes, cubes, but understand your implementation in this PR. My view is that blocks are ZX nodes including Hadamards with space and time data, and a cube is a block occupying dxdxd spacetime. 

[^1]: this is my local naming for a block "subgraph", I think similar to to what you call `BlockGraphLike` in #791.

KabirDubey:
> Btw, how do you insert footnotes? Pretty neat.

https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/

HaoTy:
> I see. If you have the chance to give a high level overview of how you envision the functions in this PR eventually passing to `compile.py`,`SequencedLayers`, etc. it might make my planned code-level review a bit faster.

Conditional cubes get compiled into `SequencedLayer`s with the top layer being a `ConditionalLayer`. This is one of the advantages of focusing on the switch between X and Z measurements, i.e., only the final layer needs to be compiled conditionally. The condition, as a partial correlation surface, gets expanded to a full correlation surface, then compiled into an `AbstractObservable`, which continues to serve as the condition. This is the part that needs additional work to generalize. While walking the `LayerTree`, upon encountering a `ConditionalLayer`, we perform the usual layout merging but create two branches of the `LayoutLayer`. This also needs generalization; otherwise it would incur exponentially many branches. Each `LayoutLayer` is compiled into a normal `Stim` circuit, where the branches are captured in a `DynamicCircuit` wrapper of `Stim` circuits (similar to `stimdx`), where the conditions are measurements that constitute the compiled observables.

> Sorry I forgot about #791, your perspective makes sense now. I'll need to do a closer read but in general different treatments (specifically separating 1.1 from the rest) sound good to me. I'm generally cautious about excessive abstractions and a first glance at a class like `ConditionalLeafNode` doesn't seem clean to me but as long as you're not runtime generating blocks that have already been executed and open things to being refactored for programmability I'm not opposed.

`ConditionalLeafCubeKind` suits type 1.2, and I'd argue that it's much friendlier to construct for an upstream compiler (or manually by a human) and much easier to implement the compilation if you separate 1.2 and 1.3. And as a matter of fact, 1.3 is not a hard requirement for universal computation, i.e., the T injection, so we can afford not having it for some time. In contrast, everything on the critical path of implementing the T injection should be a priority.

> My understanding is that all of the features are actions conditioned on some calculation between previously enumerated correlation surface possibilities and real time control system feedback (1.1 at a microscopic level, everything else at a macroscopic level). I think your proposal would support the fallback rotation synthesis gadget, but if I understand correctly it would not cover gadgets like the probabilistic T gate where the measurement and fidelity of blocks within the gadget[1](#user-content-fn-1-ae3b2638a7a41172cbb45d88b6b51a5a) inform the structure of the static gadget itself. I think this is why you distinguish between 1.2 and 1.3. It's currently unclear to me how your design would elegantly support the gap decoding example I described below, which can implicate any block.

Not sure what you mean by a probabilistic T gate. If you mean probabilistic synthesis (mixture of unitaries), it doesn't depend on runtime values. For fallback synthesis, you always prepare the fallback sequence beforehand. In the upstream dynamic logical circuit, the content of every branch is still static (we can capture multi-level conditionals), so conditionally controlling static structures should be sufficient. You can have the structure be as small as a single `Cube`.

I do agree that having more granular control is beneficial or even necessary if we want to implement certain features, like the walking surface code or your example below, but it would require refactoring how we construct a `BlockGraph` first, and then the consideration of how we place the conditions.

> > > In the case where you have a spatial arm, i.e. a lattice-surgery move, and you're running complementary gap decoding, the spatial arm and junctions would be conditional blocks. Then the compilation would emit the circuit and, during or after[2](#user-content-fn-2-66c6fb08ab93e75c4ef3b1a802877efa) execution/simulation, calculate the gap and annotate the observable with whichever of the `O(d)` spatial-to-spatial-boundaried correlation surfaces obtains the confidence satisfying the desired stability/runtime tradeoff.
> 
> Perhaps these would be covered by,
> 
> > The "delay abstraction" is realized by the additional abstraction on top of BlockGraph in my above comment and [#791 (comment)](https://github.com/tqec/tqec/issues/791#issuecomment-3701535984) (not implemented yet).

> but this is not intuitive to me because you're taking a signal obtained by a microscopic process (a physical measurement specified at the `stim` (or I guess tentatively `stimdx`) circuit generation level) that implicates a microscopic instruction (the scheduling of layers within a Block), but writing that code at a macroscopic (block graph) level. The modification of the block graph must inevitably happen at the level of blocks, but modifying `BlockGraph` is insufficient because a delay may not shift the block graph a macroscopic amount of distance. You agree here that

We are building a compiler here, whose purpose is to bridge macroscopic and microscopic abstractions. The program is written at a high-level, one way or another, then compiled down to microscopic instructions. If you think you can directly write it at the low-level, you are simply shifting the burden to the user.

> > A `Block` is a `SequencedLayers`, an IR during compilation, so it gets its own conditional version.
> 
> so I guess all I am saying is that a conditional block graph ought to be no more than a block graph consisting of conditional blocks (including ports) whose type, connectivity, and underlying circuit depend on a control system response. I believe this is the most simple implementation that covers all the cases, including making this strategy:
> 
> > The strategy I use is to have the conditional version of every intermediate representation involved in the compilation, down to dynamic physical circuits, so that at runtime the resolved observable value can lead directly to compiled low-level instructions, rather than high-level branches that would need to be compiled at runtime.
> 
> not have to serialize and evaluate any IR below the block graph IR except the conditional blocks in a block graph.

We should consider not only what's a minimalistic design within tqec, but also what's natural and user-friendly for the upstream to construct. Think about it this way: when you are writing if statements in Python, you don't write a "if" for every line in the branch. By requiring every block to be separately controlled by a condition and giving up the user-friendiliness to control many of them together, you effectively shift the burden of this compilation step to the upstream.

I agree it's debatable whether tqec should handle this step completely or if the upstream can have a say. But honestly, I think grouping instructions by shared conditions is still desirable even at the physical circuit level. See, for example, how you write a Guppy circuit. So IMO, a reasonable compilation of type 1.3 should keep the condition structure and relationship from the logical circuit all the way to the physical circuit. This also has the advantage of letting us reuse much of the current compilation pipeline, as opposed to rewriting pretty much everything. In the far future, perhaps we will do that and expose two levels of entry points, and upstream compilers can choose accordingly.

jbolns:
> While walking the LayerTree, upon encountering a ConditionalLayer, we perform the usual layout merging but create two branches of the LayoutLayer. This also needs generalization; otherwise it would incur exponentially many branches.

Q1. Is the branching carried out exhaustively all the way to the end of the computation?

Q2. And how are you planning to adjust the layers if/when there's a need to involve relatively positioning on Z?

Reason I ask... 

I saw #805 merged some weeks ago.

Streaming should allow for a version of an idea I floated a while ago, i.e., introduce some sort of time horizon. I was thinking about it in the context of rebuilding the Blockgraph entirely, which would be quite hard. However, a time horizon seems realisable if all it does is pause and reignite an existing stream.

Say, for instance, you have a Blockgraph with 7 conditionals, all requiring information from the previous one. 
1. Two branches until conditional numero 2. 
3. Pause stream until you get the outcome of conditional numero 1.
4. Two branches until numero 3. 

Does that add up to exponential? I think it stays in polynomial territory?

I'm not suggesting this is better than 
> I do agree that having more granular control is beneficial or even necessary if we want to implement certain features, like the walking surface code or your example below, but it would require refactoring how we construct a BlockGraph first, and then the consideration of how we place the conditions.

But it might be possible without having to re-conceive the BlockGraph.

HaoTy:
> > While walking the LayerTree, upon encountering a ConditionalLayer, we perform the usual layout merging but create two branches of the LayoutLayer. This also needs generalization; otherwise it would incur exponentially many branches.
> 
> Q1. Is the branching carried out exhaustively all the way to the end of the computation?

Currently. I only have a toy example compiled, which only has one branching point. The current approach will likely produce incorrect results for more than one branching points.

> Q2. And how are you planning to adjust the layers if/when there's a need to involve relatively positioning on Z?

This out of the scope of type 1.2 and this PR, but I plan to carry the relative positioning to the dynamic physical circuit level.

> 
> Reason I ask... 
> 
> I saw #805 merged some weeks ago.
> 
> Streaming should allow for a version of an idea I floated a while ago, i.e., introduce some sort of time horizon. I was thinking about it in the context of rebuilding the Blockgraph entirely, which would be quite hard. However, a time horizon seems realisable if all it does is pause and reignite an existing stream.

#805 is only for strsming circuits from a compiled `LayerTree`, so can't support what you are describing. Streaming the full compilation pipeline is very desirable but more nontrivial than #805.

> Say, for instance, you have a Blockgraph with 7 conditionals, all requiring information from the previous one. 
> 1. Two branches until conditional numero 2. 
> 3. Pause stream until you get the outcome of conditional numero 1.
> 4. Two branches until numero 3. 
> 
> Does that add up to exponential? I think it stays in polynomial territory?
> 
> I'm not suggesting this is better than 
> > I do agree that having more granular control is beneficial or even necessary if we want to implement certain features, like the walking surface code or your example below, but it would require refactoring how we construct a BlockGraph first, and then the consideration of how we place the conditions.
> 
> But it might be possible without having to re-conceive the BlockGraph.

I remember you talking about this idea. From what I can tell, I think you overcomplicated the thinking. Conceptually, it's really simple: if a program is given with polynomially many branches, it can be compiled into polynomial branches - just maintain the branch structure thoughout the compilation.

What we will do is to compile every branch into low-level instructions and keep them in the branch structure from the input. This is because 1) we want to minimize the idling time of quantum hardware so should do as much as possible at compile time, 2) we csn compile them once and use them across shots, and 3) through all the shots, we will likely go over every branch anyway.

KabirDubey:
@greptileai please review

jbolns:
> I think you overcomplicated the thinking

Thanks! And to be clear, this is not sarcasm. To me, this sounds like it would work but is less lean than your preferred approach. I take that! Any day. 

> through all the shots, we will likely go over every branch anyway.

I hadn't consider this, in which case the over-complication is indeed unnecessary even to my eyes.

src/tqec/utils/rotations.py
```diff
@@ -114,6 +117,13 @@ def rotate_block_kind_by_matrix(
     if str(block_kind) == "PORT":
         return block_kind
 
+    if isinstance(block_kind, ConditionalCubeKind):
+        rotated_branches = tuple(
+            rotate_block_kind_by_matrix(branch, rotation_matrix) for branch in block_kind.branches
+        )
+        assert all(isinstance(branch, StaticCubeKind) for branch in rotated_branches)
+        return ConditionalCubeKind(cast(tuple[StaticCubeKind, StaticCubeKind], rotated_branches))
```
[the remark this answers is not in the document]
HaoTy: Now the error is caught and re-raised.

src/tqec/computation/block_graph.py
```diff
@@ -90,9 +97,21 @@ def num_ports(self) -> int:
...
+        cube, i.e. it contributes the branch-averaged (expected) number of half Y
+        cubes, as the branch actually implemented is only known at runtime.
+        """
+        count = 0.0
+        for node in self.cubes:
+            kind = node.kind
+            if kind is LeafCubeKind.Y_HALF_CUBE:
+                count += 1
+            elif isinstance(kind, ConditionalCubeKind):
+                count += sum(branch is LeafCubeKind.Y_HALF_CUBE for branch in kind.branches) / 2
```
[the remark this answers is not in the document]
HaoTy: `len(kind.branches)` is provably always 2.

tests/computation/conditional_test.py
```diff
@@ -0,0 +1,243 @@
...
+        Cube(
+            Position3D(0, 0, 0),
+            ZXCube.ZXZ,
+            condition=_correlation_surface(Position3D(0, 0, 0), Position3D(0, 0, 1), Basis.Z),
+        )
+
+
+def test_conditional_cube_condition_must_be_in_the_past() -> None:
+    # A condition strictly in the past of the conditional cube is accepted.
+    Cube(
```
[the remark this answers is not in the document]
HaoTy: Added tests.

Zhaoyilunnn:
`fill_port("out", LeafCubeKind.PORT)` leaves a PORT cube in the graph but removes its label from the port index, so subsequent lookup by label fails. This input should be rejected before changing the graph.

```python
g = BlockGraph()
g.add_cube(Position3D(0, 0, 0), "PORT", label="out")
g.fill_port("out", LeafCubeKind.PORT)
print(g.num_ports, g.ordered_ports)  # 1, []
```

Zhaoyilunnn:
@KabirDubey @HaoTy 

I think this PR looks ready for final review and merge as the conditional-cube data-model groundwork, with compilation and runtime integration left to follow-up work. Do you see any remaining blockers, or could we merge it once the latest CI passes?

PR #830, opened 2026-02-06
Refactor the `CorrelationSurface` class to use `Position3D`s instead of vertex IDs

HaoTy:
This change is to prepare for using `CorrelationSurface` as the condition for conditional cubes. 
An additional benefit is that the correlation surface compilation and visualization code has been simplified, because the previously repetitive vertex-to-position logic is not absorbed into the conversion from the internal correlation surface representation. 
A bonus contribution of this PR is the revamped Pauli web conversion, which, as a by-product, fixes a Hadamard-edge bug overlooked in the original implementation.

A downside of using positions is that the correlation surfaces need to be shifted when the block graph is shifted, for example, in `compile_block_graph()`. Since the `CorrelationSurface` class is designed to be immutable, this means the shifted correlation surfaces will be new objects, and the references the user has won't be updated. So far, this doesn't seem to cause any issues. In the long term, the way observables are supplied should be revamped and made more user-friendly, and that's the chance we can make this truly a non-issue.


HaoTy:
This PR is ready for review. Anyone familiar with the touched code, please feel free to take a look.

The branch is based on a previously merged one, so the commit record appears longer than what's actually done in this PR. My apologies! The files changed page is of course correct, so don't be intimidated by the commits 😊

src/tqec/compile/observables/abstract_observable.py
```diff
@@ -16,7 +16,7 @@
 from tqec.utils.position import Direction3D, Position3D
```
smburdick, line 16: @HaoTy This PR is very long, and difficult to navigate comfortably. Can you make a more comprehensive list of what this PR changes?

I agree with the overall approach, since this feels stronger than storing an index. Maybe we could also map the correlation surfaces based on their position, and possibly other characteristics, as a functional ID in contexts where it might be needed.

smburdick (dismissed):

src/tqec/compile/observables/abstract_observable.py
HaoTy: As I mentioned above, this PR is based on a previously merged one, so the commit record appears longer than what's actually done in this PR. What's actually changed is exactly what the title and description say.

inmzhang (dismissed):
Overall looks good to me. One subtle thing is that we may want to also add a test for single node correlation surface/Pauli web round-trip and ensure the correct behavior.

gideonuchehara:
Hi, I'd like to help move this PR forward. I'm reviewing the changes and would be happy to assist with the remaining ask from @inmzhang (adding a round-trip test for single-node correlation surfaces). Could someone confirm this is still the main blocker before I start?

HaoTy:
Somehow I missed the notification of Yiming's review. @gideonuchehara If you want to give it a shot, go ahead. Should be fairly simple to add.

gideonuchehara:
Sure.

HaoTy:
@gideonuchehara Any update or ETA?

gideonuchehara:
> @gideonuchehara Any update or ETA?

Yes, I will be pushing my update soon. Sorry for the delay. Thanks

inmzhang (approved):

Issue #854, opened 2026-03-06
Refactor LayerTree to use detector database in memory


smburdick:
For the sake of testing, we should be able to use the detector database without having to read/write to a file, instead, simply update the database object without performing a write operation. This is because pickle (de)serialization is a large bottleneck in our code base.

PR #853, opened 2026-03-05
Optimize with_mapped_qubit_indices

smburdick:
With additional testing, this showed not to be a huge improvement. I think we can make other changes in the codebase, that will lead to a better speedup.

Here's what I did
* Modify the instruction target list in-place from the copy instead making a second list.
* Cast targets as `Stim.GateTarget` instead of invoking constructor

Per #847 I do still think it's worth circumventing this method altogether, but that would likely require some architectural. I think pre-processing the annotations might be a way to have a single global qubit map.

PR #855, opened 2026-03-07
Improve compile test runtime by memoizing database creation and reducing saving

smburdick:
Address #854 
* In `compile_test.py`, use a single `DetectorDatabase` instance for the entire suite, and wait until the suite is over to write. The suite runs in under 13 minutes now ([previously, 27 minutes](https://github.com/tqec/tqec/pull/853)), assuming the detector database is already populated
* Make writing to the database in `tree.py::_annotate_detectors` optional
* Clean up database initialization logic in `tree.py`


pyproject.toml
```diff
@@ -74,8 +74,11 @@ Issues = "https://github.com/tqec/tqec/issues"
 test = [
   "pytest>=8.4",
   "pytest-cov>=7.0",
+  "pytest-timeout>=2.4.0",
   "types-networkx>=3.5",
   "scipy-stubs>=1.15",
+  "pytest-profiling>=1.8.0",
+  "pytest-timeout>=2.4.0",
```
HaoTy: Duplicate
```suggestion
```

tests/compile/compile_test.py
```diff
@@ -134,12 +137,33 @@ def generate_circuit_and_assert(
...
 
-@pytest.mark.slow
+@pytest.fixture(scope="session")
+def filepath():
+    return _get_database_path()
+
+
+@pytest.fixture(scope="session", autouse=True)
+def detector_db(filepath: Path):
+    if filepath is not None and filepath.exists():
```
HaoTy:
```suggestion
    if filepath.exists():
```
`filepath` is never `None`

tests/compile/compile_test.py
```diff
@@ -134,12 +137,33 @@ def generate_circuit_and_assert(
...
+        return DetectorDatabase.from_file(filepath)
+    else:
+        return DetectorDatabase()
+
+
+@pytest.fixture(scope="session", autouse=True)
+def save_to_db(filepath, detector_db):
+    yield
+    if filepath is not None and filepath.exists():
+        detector_db.to_file(filepath)
```
HaoTy:
```suggestion
    detector_db.to_file(filepath)
```

`filepath` is never `None`, and its directory always exists. The previous implementation fails to save the database when the pickle file doesn't exist.

tests/compile/compile_test.py
```diff
@@ -134,12 +137,33 @@ def generate_circuit_and_assert(
 CONVENTIONS = (FIXED_BULK_CONVENTION, FIXED_BOUNDARY_CONVENTION)
 
 
-@pytest.mark.slow
+@pytest.fixture(scope="session")
+def filepath():
```
HaoTy:
```suggestion
@pytest.fixture(scope="session", name="filepath")
def fixture_filepath():
```

To avoid name shadowing, same for the other two fixtures

src/tqec/compile/tree/tree.py
```diff
@@ -341,26 +346,21 @@ def generate_circuit(
...
             k,
             manhattan_radius,
             detector_database=detector_database,
-            database_path=database_path,
             only_use_database=only_use_database,
             lookback=lookback,
             parallel_process_count=parallel_process_count,
             reschedule_measurements=reschedule_measurements,
+            update_db=update_db,
+            database_path=db_path_input,
```
HaoTy: There's no need to introduce a new `update_db` argument. We can simply set `database_path=None` to achieve `update_db=False` since `database_path` is only used for saving the database in `_generate_annotations` and `_annotate_detectors`.

src/tqec/compile/tree/tree.py
```diff
@@ -341,26 +346,21 @@ def generate_circuit(
...
-        # If we later support efficient parallelism with a populated database,
-        # we will expose the parallel_count parameter to users.
-        parallel_process_count = (
-            cpu_count() // 2 + 1
-            if (detector_database is None or len(detector_database) == 0)
-            else 1
-        )
+            parallel_process_count = 1
+        else:
+            parallel_process_count = cpu_count() // 2 + 1
```
HaoTy: In the case that the database is empty, we also want to parallelize the detector computation.

src/tqec/compile/tree/tree.py
```diff
@@ -298,30 +302,31 @@ def generate_circuit(
...
-        if detector_database is None:  # Nothing passed in,
-            if database_path.exists():  # look for an existing database at the path.
-                detector_database = DetectorDatabase.from_file(database_path)
-            else:  # if there is no existing database, create one.
-                detector_database = DetectorDatabase()
-        # If do_not_use_database is True, override the above code and reset the database to None
-        if do_not_use_database:
-            detector_database = None
-        if detector_database is not None:
+        db_path_input: Path | None = DEFAULT_DETECTOR_DATABASE_PATH
```
HaoTy: `DEFAULT_DETECTOR_DATABASE_PATH` is never `None`.
```suggestion
        db_path_input: Path = DEFAULT_DETECTOR_DATABASE_PATH
```

src/tqec/compile/tree/tree.py
```diff
@@ -298,30 +302,31 @@ def generate_circuit(
                 to be in the same moment. Since each plaquette may have its own measurement
                 schedule, setting this may be necessary for hardware that requires
                 measurements to be synchronous.
+            update_db: whether to write to the detector database upon invocation
```
HaoTy: It is preferable to expand the abbreviations for an argument/variable/method name to be self-explanatory, especially if the name is user-facing. Also to be consistent with other database argument names.
```suggestion
            update_database: whether to save (or update the saved) detector database after it is populated.
```

src/tqec/compile/detectors/database.py
```diff
@@ -298,6 +298,7 @@ def __init__(
         self,
         mapping: dict[_DetectorDatabaseKey, frozenset[Detector]] | None = None,
         frozen: bool = False,
+        path: Path | None = None,
```
HaoTy: What's the goal of introducing this?

HaoTy:
When the detector database is pre-populated, running all tests takes under 2 minutes for me. Looks very promising! Just some minor issues to be fixed.
In the case when there's no detector database, it takes about 8 minutes with the accelerated tqecd. In principle, this should be faster if the num_processes for empty database logic is fixed.

src/tqec/compile/detectors/database.py
smburdick: I was hoping to refactor this logic such that instead of passing around both the detector database *and* the filepath, we just make the filepath an instance variable, so that only the database instance needs to be tracked.

src/tqec/compile/tree/tree.py
smburdick: Okay, I will revert it back. As the comment suggests, I think it's worth exploring opportunities for parallelism with the DB in place.

src/tqec/compile/detectors/database.py
smburdick: Removed it

HaoTy (changes requested):
Sorry, I should have been clearer that `update_database` is still needed for `generate_circuit`, because the user may want to load but not update a stored database. In that case, `database_path` is not `None` and `update_database=False`.
For `_annotate_detectors` and `_generate_annotations`, we can use `database_path=None` to mean `update_database=False`, since `database_path` is only used for dumping. The current implementation looks good, but would you mind reverting the placement of the `database_path` arguments?

Otherwise looks good!

HaoTy:
Ideally, we'd want to unify the database logic of `generate_circuit` and `generate_crumble_url` and handle the loading and dumping either both in the outmost function or both in `_annotate_detectors`. I also don't like pickling, manual database versioning, and IO troubles in general. I think once tqecd is fast enough, we can adopt an alternative solution, such as a globally shared database object computed from scratch each session, at which time we can also explore the best way to parallelize things. But these are out of the scope of this PR.

smburdick:
> would you mind reverting the placement of the database_path arguments?

I'm not sure what you mean. As is, the default argument is the default path, which may not be correctly defined for the user's environment. I think the database path should be `None` by default since if the user isn't providing a path, they aren't worried about updating it.

>  I think once tqecd is fast enough, we can adopt an alternative solution, such as a globally shared database object computed from scratch each session, at which time we can also explore the best way to parallelize things

I haven't worked with tqecd personally, but the point of the detector DB is actually to prevent having to compute detectors from scratch. That's the main reason why running the compilation tests takes more than twice as long the first time you do it.

HaoTy:
> > would you mind reverting the placement of the database_path arguments?
> 
> I'm not sure what you mean. As is, the default argument is the default path, which may not be correctly defined for the user's environment. I think the database path should be `None` by default since if the user isn't providing a path, they aren't worried about updating it.

I don't mean to revert the default value (the default path should always be well-defined, though) but the order of the arguments, i.e., `database_path` was after `detector_database` and it's better to keep things where they were whenever changes are unnecessary.

> 
> > I think once tqecd is fast enough, we can adopt an alternative solution, such as a globally shared database object computed from scratch each session, at which time we can also explore the best way to parallelize things
> 
> I haven't worked with tqecd personally, but the point of the detector DB is actually to prevent having to compute detectors from scratch. That's the main reason why running the compilation tests takes more than twice as long the first time you do it.

Yes, the point of the database is to avoid duplicate computation, which not only comes from repeated calls to circuit generation, but also within a single call, where an arbitrarily large qubit patch is divided into constant-size subtemplates. Now, if tqecd is fast enough that computing all required subtemplates takes about the same time as the IO overhead, then we have good reasons to avoid the pickling.

HaoTy:
Also, when there are multiple circuit generation calls in a single Python program, a globally shared database object in memory avoids multiple pickle loads and dumps.

smburdick:
> I think you missed my point that
> 
> > `update_database` is still needed for `generate_circuit`, because the user may want to load but not update a stored database. In that case, `database_path` is not `None` and `update_database=False`.
> 
> In other words, we want your previous `generate_circuit` with the `update_database` argument (but don't add `update_database` back to `_annotate_detectors` and `_generate_annotations`).

I think we should actually avoid opening a database in `tree.py::generate_circuit` altogether. There are 3 invocations of this method (in `compile_test`, `TopologicalComputationGraph::generate_stim_circuit`, and `LayerTree::generate_crumble_url`). Each of these methods do provide a database argument, but the latter 2 also don't provide a path, which indicates that the db shouldn't be written to.

If I pass a database *and* a path, that should be enough to say, "I would like to write this to a file," otherwise the path argument serves no purpose. What do you think?

HaoTy:
On second thought, I agree with you. We should simplify the horribly complicated logic involved in whether to load/use/compute/store/update a detector database. 

Recalling when I first used tqec, these options confused me a lot and made me question whether I had used the database in the wrong way whenever my circuit was not compiled correctly. Now in reflection, we should not provide so many possibilities to the user, as the user really only cares about correctly compiling and simulating a circuit. The detector database only accelerates compilation and doesn't change the result, so options like `do_not_use_database` and `only_use_database` are almost never useful except for debugging tqec's internal implementation and should not be exposed to the user.

I will merge this PR if you are happy with the current implementation.

smburdick:
@HaoTy Sure, go ahead and merge, I'm still tinkering with the database logic. That probably needs to be its own PR anyway.

PR #857, opened 2026-03-10
Reimplement `with_mapped_qubit_indices` using regex matching

HaoTy:
About an order of magnitude faster across different circuit sizes.
```
Generating moments for distance=11... Done in 0.08 seconds
Mapping 84 original Moments... Done!
  - Average           : 0.14085632801055908
  - Standard deviation: 0.006569241556795203
  - Median            : 0.1394190788269043
Mapping 84 new Moments... Done!
Done!
  - Average           : 0.012575221061706544
  - Standard deviation: 0.00034174860829420305
  - Median            : 0.012546300888061523
Improvements of new method over old:
  - Average           : 0.9107230662667575
  - Median            : 0.9100101579093176
  - Minimum           : 0.9116193113751356
  - Maximum           : 0.9244317002181927
  ```
  
Regex matching is error-prone, but I think the speedup is worth the risk. I have tried to cover all corner cases according to [https://github.com/quantumlib/Stim/blob/main/doc/file_format_stim_circuit.md](https://github.com/quantumlib/Stim/blob/main/doc/file_format_stim_circuit.md) and have added a comprehensive test.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,12 @@
 # How to contribute to the `tqec` library
 
+## Run tests
+
+After downloading the code and setting up your development environment, try to run all the tests using `uv run pytest tests/`. If you're new to the codebase, try examining `compile_test.py::test_compile_memory` to understand the compilation pipeline that TQEC uses.
+
 ## Open issues
 
-If you found a bug or would like a missing feature to be implemented, [create an issue](https://github.com/tqec/tqec/issues/new/choose). In order for your issue to be efficiently treated, please include the appropriate labels.
+If you found a bug, an optimization, or would like a missing feature to be implemented, [create an issue](https://github.com/tqec/tqec/issues/new/choose). In order for your issue to be efficiently treated, please include the appropriate labels.
 
 ## Asking a question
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/circuit/moment_test.py [hunk 1]
```diff
@@ -234,3 +234,30 @@ def test_moment_with_mapped_qubit_indices() -> None:
         "QUBIT_COORDS(0, 0) 12\nQUBIT_COORDS(0, 1) 56\nH 12 56\nDETECTOR(0, 0) rec[-1]"
     )
     assert mapped_moment.qubits_indices == {12, 56}
+    moment = Moment(
```
smburdick, line 236: I think it would be worth having 1-2 more test cases

HaoTy: Added another complex test, which indeed caught a corner case (that would not happen in a TQEC-generated `Moment` circuit).

HaoTy:
> I do feel this regex will essentially be linear, which is a big improvement over our previous O ( n 2 ) approaches. Great to have this.

Thanks, but the previous implementation was also linear, just with a high overhead due to back-and-forth with Stim.

tests/circuit/moment_test.py
smburdick: Good

tests/circuit/moment_test.py
```diff
@@ -234,3 +233,57 @@ def test_moment_with_mapped_qubit_indices() -> None:
...
+            M 4 !5
+            CORRELATED_ERROR(0.1) X6 !Y7*!Z8
+            OBSERVABLE_INCLUDE(1) rec[-2]
+            """
+        )
+    )
+    mapped_moment = moment.with_mapped_qubit_indices(
+        {0: 10, 1: 11, 2: 12, 3: 13, 4: 14, 5: 15, 6: 16, 7: 17, 8: 18}
+    )
+    assert mapped_moment.circuit == stim.Circuit(
```
smburdick, line 279: Let's split this test into 3 separate unit tests.

smburdick: On second thought, it's probably Pythonic to keep it as is, since they all test the same behavior.

HaoTy: The new tests were deliberately created as circuits with various instructions and arguments to test the regex matching in complex situations. In case of an error, it is visually obvious to compare it to the expected outcome and identify which part of the matching went wrong, so there is no need for split unit tests.

PR #858, opened 2026-03-10
Remove do_not_use_database and compilation/simulation-level only_use_database flags

smburdick:
* If a database and a path are provided to LayerTree compilation, that means that they should be writing to the database each time the annotations are generated.
* If only the path is provided, invoke a new database from it.
* Otherwise, do not use a database (and don't write to it, either).
* Remove flags related to database usage. Rework the logic so that if a non-null database is passed through, use it.
* Remove timeouts for test runs with cache misses, or without cacheing


src/tqec/compile/tree/tree.py
```diff
@@ -304,48 +301,8 @@ def generate_circuit(
...
-                        f"running is newer (version {current_version}). The database will be "
-                        "regenerated.",
-                        TQECWarning,
-                    )
-                    detector_database = DetectorDatabase()
-        else:
-            detector_database = None
-
+        if detector_database is None and database_path is not None and database_path.exists():
+            detector_database = DetectorDatabase.from_file(database_path)
```
HaoTy: The case of `detector_database` and `database_path` both being None is not handled. And we definitely want to perform the version checking.

HaoTy (commented):
`LayerTree.generate_circuit` is still a public method of a public class, so we can't assume the input combinations are always valid. In essence, it's not really different from `TopologicalComputationGraph.generate_stim_circuit` (we should unify the naming), as the outmost user-facing method involving the detector database is `start_simulation_using_sinter`. 

My opinion is that either we remove `do_not_use_database` and `only_use_database` from all methods altogether so that we can greatly simplify the internal logic and reduce confusion for users, or we leave the current implementation as is for now. Changing these logics piece-by-piece like this requires meticulous editing and reviewing to ensure all corner cases are covered, and is a waste of effort.

src/tqec/compile/tree/tree.py
smburdick: Right now the database version is hardcoded as `1.0.0`. The problem is, any time the user updates the database (which can happen at any time), the "versioning" logic as it stands here is nil. Instead, I think it would be worth tracking the hash (or a 'snapshot') of the database itself, either of the entire thing, or of a Merkle tree of the individual elements. That way the user can know whether the database has changed, warranting a save.

smburdick:
> `LayerTree.generate_circuit` is still a public method of a public class, so we can't assume the input combinations are always valid. In essence, it's not really different from `TopologicalComputationGraph.generate_stim_circuit` (we should unify the naming), as the outmost user-facing method involving the detector database is `start_simulation_using_sinter`.

`generate_circuit` is probably fine, since we could in principle compile to other languages down the road

> My opinion is that either we remove `do_not_use_database` and `only_use_database` from all methods altogether so that we can greatly simplify the internal logic and reduce confusion for users, or we leave the current implementation as is for now. Changing these logics piece-by-piece like this requires meticulous editing and reviewing to ensure all corner cases are covered, and is a waste of effort.

I vote in favor of the former. It will change a lot of files, yes, but I don't think it will fundamentally be difficult to test, seeing as we already have a large number of tests in place that pass the same database through. In the long run, reducing arguments will make the code easier to maintain.

src/tqec/compile/tree/tree.py
HaoTy: The versioning is for when the definition of the database class (or its dependency classes) has changed in TQEC, not for the content changes.

HaoTy:
> `generate_circuit` is probably fine, since we could in principle compile to other languages down the road

Agreed.

> I vote in favor of the former. It will change a lot of files, yes, but I don't think it will fundamentally be difficult to test, seeing as we already have a large number of tests in place that pass the same database through. In the long run, reducing arguments will make the code easier to maintain.

Great, feel free to start working on it. Or better, wait a little bit and gather others' options in tomorrow's meeting.

src/tqec/compile/tree/tree.py
smburdick: > The versioning is for when the definition of the database class (or its dependency classes) has changed in TQEC, not for the content changes.

Ah, that makes a lot more sense.

smburdick: > The case of detector_database and database_path both being None is not handled.

Actually, in that case, we switch to using parallel processing (providing a speedup as there is no db), and in the detector annotation, it will not write to a file:

In `tree.py::_annotate_detectors`:
```py
        # The database will have been updated inside the above function, and here at
        # the end of the computation we save it to file.
        if detector_database is not None and database_path is not None:
```

HaoTy: I meant that an empty database should be instantiated and passed in.

HaoTy: Oh I see it will be created in `AnnotateDetectorsOnLayerNode`. Never mind.

src/tqec/compile/detectors/compute.py
```diff
@@ -410,19 +377,13 @@ def compute_detectors_at_end_of_situation(
         detectors = database.get_detectors(subtemplates, plaquettes_by_timestep)
         # If not found and only detectors from the database should be used, this
         # is an error.
```
HaoTy, line 412: Be careful with the comments too :)

src/tqec/simulation/generation.py
```diff
@@ -18,8 +18,6 @@ def generate_stim_circuits_with_detectors(
     manhattan_radius: int,
     detector_database: DetectorDatabase | None = None,
     database_path: str | Path = DEFAULT_DETECTOR_DATABASE_PATH,
```
HaoTy:
```suggestion
    database_path: str | Path | None = DEFAULT_DETECTOR_DATABASE_PATH,
```

src/tqec/simulation/generation.py
```diff
@@ -64,10 +62,6 @@ def generate_stim_circuits_with_detectors(
             not specified. If ``detector_database`` is not passed in, the code attempts to
             retrieve the database from this location. The user may pass in the path
             either in str format, or as a Path instance.
```
HaoTy:
```suggestion
            either in str format, or as a Path instance. The computed database will not be
            saved to the disk if this argument is `None`.
```

src/tqec/simulation/generation.py
```diff
@@ -102,8 +94,6 @@ def generate_sinter_tasks(
     manhattan_radius: int,
     detector_database: DetectorDatabase | None = None,
     database_path: str | Path = DEFAULT_DETECTOR_DATABASE_PATH,
```
HaoTy:
```suggestion
    database_path: str | Path | None = DEFAULT_DETECTOR_DATABASE_PATH,
```

src/tqec/simulation/generation.py
```diff
@@ -136,10 +126,6 @@ def generate_sinter_tasks(
             not specified. If ``detector_database`` is not passed in, the code attempts to
             retrieve the database from this location. The user may pass in the path
             either in str format, or as a Path instance.
```
HaoTy:
```suggestion
            either in str format, or as a Path instance. The computed database will not be
            saved to the disk if this argument is `None`.
```

HaoTy (changes requested):
Looks good. There's a `database_path` argument in `start_simulation_using_sinter` that should also be allowed to be `None`. Please edit it and its docstring.

Now with the cluttered logic removed, we can aim to 
1. Unify the detector database logic in generate_circuit and generate_crumble_url methods
2. Handle the saving and loading at the same method level
3. Avoid saving to disk if the database has not been changed.

src/tqec/simulation/generation.py
smburdick: I actually don't think `str` paths should be allowed.

smburdick:
> 3. Avoid saving to disk if the database has not been changed.

This gets back to my earlier comment about saving a hash of the database (both the object and the file), and using that to decide if we need to save to disk.

BSchelpe:
I'm sorry if I missed the part of the meeting where this was discussed, but could you summarise here why you're removing do_not_use_database and only_use_database flags?

smburdick:
> I'm sorry if I missed the part of the meeting where this was discussed, but could you summarise here why you're removing do_not_use_database and only_use_database flags?

The flags aren't actually logically needed. The goal is in cleaning up the API and underlying logic, and serving a longer-term goal of improving how the database works and is being used. If you like, you can run the tests locally to see how the codebase works without them.

[In general, boolean parameters are not good practice.](https://stackoverflow.com/questions/45641284/is-it-bad-practice-to-put-booleans-as-method-parameters)

HaoTy:
> I'm sorry if I missed the part of the meeting where this was discussed, but could you summarise here why you're removing do_not_use_database and only_use_database flags?

In addition to Sam's nice summary above, see also my reasoning from my past experience as a new user [here](https://github.com/tqec/tqec/pull/855#issuecomment-4027482292). Of course, open to discussions if you have different opinions.

src/tqec/simulation/generation.py
HaoTy: Why? It's usually more friendly to the users if they can specify the arguments using built-ins, without importing anything. We can always check and cast the input.

src/tqec/compile/detectors/compute.py
```diff
@@ -336,38 +334,11 @@ def _compute_detectors_at_end_of_situation(
...
-                    get_template_representation_from_instantiation(subtemplate),
-                    "Plaquettes:",
-                    json.dumps(plaquettes.to_name_dict()),
-                )
-            )
-            for subtemplate, plaquettes in zip(subtemplates, plaquettes_by_timestep)
-        )
-    )
-
-
```
HaoTy, line 364: It seems some of my comments in VSCode were not synced to GitHub. In short, I feel this method is useful for internal debugging and should be kept somewhere. Maybe let's keep the `only_use_database` logic in `compute_detectors_at_end_of_situation` for debugging purposes? What do you think?

HaoTy:
Also note my comment above regarding allowing `str`.

src/tqec/simulation/generation.py
smburdick: I guess that was my purist's instinct. I think string paths are okay for the public API, but internally they should be cast to Paths right away.

src/tqec/compile/detectors/compute.py
smburdick: If the user invokes `compute_detectors_at_end_of_situation`, I don't think it's unreasonable to assume, based on the name alone, that if the detectors aren't in the DB yet, they should be. If a database isn't provided, just compute them and return them. Again, going on the theme of simplifying the logic.

HaoTy: The user won't use this argument, because it's removed from the higher-level interfaces. But I'm imagining that in the future development of TQEC, if something's wrong with the detector computation/database, this method would be useful for debugging.

src/tqec/simulation/generation.py
HaoTy: Yeah that makes sense. What level is considered internal is debatable, though.

BSchelpe:
do_not_use_database and only_use_database were both intended as debugging arguments. I agree they are confusing to a first time user. I have not done very much debugging of detector calculations. Are the cases where the flags would be useful rare enough that it is better to remove them for clarity? Or perhaps more importantly can the functionality be very easily coded up adhocly by someone wanting it for debugging purposes in the future?

src/tqec/compile/detectors/compute.py
smburdick: Fair enough, but if you're debugging the database, I think it should be moved and adapted to `database.py`.

smburdick:
> do_not_use_database and only_use_database were both intended as debugging arguments. I I agree they are confusing to a first time user. I have not done very much debugging of detector calculations. Are the cases where the flags would be useful rare enough that it is better to remove them for clarity?

I would argue they are not useful at all. I removed them and didn't really change any of the tests (aside from deleting the removed args and exception handling), and they all passed.

> Or perhaps more importantly can the functionality be very easily coded up adhocly by someone wanting it for debugging purposes in the future?

Debugging usually involves using `pdb` or, in this case, perhaps creating a different kind of `DetectorDatabase`, not changing a public API.

BSchelpe:
> I would argue they are not useful at all. 

Have you checked with @nelimee that he is happy with the functionality of the flags being removed? I think he originally had use cases where he thought they would be useful.

HaoTy:
> do_not_use_database and only_use_database were both intended as debugging arguments. I agree they are confusing to a first time user. I have not done very much debugging of detector calculations. Are the cases where the flags would be useful rare enough that it is better to remove them for clarity? Or perhaps more importantly can the functionality be very easily coded up adhocly by someone wanting it for debugging purposes in the future?

I haven't used them for debugging but can imagine them being handy. My view is that we leave these options in the internal method (e.g., `compute_detectors_at_end_of_situation`), so that they can be set to True when debugging but not exposed to the user. 

I will leave this open for everyone to reach a consensus.

BSchelpe:
>My view is that we leave these options in the internal method (e.g., `compute_detectors_at_end_of_situation`), so that they can be set to True when debugging but not exposed to the user.

That sounds like a good plan to me.

smburdick:
> My view is that we leave these options in the internal method (e.g., compute_detectors_at_end_of_situation), so that they can be set to True when debugging but not exposed to the user.

The only thing `compute_detectors_at_end_of_situation` did as it pertained to the flag `only_use_database` is that if the flag is set to true, and the provided database has no detectors for the given sub-templates and plaquettes, it throws an error. The client of `compute_detectors_at_end_of_situation` already has the ability to compute the detectors from the arguments to the method, so this flag serves no real purpose. If the plan is to keep it false for all of our own internal test cases, and it is false by default, why keep it in place?

Also, the method doesn't do anything special that would require this flag for any debugging. If the actual database logic, or the private `_compute_detectors_at_end_of_situation` method (which is far more complex) needs to be debugged, then `compute_detectors_at_end_of_situation` is not the right place to do that.

HaoTy:
> The only thing `compute_detectors_at_end_of_situation` did as it pertained to the flag `only_use_database` is that if the flag is set to true, and the provided database has no detectors for the given sub-templates and plaquettes, it throws an error. 

Exactly, it identifies which sub-templates and plaquettes are missing from the database and saves them—for debugging (or rather, to find the starting point). I don't have enough experience to say how useful this is in practice; perhaps it's best if we consult someone with that experience.

smburdick:
> > The only thing `compute_detectors_at_end_of_situation` did as it pertained to the flag `only_use_database` is that if the flag is set to true, and the provided database has no detectors for the given sub-templates and plaquettes, it throws an error.
> 
> Exactly, it identifies which sub-templates and plaquettes are missing from the database and saves them—for debugging (or rather, to find the starting point). I don't have enough experience to say how useful this is in practice; perhaps it's best if we consult someone with that experience.

Okay, by popular demand, I restored the flag.

KabirDubey:
It's possible that removing the boolean `do_not_use_database` flag closes #724 and #681, leaving a note here for me to check this. 

I am invested in debugging detector annotations but I also was unaware of this  `compute_detectors_at_end_of_situation` method. I'll keep it in mind.

smburdick:
> It's possible that removing the boolean `do_not_use_database` flag closes #724 and #681, leaving a note here for me to check this.
> 
> I am invested in debugging detector annotations but I also was unaware of this `compute_detectors_at_end_of_situation` method. I'll keep it in mind.

As a mac user, I haven't had any problems with the database myself.

smburdick:
Going to disable timeouts in a different PR if needed.

HaoTy:
Is everyone okay with the current solution?

smburdick:
> Is everyone okay with the current solution?

I believe @BSchelpe and possibly @KabirDubey had concerns about this

KabirDubey:
I wanted to see if this resolved my earlier database issues, but if it's blocking smth and looks good to you then go ahead and merge. I'll try to take a look this weekend, will prioritize.

BSchelpe:
>My view is that we leave these options in the internal method (e.g., `compute_detectors_at_end_of_situation`), so that they can be set to True when debugging but not exposed to the user.

Is this the current solution? If so, I am happy with it.

smburdick:
> > My view is that we leave these options in the internal method (e.g., `compute_detectors_at_end_of_situation`), so that they can be set to True when debugging but not exposed to the user.
> 
> Is this the current solution? If so, I am happy with it.

`compute_detectors_at_end_of_situation` is unchanged by this PR

BSchelpe:
So is the PR still currently just removing those options, or are they still there on a different internal method?

smburdick:
> So is the PR still currently just removing those options, or are they still there on a different internal method?

It's more about simplifying the invocation of LayerTree (making it less bug prone) and its downstream invocations, by removing the extra booleans. You can still use them in some of the methods like the aforementioned compute_detectors_at_end_of_situation, though.

BSchelpe:
> > So is the PR still currently just removing those options, or are they still there on a different internal method?
> 
> It's more about simplifying the invocation of LayerTree (making it less bug prone) and its downstream invocations, by removing the extra booleans. You can still use them in some of the methods like the aforementioned compute_detectors_at_end_of_situation, though.

Apologies for the delay. I missed that you had restored the only_use_database flag in compute_detectors_at_end_of_situation. And do_not_use_database is covered by sending in database=None, so I'm happy with the PR as it stands 👍

src/tqec/compile/tree/tree.py
```diff
@@ -286,12 +280,8 @@ def generate_circuit(
                 ``database_path``.
             database_path: specify where to save to after the calculation.
                 This defaults to :data:`.DEFAULT_DETECTOR_DATABASE_PATH` if
-                not specified. If detector_database is not passed in, the code attempts to
+                not specified. If detector_database None and the code attempts to
```
HaoTy: Typo here

HaoTy (changes requested):
Apart from the typo, LGTM.

HaoTy:
Come on, can you actually read the sentence and see if it makes sense (and is true to the argument's functionality)?

smburdick:
> Come on, can you actually read the sentence and see if it makes sense (and is true to the argument's functionality)?

I apologize, was context switching between too many things. I'll check it again tomorrow. Thanks.

smburdick:
> Come on, can you actually read the sentence and see if it makes sense (and is true to the argument's functionality)?

The docstring should be ok now.

HaoTy (approved):
LGTM

PR #864, opened 2026-03-13
Preliminary bgraph import implementation

jbolns:
Quick import feature to enable import of `.bgraph` files produced by both Topologiq and TopoLS. 
- Example `.bgraph` with standardised format added to assets files. 
- New interops folder/file with related import feature.
- Some changes to other import files/folders to re-use existing functions. 

Supersedes #720 with a muuuuuch lighter approach.


src/tqec/interop/shared.py
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                )
+        except Exception as e:
+            raise TQECError(f"Error repositioning pipes from parsed data: {e}.")
+
+        # Build and return blockgraph
+        try:
+            block_graph = BlockGraph.from_dict(blockgraph_dict)
+        except Exception as e:
+            raise TQECError(f"Error creating blockgraph from parsed data: {e}.")
```
HaoTy: Why do we need another layer of dict format and `BlockGraph.from_dict` instead of directly using `BlockGraph.add_cube` and `BlockGraph.add_pipe`? If it's for reusing the Collada IO code, I don't think it's necessary.

src/tqec/interop/shared.py
```diff
@@ -0,0 +1,191 @@
...
+class LoadFromAnywhere(ABC):
+    """ABC template for subclasses that create :py:class:`~.BlockGraph` from external source."""
+
+    _instance = None
+
+    def __new__(cls):
+        """Instantiate ABC class if/when ever called first."""
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
```
HaoTy: Why the singleton behavior here? In fact, we don't need a class for a stateless loader. `BlockGraph` uses separate `from_xxx` methods anyway. 

It's also not wise to require all loaders to conform to the `parse()` method and the specified dict format, because
1. Using the native `BlockGraph.add_cube` and `BlockGraph.add_pipe` is preferred whenever possible, and
2. We want to avoid having yet another loader for yet another file format.

So really, the loaders can be implemented as plain methods, and should be able to choose whether to leverage `parse()` or not.

tests/interop/shared_test.py
```diff
@@ -0,0 +1,56 @@
...
+    with pytest.raises(TypeError) as e:
+        parse_method = getattr(LoadFromAnywhere, "parse")
+        _ = parse_method(filepath=filepath)
+    assert "LoadFromAnywhere" in str(e.value)
+
+    # Concrete method inside ABC class also won't trigger despite being fully defined
+    with pytest.raises(TypeError) as e:
+        load_method = getattr(LoadFromAnywhere, "load")
+        _ = load_method(filepath=filepath)
+    assert "LoadFromAnywhere" in str(e.value)
```
HaoTy: Not really necessary.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,129 @@
...
+- kind: The kind of the cube.
+- label (optional): An optional annotation that is typically used to denote when a cube is a PORT.
+
+Each CUBE item should be given as a CSV-separated sequence.
+
+::
+    CUBES: index;x;y;z;kind;label;
+    4;0;0;0;zxz;;
+    3;3;0;0;xxz;;
+    5;0;0;3;zxx;;
```
HaoTy: Give an example of an open port.

How to represent a Y half cube?

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,129 @@
...
+- kind: The kind of the cube.
+- label (optional): An optional annotation that is typically used to denote when a cube is a PORT.
+
+Each CUBE item should be given as a CSV-separated sequence.
+
+::
+    PIPES: src;tgt;kind;
+    4;3;oxz;
+    4;5;zxo;
+    4;6;oxz;
```
HaoTy, line 94: Specify the Hadamard like in `PipeKind.from_str`.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,129 @@
...
+~~~~~
+Contains the information needed to translate each line of this section into a :code:`tqec` `pipe <./terminology.rst>`.
+
+The information in this section is meant to be parsed and must contain, at a minimum:
+- index: the ID of the cube, typically an integer.
+- position: The (x, y, z) position of the cube.
+- kind: The kind of the cube.
+- label (optional): An optional annotation that is typically used to denote when a cube is a PORT.
+
+Each CUBE item should be given as a CSV-separated sequence.
```
HaoTy: Forgot to change the description for pipes.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,129 @@
...
+        circuit_name; CNOTs;
+
+CUBES
+~~~~~
+Contains the information needed to translate each line of this section into a :code:`tqec` `cube <./terminology.rst>`.
+
+The information in this section is meant to be parsed and must contain, at a minimum:
+- index: the ID of the cube, typically an integer.
+- position: The (x, y, z) position of the cube.
+- kind: The kind of the cube.
```
HaoTy: Specify what is allowed, or reference an existing doc page.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,129 @@
...
+- name of the circuit: used to give the blockgraph a reference name.
+
+Additional fields possible, although not necessarily parsed by default by TQEC.
+
+Each METADATA item should be given as a CSV-separated pair.
+
+::
+    METADATA: attr_name; value;
+        source; name_of_tool_that_created_the_file;
+        pipe_length; 2.0;
```
HaoTy, line 46: Pipe length is a choice made at visualization, not an inherent property of the lattice surgery computation, so should not be here.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,129 @@
+BGRAPH
+========================
+
+BGRAPH is a work-in-progress file format (`.bgraph`) that can be used to facilitate interoperability between external lattice surgery (LS) tools.
+
+While only a preliminary minimum baseline in need of improvement, the current BGRAPH specification already enjoys the following advantages:
+- Common format both `Topologiq <https://github.com/pycollada/pycollada>` and `TopoLS <https://github.com/pycollada/pycollada>` can produce
```
HaoTy: Forgot to change links.

src/tqec/computation/block_graph.py
```diff
@@ -827,6 +827,19 @@ def from_dict(data: dict[str, Any]) -> BlockGraph:
             )
         return graph
 
+    @staticmethod
+    def from_bgraph(
+        bgraph_str: str | None = None,
+        filepath: str | pathlib.Path | None = None,
+        graph_name: str | None = None,
+    ) -> BlockGraph:
```
HaoTy, line 835: Also implement a `to_bgraph`.

src/tqec/interop/bgraph/read.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # Meta
+        graph_name_match = re.search(r"(?<=circuit_name; )(.*\b)", bgraph_str)
+        pipe_length_match = re.search(r"(?<=pipe_length; )(.*\b)", bgraph_str)
+        graph_name = graph_name_match.group(0) if graph_name_match else "circuit"
+        pipe_length = float(pipe_length_match.group(0)) if pipe_length_match else 0.0
+
+        # Find all cubes and pipes in `.bgraph`
+        cube_matches = re.finditer(r"(?<=\n)(?:\-*\d*;){3,}.*", bgraph_str)
+        pipe_matches = re.finditer(r"(?<=\n)(?:\d*;){2}\w{3};", bgraph_str)
```
HaoTy, line 80: Is this robust when the circuit name or port name contains the keywords?

HaoTy (changes requested):

tests/interop/shared_test.py
jbolns: OK

src/tqec/interop/shared.py
jbolns: OK

src/tqec/interop/shared.py
jbolns: Removed instantiation.

docs/user_guide/bgraph.rst
jbolns: OK. Added reference links and clarifying notes to RST.

docs/user_guide/bgraph.rst
jbolns: OK. Added reference links and clarifying notes to RST.

docs/user_guide/bgraph.rst
jbolns: Lol. Yup.

docs/user_guide/bgraph.rst
jbolns: OK. Added reference links and clarifying notes to RST.

docs/user_guide/bgraph.rst
jbolns: Clarified optionality of specific metadata fields. 
If not given, it defaults to 0.0, i.e., TQEC. 
Needs to stay there, though. It's a choice for visualisation for TQEC. In SketchUp, pipes are real. Topologiq might eventually get rid of them, but for now it also uses them to guide placements.

docs/user_guide/bgraph.rst
jbolns: Yikes.

src/tqec/computation/block_graph.py
jbolns: This I didn't do. I have a function in Topologiq that does it, so of course can be done. 

But is there an use case for it?

I see why TQEC would want to take as many formats as it can, but I can't really say it needs to produce all formats it takes. Truth be said, the only format TQEC will really need to produce at some point is whatever is used for databasing blockgraphs. This may or may not be .bgraph, but we're not at that bridge just yet.

src/tqec/interop/bgraph/read.py
jbolns: Yes. Can't take credit for it. Didn't even think about it. But I just checked and it is, thanks to the ";". 
I suppose one could break it if one tries, by adding the keyword and a semi-colon, but that would be pushing the boundaries of reason.

jbolns:
Thanks @HaoTy . 
Most comments addressed exactly as suggested, and those that weren't were justified.
Should at the very least reduce the number of open matters so we can focus on any remaining cleavage.

src/tqec/computation/block_graph.py
HaoTy: Yes, for storing block graphs from TQEC in a text-based format. Believe it or not, many users at this moment build block graphs directly (programmatically) in TQEC, without using Topologiq or SketchUp. But I'm okay with adding it later.

jbolns: OK, let's pin this for last after ensuring no other cleavages are pending.

HaoTy: Sorry, was a little busy yesterday preparing the correlation surface compilation tutorial.

This can be added (optionally by someone else) after this PR is merged.

docs/user_guide/bgraph.rst
HaoTy: I'm okay with having optional fields in the METADATA, but it's not a choice for visualization only for TQEC. It has nothing to do with the actual computation. If your algorithm depends on the pipe length, you should get rid of this dependency.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,137 @@
...
+    CUBES: index;x;y;z;kind;label;
+    4;0;0;0;zxz;;
+    3;3;0;0;xxz;;
+    0;6;0;0;ooo;in_0;
+    8;-6;0;0;ooo;out_0;
+
+
+.. note::
+
+    There is no fully-agreed denomination for open boundaries/ports or Y-cubes. For the time being, it is possible to use "ooo" or "P" for ports and any arbitrary string containing a "y" to denote a Y-cube.
```
HaoTy: How to denote the placement of the Y cube? It can be in either half of a full cube.
Also would be good to think about how to represent an arbitrarily long T cultivation block. We don't have it yet, but it's necessary for universal computation.

HaoTy: I'd also suggest not allowing an "arbitrary string containing a 'y'". That's too error-prone. A typo can easily lead to an unnoticed wrong computation.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,137 @@
...
+::
+    METADATA: attr_name; value;
+        source; name_of_tool_that_created_the_file;
+        pipe_length; 2.0;
+        circuit_name; CNOTs;
+
+
+.. note::
+
+    All pipes must measure exactly the same. In TQEC, cubes are immediately adjacent, which means `pipe_length` is 0.0. However, this is a TQEC convention. Other software can and do use longer pipes.*
```
HaoTy: Factually incorrect as noted above.

docs/user_guide/bgraph.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Schema
+-------
+A BGRAPH file can be conceived as a QASM-like representation of a lattice surgery.
+
+Any BGRAPH file should be divided into four main sections:
+- HEADER: The title of the document, with information about the source tool used to produce it.
+- METADATA: A short section with meta information on top
+- CUBES: A subsequent section with detailed information about the cubes in the blockgraph
+- PIPES: A final section with detailed information about the cubes in the blockgraph.
```
HaoTy: "about the *pipes"

Please check the preview of the doc page: https://tqec.github.io/tqec/pull/864/user_guide/bgraph.html. There are quite a few typos and syntax issues that cause links, references to other doc pages, and newlines to not render.

src/tqec/interop/bgraph/read.py
```diff
@@ -0,0 +1,135 @@
+"""Read block graphs contained in a lattice surgery (LS) `.bgraph` file."""
```
HaoTy: Missing `__init__.py` in this submodule.

src/tqec/interop/bgraph/read.py
HaoTy: Pipe regex doesn't match the additional Hadamard character.

In general, regex matching the whole file is error-prone. Better to add comprehensive tests that try to cover all edge cases.

src/tqec/interop/bgraph/read.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        # Validate input types
+        if io_str or input_in_other_format:
+            raise TQECError("The parsing method is currently only for `.bgraph` files.")
+
+        if not raw_str and not filepath:
+            raise TQECError("LoadFromBgraph requires a `.bgraph` string or filepath.")
+
+        # Read file
+        bgraph_str: str = raw_str if raw_str else ""
```
HaoTy: Make `raw_str` and `filepath` mutually exclusive. My preference is to leverage duck typing or have two separate methods.

src/tqec/interop/shared.py
```diff
@@ -0,0 +1,177 @@
...
+from tqec.computation.block_graph import BlockGraph, block_kind_from_str
+from tqec.computation.cube import YHalfCube
+from tqec.utils.exceptions import TQECError
+from tqec.utils.position import FloatPosition3D, Position3D
+from tqec.utils.scale import round_or_fail
+
+
+# ABC TEMPLATES
+class LoadFromAnywhere(ABC):
+    """ABC template for subclasses that create :py:class:`~.BlockGraph` from external source."""
```
HaoTy: What I mean is to not have loader classes at all. They are stateless, i.e., don't store anything, and we have established that there's no reason to enforce the loader to use the `parse` interface (which has no enforcement since the output dict is not typed), so there's no reason to have classes. Avoiding them also reduces the unnecessary input arguments for unsupported loaders.

HaoTy: Then you can directly use `add_pipe` and `add_cube` to build a `BlockGraph` as you parse a `bgraph`, instead of parsing them to another dict and using `from_dict`.

src/tqec/interop/bgraph/read.py
```diff
@@ -0,0 +1,135 @@
...
+        if filepath:
+            with open(filepath) as f:
+                bgraph_str = f.read()
+                f.close()
```
HaoTy: No need to close inside `with`.

src/tqec/interop/bgraph/read.py
```diff
@@ -0,0 +1,135 @@
...
+            raise TQECError("Error parsing cubes from `.bgraph` file.")
+
+        # Pipes
+        parsed_pipes: dict[tuple[int, int], dict[str, tuple[int, int, int] | str]] = {}
+        try:
+            for match in pipe_matches:
+                src_id, tgt_id, kind, _ = match.group(0).strip().split(";")
+                parsed_pipes[(int(src_id), int(tgt_id))] = {"kind": kind.upper()}
+        except (ValueError, TypeError, IndexError, KeyError):
+            raise TQECError("Error parsing pipes from `.bgraph` file.")
```
HaoTy: raise from

src/tqec/interop/shared.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Some tools do not allow saving name explicitly so parser needs a default
+        if override_graph_name:
+            parsed_data["name"] = override_graph_name
+
+        # Build blockgraph
+        block_graph = BlockGraph(parsed_data["name"])
+
+        # Add cubes
+        pipe_length = parsed_data["pipe_length"] if "pipe_length" in parsed_data else 0.0
```
HaoTy: `parsed_data.get("pipe_length", 0.0)`

HaoTy (changes requested):

docs/user_guide/bgraph.rst
jbolns: Concerning placement and string. Not pushback. An actual question. Is this a visualisation matter, i.e., people will sometimes use half and sometimes full, or is there really a situation where the same circuit will contain full and half blocks?

Concerning only string. Specific suggestions? In Topologiq, I'll probably call them "yoy". So, I don't need such degree of flexibility. It's now completely free because I don't really know the extent to which particularities might arise in other contexts.

docs/user_guide/bgraph.rst
jbolns: OK, then let's resolve this one because it is optional.

docs/user_guide/bgraph.rst
jbolns: Will, but real problem here is that I couldn't quickly figure out how to produce this preview locally as to review iteratively. Any tips?

src/tqec/interop/bgraph/read.py
jbolns: yup. pushes in next commit

src/tqec/interop/bgraph/read.py
jbolns: OK. Mutual exclusiveness pushes on next commit.

src/tqec/interop/shared.py
jbolns: Say I agree with your comment just above, why to even use classes at all?

Therefore, I feel the logical conclusion is to revert back to a simple import / export function based approach that is additional to those we already have? 

I know you don't like diversity of inputs, but it is also not in itself a huge problem. Numerous software have many in/out methods to cater to various audiences. They do just fine.

And with a simpler import / export, once the export is dealt with, maybe we can talk about getting rid of JSON? It would require a separate issue re-implementing the SketchUp plugin to use BGRAPH, but with an alternative in place, there might be someone willing to take that up.

src/tqec/interop/bgraph/read.py
jbolns: Pushes on next commit.

src/tqec/interop/shared.py
jbolns: pushes on next commit

src/tqec/interop/bgraph/read.py
jbolns: OK

src/tqec/interop/bgraph/read.py
jbolns: ok. pushes on next commit. also changed similar raises in shared.py to match

src/tqec/interop/bgraph/read.py
jbolns: Variety of strings is very limited, so I think ReGex can do just fine. 
But will adjust to cover for any possible kind and will add a more complex test. 
Will not add a Y cube. None of the input tests include a Y cube yet. Best to wait until that is implemented fully to add across.

jbolns:
All fixes except the big one ready to push. Let me know what to do about classes and I'll adapt code accordingly. 
My recommendation is to simply use functions, as originally intended. If you make a Protocol in the future, you can point to your preferred import method. I don't see a point for using classes at all if there's no ABC class.

HaoTy:
> All fixes except the big one ready to push. Let me know what to do about classes and I'll adapt code accordingly. My recommendation is to simply use functions, as originally intended. If you make a Protocol in the future, you can point to your preferred import method. I don't see a point for using classes at all if there's no ABC class.

I see we had a misunderstanding. The Protocol I suggested was not for importing text-based files but for integrating TQEC and upstream compilers natively with Python, without relying on files. For file import and export, if your original intention is to simply use functions, then we are on the same page here.

docs/user_guide/bgraph.rst
HaoTy: I think you may need to build the docs locally. Maybe see how it's done in the GitHub workflow?

jbolns:
Right, excellent! I'll migrate to functions tomorrow. Since it reduces complexity, I might then be able to include both import and export being that there's applicable write function in topologiq. 

Ps. I got a request to include raw_string in the importer for a preliminary level of interoperability. That will need to stay. I can't imagine it being a blocker for better integration. Or, perhaps said differently, a minimal expectation one can reasonably have for a broader / more holistic integration method, approach, or whatnot is that it beats the efficiency of passing a string through a function.

docs/user_guide/bgraph.rst
HaoTy: The Y-basis initialization/measurement only takes d/2 depth, i.e., half a block temporal-wise, so in principle, you can run the subsequent computation directly after the half-cube. This can be useful when we want to minimize the idling in the computation. The d-unit abstraction in `BlockGraph` makes this (and the walking surface code and cultivation) a bit hard to support.

Nonetheless, at least there will be two versions for the Y half cube, located at the bottom and top half of a cube, respectively. We should distinguish them in `bgraph`. Maybe "yb" and "yt", utilizing the cube kind field? Just a preliminary proposal, open to discussion.

docs/user_guide/bgraph.rst
jbolns: Takes sooo long. 
Got this sorted out with something called restview. Won't generate the jupyter bits, but it seems to work for everything else. 
No worries, deleted from dependencies after use. I won't change anything in project files.

docs/user_guide/bgraph.rst
HaoTy: To be closer to the nature, the two versions should be distinguished by facing up vs facing down. So perhaps "yu" and "yd"?

src/tqec/interop/bgraph/read.py
HaoTy: Okay to not test the Y cube. The edge cases I have in mind are the free-form fields containing space, symbols (except ";"), and all kinds of keywords

jbolns:
OK, now using functions. 

Might still need some work, but I think it covers all points you've made, including:
- Reduced usage of RegEx (mainly because I didn't know TQEC doesn't actually have clear IDs for cubes and parsing what it has is a whole other level of RegEx)
- I tried to do the duck taping (lol) thing (but since I've never really done that, not entirely confident about it).

But let me know @HaoTy .

docs/user_guide/bgraph.rst
HaoTy:
```suggestion
- PIPES: A final section with detailed information about the pipes in the blockgraph.
```

src/tqec/interop/bgraph/read_write.py
```diff
@@ -0,0 +1,260 @@
...
+            if "Y" in kind.upper():
+                if kind.upper() in ["Y", "YI", "YM"]:
+                    kind = "Y"
+                    if isinstance(block_kind_from_str(kind), YHalfCube):
+                        position = int_position_before_scale(
+                            offset_y_cube_position(
+                                FloatPosition3D(*(int(x), int(y), int(z))), pipe_length
+                            ),
+                            pipe_length,
+                        )
```
HaoTy, line 58: Ideally, we should always make the kind of Y explicit, so a single "Y" shouldn't be allowed. But at this moment, main doesn't distinguish between Y initialization and measurement, and there's the problem of how to represent two Y half cubes in the same cube (if that's possible) in TQEC, since `BlockGraph` keys every cube with an integer-valued `Position3D` instance. I made some attempts on this, hoping this PR can get to use the right TQEC representations of Y, but decided it's better to address this in coordination with #548 and #829. I will try to make some progress there; in the meantime, I think what's in the PR is acceptable.

src/tqec/interop/bgraph/read_write.py
```diff
@@ -0,0 +1,260 @@
...
+    bgraph_lines.append(f"circuit_name; {graph_name};\n")
+
+    # Write cubes into lines array
+    bgraph_lines.append("\nCUBES: index;x;y;z;kind;label;\n")
+    for cube in block_graph.cubes:
+        scaled_position = scale_position(cube.position, pipe_length=pipe_length)
+        if cube.is_y_cube and block_graph.has_pipe_between(
+            cube.position, cube.position.shift_by(dz=1)
+        ):
+            scaled_position = scaled_position.shift_by(dz=0.5)
```
HaoTy: I see why you need the pipe length; it's for representing the cube coordinates as in 3D model files. However, in the actual computation, pipes do not take any spacetime volume, i.e., they are non-existent, and we should keep the `.bgraph` cube coordinates true to their positions in the computation, like what TQEC `BlockGraph` does. In this case, pipe length never needs to be stored and is only needed when you want to visualize things. Nothing needs to be scaled. In the Collada `.dae` case, pipe length is unavoidable, as the file is a 3D model, which is a visualization. For `.bgraph`, however, we have control over the fields, and should respect the actual computation.

docs/user_guide/bgraph.rst
```diff
@@ -134,4 +138,34 @@ Additionally, you can attach the correlation surface to the model to visualize w
...
-And, of course, you can follow the same instructions available in several other docs to produce circuits and simulate the blockgraph.
+And, of course, you can follow the same instructions available in several gallery docs (for instance, this `Steane code example <https://tqec.github.io/tqec/gallery/steane_encoding.html>`_) to produce circuits and simulate the blockgraph.
+
+
+Producing a BGRAPH
+------------------
+
+It is also possible to produce a BGRAPH of any TQEC blockgraph.
+
+In the example below, we print the BGRAPH string instead of saving it to file, but the same procedure can be used to save to a file with :code:`.bgraph` extension by removing :code:`save_to_file=False` or changing it to :code:`save_to_file=True`.
```
HaoTy: Why don't we make `path_to_output_file` optional and get rid of `save_to_file`? i.e., when `path_to_output_file=None`, we don't save the file.

tests/interop/bgraph/read_write_test.py
```diff
@@ -0,0 +1,48 @@
...
+    # Write to string
+    bgraph_out_str = write_bgraph(
+        graph,
+        assets_folder / "dummy.bgraph",
+        pipe_length=2.0,
+        graph_name=graph_name,
+        save_to_file=False,
+    )
+
+    # Re-load and compare (direct string comparison not possible: IDs change)
```
HaoTy: This behavior is ok but undesirable. `BlockGraph` doesn't have cube IDs (corresponding ZX graph does have vertex IDs). Without refactoring how `BlockGraph` stores things, we can get rid of cube IDs in `.bgraph` and represent pipes with positions too. But I feel for human-readable files, IDs are sometimes useful. Perhaps it's best to keep things as is for the moment.

tests/interop/bgraph/read_write_test.py
```diff
@@ -0,0 +1,48 @@
...
+        _ = load_bgraph(input)
+
+
+@pytest.mark.parametrize("test_type", ["filepath", "raw_str", "str_of_filepath"])
+def test_bgraph_load_write(test_type: str) -> None:
+
+    # Parse from assets file
+    graph_name = "cnots"
+    assets_folder = Path(__file__).parent.parent.parent.parent / "assets"
+    filepath = assets_folder / f"{graph_name}.bgraph"
```
HaoTy, line 25: We should definitely have more tests for edge cases. They don't have to be `.bgraph` files; the strings can be hardcoded in the test files.

HaoTy: To clarify, we should preserve the current `.bgraph` file, as at least one file-read test on a non-`to_bgraph`-generated `.bgraph` file is needed. And it's good to have an example.

HaoTy (changes requested):
The changes are quite positive. Almost there!

afowler:
I'm pretty sure the y half cube is actually slightly more than half to
achieve full distance so two don't fit into the same space. Provided we
restrict a Y cube to have exactly one up or down connection, additional
specification shouldn't be necessary.

src/tqec/interop/bgraph/read_write.py
jbolns: Not sure where this landed given the comment below by @afowler. If my maths are correct, the v in current PR works at least until the dust settles on how Y will be implemented in TQEC?

src/tqec/interop/bgraph/read_write.py
jbolns: OK, I did the following to try and bridge the distance. 
When not explicitly given, it is not written to bgraph string or file. When given explicitly, i.e., pipes have a length for whatever reason that might happen, then there's a need to write it. Since one doesn't control if others use or do not use pipes (fine, that's a little bit of a cop out being that I do have some control over that, but is applicable in the abstract), the reader should try and find it, defaulting to 0.0 if not there.

docs/user_guide/bgraph.rst
jbolns: Done, I think.

tests/interop/bgraph/read_write_test.py
jbolns: My immediate acerbic comeback is, IDs hurt no one.
More generally, if one works on the basis that the BGRAPH might be produced by one tool, imported by another, and then exported again, one cannot assume the IDs will be handled equally. 
Having said that, I adapted the export so that it uses positional tuples when writing. 
So, in summary, on the way in, it's all a string so as long as the string is unique, all should work, and on the way out it writes position tuples as IDs. A tiny bit redundant, but potentially more in line with the principle that the position in the blockgraph is the ID.

tests/interop/bgraph/read_write_test.py
jbolns: Not against this. But to not go back and forth with trial and error, what are edge cases?
Is the primary worry that invalid BGRAPHs will make it through, or that valid BGRAPHS will be rejected?
- In theory, an invalid BGRAPH would be rejected at build even if it makes it through: invalid blocks or pipes.
- What kind of variety would one need to test valid BGRAPHs are not rejected?

jbolns:
Except for the tests, I think everything else is covered in last couple commits. @HaoTy 
Didn't click re-review because I agree more tests are possible, but would welcome more detail as requested in the applicable comment.

src/tqec/interop/bgraph/read_write.py
HaoTy: Yes.

src/tqec/interop/bgraph/read_write.py
HaoTy: Not "use or do not use pipes", pipes do not occupy spacetime volume in the actual computation, so should not  take any space in `.bgraph`, a truthful representation of the computation, as well. I feel that if we allow the pipes to optionally take space in `.bgraph`, it will cause confusion and misunderstandings, just like what we had. Perhaps @afowler can rule things out here?

tests/interop/bgraph/read_write_test.py
HaoTy: There's also the possibility of a valid graph getting imported as a different, but valid, graph.
Mostly I want to test valid graphs that have `bgraph` keywords in the free-form fields. I think you can feed the `.bgraph` specification to an agent and let it generate a comprehensive test suite. This also serves as a verification of the clarity of the specification. For example, I don't think it currently specifies that ";" is not allowed in a circuit name.

jbolns: OK, I understand the concern better now. Will try and get it done tomorrow.

src/tqec/interop/bgraph/read_write.py
jbolns: Yes, in my book, @afowler always has the last word. It does seem important to clarify the alternatives.

If the case is that having the line in the file risks confusion.
- If I remove the possibility of an optional field for `pipe_length`, one needs an additional subroutine to infer pipe lengths from the distances between the cubes. Seems an unnecessary source of complexity. 

If the case is that no tool should consider giving lengths to pipes or produce blockgraphs where pipes have an explicit or explicit length:
- In such case, "let's merge and close this as an imperfect interoperability format" and then start discussing how to improve BGRAPH so that it can one day perhaps become an industry standard seems preferable to "let's keep this PR open until we have an industry standard".

afowler:
It's good to have to option of an adjustable pipe/connector length.
Definitely useful to be able to reduce the pipe length in displays where
you want to show a structure closer to what actually runs on the hardware.

tests/interop/bgraph/read_write_test.py
HaoTy: This is a nice idea. I wonder if we can leverage `to_zx_graph` at export and assign the correct ZX vertex IDs to cubes. Would be useful for debugging whenever the corresponding ZX graph is involved.

src/tqec/interop/bgraph/read_write.py
HaoTy: Ok, I will accept the current state then. But seems you still don't get my point -

> * If I remove the possibility of an optional field for `pipe_length`, one needs an additional subroutine to infer pipe lengths from the distances between the cubes. Seems an unnecessary source of complexity.

No need to infer the pipe length because it's always 0. Two (computationally) adjacent cubes are always adjacent coordinate-wise. If a tool concerns the visualization of the computation, it can choose whatever pipe length it deems appropriate for the visualization.

> * In such case, "let's merge and close this as an imperfect interoperability format" and then start discussing how to improve BGRAPH so that it can one day perhaps become an industry standard seems preferable to "let's keep this PR open until we have an industry standard".

IMO, quite the opposite, having the option of a non-zero pipe length makes the file format flawed, because it may not represent the actual computation anymore, and one has to convert between two coordinate standards. A file format for the block graph should stay true to the computation it is tasked to represent, not serving as a 3D model/visualization file.

afowler:
When it comes to displaying block graphs with different pipe lengths, I
agree that this should be a global choice at display time and the length
doesn't need to be encoded in every pipe.

tests/interop/bgraph/read_write_test.py
jbolns: Latest v. adds tests. I took your advice and asked an AI to gauge clarity and suggest potential strings that could have unintended consequences. It said it's quite alright an explanation and suggested a few strings that are included in tests.

tests/interop/bgraph/read_write_test.py
jbolns: That's exactly how I used them in Topologiq. If TQEC ever gets a way to specify IDs, you need only to take the IDs given in the blockgraph and they'll match the input ZX graph (any not in the graph would be extra due to the need to negotiate the 3D space).

src/tqec/interop/bgraph/read_write.py
jbolns: I get your point, mate. I typically do. We just happen to have very different views. But that's alright. At least on my end. I actually kind of like you. 

Concerning the pipe_length. I removed it from write_bgraph. It will never be in the BGRAPH is the file is produced by TQEC. It will never be in the BGRAPH is the file is produced by another tool that effectively has adjacent cubes.

jbolns:
@afowler 
Latest v. removes the optional declaration from write_bgraph, meaning no BGRAPH ever created by TQEC will have a pipe_length. It maintains the ability to take BGRAPH with pipe_lengths, which is necessary for interoperability. This may or may not change in the future, but that's in the future.

src/tqec/interop/bgraph/read_write.py
HaoTy: Then we must have very different ways of thinking. I'm okay with the current compromise.

tests/interop/bgraph/read_write_test.py
HaoTy: I'm referring to the ZX graph that is mapped from the block graph. From my understanding, it is equivalent to the original ZX graph but typically not the same. Does Topologiq not add any additional cubes that are originally not in the ZX graph?

jbolns: (1) If cube ID in the input ZX graph, cube can be conceived as materialising the spider in 3D. (2) If cube ID not in the input ZX graph, cube is the result of the LS. Additionally, and I might ask you about this later, if you blockgraph -> positioned ZX - full reduce then in theory the reduced version should match the reduction of the original ZX input. 

But here, I'm actually not understanding you, because I'm not sure what the underlying question is.

src/tqec/computation/block_graph.py
```diff
@@ -827,6 +827,28 @@ def from_dict(data: dict[str, Any]) -> BlockGraph:
             )
         return graph
 
+    def to_bgraph(
+        self,
+        filepath: str | pathlib.Path | None = None,
+        graph_name: str = "circuit",
+    ) -> str:
+        """Construct a block graph from a dictionary representation."""
```
HaoTy: Copy-pasted.

src/tqec/computation/block_graph.py
```diff
@@ -827,6 +827,28 @@ def from_dict(data: dict[str, Any]) -> BlockGraph:
...
+        bgraph_str = write_bgraph(self, filepath, graph_name=graph_name)
+
+        return bgraph_str
+
+    @staticmethod
+    def from_bgraph(
+        bgraph_str_or_filepath: str | pathlib.Path,
+        graph_name: str = "",
+    ) -> BlockGraph:
+        """Construct a block graph from a dictionary representation."""
```
HaoTy: Copy-pasted.

src/tqec/computation/block_graph.py
```diff
@@ -827,6 +827,28 @@ def from_dict(data: dict[str, Any]) -> BlockGraph:
             )
         return graph
 
+    def to_bgraph(
+        self,
+        filepath: str | pathlib.Path | None = None,
+        graph_name: str = "circuit",
```
HaoTy, line 833: Use `BlockGraph`'s name?

src/tqec/interop/bgraph/read_write.py
```diff
@@ -0,0 +1,259 @@
...
+    # Reject if input string remains empty
+    if bgraph_str_or_path == "":
+        raise TQECError("Empty BGRAPH string or filepath.")
+
+    # Looks like a regular string with BGRAPH syntax
+    bgraph_str = ""
+    if (
+        not isinstance(bgraph_str_or_path, Path)
+        and bgraph_str_or_path.startswith("BLOCKGRAPH ")
+        and ".bgraph" not in bgraph_str_or_path
```
HaoTy: I can have a directory whose name starts with "BLOCKGRAPH " and this step would fail. Can we use `Path` to check if the string is a valid file path first?

HaoTy: In the `.bgraph` file, I can also call my circuit `xxx.bgraph`, and the third check would fail its purpose. Both scenarios are very unlikely, but still possible.

src/tqec/interop/bgraph/read_write.py
```diff
@@ -0,0 +1,259 @@
...
+    graph_name: str = "circuit",
+) -> str:
+    """Write a :filetype:`.bgraph` from a :class:`.BlockGraph`.
+
+    Args:
+        block_graph: The blockgraph to be written into a BGRAPH file.
+        filepath: The output file path or file-like object that supports binary write.
+        pipe_length (optional): The length of pipes to use for distancing cubes out.
+        graph_name (optional): Name to give the BGRAPH file and add to BGRAPH metadata.
+        save_to_file (optional): Flag to trigger file writing (BGRAPH is always returned as string).
```
HaoTy: Need to be updated.

tests/interop/bgraph/read_write_test.py
HaoTy: Can we change the circuit names and cube names to something more dangerous, such as "pipe_length", "circuit_name"?

HaoTy: Also the regex uses `\b`, so it's worth checking if circuit names with spaces and punctuation work correctly. Especially at the beginning or the end of the circuit name.

tests/interop/bgraph/read_write_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Load
+    if test_type == "filepath":
+        graph = load_bgraph(filepath)
+    elif test_type == "str_of_filepath":
+        graph = load_bgraph(str(filepath))
+    else:
+        with open(filepath) as f:
+            bgraph_str = f.read()
+            f.close()
```
HaoTy: Another unnecessary close here.

tests/interop/bgraph/read_write_test.py
HaoTy: 1. I just wonder, if my understanding is correct that Topologiq does add cubes without a one-to-one correspondence with the original ZX diagram vertices, what is the cube ID?
2. Can we use the vertex IDs in `BlockGraph.to_zx_graph().g` as cube IDs in the TQEC writer?

HaoTy (changes requested):
LGTM otherwise.

src/tqec/computation/block_graph.py
jbolns: Fixed. In last PR.

src/tqec/computation/block_graph.py
jbolns: Fixed. In last PR.

src/tqec/computation/block_graph.py
jbolns: Added something similar. It uses graph name if explicit name not given. Maybe I want to save several versions of a blockgraph with different name.

src/tqec/interop/bgraph/read_write.py
jbolns: Now checking for Path first, with some additional checks to prompt user explicitly if a corresponding file is not found. 
Also explicitly banned any BGRAPH using the word "bgraph" , with corresponding error and error message.

src/tqec/interop/bgraph/read_write.py
jbolns: Done.

tests/interop/bgraph/read_write_test.py
jbolns: Removed.

jbolns:
@HaoTy I think last PR addresses all pending 

Can't comment on several seemingly-still-relevant points because they're tagged as outdated, so a couple related notes below.

**String tests**
Added a few additional string tests to consider all possibilities noted.

I'm NOT saying I have thought of all possible situations under the sun. However, none of the concerns noted worry me because the ReGeX checks are sufficiently lenient as to, worse case scenario, give you an ugly name.

So, for instance, punctuation:
- full punctuated name would be picked up if punctuation happens between characters. 
- if punctuation happens at the end, the parsed name would be everything that comes before it
- if the name given is quite literally only punctuation, the default "circuit" would kick in. 

Additionally, cube names will not go through unless they're specifically made of valid characters because even if they're accidentally parsed (unlikely), they'd be rejected at construction. 

**IDs in Topologiq**
I have been meaning to ask you about correlation surface finding and, more generally, comparing correlation surfaces as means to check ZX graph equivalence. Maybe we can add IDs to the conversation whenever we do that. 

Meanwhile, I think it would be more helpful to think of Topologiq in terms of edges, not spiders/cubes.

The raw logic in Topologiq ensures that all *edges* in the ZX graph are honoured (some related changes came in recently, but we can ignore them for the purpose of discussing IDs).
- Where possible, Topologiq maintains a one-to-one correspondence: Z <-> Z becomes Z-cube <-> Z-cube.
- Where not possible, it maintains a one-to-one correspondence by ensuring target and source for edge have a one-to-one correspondent AND adds intermediate cubes to make the connection possible in 3D.

As a result, 100% of the spiders in the input ZX graph will have a one-to-one correspondent, assigned the same ID as in the input ZX. Additionally, the blockgraph will also have additional intermediate cubes in between some (and not all) of the original source/targets, and the IDs for these intermediate cubes follow a `max_id + 1` rationale.

HaoTy (approved):
My personal principle on names is to stay true to what is given, even if the name is an empty string or just punctuation. This applies to both reading and writing. Sometimes, missing punctuation changes the semantic meaning. For example, if a circuit is to simulate a molecule that ends with "+", with and without the "+" mean different things.

But not a deal-breaker for now. Maybe I will make some edits myself later, which is probably more efficient than us going back and forth.

Issue #860, opened 2026-03-11
Remove slow test marks from formerly slow tests


smburdick:
With speedups to our compilation test suite in #857 and #855 , many of the "slow" tests now run in a matter of seconds.

* Decide on a threshold for what constitutes a "slow" test.
* Remove slow tags from tests that run below the threshold.

smburdick:
Based on the [test duration data](https://pastebin.com/1FXvMumF), I feel like we can mark these tests as slow:

`test_compile_steane_encoding` (>= 3.78s)
`test_compile_three_cnots` (>= 2.51s)

PR #870, opened 2026-03-16
Revise slow marker placement

smburdick:
Address #860 
* Nonslow tests take <1s, or 40s total, on M1 mac.


jbolns:
Remind me. The slow marker meant that they did not run by default? and, with this PR, they're being converted into tests that always run?

smburdick:
> Remind me. The slow marker meant that they did not run by default?

Correct, the not-slow tests are what run in our [CI pipeline](https://github.com/smburdick/tqec/blob/44b36a8ad6e2a6b5d9d35c130f3fab820f696684/.github/workflows/ci.yml#L56).

> and, with this PR, they're being converted into tests that always run?

Yes, since they run much faster now, I think they should be part of CI again. I actually put them there recently, and we realized that went too far. With the performance improvements, they should be safe to run now.

PR #852, opened 2026-03-01
Raise an error for block graphs without deterministic observables

HaoTy:
Adds a clarifying error message when the block graph doesn't have any deterministic observables. Addresses #850.

Issue #850, opened 2026-02-27
Valid block graph fails correlation surface finding


KabirDubey:
Consider the block graph pictured [here](https://github.com/tqec/tqec/pull/719#issuecomment-3969660376).
The cube in the middle is ZXZ, it's connected to opposite nodes along one spatial axis but like nodes along the other.

<details>
<summary><b>block_graph_4.json</b></summary>

```
{
  "name": "Block Graph #4",
  "cubes": [
    {
      "position": [
        12,
        1,
        0
      ],
      "kind": "ZZX",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    },
    {
      "position": [
        13,
        0,
        0
      ],
      "kind": "ZXX",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    },
    {
      "position": [
        13,
        0,
        1
      ],
      "kind": "Y",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    },
    {
      "position": [
        13,
        1,
        0
      ],
      "kind": "ZZX",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    },
    {
      "position": [
        13,
        2,
        0
      ],
      "kind": "ZXX",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    },
    {
      "position": [
        13,
        2,
        1
      ],
      "kind": "Y",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    },
    {
      "position": [
        14,
        1,
        0
      ],
      "kind": "XZX",
      "label": "",
      "transform": [
        [
          1,
          0,
          0
        ],
        [
          0,
          1,
          0
        ],
        [
          0,
          0,
          1
        ]
      ]
    }
  ],
  "pipes": [
    {
      "u": [
        12,
        1,
        0
[...]

Issue #863, opened 2026-03-12
Documentation building fails with invalid database path


smburdick:
[Build](https://github.com/tqec/tqec/actions/runs/23027365699/job/66878184106?pr=858)

I'm not sure why the documentation build step is running code like this. I can work around it for now, but, it's worth investigating.

PR #881, opened 2026-03-27
Updating correlation surface docs for nondeterminism

KabirDubey:
I ran some test block graphs with nondeterministic observables and confirmed that the error message added in `fix/clarify-no-deterministic-observable` was triggered. I wanted to add these test cases into this PR, but then I noticed that might cause some issues with PR #830, which updates the `tests/computation/correlation_test.py` file. I now feel familiar enough with that area of the code to review that PR so I will take a look and leave a comment with the test cases that I think should be included. 

I updated the terminology section of the documentation to describe how the correlation surface concept relates to other terms like detectors, detecting regions, Pauli frames, and Pauli webs. Around the end, I explain our desired handling for nondeterministic observables based on the discussion in issue #850. I think this is useful to document just so everyone is on the same page. 

Please review and then I can merge both this PR and PR #852. 

I had some hesitation with this PR because I was not sure if I agreed with the statement that [An X initialization followed by a Z measurement doesn't violate the pipe logic](https://github.com/tqec/tqec/issues/850#issuecomment-3981038096). Such a block graph may pass the _structural_ validations of [fig 9 in the LaSsynth paper](https://arxiv.org/pdf/2404.18369), but it doesn't pass the _functional_ validations in fig 11.  Although the correlation-surface-finding algorithm is fast, it could be inefficient to annotate the detectors and look for the correlation surfaces for a computation that never had any deterministic observables to begin with. This is good enough for now though, I will make a separate issue to discuss further.


KabirDubey:
Having a tough time debugging the failing GitHub actions. Fresh venv set up and dev dependencies installed but `make html` in `docs` directory failing:

```
Traceback
=========

      File "/Users/kabirdubey/.pyenv/versions/.tqec-venv/lib/python3.12/site-packages/nbsphinx/__init__.py", line 662, in parse
        raise NotebookError(
    nbsphinx.NotebookError: PandocMissing in gallery/cnot.ipynb:
    Pandoc wasn't found.
    Please check that pandoc is installed:
    https://pandoc.org/installing.html


The full traceback has been saved in:
/var/folders/j8/wb4qbgyj31l2f745vj9jk5sm0000gn/T/sphinx-err-t8eh06c5.log
```

This is probably due to [this pandoc thing](https://stackoverflow.com/questions/62398231/building-docs-fails-due-to-missing-pandoc) which we've already linked in the dev installation docs but none of the `pip` options in the stackexchange post worked for me. Will have to try conda later? Please let me know if you use Mac OS and have a way to avoid these annoyances. Probably related to #863.

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
 
 You can think of constructing the correlation surface as moving a line of logical operators through the structure,
 only allowing the logical operators to attach to walls with the same basis.
-The physical qubit measurements and stabilizer measurements in the correlation surface determine the sign relationship between the logical operators at the input and output.
+The physical qubit measurements and stabilizer measurements in the correlation surface determine the sign relationship between the logical operators at the input and output. Any required Pauli operator corrections are tracked in a classical data structure called a Pauli frame. The tracking does not delay the circuit, unless there is an operation which needs the correct Pauli frame in real time.
+
+
+Related concepts
```
smburdick, line 196: You need to have an equal number of `+` signs underneath this, it is breaking the doc build

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
 
 You can think of constructing the correlation surface as moving a line of logical operators through the structure,
 only allowing the logical operators to attach to walls with the same basis.
-The physical qubit measurements and stabilizer measurements in the correlation surface determine the sign relationship between the logical operators at the input and output.
+The physical qubit measurements and stabilizer measurements in the correlation surface determine the sign relationship between the logical operators at the input and output. Any required Pauli operator corrections are tracked in a classical data structure called a Pauli frame. The tracking does not delay the circuit, unless there is an operation which needs the correct Pauli frame in real time.
```
smburdick: Might want to give the Pauli frame its own page/section

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+The exception are computations involving :math:`T` gates. Although all of the measurements associated with the :math:`T` gate teleportation would be random, the random results signify whether an :math:`S` gate correction is needed or not, and therefore must observed with confidence and responded to with the appropriate classical feedback. As of March 2026, the simulation of nondeterministic observables that decide the conditional measurement basis for Pauli corrections are not supported by ``tqec``.
```
smburdick: Remove "As of March 2026," and change the last sentence to the active voice

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+Two quantum computations are logically equivalent if they both implement the same set of correlation surfaces. The process of rewriting a block graph into one that has the same correlation surfaces is essentially a topological deformation because maintaining the connectivity between logical blocks is necessary and sufficient for preserving the list of correlation surfaces. The stabilizer ZX calculus is a mathematically rigorous diagrammatic language for reasoning about these deformations; correlation surfaces roughly correspond to what the ZX calculus research community calls Pauli webs. The correspondence between ``tqec`` block graphs and ZX graphs is sufficiently accurate for ``tqec`` to use ZX graphs as an intermediate representation, but one may find subtle differences in terms depending on the class of ZX graphs one is analyzing.
```
smburdick: * add a comma: "... is essentially a topological deformation, because ..."
* " ... correspond to Pauli webs in the ZX calculus."

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+It is possible for a block graph to support a logical observable that is non-deterministic. This occurs when the detectors which support the logical observable, despite potentially maintaining its fault tolerance, do not have deterministic parity in the absence of errors. In other words, the data provided by the observable is random. For example, consider a surface code patch initialized in the :math:`Z` basis and then measured in the :math:`X` basis. Tracing the :math:`X` observable back to a :math:`Z` initialization would specify a totally random event. Generally speaking, this makes it difficult to tell whether the computation was fault tolerant or not because there is no deterministic value that could serve as a ground truth. For this reason, the ``tqec`` compiler avoids tracing correlation surfaces corresponding to non-deterministic observables, and raises an error when no deterministic correlation surfaces are found.
```
smburdick: "was fault tolerant or not, because"

smburdick (changes requested):
* Need to address the docs breaking.
* As for the actual content, aside from minor grammatical points, it might be worth defining some more terms or otherwise linking to newly introduced concepts (eg, Pauli frames, ZX calculus, observables)

smburdick:
Also be sure to coordinate with @HaoTy as you are merging into his branch

docs/user_guide/terminology.rst
KabirDubey: I'm just going to delete this line because we are actively trying to support nondeterministic observables that determine the conditional measurement basis for Pauli corrections.

docs/user_guide/terminology.rst
KabirDubey: New section added

docs/user_guide/terminology.rst
KabirDubey: Dealing with this stuff is kind of annoying, is there an automated way to enforce these formatting requirements?

KabirDubey:
> Having a tough time debugging the failing GitHub actions. Fresh venv set up and dev dependencies installed but `make html` in `docs` directory failing:
> 
> ```
> Traceback
> =========
> 
>       File "/Users/kabirdubey/.pyenv/versions/.tqec-venv/lib/python3.12/site-packages/nbsphinx/__init__.py", line 662, in parse
>         raise NotebookError(
>     nbsphinx.NotebookError: PandocMissing in gallery/cnot.ipynb:
>     Pandoc wasn't found.
>     Please check that pandoc is installed:
>     https://pandoc.org/installing.html
> 
> 
> The full traceback has been saved in:
> /var/folders/j8/wb4qbgyj31l2f745vj9jk5sm0000gn/T/sphinx-err-t8eh06c5.log
> ```
> 
> This is probably due to [this pandoc thing](https://stackoverflow.com/questions/62398231/building-docs-fails-due-to-missing-pandoc) which we've already linked in the dev installation docs but none of the `pip` options in the stackexchange post worked for me. Will have to try conda later? Please let me know if you use Mac OS and have a way to avoid these annoyances. Probably related to #863.

Resolved.

KabirDubey:
Thank you, Sam!

> it might be worth defining some more terms or otherwise linking to newly introduced concepts (eg, Pauli frames, ZX calculus, observables)

I created a new section for Pauli frames. I agree that it would be nice to have more thorough explanations and also citations to guide some additional reading for newcomers but I'd like to prioritize other `tqec` work right now. My primary goal here is to document the distinction between deterministic and nondeterministic observables.

KabirDubey:
> Also be sure to coordinate with @HaoTy as you are merging into his branch

@HaoTy please review and if you approve I can merge both our branches (or you can merge your branch and I can merge mine to main, idk how it's supposed to work, don't have a preference).

docs/user_guide/terminology.rst
smburdick: I don't know if there is. I would look into reStructuredText and Sphinx since that's what we use.

smburdick (dismissed):
LGTM. I created a new issue #892 to track developments in the docs.

Waiting on @HaoTy 's update for merge.

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
 
 You can think of constructing the correlation surface as moving a line of logical operators through the structure,
 only allowing the logical operators to attach to walls with the same basis.
-The physical qubit measurements and stabilizer measurements in the correlation surface determine the sign relationship between the logical operators at the input and output.
+The physical qubit measurements and stabilizer measurements in the correlation surface determine the sign relationship between the logical operators at the input and output. Any required Pauli operator corrections are tracked in a classical data structure called the :ref:`Pauli frame <pauli_frame>`. The tracking does not delay the circuit, unless there is an operation which needs the correct Pauli frame in real time.
```
HaoTy: Correlation surfaces don't produce Pauli corrections. The explanations themselves are correct though, maybe move to the Pauli frame section?

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+
+Related concepts
+~~~~~~~~~~~~~~~~
+
+A set of measurements with predictable parity in the absence of errors is called a detector. The detectors which inform the Pauli frame of a logical observable constitute a correlation surface. The detecting regions highlighted in Crumble and annotated in Stim are a labeling of the spacetime stabilizers manifested by detectors at a physical circuit level.
```
HaoTy: Correlation surfaces can not be viewed as a set of detectors.
```suggestion
A set of measurements with predictable parity in the absence of errors is called a detector. The detecting regions highlighted in Crumble and annotated in Stim are a labeling of the spacetime stabilizers manifested by detectors at a physical circuit level.
```

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+Two quantum computations are logically equivalent if they both implement the same set of correlation surfaces. The process of rewriting a block graph into one that has the same correlation surfaces is essentially a topological deformation, because maintaining the connectivity between logical blocks is necessary and sufficient for preserving the list of correlation surfaces. The stabilizer ZX calculus is a mathematically rigorous diagrammatic language for reasoning about these deformations; correlation surfaces roughly correspond to Pauli webs in the ZX calculus. The correspondence between ``tqec`` block graphs and ZX graphs is sufficiently accurate for ``tqec`` to use ZX graphs as an intermediate representation, but one may find subtle differences in terms depending on the class of ZX graphs one is analyzing.
```
HaoTy: Two Clifford quantum computations are logically equivalent if they both implement the same set of Pauli operator maps / stabilizer flows from inputs to outputs. Of course, correlation surfaces tell you this relationship, so the first sentence is also correct. 

Maintaining connectivity alone is not sufficient; the boundary basis matches (the color rules) also need to be preserved.

> stabilizer ZX calculus

I think it's formally called the phase-free ZX calculus? Not sure, could be a slightly different thing.

> ZX graph

Formally ZX diagram.

> The correspondence between ``tqec`` block graphs and ZX graphs

We can cite https://arxiv.org/abs/1704.08670 if we want.

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+In particular, since post-selection is not a physically-realistic substitute for measurement, the ``tqec`` compiler views measurement operations that are not explicitly paired with a classical feedforward instruction by the user as either :math:`T` gates or a "discard" indicating that the measurement outcome has no effect on the program.
```
HaoTy: Pardon my ignorance, where does this come from?

docs/user_guide/terminology.rst
```diff
@@ -194,7 +194,21 @@ Tracking the process of logical operator movement above, we can get the followin
...
+
+It is possible for a block graph to support a logical observable that is non-deterministic. This occurs when the detectors which support the logical observable, despite potentially maintaining its fault tolerance, do not have deterministic parity in the absence of errors. In other words, the data provided by the observable is random. For example, consider a surface code patch initialized in the :math:`Z` basis and then measured in the :math:`X` basis. Tracing the :math:`X` observable back to a :math:`Z` initialization would specify a totally random event. Generally speaking, this makes it difficult to tell whether the computation was fault tolerant or not, because there is no deterministic value that could serve as a ground truth. For this reason, the ``tqec`` compiler avoids tracing correlation surfaces corresponding to non-deterministic observables, and raises an error when no deterministic correlation surfaces are found.
```
HaoTy:
```suggestion
It is possible for a block graph to support a logical observable that is non-deterministic. This occurs when the detectors which support the logical observable, despite potentially maintaining its fault tolerance, do not have deterministic parity in the absence of errors. In other words, the data provided by the observable is random. For example, consider a surface code patch initialized in the :math:`Z` basis and then measured in the :math:`X` basis. Tracing the :math:`X` observable back to a :math:`Z` initialization would specify a totally random event. Generally speaking, this prohibits Stim's compiled-sampler-like simulators from estimating the logical error rate, because there is no deterministic value that could serve as a ground truth. For this reason, for now, the ``tqec`` compiler avoids tracing correlation surfaces corresponding to non-deterministic observables, and raises an error when no deterministic correlation surfaces are found. Nonetheless, non-deterministic correlation surfaces can be simulated with less efficient simulators and appear in hardware executions, so ``tqec`` plans to support them in the future.
```

docs/user_guide/terminology.rst
```diff
@@ -360,11 +374,28 @@ The examples below utilize the ``Z`` shape.
...
+.. _pauli_frame:
+
+Pauli Frame
+-----------
+
+The Pauli frame is a classical data structure that, in each execution of a quantum program,
+stores the effect of the Pauli operations that were determined to be necessary by the decoder
+and program specification. If there were no logical errors during circuit execution, the decoding
+system updates the Pauli frame with the value of the Pauli operator that corrects the error which
+the decoding algorithm outputs as the most likely explanation for each detection event. A detection
```
HaoTy:
```suggestion
and program specification. The decoding system updates the Pauli frame with the
value of the Pauli operator that corrects the error which the decoding algorithm outputs
as the most likely explanation for the observed set of detection events. A detection
```
Most decoders look at the full syndrome. The absence of the logical error is not a condition that prevents the decoder from updating the Pauli frame but rather a consequence of the decoder making a wrong guess.

docs/user_guide/terminology.rst
```diff
@@ -360,11 +374,28 @@ The examples below utilize the ``Z`` shape.
...
 --------
 
-In the ``tqec`` library, a detector is a set of one or more measurements that are
-supposed to have a deterministic parity in the absence of errors.
+A detector is a set of one or more measurements that are
+supposed to have a deterministic parity in the absence of errors. :footcite:`McEwen_2023`
+
+.. _pauli_frame:
+
+Pauli Frame
```
HaoTy, line 391: Can cite https://www.nature.com/articles/nature03350 for Pauli frame, if we want.

HaoTy (changes requested):

docs/user_guide/terminology.rst
inmzhang:
```suggestion
It is possible for a block graph to support a logical observable that is non-deterministic. This occurs when the measurements which support the logical observable do not have deterministic parity, even in the absence of errors. For example, consider a surface code patch initialized in the :math:`Z` basis and then measured in the :math:`X` basis. Tracing the :math:`X` observable back to a :math:`Z` initialization would specify a totally random event. Generally speaking, this makes it difficult to tell whether the computation was fault tolerant or not, because there is no deterministic value that could serve as a ground truth. For this reason, the ``tqec`` compiler avoids tracing correlation surfaces corresponding to non-deterministic observables, and raises an error when no deterministic correlation surfaces are found.
```

docs/user_guide/terminology.rst
inmzhang:
```suggestion
Two Clifford quantum computations are logically equivalent if they both implement the same set of correlation surfaces. The process of rewriting a block graph into one that has the same correlation surfaces is essentially a topological deformation, because maintaining the connectivity between logical blocks is necessary and sufficient for preserving the list of correlation surfaces. The stabilizer ZX calculus is a mathematically rigorous diagrammatic language for reasoning about these deformations; correlation surfaces roughly correspond to Pauli webs in the ZX calculus. The correspondence between ``tqec`` block graphs and ZX graphs is sufficiently accurate for ``tqec`` to use ZX graphs as an intermediate representation, but one may find subtle differences in terms depending on the class of ZX graphs one is analyzing.
```

Should emphasize Clifford computation. When non-Clifford gate comes into play, two computation might be not equivalent even they have the same correlation surfaces.

afowler:
Do you have a small example to illustrate your comment?

inmzhang:
> Do you have a small example to illustrate your comment?

The minimal example in my mind is just the T teleportation. It depends on how we compute correlation surface, if we treat the T state as an open port during correlation surface computation then we can replace the T state with arbitrary state while getting exactly the same correlation surface set. In my mind the correlation surface computation process is non-Clifford agnostic, correct me if I'm wrong.

afowler:
Thanks, now I understand your comment. Yes, when you have non Clifford
states the correlation surfaces alone won't tell you the output probability
distribution, you also need to know what the non Clifford states are.

smburdick:
@KabirDubey are you still interesting in having this merged? If so, please bring it up-to-date, and address reviewer comments.

KabirDubey:
> @KabirDubey are you still interesting in having this merged? If so, please bring it up-to-date, and address reviewer comments.

Yes, will do today, apologies for the delay, thanks for the reminder.

docs/user_guide/terminology.rst
KabirDubey: Moved to Pauli frame section and changed correlation surface section to be about producing a PPM rather than a Pauli correction.

docs/user_guide/terminology.rst
KabirDubey: suggested change applied

docs/user_guide/terminology.rst
KabirDubey: > Maintaining connectivity alone is not sufficient; the boundary basis matches (the color rules) also need to be preserved.

Ended up deleting the sentence. I think block graph rewrite semantics will require a separate section of documentation. 

> I think it's formally called the phase-free ZX calculus? Not sure, could be a slightly different thing.

Okay I looked into this. Although the phase free ZX calculus captures the surface code (and all CSS codes), this is only in the sense of their encoding maps. Our diagrams include T state initialization and Y basis initialization and measurement which are nodes of phase pi/4 and pi/2 respectively. My new writing states that blocks graphs are an instantiation of the Clifford+T fragment, meaning all nodes are integer multiples of pi/4.

Although Clifford+T fragment is the most accurate classification, this fragment seems to lack less provable guarantees than the stabilizer fragment (which is just integer multiples of pi/2 nodes, (excludes T nodes)). In practice, ``tqec`` uses the rewrite rules of the stabilizer ZX calculus. Because of the equivalences in this [this paper](https://arxiv.org/pdf/2603.09580), specifically section 4, I don't expect our introduction of pi/4 nodes to break anything, especially because they will always be in the form of a teleportation gadget whose flow propagation rules are known and able-to-be-defined independently of the ZX calculus.  

> Formally ZX diagram.

Added an explicit definition of a ZX graph. 

> We can cite https://arxiv.org/abs/1704.08670 if we want.

Done. Also added a bunch of other citations from the ZX calculus community, mainly because I want readers interested in formal support for the IR to refer to those results.

> Should emphasize Clifford computation. When non-Clifford gate comes into play, two computation might be not equivalent even they have the same correlation surfaces.

Fixed, also included Austin's clarification below.

docs/user_guide/terminology.rst
KabirDubey: The motivation for writing this came from some confusion I experienced when reviewing David's `pyzx` parser. I have added some explicit references that came up during that conversation which cleared my doubts. 

Post-selection formally means something stronger and less physical than measurement and feedforward. It means you measure the qubit, and pretend somehow that you only keep the branch where the measurement result was the one you wanted. That is not a realistic replacement for actual measurement control, because in hardware you cannot force the desired outcome; you can only observe the outcome and adapt, or discard failed runs.

The key point I a want to add to the docs is that measurement-dependent corrections must be explicit. Otherwise, the compiler assumes the measurement has no program-level classical consequence, rather than silently assuming an idealized post-selected branch. 

I expect this terminology and logic to not change even after we build proper control system infrastructure to ``tqec``.

docs/user_guide/terminology.rst
KabirDubey: Exact change applied.

docs/user_guide/terminology.rst
KabirDubey: Thanks, exact change applied.

docs/user_guide/terminology.rst
KabirDubey: Done

docs/user_guide/terminology.rst
KabirDubey: Exact change applied.

KabirDubey:
@smburdick @HaoTy @inmzhang @afowler this PR is ready for re-review. Once again apologies for the delay--needed some time to ponder these concepts and read papers. I responded to each open comment line by line and left the decision to resolve up to the reviewer. 

As an additional minor change, I gave cell-ids to each jupyter notebook executable--this allows me to configure sphinx to skip running code when building the docs. Otherwise I must wait 30 minutes for these cells to run after `make html`. This is too long to check if your PR which doesn't modify the gallery examples will pass the docs build tests. I'll make a note on #981 to look into this further and document a good workflow in the contributing guide.

docs/user_guide/terminology.rst
```diff
@@ -360,12 +382,24 @@ The examples below utilize the ``Z`` shape.
...
+supposed to have a deterministic parity in the absence of errors :footcite:`McEwen_2023`.
+
+.. _pauli_frame:
+
+Pauli Frame
+-----------
+
+The Pauli frame is a classical data structure that, in each execution of a quantum program,
+stores the effect of the Pauli operations that were determined to be necessary by the decoder
+and program specification :footcite:`Knill_2005`. The tracking does not delay the circuit, unless there is an operation which needs the correct Pauli frame in real time. If there were no logical errors during circuit execution, the decoding system updates the Pauli frame with the value of the Pauli operator that corrects the error which the decoding algorithm outputs as the most likely explanation for the observed set of detection events. A detection event is a deviation in the expected parity of a :ref:`detector <detector>`.
```
HaoTy: > If there were no logical errors during circuit execution

Seems unnecessary? Logical errors are unknown in real execution, and the Pauli frame will be updated with the decoder's prediction regardless.

HaoTy (dismissed):
Except for the minor Pauli frame point, LGTM!

afowler:
Note that propagating a physical Pauli frame is not necessary. We only care
about the parity of measurements on correlation surfaces and this parity is
directly flipped when decoding rather than unnecessarily inserting Pauli
corrections.

Best,
Austin.

KabirDubey:
Copilot and Pauli frame comments addressed

docs/user_guide/terminology.rst
```diff
@@ -360,11 +378,22 @@ The examples below utilize the ``Z`` shape.
...
+supposed to have a deterministic parity in the absence of errors :footcite:`McEwen_2023`.
+
+.. _pauli_frame:
+
+Pauli Frame
+-----------
+
+The Pauli frame is a classical data structure that, in each execution of a quantum program,
+stores the effect of the Pauli operations that were determined to be necessary by the decoder
+and program specification :footcite:`Knill_2005`. The Pauli correction is given by the parity of measurements on :ref:`correlation surfaces <correlation_surface>`. The parity is directly flipped when decoding. The tracking only delays the circuit if an operation which needs the correct Pauli frame is scheduled in a blocking manner.
```
HaoTy, line 396: Sorry, my previous comment was inaccurate. Even with windowed decoding, the boundary condition being passed is not exactly a physical Pauli frame. Fortunately Austin clarified this (that we don't need a physical Pauli frame). As for the logical Pauli frame, I don't think it's needed if there's no Pauli in the logical circuit (or we apply them physically instead of virtually) and we track correlation surfaces instead of byproduct operators. That said, if you do have a logical Pauli frame, I believe it should be applied after computing the correlation surface parity.

afowler:
You also don't need a logical Pauli frame. This is difficult to even define
when you have a tangled structure. You do have observables, namely the
correlation surfaces you are going to take the parity of when reporting a
measurement.

Imagine a memory experiment with a single Z observable, if you have a
single qubit circuit and it calls for a bit flip you would flip a bit
associated with the observable that tracks the net effect of such
operations.

Similar to a logical Pauli frame, but focused on observables.

Best,
Austin.

PR #901, opened 2026-04-09
Add pull request template

smburdick:
* Provide a clear process for pull requests.
* Reinforce coding standards (eg, testing, commenting, documentation)
* Also provide a space for OS/arch for tracking any platform-centric bugs

Borrowed from https://github.com/axolo-co/pull_request_template


smburdick:
I think we should also consider the [Linux kernel guidelines on AI assisted commits](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-assistants.rst).

Zhaoyilunnn (approved):

PR #924, opened 2026-05-02
Fix reschedule_measurements() modifies globally shared Plaquette instances

HaoTy:
Plaquette instances are shared globally, but `LayoutLayer.reschedule_measurements()` modifies them in-place without creating a copy. Thus, repeated `LayoutLayer` compilation will reschedule the already rescheduled Plaquette instances, pushing the measurements to later and later moments. This is the root cause of the extra TICK issue in the failed `test_compile_move_rotation` test in #805. Benchmarking shows that this change doesn't incur a noticeable performance overhead.


inmzhang (approved):
LGTM.

PR #926, opened 2026-05-03
fix: single-node CorrelationSurface PauliWeb round-trip

gideonuchehara:
Addresses the reviewer request in #830 to add a test for single-node correlation surface / Pauli web round-trip and ensure correct behavior.

## Problem

The round-trip `CorrelationSurface → PauliWeb → CorrelationSurface` was silently broken for single-node (self-loop) surfaces. Three interrelated bugs caused this:

**Bug 1 — `_graph_view` doubled self-loop edges** (`correlation.py`)

For a self-loop edge where `u == v`, the reverse insertion was identical to the forward one, so `[edge, edge]` ended up in the same slot. The XOR reduction in `_to_mutable_graph_representation` then cancelled: `Pauli.Z ^ Pauli.Z = Pauli.I`.

**Bug 2 — `_to_mutable_graph_representation` called `is_hadamard` on a ghost edge** (`correlation.py`)

A self-loop `(u, u)` does not exist as an edge in the underlying pyzx graph. Calling `is_hadamard(zx_graph, (u, u))` would raise an error once Bug 1 was fixed.

**Bug 3 — pyzx's `PauliWeb` silently drops self-loop half-edges** (`interop/pyzx/correlation.py`)

Even with Bugs 1 & 2 fixed, `to_pauli_web()` calls `pauli_web.add_half_edge((0, 0), ...)` which pyzx silently ignores. `pauli_web.half_edges()` then returns `{}`, and the reconstruction produces `CorrelationSurface(span=frozenset())`.

## Fixes

- **`_graph_view`**: skip the reverse insertion when `u == v`.
- **`_to_mutable_graph_representation`**: use `edge_is_hadamard = False` when `u == v`.
- **`pauli_web_to_correlation_surface`**: detect single-node graphs (no edges) and reconstruct the surface directly from the vertex type, bypassing `PauliWeb` entirely (X-type spider → `Basis.Z`, Z-type → `Basis.X`).
- **`_to_immutable_public_representation`**: handle self-loop entries in `_CorrelationSurface` for defensive completeness.

## Test

Added `test_single_node_correlation_pauliweb_round_trip` parametrized over `VertexType.X` and `VertexType.Z`. Both cases now pass. Full suite: 691 passed, 8 skipped.

### DIFF
--- a/src/tqec/computation/_correlation.py
+++ b/src/tqec/computation/_correlation.py
@@ -111,7 +111,19 @@ class _CorrelationSurfaceBase(MutableMapping[int, dict[int, Pauli]]):
                     node_u = zx_nodes.setdefault((u, basis_u), ZXNode(pos_u, basis_u))
                     node_v = zx_nodes.setdefault((v, basis_v), ZXNode(pos_v, basis_v))
                     span.append(ZXEdge.sorted(node_u, node_v))
+        # Handle single-node (self-loop) case: a vertex with no graph edges
+        # whose correlation surface is represented as a self-loop entry.
+        for u, edges in self.items():
+            if u in edges:
+                pauli = edges[u]
+                pos_u = graph[u]
+                for xz in Pauli.iter_xz():
+                    if xz in pauli:
+                        basis = bases[xz.value >> 1]
+                        node = zx_nodes.setdefault((u, basis), ZXNode(pos_u, basis))
+                        span.append(ZXEdge.sorted(node, node))
         return CorrelationSurface(frozenset(span))
+
 
 
 _CorrelationSurfaceType = TypeVar("_CorrelationSurfaceType", bound="_CorrelationSurfaceBase")
--- a/src/tqec/computation/correlation.py
+++ b/src/tqec/computation/correlation.py
@@ -128,7 +128,8 @@ class CorrelationSurface:
         for edge in self.span:
             u, v = edge.u.position, edge.v.position
             edges.setdefault(u, {}).setdefault(v, []).append(edge)
-            edges.setdefault(v, {}).setdefault(u, []).append(edge)
+            if u != v:  # Don't duplicate self-loop edges
+                edges.setdefault(v, {}).setdefault(u, []).append(edge)
             bases.setdefault(u, set()).add(edge.u.basis)
             bases.setdefault(v, set()).add(edge.v.basis)
         return edges, bases
@@ -275,10 +276,13 @@ class CorrelationSurface:
             u = p2v[pos_u]
             for pos_v, edge in edges.items():
                 v = p2v[pos_v]
+                # Self-loop edges (single-node case) don't exist in the ZX graph,
+                # so is_hadamard cannot be called on them.
+                edge_is_hadamard = False if u == v else is_hadamard(zx_graph, (u, v))
                 surface._add_pauli_to_edge(
                     (u, v),
                     reduce(xor, (e.get_basis(pos_u).to_pauli() for e in edge)),
-                    is_hadamard(zx_graph, (u, v)),
+                    edge_is_hadamard,
                 )
         for u, v in zx_graph.edges():
             if u not in surface or v not in surface[u]:
--- a/src/tqec/interop/pyzx/correlation.py
+++ b/src/tqec/interop/pyzx/correlation.py
@@ -42,7 +42,7 @@ def pauli_web_to_correlation_surface(
     if len(vertices) == 1 and not list(zx_graph.edges()):
         u = vertices[0]
         pos = g[u]
-        basis = vertex_type_to_pauli(zx_graph.type(u)).to_basis().flipped()
+        basis = vertex_type_to_pauli(zx_graph.type(u), zx_graph.phase(u)).to_basis().flipped()
         node = ZXNode(pos, basis)
         return CorrelationSurface(frozenset([ZXEdge.sorted(node, node)]))
     return _pauli_web_to_correlation_surface(pauli_web)._to_immutable_public_representation(g)
--- a/src/tqec/interop/pyzx/correlation.py
+++ b/src/tqec/interop/pyzx/correlation.py
@@ -6,9 +6,9 @@ from pyzx.graph.graph_s import GraphS
 from pyzx.pauliweb import PauliWeb
 
 from tqec.computation._correlation import _CorrelationSurface
-from tqec.computation.correlation import CorrelationSurface
+from tqec.computation.correlation import CorrelationSurface, ZXEdge, ZXNode
 from tqec.interop.pyzx.positioned import PositionedZX
-from tqec.interop.pyzx.utils import is_hadamard
+from tqec.interop.pyzx.utils import is_hadamard, vertex_type_to_pauli
 from tqec.utils.enums import Pauli
 
 
@@ -34,6 +34,17 @@ def pauli_web_to_correlation_surface(
     pauli_web: PauliWeb[int, tuple[int, int]], g: PositionedZX
 ) -> CorrelationSurface:
     """Create a correlation surface from a Pauli web."""
+    # pyzx's PauliWeb does not support self-loop half-edges, so single-node
+    # correlation surfaces (which have no graph edges) cannot be encoded in a
+    # PauliWeb.  Reconstruct them directly from the vertex type.
+    zx_graph = g.g
+    vertices = list(zx_graph.vertices())
+    if len(vertices) == 1 and not list(zx_graph.edges()):
+        u = vertices[0]
+        pos = g[u]
+        basis = vertex_type_to_pauli(zx_graph.type(u)).to_basis().flipped()
+        node = ZXNode(pos, basis)
+        return CorrelationSurface(frozenset([ZXEdge.sorted(node, node)]))
     return _pauli_web_to_correlation_surface(pauli_web)._to_immutable_public_representation(g)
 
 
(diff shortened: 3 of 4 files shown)


tests/interop/pyzx/correlation_test.py
```diff
@@ -10,3 +16,16 @@ def test_correlation_pauliweb_conversion() -> None:
         _check_correlation_surface_validity(surface, g)
         pauli_web = surface.to_pauli_web(g)
         assert CorrelationSurface.from_pauli_web(pauli_web, g) == surface
+
+
+@pytest.mark.parametrize("ty", [VertexType.X, VertexType.Z])
```
smburdick, line 21: I suggest adding more tests (happy/sad paths)

src/tqec/computation/_correlation.py
```diff
@@ -111,9 +111,21 @@ def _to_immutable_public_representation(self, graph: PositionedZX) -> Correlatio
...
+        # whose correlation surface is represented as a self-loop entry.
+        for u, edges in self.items():
+            if u in edges:
+                pauli = edges[u]
+                pos_u = graph[u]
+                for xz in Pauli.iter_xz():
+                    if xz in pauli:
+                        basis = bases[xz.value >> 1]
+                        node = zx_nodes.setdefault((u, basis), ZXNode(pos_u, basis))
+                        span.append(ZXEdge.sorted(node, node))
```
smburdick, line 124: I feel this isn't going to scale well to large graphs. Consider an alternate approach using e.g. sets or dictionaries. And provide some test cases with large graphs to show whether the polynomial growth is problematic.

src/tqec/computation/correlation.py
```diff
@@ -275,10 +276,13 @@ def _to_mutable_graph_representation(self, graph: PositionedZX) -> _CorrelationS
             u = p2v[pos_u]
             for pos_v, edge in edges.items():
                 v = p2v[pos_v]
+                # Self-loop edges (single-node case) don't exist in the ZX graph,
+                # so is_hadamard cannot be called on them.
+                edge_is_hadamard = False if u == v else is_hadamard(zx_graph, (u, v))
```
smburdick, line 281: Can it be `False` if the like edges have already been filtered out?

src/tqec/interop/pyzx/correlation.py
```diff
@@ -34,6 +34,17 @@ def pauli_web_to_correlation_surface(
     pauli_web: PauliWeb[int, tuple[int, int]], g: PositionedZX
 ) -> CorrelationSurface:
     """Create a correlation surface from a Pauli web."""
+    # pyzx's PauliWeb does not support self-loop half-edges, so single-node
+    # correlation surfaces (which have no graph edges) cannot be encoded in a
+    # PauliWeb.  Reconstruct them directly from the vertex type.
+    zx_graph = g.g
+    vertices = list(zx_graph.vertices())
+    if len(vertices) == 1 and not list(zx_graph.edges()):
+        u = vertices[0]
```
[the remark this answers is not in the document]
smburdick: Would want to check if `phase()` sufferers from rounding error problems

tests/interop/pyzx/correlation_test.py
```diff
@@ -10,3 +16,16 @@ def test_correlation_pauliweb_conversion() -> None:
...
+def test_single_node_correlation_pauliweb_round_trip(ty: VertexType) -> None:
+    g = GraphS()
+    g.add_vertex(ty)
+    pg = PositionedZX(g, {0: Position3D(0, 0, 0)})
+    surfaces = find_correlation_surfaces(pg)
+    assert len(surfaces) == 1
+    surface = surfaces[0]
+    assert surface.is_single_node
+    pauli_web = surface.to_pauli_web(pg)
```
smburdick, line 30: Can you also check whether there are repeated edges?

HaoTy:
I think the fix can be simpler. I will take a closer look tonight or tomorrow.

gideonuchehara (commented):
> Can it be `False` if the like edges have already been filtered out? 

---

> The two guards serve different purposes and both are needed:
>
> - The guard in `_graph_view` (`if u != v`) prevents the self-loop edge from being **inserted twice** into the index, which was causing the XOR reduction `Pauli.Z ^ Pauli.Z = Pauli.I` to incorrectly cancel.
>
> - The guard in `_to_mutable_graph_representation` (`False if u == v`) prevents a **crash** when calling `is_hadamard(zx_graph, (u, u))`. A self-loop `ZXEdge(node, node)` still appears in `_graph_view` as a `pos_u == pos_v` entry — it is not filtered out, just no longer duplicated. After the `p2v` lookup, `u == v` is still reachable, but that edge doesn't exist in the underlying pyzx graph, so `is_hadamard` would fail.

---

src/tqec/interop/pyzx/correlation.py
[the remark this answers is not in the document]
gideonuchehara: ---

> Good point to flag. pyzx stores phases as Python's `fractions.Fraction`, which is exact rational arithmetic, no floating-point representation. So `zx_graph.phase(u)` returns e.g. `Fraction(1, 2)` for an S node, not `0.5`. The `vertex_type_to_pauli` function matches on `Fraction(numerator=1, denominator=2)` explicitly, so there's no rounding risk here. The call is safe.

---

tests/interop/pyzx/correlation_test.py
gideonuchehara: ---

> Added in commit `b118b2b`:
>
> - **Happy paths**: `test_two_node_correlation_pauliweb_round_trip` — parametrized over four combinations of vertex types and edge types (normal and Hadamard edges), verifying the full round-trip for two-node graphs.
> - **Repeated-edge check**: `test_single_node_span_has_no_repeated_edges` — verifies the self-loop edge appears exactly once in the span, guarding against the original `_graph_view` duplication bug regressing.
> - **Sad path**: `test_from_pauli_web_wrong_graph_raises` — verifies that reconstructing a `CorrelationSurface` from a `PauliWeb` with a mismatched graph (different vertex type) does not silently return the same surface.

src/tqec/computation/correlation.py
gideonuchehara: > ---
> 
> > The two guards serve different purposes and both are needed:
> >
> > - The guard in `_graph_view` (`if u != v`) prevents the self-loop edge from being **inserted twice** into the index, which was causing the XOR reduction `Pauli.Z ^ Pauli.Z = Pauli.I` to incorrectly cancel.
> >
> > - The guard in `_to_mutable_graph_representation` (`False if u == v`) prevents a **crash** when calling `is_hadamard(zx_graph, (u, u))`. A self-loop `ZXEdge(node, node)` still appears in `_graph_view` as a `pos_u == pos_v` entry — it is not filtered out, just no longer duplicated. After the `p2v` lookup, `u == v` is still reachable, but that edge doesn't exist in the underlying pyzx graph, so `is_hadamard` would fail.
> 
> ---

src/tqec/computation/_correlation.py
gideonuchehara: > The self-loop block is O(V), it iterates each vertex once. The inner `for xz in Pauli.iter_xz()` is a constant-size loop (X and Z only, always 2 iterations). Similarly, the main edge loop above it is O(E). Neither has polynomial growth relative to graph size. Happy to add a large-graph test to demonstrate this empirically if it would help confirm.

tests/interop/pyzx/correlation_test.py
gideonuchehara: > Addressed in `test_single_node_span_has_no_repeated_edges` (commit `b118b2b`): asserts `len(surface.span) == 1`, confirming the self-loop edge appears exactly once. This directly guards against the original `_graph_view` duplication bug where the same edge was inserted twice, which would have been masked by `frozenset` deduplication but caught here by checking the span count.

HaoTy:
Hey, upon checking, I find that the fixes are not exactly the most appropriate and a bit bloated. The single-node case is really a special case that we specifically added for the single-cube memory experiments and it's super simple, so it's better handled with short paths instead of adding an if statement in each iteration of the loop or a whole other loop. Since it's not really a thing in PyZX's `PauliWeb`, I think it's more reasonable to raise an error rather than force a conversion that doesn't make sense on the PyZX side. Even if we were to do a Pauli web conversion short path, it should happen in the underscored methods, not the public ones. The new tests are also somewhat redundant. AI would do anything you tell it to do, so use your own judgment and stop it before it makes the codebase a pile of spaghetti. I think it's easier if I implement what I have in mind myself; thanks for the PR though!

gideonuchehara:
> Hey, upon checking, I find that the fixes are not exactly the most appropriate and a bit bloated. The single-node case is really a special case that we specifically added for the single-cube memory experiments and it's super simple, so it's better handled with short paths instead of adding an if statement in each iteration of the loop or a whole other loop. Since it's not really a thing in PyZX's `PauliWeb`, I think it's more reasonable to raise an error rather than force a conversion that doesn't make sense on the PyZX side. Even if we were to do a Pauli web conversion short path, it should happen in the underscored methods, not the public ones. The new tests are also somewhat redundant. AI would do anything you tell it to do, so use your own judgment and stop it before it makes the codebase a pile of spaghetti. I think it's easier if I implement what I have in mind myself; thanks for the PR though!

Fair points, HaoTy. I can see how the changes ended up more scattered than they needed to be. The error-raising approach on the PyZX side is a much cleaner way to think about it. Thanks for taking the time to review and for offering to implement a better version. I'll keep an eye on it and learn from how you approach it.

HaoTy:
Thanks for your understanding! Feel free to contribute to other issues!

PR #939, opened 2026-05-19
Fix Y half cube z double-scaling in DAE/JSON/bgraph importers

PaulCharon:
## Summary

`BlockGraph.from_dae_file` (and the JSON and bgraph importers) mis-scale every Y half cube's `z` by an extra factor of `(1 + pipe_length)` during import. The Y silently lands at the wrong grid cell, and surfaces as `TQECError: Cube already exists at position ...` whenever the misplaced cell happens to be occupied.

Minimal repro:

```python
from tqec.computation.block_graph import BlockGraph
from tqec.computation.cube import YHalfCube, ZXCube
from tqec.computation.pipe import PipeKind
import tempfile

g = BlockGraph()
g.add_cube(Position3D(1, 1, 1), ZXCube.from_str("XZZ"))
g.add_cube(Position3D(1, 1, 2), ZXCube.from_str("XZZ"))
g.add_cube(Position3D(1, 1, 3), YHalfCube())
g.add_pipe(Position3D(1, 1, 1), Position3D(1, 1, 2), PipeKind.from_str("XZO"))
g.add_pipe(Position3D(1, 1, 2), Position3D(1, 1, 3), PipeKind.from_str("XZO"))

with tempfile.NamedTemporaryFile(suffix=".dae") as f:
    g.to_dae_file(f.name)
    BlockGraph.from_dae_file(f.name)
# -> TQECError: Cube already exists at position (1,1,1).
```

## Root cause

`offset_y_cube_position(pos, pipe_length)` performs two operations: it undoes the writer's `+0.5` z-shift for Y cubes that had a pipe directly above, *and* it divides `z` by `(1 + pipe_length)`. Every caller then passes the result to `int_position_before_scale(..., pipe_length)`, which divides every coordinate by `(1 + pipe_length)`. The z-division therefore happens twice for Y cubes; `x` and `y` are correctly divided once.

The previous docstring described `pipe_length` as governing the offset itself, but the writer's `+0.5` is a hard-coded constant in DAE space — `pipe_length` only ever drove the redundant scaling.

## Fix

Drop the `pipe_length` argument from `offset_y_cube_position` entirely; remove the internal division. Scaling now lives exclusively in `int_position_before_scale`, which has the unambiguous job of mapping DAE coordinates onto the TQEC grid. The helper now does only what its name says — reverse the writer's `+0.5` shift.

Three call sites updated:

- `src/tqec/interop/collada/read_write.py:150` — DAE importer Y cube
- `src/tqec/interop/collada/read_write.py:373` — JSON importer Y cube
- `src/tqec/interop/bgraph/read_write.py:53` — bgraph importer Y cube (latently affected: any bgraph file with non-zero `pipe_length`)

## Tests

- `tests/interop/shared_test.py`: `test_offset_y_cube_position` was parametrising only the x dimension at z=0, so it never exercised the shift or the bug. Original cases kept verbatim; added cases that cover the +0.5 shift path and confirm x/y are never touched.
- `tests/interop/collada/read_write_test.py`: new `test_dae_roundtrip_preserves_y_cube_position_above_origin` builds a Y-on-top stack, round-trips it through DAE, and asserts the Y cube position is preserved.

## Reference

Reproducer notebook (arithmetic walk-through, isolated misplacement demo, verification of the fix via monkey-patch): https://gist.github.com/PaulCharon/2f9c8e3e12ca553cbf273e84739302a8

Original downstream report: https://github.com/Walrus-Computing/piper-draw/issues/320

## Related observation (not addressed in this PR)

There is a separate geometric oddity worth flagging but not fixing here: the writer in `write_block_graph_to_dae_file` hard-codes the Y-cube shift to `dz=0.5` in DAE space, regardless of `pipe_length`. At `pipe_length != 2.0` this produces a visually misplaced Y in the DAE rendering, even though round-trips are self-consistent.

---

*AI assistance disclosure: bug investigation, fix design, and PR drafting were done in collaboration with Claude (Anthropic). All code changes were reviewed, applied, and verified locally by the author before pushing; CI runs against the actual diff.*

### DIFF
--- a/src/tqec/interop/bgraph/read_write.py
+++ b/src/tqec/interop/bgraph/read_write.py
@@ -51,9 +51,7 @@ def load_bgraph(bgraph_str_or_path: str | Path, graph_name: str = "") -> BlockGr
                     kind = "Y"
                     if isinstance(block_kind_from_str(kind), YHalfCube):
                         position = int_position_before_scale(
-                            offset_y_cube_position(
-                                FloatPosition3D(*(int(x), int(y), int(z))), pipe_length
-                            ),
+                            offset_y_cube_position(FloatPosition3D(*(int(x), int(y), int(z)))),
                             pipe_length,
                         )
                 else:
--- a/src/tqec/interop/collada/read_write.py
+++ b/src/tqec/interop/collada/read_write.py
@@ -147,7 +147,7 @@ def read_block_graph_from_dae_file(
     for pos, cube_kind, axes_directions in parsed_cubes:
         if isinstance(cube_kind, YHalfCube):
             graph.add_cube(
-                int_position_before_scale(offset_y_cube_position(pos, pipe_length), pipe_length),
+                int_position_before_scale(offset_y_cube_position(pos), pipe_length),
                 cube_kind,
             )
         else:
@@ -369,9 +369,7 @@ def read_block_graph_from_json(
     # Add cubes
     for pos, cube_kind, axes_directions in parsed_cubes:
         if isinstance(cube_kind, YHalfCube):
-            graph.add_cube(
-                int_position_before_scale(offset_y_cube_position(pos, 0.0), 0.0), cube_kind
-            )
+            graph.add_cube(int_position_before_scale(offset_y_cube_position(pos), 0.0), cube_kind)
         else:
             graph.add_cube(int_position_before_scale(pos, 0.0), cube_kind)
     port_index = 0
--- a/src/tqec/interop/shared.py
+++ b/src/tqec/interop/shared.py
@@ -25,12 +25,22 @@ def int_position_before_scale(pos: FloatPosition3D, pipe_length: float) -> Posit
     )
 
 
-def offset_y_cube_position(pos: FloatPosition3D, pipe_length: float) -> FloatPosition3D:
-    """Offsets the position of a Y-cube according to the length of pipes in the model.
+def offset_y_cube_position(pos: FloatPosition3D) -> FloatPosition3D:
+    """Undo the writer's +0.5 z-shift for Y half cubes that had a pipe above.
+
+    The DAE writer shifts a Y half cube's DAE z by +0.5 when there is a pipe
+    directly above it, so that the upper-half geometry sits flush against the
+    pipe. This helper detects that fractional z (z ≈ floor(z) + 0.5) and
+    reverses the shift, leaving an integer DAE z that
+    :func:`int_position_before_scale` then maps onto the TQEC grid.
+
+    The previous signature took a ``pipe_length`` argument and divided z by
+    ``(1 + pipe_length)``. That scaling was redundant with
+    :func:`int_position_before_scale` and caused a double-scaling bug for any
+    Y cube at z > 0.
 
     Args:
         pos: (x, y, z) position where x, y, and z are floats.
-        pipe_length: the length of the pipes in the model.
 
     Returns:
         An offset (x, y, z)  position where x, y, and z are floats.
@@ -38,7 +48,7 @@ def offset_y_cube_position(pos: FloatPosition3D, pipe_length: float) -> FloatPos
     """
     if np.isclose(pos.z - 0.5, np.floor(pos.z), atol=1e-9):
         pos = pos.shift_by(dz=-0.5)
-    return FloatPosition3D(pos.x, pos.y, pos.z / (1 + pipe_length))
+    return FloatPosition3D(pos.x, pos.y, pos.z)
 
 
 def scale_position(pos: Position3D, pipe_length: float = 0.0) -> FloatPosition3D:
--- a/tests/interop/collada/read_write_test.py
+++ b/tests/interop/collada/read_write_test.py
@@ -5,7 +5,8 @@ from pathlib import Path
 import pytest
 
 from tqec.computation.block_graph import BlockGraph
-from tqec.computation.cube import ZXCube
+from tqec.computation.cube import YHalfCube, ZXCube
+from tqec.computation.pipe import PipeKind
 from tqec.gallery.cnot import cnot
 from tqec.gallery.three_cnots import three_cnots
 from tqec.utils.enums import Basis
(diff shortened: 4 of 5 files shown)


nelimee (commented):
Not checking the actual PR content, I just wanted to say that this is a perfect example of what should be a (at-least-partially-)LLM-generated PR:

- LLM usage is disclosed clearly with (nearly) sufficient information,
- code change does not seem to be overly verbose,
- diff is small.

Because that's the first PR with LLM-generated code involved that is worth reviewing, you are the first to get probably the most annoying question: can you ensure that the code you submit is compatible with the Apache 2.0 license this repository is under? 

That is trivial for hand-generated PRs as the author is deliberately submitting code he has full rights on. That is less trivial (at least in my mind) for LLM-generated code that might contain copy-pasted code under an incompatible license without warning the user.

PaulCharon:
Thanks. Is there any additional information beyond Apache 2.0 compatibility that you would like included in the LLM disclosure?

For this PR specifically, I am very confident the result is Apache 2.0–compatible since every edit was applied by hand, and the changes are deliberately minimal. The source-code changes are net deletions (removing one helper parameter, the redundant `pos.z / (1 + pipe_length)` line, and the matching arg at three call sites). The additions are a rewritten docstring describing the new helper behavior, four new parametrize cases extending the existing `test_offset_y_cube_position`, and one new round-trip regression test in the same style as the surrounding file, and which I am confident were not cribbed from elsewhere.

More broadly, it might be worth including an explicit AI-code statement in `CONTRIBUTING.md` outlining how LLM-assisted PRs should be flagged, what verification contributors should do, and what artifacts (similarity scans, license scans, etc.) you would like attached.

starktech23:
@nelimee please add me as a reviewer, thanks.

src/tqec/interop/bgraph/read_write.py
```diff
@@ -51,9 +51,7 @@ def load_bgraph(bgraph_str_or_path: str | Path, graph_name: str = "") -> BlockGr
                     kind = "Y"
                     if isinstance(block_kind_from_str(kind), YHalfCube):
                         position = int_position_before_scale(
-                            offset_y_cube_position(
-                                FloatPosition3D(*(int(x), int(y), int(z))), pipe_length
```
starktech23, line 55: reasoning explained in the summary, removing additional `pipe_length` argument gives fix as suggested by PR and tests.

src/tqec/interop/collada/read_write.py
```diff
@@ -147,7 +147,7 @@ def read_block_graph_from_dae_file(
     for pos, cube_kind, axes_directions in parsed_cubes:
         if isinstance(cube_kind, YHalfCube):
             graph.add_cube(
-                int_position_before_scale(offset_y_cube_position(pos, pipe_length), pipe_length),
```
starktech23, line 150: how i understand it is that `int_position_before_scale` already uses `pipe_length` once, so removal from within `offset_y_cube_position` fixes redundancy.

src/tqec/interop/shared.py
```diff
@@ -25,20 +25,30 @@ def int_position_before_scale(pos: FloatPosition3D, pipe_length: float) -> Posit
...
+    The DAE writer shifts a Y half cube's DAE z by +0.5 when there is a pipe
+    directly above it, so that the upper-half geometry sits flush against the
+    pipe. This helper detects that fractional z (z ≈ floor(z) + 0.5) and
+    reverses the shift, leaving an integer DAE z that
+    :func:`int_position_before_scale` then maps onto the TQEC grid.
+
+    The previous signature took a ``pipe_length`` argument and divided z by
+    ``(1 + pipe_length)``. That scaling was redundant with
+    :func:`int_position_before_scale` and caused a double-scaling bug for any
+    Y cube at z > 0.
```
starktech23, line 40: additional background here for changes made helps to understand context and usage. 👍

tests/interop/collada/read_write_test.py
```diff
@@ -127,3 +128,27 @@ def test_collada_write_read_with_correlation_surface() -> None:
...
+    g.add_pipe(Position3D(1, 1, 1), Position3D(1, 1, 2), PipeKind.from_str("XZO"))
+    g.add_pipe(Position3D(1, 1, 2), Position3D(1, 1, 3), PipeKind.from_str("XZO"))
+
+    with tempfile.NamedTemporaryFile(suffix=".dae", delete=False) as f:
+        g.to_dae_file(f.name)
+        g2 = BlockGraph.from_dae_file(f.name)
+
+    y_cubes = [c for c in g2.cubes if isinstance(c.kind, YHalfCube)]
+    assert len(y_cubes) == 1
+    assert y_cubes[0].position == Position3D(1, 1, 3)
```
starktech23, line 154: I'd imagine that the test extension is similar to the notebook mentioned in the solution summary to check if the result is as expected or deviating. 
Note that functions YHalfCube and PipeKind have been imported.
As long as this helps verify behavior then LGTM.

tests/interop/shared_test.py
```diff
@@ -22,13 +22,22 @@ def test_int_position_before_scale(pos: tuple[float, float, float], expected: tu
 @pytest.mark.parametrize(
     "pos, expected",
     [
+        # Original cases — identity for any z that doesn't trigger the +0.5 shift.
```
starktech23, line 25: nice to point out that original test did not test for the +0.5 shift.

tests/interop/shared_test.py
```diff
@@ -22,13 +22,22 @@ def test_int_position_before_scale(pos: tuple[float, float, float], expected: tu
...
         [(1.0, 0.0, 0.0), (1.0, 0.0, 0.0)],
         [(2.0, 0.0, 0.0), (2.0, 0.0, 0.0)],
         [(2.5, 0.0, 0.0), (2.5, 0.0, 0.0)],
+        # New cases — Integer z at z > 0: still no shift.
+        [(3.0, 3.0, 9.0), (3.0, 3.0, 9.0)],
+        # z = floor(z) + 0.5: the writer's pipe-above marker -> subtract 0.5.
+        [(3.0, 3.0, 9.5), (3.0, 3.0, 9.0)],
+        [(0.0, 0.0, 0.5), (0.0, 0.0, 0.0)],
+        # x and y are never touched, even when z triggers the shift.
+        [(1.25, -7.0, 4.5), (1.25, -7.0, 4.0)],
```
starktech23, line 36: new test cases that check for the shift bug and makes sure that x and y are not touched.

tests/interop/shared_test.py
```diff
@@ -22,13 +22,22 @@ def test_int_position_before_scale(pos: tuple[float, float, float], expected: tu
...
     ],
 )
-def test_offset_y_cube_position(pos: tuple[float, float, float], expected: tuple[int, int, int]):
-    # NB! Varying pipe lengths already tested via COLLADA tests
+def test_offset_y_cube_position(
+    pos: tuple[float, float, float], expected: tuple[float, float, float]
+):
     pos_3d, expected_3d = (FloatPosition3D(*pos), FloatPosition3D(*expected))
-    assert expected_3d == offset_y_cube_position(pos=pos_3d, pipe_length=2.0)
+    assert expected_3d == offset_y_cube_position(pos=pos_3d)
```
starktech23, line 43: `pipe_length` removed again 👍 
might be worthwhile to see if this is present anywhere else in the code.

starktech23 (approved):
The logic is explained well. I left some comments to follow along the changes so I'm able to understand it better since I'm fairly new to the codebase, but as a new contributor to the project this is a very well defined PR stating the issues and reproduction of the issue well and providing minimal, effective solutions. Also appreciate the disclosure of AI usage and the way it was implemented in the solution, also helps set a baseline for what good AI assisted code contributions look like. I also had some AI assistance to better understand what the issue was and what the fixes were. At least the logic seems reasonable to me. If any additional bugs occur and could be reproduced then let's revisit this PR. 
@PaulCharon next step and a future PR could be fixing the hard-coded `dz=0.5` issue listed. Also, if you could list out your thought process and workflow while using AI tools for preparing this solution all the way to the final edit, IMO it could help with adding the guidelines to the `CONTRIBUTING.md` file for future contributors. Thanks!

PaulCharon:
Thanks for your review and the inline comments!

The bug surfaced first downstream in our project (Walrus-Computing/piper-draw#320), and I used Claude to trace through the call chain and help identify the issue. After noticing that the roundtrip worked in piper-draw, I tried replicating the roundtrip purely in tqec and found the double-scaling issue (again with AI assistance). I then used Claude to suggest a fix, create the notebook outlining the issue, and draft the PR. At each step, I double-checked the result and made small adjustments for quality and coherence. Before pushing, I ran `pre-commit`, `ty check`, and the non-slow pytest suite locally, and executed the notebook against unmodified `upstream/main` so that its outputs reflect the actual bug.

I’ll happily address the `dz=0.5` issue in a future PR.

nelimee (approved):
Following @starktech23 feedback, approving. @PaulCharon I can't remember if you will be able to click on "merge" yourself or not. If not, ping me to state that you are fine with me merging that PR, and I'll click.

Thanks for that contribution, and your responsible and disclosed use of LLM :)

PaulCharon:
Thank you! Unfortunately, I am unable to click merge since I don't have write access to tqec, but I am fine with you merging @starktech23.

starktech23:
oh I actually don't have merge rights for this either, @nelimee would be able to do it 🙌

nelimee:
Waiting for CI to pass after merging `main` here, once done I will click on merge. Well done for your first contribution! 👍

PR #948, opened 2026-06-03
Stabilize gallery circuit generation order

### LINKED ISSUE #825: Gallery curves show duplicate points for each physical error rate
## Describe the bug

## Expected behavior

Samfresh-ai:
Closes #825

## Summary
- store layout layers in coordinate order so generated layout circuits do not depend on upstream mapping insertion order
- sort observable qubits before resolving measurement records
- add regression coverage for both deterministic-ordering paths

## Validation
- `uv run --no-default-groups --group test pytest tests/compile/observables/builder_test.py tests/compile/blocks/layers/merge_test.py`
- `uvx ruff==0.15.15 check src/tqec/compile/observables/builder.py src/tqec/compile/blocks/layers/atomic/layout.py tests/compile/observables/builder_test.py tests/compile/blocks/layers/merge_test.py`
- `uvx ruff==0.15.15 format --check src/tqec/compile/observables/builder.py src/tqec/compile/blocks/layers/atomic/layout.py tests/compile/observables/builder_test.py tests/compile/blocks/layers/merge_test.py`
- generated memory, move-rotation, CNOT, and three-CNOT gallery circuits with `PYTHONHASHSEED=1` and `PYTHONHASHSEED=2`; SHA-256 hashes matched for every checked case

## AI assistance disclosure
I used OpenAI Codex to help inspect the deterministic-ordering paths and draft the small regression tests. The code path was manually reviewed and validated locally with the commands above.

### DIFF
--- a/src/tqec/compile/blocks/layers/atomic/layout.py
+++ b/src/tqec/compile/blocks/layers/atomic/layout.py
@@ -57,7 +57,7 @@ class LayoutLayer(BaseLayer):
 
         """
         super().__init__(frozenset())
-        self._layers = layers
+        self._layers = dict(sorted(layers.items(), key=lambda item: (item[0]._x, item[0]._y)))
         self._element_shape = element_shape
         self._post_init_check()
 
--- a/src/tqec/compile/observables/builder.py
+++ b/src/tqec/compile/observables/builder.py
@@ -265,7 +265,7 @@ def get_observable_with_measurement_records(
 
     measured_qubits = [
         q
-        for q in qubits
+        for q in sorted(qubits)
         # Ignore those qubits that are not measured in the circuit.
         # This is required because the some observable builders
         # include the qubits that are not in the circuit like qubits
--- a/src/tqec/compile/observables/builder.py
+++ b/src/tqec/compile/observables/builder.py
@@ -263,14 +263,9 @@ def get_observable_with_measurement_records(
             "ignore_qubits_with_no_measurement to True to ignore them."
         )
 
-    measured_qubits = [
-        q
-        for q in sorted(qubits)
-        # Ignore those qubits that are not measured in the circuit.
-        # This is required because the some observable builders
-        # include the qubits that are not in the circuit like qubits
-        # in the scretched stabilizers to simplify the calculation.
-        if q in measurement_records
-    ]
+    # Ignore qubits that are not measured in the circuit. Some observable
+    # builders include out-of-circuit qubits in stretched stabilizers to
+    # simplify the calculation.
+    measured_qubits = [q for q in sorted(qubits) if q in measurement_records]
     measurement_offsets = [measurement_records[q][-1] for q in measured_qubits]
     return Observable(observable_index, measured_qubits, measurement_offsets)
--- a/tests/compile/blocks/layers/merge_test.py
+++ b/tests/compile/blocks/layers/merge_test.py
@@ -71,6 +71,22 @@ def test_merge_base_layers(
     assert merged_layer.element_shape == logical_qubit_shape
 
 
+def test_layout_layer_stores_layers_in_coordinate_order(
+    base_layers: list[BaseLayer], logical_qubit_shape: PhysicalQubitScalable2D
+) -> None:
+    plaquette_layer, plaquette_layer2, raw_layer = base_layers
+    b00 = LayoutPosition2D.from_block_position(BlockPosition2D(0, 0))
+    b01 = LayoutPosition2D.from_block_position(BlockPosition2D(0, 1))
+    b11 = LayoutPosition2D.from_block_position(BlockPosition2D(1, 1))
+
+    layout = LayoutLayer(
+        {b11: raw_layer, b01: plaquette_layer2, b00: plaquette_layer},
+        logical_qubit_shape,
+    )
+
+    assert list(layout.layers) == [b00, b01, b11]
+
+
 def test_merge_repeated_layers(
     base_layers: list[BaseLayer], logical_qubit_shape: PhysicalQubitScalable2D
 ) -> None:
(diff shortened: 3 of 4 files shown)


src/tqec/compile/blocks/layers/atomic/layout.py
```diff
@@ -57,7 +57,7 @@ def __init__(
 
         """
         super().__init__(frozenset())
-        self._layers = layers
+        self._layers = dict(sorted(layers.items(), key=lambda item: (item[0]._x, item[0]._y)))
```
smburdick, line 60: Does this impact anything in `compile_test.py` in any way? You might want to update that as well.

src/tqec/compile/observables/builder.py
```diff
@@ -265,7 +265,7 @@ def get_observable_with_measurement_records(
 
     measured_qubits = [
         q
-        for q in qubits
+        for q in sorted(qubits)
```
smburdick: Style point: I wonder if it would be better to put the entire list comprehension in the same line. Eg, `q for q in  ....`

tests/compile/blocks/layers/merge_test.py
```diff
@@ -71,6 +71,22 @@ def test_merge_base_layers(
     assert merged_layer.element_shape == logical_qubit_shape
 
 
+def test_layout_layer_stores_layers_in_coordinate_order(
+    base_layers: list[BaseLayer], logical_qubit_shape: PhysicalQubitScalable2D
+) -> None:
+    plaquette_layer, plaquette_layer2, raw_layer = base_layers
+    b00 = LayoutPosition2D.from_block_position(BlockPosition2D(0, 0))
+    b01 = LayoutPosition2D.from_block_position(BlockPosition2D(0, 1))
+    b11 = LayoutPosition2D.from_block_position(BlockPosition2D(1, 1))
```
smburdick, line 81: I would add a few more layers

smburdick: ... and what happens if they aren't already sorted?

tests/compile/observables/builder_test.py
```diff
@@ -33,3 +37,19 @@ def test_transform_coords_into_grid() -> None:
     x = -1 + 3 * 2
     y = (12 * 2 + 2) * 2 - 1 + 1 * 2
     assert qubits == {GridQubit(x, y)}
+
+
+def test_get_observable_with_measurement_records_sorts_qubits() -> None:
+    q0 = GridQubit(0, 0)
+    q1 = GridQubit(1, 0)
+    unmeasured_qubit = GridQubit(2, 0)
+    measurement_records = MeasurementRecordsMap({q1: [-1], q0: [-2]})
```
smburdick: Again I would go for a larger set of GridQubits and add additional cases that break the functionality.

smburdick (commented):
Mostly CI points.

Samfresh-ai:
Addressed in `aecb354`.

Changes:
- expanded the `LayoutLayer` ordering regression to use a deliberately unsorted five-position input,
- expanded the observable qubit ordering test with more measured qubits plus an unmeasured qubit,
- cleaned up the sorted-qubit list comprehension style.

Local checks passed: focused merge/observable tests (`12 passed`), ruff check, ruff format check, compileall, and `git diff --check`.

nelimee (commented):
This is very close to what I thought would solve the issue initially. Discussions in https://github.com/tqec/tqec/issues/825 hints that this might not be the root problem of the issue, but I think that ensuring deterministic circuit generation should be merged in TQEC independently of the resolution of the above issue.

A quick glance at the code showed no clear issues, but I couldn't invest a lot of time reviewing it, so I only "comment" and let others decide.

massinzenia-source:
I would like to review this PR

src/tqec/compile/observables/builder.py
```diff
@@ -263,14 +263,9 @@ def get_observable_with_measurement_records(
...
-        # Ignore those qubits that are not measured in the circuit.
-        # This is required because the some observable builders
-        # include the qubits that are not in the circuit like qubits
-        # in the scretched stabilizers to simplify the calculation.
-        if q in measurement_records
-    ]
+    # Ignore qubits that are not measured in the circuit. Some observable
+    # builders include out-of-circuit qubits in stretched stabilizers to
+    # simplify the calculation.
+    measured_qubits = [q for q in sorted(qubits) if q in measurement_records]
```
dlyongemallo, line 269: Is this actually intended to have an effect today (which seems to be implied by the PR description), or is it a defensive guard against future changes?

On the currently only path from an `Observable` to a `stim.Circuit`, sorting is already done in `to_instruction` (on [this line](https://github.com/tqec/tqec/blob/63af23298ddb4c2531e0b674e920b00bff369c51/src/tqec/compile/observables/builder.py#L54)). So this change would seem to have no effect on the final output.

src/tqec/compile/blocks/layers/atomic/layout.py
dlyongemallo: Does this have any effect on the final indices at all? These are already sorted in `relabel_circuits_qubit_indices` (on [this line](https://github.com/tqec/tqec/blob/63af23298ddb4c2531e0b674e920b00bff369c51/src/tqec/circuit/schedule/manipulation.py#L343)).

dlyongemallo (commented):
The changes in this PR seem fine as defensive measures to guarantee determinism at the API level in some intermediate data, but it looks to me like they wouldn't have any effect on the final compiled `stim.Circuit`, so I don't understand how this is related to fixing #825. 

> generated memory, move-rotation, CNOT, and three-CNOT gallery circuits with PYTHONHASHSEED=1 and PYTHONHASHSEED=2; SHA-256 hashes matched for every checked case

IIUC, neither `set[GridQubit]` nor `dict[...]` (the data types under test in this PR) are types covered by [`PYTHONHASHSEED`](https://docs.python.org/3.12/using/cmdline.html#envvar-PYTHONHASHSEED), so this doesn't validate your change.

nelimee:
Hey! Following reviews here and on #952 I think that #952 is on track to be merged.

smburdick:
> Hey! Following reviews here and on #952 I think that #952 is on track to be merged.

Yes, I agree that #952 is the more cohesive solution.

nelimee:
#952 has been merged, closing this PR. Thanks for your submission!

PR #952, opened 2026-06-05
fix: canonicalize clique ordering to stabilize gallery circuit generation (#825)

### LINKED ISSUE #825: Gallery curves show duplicate points for each physical error rate

Hope-Alemayehu:
Closes #825. Following the discussion in the issue.

## Problem
The clique grouping in `fill_ports_for_minimal_simulation` comes from `nx.greedy_color`, whose order depends on networkx's internal strategy and can shift across versions. That order leaks into the circuit text via OBSERVABLE_INCLUDE indices, changing the sinter strong_id of logically-identical circuits and producing duplicate points over time. Forcing different coloring strategies confirmed this, the circuit text changed at the same (d, r, p), though the partition itself stayed stable.

## Fix
Canonically sort the cliques and the stabilizers within each clique right after `greedy_color`, so the circuit depends only on the logical problem, not on networkx's coloring order. (A genuine change in the partition, which sorting can't absorb, is caught by the golden-snapshot test below.) Localized to `open_graph.py`; no changes to caching or persistence.

## Tests / Changes
- `test_fill_ports_deterministic_across_coloring_strategies`: forces 7 strategies and asserts identical circuit text across all (guards reordering).
- `test_fill_ports_circuit_golden_snapshot`: pins canonical circuit SHAs so a future repartition fails loudly in CI instead of silently duplicating points.
- Updated `three_cnots`/`steane_encoding` port-filling tests to order-independent assertions for the new canonical clique order (same content, decoupled from index).

## AI disclosure
I used Claude Code to help write the regression tests and investigate the root cause. The diagnosis, approach, and implementation were reviewed and verified by me; tests, ruff, and ty pass locally.


tests/computation/open_graph_test.py
```diff
@@ -20,3 +61,80 @@ def test_fill_ports_for_minimal_simulation(search_small_area_observables: bool)
...
+        + "\n".join(f"  {s}: {fp}" for s, fp in fingerprints.items())
+    )
+
+
+# Golden circuits. Regenerate (and review the diff) only on intended changes.
+_GOLDEN_CIRCUITS: dict[str, list[tuple[tuple[str, ...], str]]] = {
+    "cnot": [
+        (
+            ("XIXX", "XXXI"),
+            "b5e8fe52188e80ef5fa6ab27afad578d81d56615589f679877130a371ec08e72",
```
smburdick, line 100: If there was any change to how `str(circuit)` worked (which I think is very possible, but I didn't design it), then these test cases would all break.

tests/computation/open_graph_test.py
```diff
@@ -20,3 +61,80 @@ def test_fill_ports_for_minimal_simulation(search_small_area_observables: bool)
...
+        def patched(g: nx.Graph, *args: object, **kwargs: object) -> dict[int, int]:
+            if strategy == "random_sequential":
+                random.seed(12345)
+            return original_greedy_color(g, strategy=strategy)
+
+        return patched
+
+    fingerprints = {}
+    for strategy in COLORING_STRATEGIES:
+        monkeypatch.setattr(nx.algorithms.coloring, "greedy_color", make_patched(strategy))
```
smburdick, line 82: Can this be tested just how NX is being used in TQEC normally?

src/tqec/computation/open_graph.py
```diff
@@ -132,9 +132,11 @@ def fill_ports_for_minimal_simulation(
             g.add_edge(i, j)
     # Solve with heuristic greedy coloring
     coloring = nx.algorithms.coloring.greedy_color(g)  # type: ignore[invalid-argument-type]
-    cliques: dict[int, list[str]] = {}
+    clique_map: dict[int, list[str]] = {}
     for node, color in coloring.items():
-        cliques.setdefault(color, []).append(generators[node])
+        clique_map.setdefault(color, []).append(generators[node])
+
+    cliques: list[list[str]] = sorted(sorted(clique) for clique in clique_map.values())
```
smburdick: I think we should specifiy the key in the `sorted()` invocation(s) just to be safe.

tests/computation/open_graph_test.py
```diff
@@ -1,7 +1,48 @@
...
+from tqec.gallery.three_cnots import three_cnots
+
+OPEN_PORT_EXAMPLES = {
+    "cnot": cnot,
+    "three_cnots": three_cnots,
+    "move_rotation": move_rotation,
+}
+
+# "largest_first" is the production default; the rest probe ordering drift.
+COLORING_STRATEGIES = [
```
smburdick, line 20: I'm fairly sure this enum is already specified someplace?

smburdick (changes requested):
I don't feel good about using hashes of `str()` to compare circuits this way at this time. Consider comparing them elementwise instead.

nelimee:
One quick comment for the moment:

> I don't feel good about using hashes of str() to compare circuits this way at this time.

For #825 specifically, that's what we need. `sinter` is hashing the string representation of the `stim.Circuit` provided to generate a unique ID, and that's how `sinter` checks if a circuit already contains simulation data or not. But be wary of `hash()` (the Python built-in) as it is not deterministic across runs.

tests/computation/open_graph_test.py
```diff
@@ -1,7 +1,48 @@
...
+    "smallest_last",
+    "independent_set",
+    "connected_sequential_bfs",
+    "connected_sequential_dfs",
+    "saturation_largest_first",
+    "random_sequential",
+]
+
+
+def _circuit_sha(filled) -> str:
```
nelimee: Missing typing information (even on a private helper, we try to maitain 100% correct typing everywhere)

tests/computation/open_graph_test.py
```diff
@@ -1,7 +1,48 @@
...
+    ``manhattan_radius=-1`` skips detectors (fast, no multiprocessing) while
+    keeping the ``OBSERVABLE_INCLUDE`` ordering that the coloring drift perturbs.
+    """
+    circuit = compile_block_graph(
+        filled.graph, observables=filled.observables
+    ).generate_stim_circuit(1, manhattan_radius=-1, database_path=None)
+    return hashlib.sha256(str(circuit).encode()).hexdigest()
+
+
+def _fingerprint(filled_graphs) -> tuple[tuple[tuple[str, ...], str], ...]:
```
nelimee: Typing information on the parameter

tests/computation/open_graph_test.py
nelimee: Yes, any change in TQEC generation or `stim.Circuit.__str__` would lead to change here. This is fairly similar to a test made for the data-base. I expect this to be manageable, because it shouldn't break that often, it is a good canary test for unwanted changes, and fixing that test when it is expected to change shouldn't be too long (5 minutes, the time to run the test suite once and copy-paste the new values from the assert failures).

Up to you if that convinces you this is fine. We can also reduce the number of explicit hashes so that it is less cumbersome to update.

nelimee (commented):
As said in the issue, I think this solves a legitimate issue. I have made a few comments, but nothing too serious.

jbolns:
It looks like @smburdick and @nelimee got this one very well covered. Nothing to add.

Hope-Alemayehu:
Hi, thanks for the feedback, I pushed the changes you suggested.

Previously @jbolns asked me whether the fix is general for circuit generation. I tried to dig into that and here's what I found, though I could be missing something.

From what I can tell, three of the four ordering categories from #825 (gate targets, detector/observable record targets, qubit coords/indices) already have explicit sorted() calls in the existing code. #952 seems to close the one that didn't, greedy_color's clique ordering leaking into observable indices. I ran every gallery circuit across multiple runs and PYTHONHASHSEED values and they all came back byte-identical, which was nice.

I put the full investigation with file references and test outputs on the determinism-audit branch if anyone wants to check my reasoning.

jbolns:
I think this one's ready? @smburdick @nelimee 
- Don't see any risk for collateral downsides, i.e., nothing is likely to break as a result of operations in PR. 
- Imperfections in tests are justified on the specifics of the task.
- I don't think there's time for anyone else to undertake a similar level of diagnostics before the end of the hackathon. 
- It's the kind of thing where time is the best testing strategy: if duplicates return we will know something was missed.

tests/computation/open_graph_test.py
dlyongemallo: `stim.Circuit` defines an order-sensitive `__eq__` so a circuit-level comparison can be performed using just `==` without resorting to comparing hashes based on `str()`. That gives you the same test functionality but isn't brittle. (But this is an improvement to the tests which can be made later and shouldn't block merge.)

tests/computation/open_graph_test.py
nelimee: This is in https://github.com/networkx/networkx/blob/4ba50db2d08c828f1cfcb06394b895aac8bfeecf/networkx/algorithms/coloring/greedy_coloring.py#L252 but:

1. This is not re-exported (small hint that this might be "less" public than other methods, even though it is a public constant).
2. With the last commit, the two lists are actually different because most of the colouring strategies have been removed.

I'll resolve that comment, let's address that in a follow-up PR if you think we should change it.

nelimee:
> I think this one's ready? @smburdick @nelimee
> 
>     * Don't see any risk for collateral downsides, i.e., nothing is likely to break as a result of operations in PR.
> 
>     * Imperfections in tests are justified on the specifics of the task.
> 
>     * I don't think there's time for anyone else to undertake a similar level of diagnostics before the end of the hackathon.
> 
>     * It's the kind of thing where time is the best testing strategy: if duplicates return we will know something was missed.

I agree, from maintainer discussions and reviews here, I think this PR can be merged. I'll do so before the end of today, well done @Hope-Alemayehu 🎉 And sorry for the long review process! If you have not done so already, please comment on the issue pinging me so that I can assign you.

nelimee (approved):
To me this looks nice, there are a few comments I will leave open in case we want to address them later.

smburdick (approved):
I think it's good that the hashing logic stays within the test suite for now. I'd be interested to know how much this ends up being scaled. If there aren't many gallery circuits in the long run, this should be fine to remain as-is.

nelimee:
Well done for your contribution! 🎉 And I am sorry for some delays in my answers, I was pretty busy last week.

Hope-Alemayehu:
Thank you for the review and for helping me through the process. I really enjoyed working on this issue and learned a lot from it.

I'd love to continue contributing to and learning more about the project, and I'd like to build a deeper understanding of the codebase and the underlying concepts. If there are any areas of the project, topics, or resources you would recommend exploring, I'd really appreciate it.

Thanks again!

nelimee:
> I'd love to continue contributing to and learning more about the project, and I'd like to build a deeper understanding of the codebase and the underlying concepts. If there are any areas of the project, topics, or resources you would recommend exploring, I'd really appreciate it.

I think you can start by watching the second part of https://drive.google.com/file/d/1KfY0UMH-67Rl_q3IPB0n598v-pljzWj7/view. Then maybe start reviewing PRs and reading the code, as well as asking questions in the weekly meetings :) And when you feel more comfortable, pick an issue and try to solve it would be wonderful!

PR #958, opened 2026-06-11
fix: stabilize filled graph circuit ordering

### LINKED ISSUE #825: Gallery curves show duplicate points for each physical error rate

dannyward630:
## Summary

- canonicalize correlation surfaces before NetworkX greedy coloring
- canonicalize stabilizers within each clique and preserve stable X/Y/Z output ordering
- add a regression test that reverses correlation-surface iteration and compares the generated Stim circuits structurally

## Why

Sorting only after coloring is too late: unordered surface iteration can change graph node insertion order and therefore the greedy-color result. That produces different circuit strong IDs for equivalent gallery circuits and leaves duplicate points in the error-rate curves.

This takes the pre-coloring canonicalization approach discussed in #825 and avoids string or hash snapshots in the regression test.

## Checks

- `uv run pytest tests -q` (758 passed, 257 deselected)
- `uv run ruff check src/tqec/computation/open_graph.py tests/computation/open_graph_test.py` (passed)
- `uv run ty check src/tqec/computation/open_graph.py tests/computation/open_graph_test.py` (passed)
- `git diff --check` (passed)

Closes #825.

## AI disclosure

I used an LLM as a collaborative tool for repository triage, test design, and reviewing the ordering edge cases. I reproduced the nondeterminism before the fix, implemented and reviewed the patch locally, and ran the checks listed above.

### DIFF
--- a/src/tqec/computation/open_graph.py
+++ b/src/tqec/computation/open_graph.py
@@ -96,9 +96,12 @@ def fill_ports_for_minimal_simulation(
         )
 
     correlation_surfaces = graph.find_correlation_surfaces()
-    stab_to_surface: dict[str, CorrelationSurface] = {
-        s.external_stabilizer_on_graph(graph): s for s in correlation_surfaces
-    }
+    stab_to_surface: dict[str, CorrelationSurface] = dict(
+        sorted(
+            (surface.external_stabilizer_on_graph(graph), surface)
+            for surface in correlation_surfaces
+        )
+    )
     generators = list(stab_to_surface.keys())
 
     if search_small_area_observables:
@@ -132,9 +135,17 @@ def fill_ports_for_minimal_simulation(
             g.add_edge(i, j)
     # Solve with heuristic greedy coloring
     coloring = nx.algorithms.coloring.greedy_color(g)  # type: ignore[invalid-argument-type]
-    cliques: dict[int, list[str]] = {}
+    clique_map: dict[int, list[str]] = {}
     for node, color in coloring.items():
-        cliques.setdefault(color, []).append(generators[node])
+        clique_map.setdefault(color, []).append(generators[node])
+    pauli_order = {"X": 0, "Y": 1, "Z": 2}
+    cliques = sorted(
+        (sorted(clique) for clique in clique_map.values()),
+        key=lambda clique: (
+            min(pauli_order[p] for stabilizer in clique for p in stabilizer if p != "I"),
+            clique,
+        ),
+    )
 
     def ports_basis_for_clique(
         supported_stabilizers: list[str],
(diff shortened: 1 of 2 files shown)


src/tqec/computation/open_graph.py
```diff
@@ -96,9 +96,12 @@ def fill_ports_for_minimal_simulation(
...
     correlation_surfaces = graph.find_correlation_surfaces()
-    stab_to_surface: dict[str, CorrelationSurface] = {
-        s.external_stabilizer_on_graph(graph): s for s in correlation_surfaces
-    }
+    stab_to_surface: dict[str, CorrelationSurface] = dict(
+        sorted(
+            (surface.external_stabilizer_on_graph(graph), surface)
+            for surface in correlation_surfaces
+        )
+    )
```
nelimee, line 104: Could you elaborate on why you made that change? I do not understand why this is needed.

src/tqec/computation/open_graph.py
```diff
@@ -132,9 +135,17 @@ def fill_ports_for_minimal_simulation(
...
-    cliques: dict[int, list[str]] = {}
+    clique_map: dict[int, list[str]] = {}
     for node, color in coloring.items():
-        cliques.setdefault(color, []).append(generators[node])
+        clique_map.setdefault(color, []).append(generators[node])
+    pauli_order = {"X": 0, "Y": 1, "Z": 2}
+    cliques = sorted(
+        (sorted(clique) for clique in clique_map.values()),
+        key=lambda clique: (
+            min(pauli_order[p] for stabilizer in clique for p in stabilizer if p != "I"),
```
nelimee, line 145: Why is this sorting key interesting to include here?

nelimee (commented):
I asked a few questions on some of the changes because I do not understand their added value. Initial opinion is that this PR is less interesting than PRs on the same issue but opened before, but the answers to my questions might change that initial opinion.

src/tqec/computation/open_graph.py
jbolns: Contributor noted on #825 that he/she started by replicating the matter. Is there anything specific in the replication that suggests this pre-sorting is needed?

HaoTy:
Note that `fill_ports_for_minimal_simulation` calls `find_correlation_surfaces` for correlation surfaces instead of taking them from an input argument. I believe the current correlation surface algorithm is deterministic, and the returned list is sorted by the span:
https://github.com/tqec/tqec/blob/d23a69c0664d8db142c935bdea664a6e255a09e6/src/tqec/computation/correlation.py#L373-L381

Do you have a reproducible example of the algorithm returning non-deterministic orders?
The `CorrelationSurface` class uses `frozenset`, which may exhibit non-deterministic behaviors in certain cases, but should not matter here.

dannyward630:
Thanks for the review. After checking the current implementation and the other proposals, I agree this branch should not move forward. `find_correlation_surfaces` already returns surfaces sorted by span, so reversing that result in the test does not reproduce a production ordering path. The remaining clique canonicalization overlaps #952, while this branch adds an extra Pauli-priority ordering without a demonstrated need. Closing this rather than adding more speculative ordering code.

PR #976, opened 2026-06-22
Fix #690: Implement an option to control the opacity of blockgraph structures

### LINKED ISSUE #690: Implement an option to control the opacity of blockgraph structures

Mr-Neutr0n:
# Description

Adds an `opacity` parameter (default `1.0`, i.e. unchanged behaviour) to `BlockGraph.to_dae_file`, `BlockGraph.view_as_html`, and `write_block_graph_to_dae_file`. The value scales COLLADA material transparency so block faces can be rendered semi-transparent, which is what makes a single-view 3D export readable when blocks occlude each other.

Correlation surfaces are deliberately excluded from the scaling: `_add_materials` keeps their intrinsic alpha regardless of the block opacity, so lowering block opacity reveals the surfaces instead of fading them out too. That was the issue @smburdick caught in the first review.

Fixes #690

# How Has This Been Tested?

- Added `test_write_block_graph_to_dae_file_opacity` in `tests/interop/collada/read_write_test.py`, which writes a graph at several opacity values and asserts the resulting COLLADA `<transparency>` values, including that correlation-surface materials are unaffected.
- Ran the existing `tests/interop/collada/` suite to confirm no regression in the default (`opacity=1.0`) path.
- Manually exported `cnot` and `three_cnots` from `tqec.gallery` at `opacity=0.3` and confirmed the interior blocks are visible in the rendered HTML view.

**Test Configuration**:
* Python version: 3.12
* Operating system and system architecture: Ubuntu 24.04, x86_64

# Checklist:

* &nbsp; [x] My code follows the style guidelines of this project
* &nbsp; [x] I have commented my code, particularly in hard-to-understand areas
* &nbsp; [x] My code and tests pass locally and have at least 80% line coverage
* &nbsp; [x] Any dependent changes have been merged and published in downstream modules

---
This change was prepared with AI assistance under human direction and review.


smburdick (changes requested):
* I don't think this fully implements #690 because I am unsure whether this will show the correlation surfaces. This code just allows opacity of any block, which I don't think is what we want. You may want to consult @purva-thakre or @nelimee on the full extent of this task though.
* A comment in #690 mentions opacity implemented in piper draw, I think checking against that should be added to this task, since visual aspects of block graphs will be shown first in that service.
* Please add test cases and show the results.

Mr-Neutr0n:
Hi @smburdick, thanks for the detailed review.

You were right that the first cut applied opacity to every material, including the correlation surfaces. I updated `_add_materials` so correlation surface colors keep their intrinsic alpha regardless of the block opacity. That way lowering block opacity lets you see the internal structure without washing out the correlation surfaces.

I also added tests that parse the generated DAE and check: default opacity is fully opaque; block material transparency scales with the opacity parameter; out-of-range opacity values are clamped; and correlation surface transparency stays at 0.8 even when block opacity is 0.5.

On the piper-draw reference: I read the issue comments and it looks like piper draw handles its own rendering. This PR is scoped to the DAE export path that the gallery and docs embed directly in web pages, so the two can coexist. If you want parity with piper draw's exact opacity behavior I am happy to look at that next.

I still was not able to reproduce the original duplicate-key bug on a live instance, so I am relying on the issue report and screenshots for that part. The opacity feature itself is what these tests cover.

Let me know if you would like me to add anything else or split this further.

smburdick:
Another thing: can you fill out the rest of the [pull request template](https://github.com/tqec/tqec/blob/main/.github/pull_request_template.md), and please do so for all future reviews.

Will look at your tests once that's in place.

KabirDubey:
Thanks @Mr-Neutr0n, are you still interested in getting this merged? If so, please address @smburdick 's comment, or at least demonstrate that your implementation works.

Mr-Neutr0n:
@KabirDubey yes, still interested, sorry for the delay on this.

@smburdick I've filled out the full PR template in the description above, and I'll use it from the start on anything future.

Two things also sorted just now:
- pre-commit-ci was red on a single ruff-check finding (a stray blank line after the imports in the test file). Fixed in `2fc6268`, so CI should go green on this run.
- On demonstrating it works: `test_write_block_graph_to_dae_file_opacity` covers it directly, asserting the COLLADA `<transparency>` values at several opacity levels and checking that correlation-surface materials keep their own alpha rather than inheriting the block opacity. That last part is the fix for the problem you flagged in the first review.

Happy to add a rendered before/after screenshot if that's more useful than the test assertions.

Mr-Neutr0n:
Closing this in favour of #997.

@rasumeng asked for this issue on 2026-06-21 and I opened this PR the next day, which I hadn't registered at the time — that was my mistake, and it left two near-identical PRs in the queue for one issue. Given @nelimee's note on #690 about LLM-assisted PRs being a burden to review, two of them for the same feature is precisely the burden worth removing.

It isn't only a question of who asked first: having read both, **#997 is the better patch**. Three concrete differences, all in its favour:

1. **Correlation-surface detection.** I matched on the colour name suffix (`face_color.value.endswith("_CORRELATION")`); #997 uses a `frozenset` of `TQECColor` members. Enum identity won't break the day someone renames a colour.
2. **Range handling.** I silently clamped with `min(max(opacity, 0.0), 1.0)`, so `opacity=5` quietly became `1.0`. #997 raises `TQECError` on out-of-range input, which matches how the rest of the codebase reports bad arguments and surfaces the mistake instead of hiding it.
3. **Tests.** I hand-rolled an ElementTree reader with namespace fallbacks to pull `<transparency>` out of the DAE. #997 just uses `collada.Collada(...)` and reads `effect.transparency` — same assertion, a fraction of the code, and one less thing to maintain.

@smburdick, on your review here: you were right that the first cut washed out the correlation surfaces, and both PRs now preserve their intrinsic alpha. Your other two points still apply to #997 and are worth carrying over — whether scalar block opacity is the whole of #690 or whether correlation-surface display is the real ask, and the comparison against PiperDraw's flows view.

On that first one: I asked the same scoping question on #690 back in June and it's still open. It's the one thing that would benefit from a maintainer's steer before more work goes in, since it decides whether #997 is complete or just the first half.

@rasumeng — sorry for stepping on this one, it's yours. Happy to review #997 if that's useful to you.

nelimee:
> @ rasumeng — sorry for stepping on this one, it's yours. Happy to review #997 if that's useful to you.

That would be super nice of you if you have the time, feel free to review anything you want, the more reviewers we have, the better :)

Issue #967, opened 2026-06-16
dae import failing for rotated Y half cube structures


KabirDubey:
## Describe the bug clearly and concisely

`dae` files containing these structures fail to be imported/parsed as a block graph

<img width="584" height="145" alt="Image" src="https://github.com/user-attachments/assets/36834d24-0eb6-4d51-97bd-9e6a154b31a6" />

`dae` files for each of these structures can be found [here](https://github.com/KabirDubey/tqec/tree/dae-component-reproduction/_reproduction/inputs/dae). The number in the filename `g0{number}` is the structure you see above numbered left to right.  

## Steps to reproduce the bug

I have provided a script to reproduce the issue [here](https://github.com/KabirDubey/tqec/blob/dae-component-reproduction/_reproduction/reproduce.py).

Here's the error message:
```
❯ uv run python reproduce.py
g01 DAE: TQECError: Rounding from 14.444466590881348 to integer failed.
g02 DAE: TQECError: There is an invalid rotation for Y block.
Cultivation and Y blocks should only allow rotation around Z axis.
g03 DAE: TQECError: Rounding from 15.354693412780762 to integer failed.
g04 DAE: TQECError: There is an invalid rotation for Y block.
Cultivation and Y blocks should only allow rotation around Z axis.
g05 DAE: TQECError: There is an invalid rotation for Y block.
Cultivation and Y blocks should only allow rotation around Z axis.
g06 DAE: TQECError: There is an invalid rotation for Y block.
Cultivation and Y blocks should only allow rotation around Z axis.
g07 DAE: TQECError: There is an invalid rotation for Y block.
Cultivation and Y blocks should only allow rotation around Z axis.
g08 DAE: TQECError: There is an invalid rotation for Y block.
Cultivation and Y blocks should only allow rotation around Z axis.
Wrote 0 HTML files to {filepath ommitted}
```

## Describe the expected behavior

These are all valid uses of the Y half cube, and we support import of rotated structures, so there should be no parsing issue. If the parsing issue is fixed, the `reproduce.py` script above should automatically generate an `html` file which you can view to inspect that the block graphs are correct. 

## Additional information

[...]

Issue #934, opened 2026-05-09
Add benchmarking against complex circuits and corner cases


smburdick:
* With new circuit builder tools like Piper Draw and Entropica, it will be easier to make new test cases for tqec. I think larger and more complex circuits for load testing will help us identify more bottlenecks in our compilation pipeline.
* Add `pytest-benchmark` to our testing dependencies.

aks8134:
Hi @smburdick, One of the benchmarks containing circuits for a lot of quantum algorithms is MQT Bench. 
https://github.com/munich-quantum-toolkit/bench. Do you want to meet sometime to discuss on which circuits might be worth adding to the pytest-benchmark?

jbolns:
Hi @smburdick and @aks8134 
I'd be interested in joining if you end up meeting. 
With an incoming update, Topologiq should be able to handle many MQT benchmarks. And TQEC/tqec still needs a few or at least one document delineating how to do the end-to-end logical circuit -> simulation pipeline.

For context. Unless MQT Bench has addded full lattice surgery support, I don't see a path for direct MQT Bench -> TQEC/tqec usage.

smburdick:
> Hi [@smburdick](https://github.com/smburdick), One of the benchmarks containing circuits for a lot of quantum algorithms is MQT Bench. https://github.com/munich-quantum-toolkit/bench. Do you want to meet sometime to discuss on which circuits might be worth adding to the pytest-benchmark?

I think it would be worth bringing up in the TQEC meeting for a few minutes for starters

afowler:
+1 for discussing this in group meeting

PR #978, opened 2026-06-23
Small fixes to Y half cube input and output

KabirDubey:
### Description

Although the `.bgraph`/BGRAPH file type was originally introduced in #864 as a unifying file type for upstream compilers to emit, I've found it useful to specify isolated block graphs which serve as unit tests of faulty circuits (circuits which, in particular configurations, do not achieve desirable circuit distance)[^1]. The existing raw `.dae` and `.json` are difficult to inspect.

This PR merges some methods which I've found useful to have locally while running distance experiments with the Y half cube. 

Specifically, this PR:

1. fixes a minor bug where block graphs involving Y basis initializations failed BGRAPH to COLLADA round trips.

```python
from tqec.computation.block_graph import BlockGraph


graph = BlockGraph("downward_y")
graph.add_cube(Position3D(0, 0, 0), "XZX")
graph.add_cube(Position3D(0, 0, -1), "Y")
graph.add_pipe(Position3D(0, 0, 0), Position3D(0, 0, -1))

bgraph = graph.to_bgraph()
round_tripped = BlockGraph.from_bgraph(bgraph)

assert round_tripped == graph
```

2. fixes the Y half cube i/o beyond the monkey patch in #939. Adds docstrings to cement the standards created in the discussion in #864: half-integer scaling is just for macroscopic visualization, and not an attribute of the circuit represented by a `BlockGraph` object or `.bgraph` file. All that matters for block graphs are block connectivity and kind. Therefore no `read` method (outputting a `BlockGraph` type object) ever needs `offset_y_half_cube` position. 

Such blocks simply need to rounded so that they are (a) distinguished from their neighbors in a graph data structure and (b) don't fail round-trips with representations for visualization[^2]. I think this is true for all blocks, including sliding, MSC, patch rotation. Downstream circuit generation is more complicated and handled separately. 

### Other notes

I/O can def be made more elegant (e.g. we can probably eliminate `shared.py`) and more functional (e.g. `int_position_before_scaling` probably needs to be modified or receive a normalized input to solve #967), but that's beyond the scope of my interests for this PR. 

#### Additional minor features

3. `BlockGraph` can now automatically add missing pipes between adjacent cubes
4. added DAE ⇄ BGRAPH file conversion helpers for interop; the implementation and testing here trivially goes through an instantiated `BlockGraph` but can be made more efficient by going through a text-based algorithm.
5. Renamed `load_bgraph` to `read_bgraph` because the API module is called `read_write`.

### How Has This Been Tested?

I made sure existing tests passed and added some new ones. One of the new tests stacks odd and even number of Y half cubes in the same x,y coordinate to try to test if integer DAE coordinates breaks the read/write but I didn't think too much if this makes sense or if I tested it appropriately.  

The following visual problem flagged in #939:

> There is a separate geometric oddity worth flagging but not fixing here: the writer in write_block_graph_to_dae_file hard-codes the Y-cube shift to dz=0.5 in DAE space, regardless of pipe_length. At pipe_length != 2.0 this produ an ces a visually misplaced Y in the DAE rendering, even though round-trips are self-consistent.

requires human eyes to test. I did so using `view_as_html` over block graphs with `pipe_lengths != 2.0` and observed a problem that this PR does not resolve, and actually seems unrelated to `pipe_length`. Will make a separate issue. EDIT: See #979 

**Test Configuration**:
* Python version: 3.12.5
* Operating system and system architecture: macOS 14.7.8, arm64

# Checklist:

* &nbsp; [x] My code follows the style guidelines of this project
* &nbsp; [x] I have commented my code, particularly in hard-to-understand areas
* &nbsp; [x] My code and tests pass locally and have at least 80% line coverage
* &nbsp; [x] Any dependent changes have been merged and published in downstream modules

Played around with copilot and CodeRabbit for code review on my fork but was not able to do anything productive. Used LLM to fix failing tests.

[^1]: I call these special packagings "gadgets," but just to keep things light, this PR doesn't introduce that terminology/abstraction.

[^2]: it's an injection from your visualization space to a 3d lattice graph (lattice meaning integer valued).


jbolns:
@KabirDubey Looks good at first sight. To save time, I'll review more closely after the following clarification questions, mostly for context to know how deep to go. 
- Did you test this using the JSON importer in https://github.com/tqec/tqec-integrations ? That thing is so dated I'd be surprised if it still works. So essentially, I'm just curious as to whether you could materialise in SketchUp a blockgraph imported from a BGRAPH file. 
- Is this sync-ed this to the changes in #815 ?

smburdick:
This isn't a full review, but I think I'd feel a lot better if we demonstrated the Collada <-> .bgraph interoperability at scale. AFAIK, tools like Piper Draw and Entropica export to DAE; has there been a discussion as to whether bgraph will be a preferred format? (which I think it should be, given that it's a lot easier to parse and possibly less error prone than DAE, which was made for non-quantum use cases)

jbolns:
@smburdick 

> has there been a discussion as to whether bgraph will be a preferred format?

I remember Piper Draw folks saying they were looking into BGRAPH, but don't know what happened with that.

The json importer/exporter SketchUp plugin could be updated to use BGRAPH instead of JSON. This might also allow getting rid of JSON import/exports in TQEC, which I think @HaoTy would like. It is a small repository currently forgotten, but it deserves some love.

KabirDubey:
Thank you.

> @KabirDubey Looks good at first sight. To save time, I'll review more closely after the following clarification questions, mostly for context to know how deep to go.
> 
>     * Did you test this using the JSON importer in https://github.com/tqec/tqec-integrations ? That thing is so dated I'd be surprised if it still works. So essentially, I'm just curious as to whether you could materialise in SketchUp a blockgraph imported from a BGRAPH file.

I did not, but will do today. Since we support `.bgraph` to `.dae` and user of SketchUp should be able to import `.dae`, I would expect that to be the workflow. Or `.dae` import to SketchUp not available to SketchUp 8 users? No worries if you don't know off the top of your head, I can check later on my own time. My current inclination is that `tqec-integrations` repo should be depreciated and that `tqec` should host the `.bgraph` API as part of its design automation service. 

>     * Is this sync-ed this to the changes in [Update sketchup template #815](https://github.com/tqec/tqec/pull/815) ?

The two should not conflict as PRs. The `.bgraph` file type needs to be updated to support patch rotation and MSC--I'm keeping track of this stuff and will make a separate issue as part of my summer cleanup.

jbolns:
I'm going to have to leave actual review for tomorrow or Friday, but general reactions/replies below. 

> My current inclination is that tqec-integrations repo should be depreciated and that tqec should host the .bgraph API as part of its design automation service.

Yeah, I'd support that. 

> Since we support .bgraph to .dae and user of SketchUp should be able to import .dae, I would expect that to be the workflow.

That is the ideal workflow, yes. 

No need to test the JSON importer plugin for this issue then. In any case, agree that if not an update then to work with BGRAPH then deprecation is appropriate. 

> .dae import to SketchUp not available to SketchUp 8 users?

I would almost support deprecation, except, indeed, DAE imports are not available on v8. The plugin is the only way to put a blockgraph into v8. It's a very short plugin, tho, so it can also be put in the assets folder.

> The .bgraph file type needs to be updated to support patch rotation and MSC--I'm keeping track of this stuff and will make a separate issue as part of my summer cleanup.

Yes, then defo not relevant to this issue.

KabirDubey:
Thanks Sam!!

> This isn't a full review, but I think I'd feel a lot better if we demonstrated the Collada <-> .bgraph interoperability at scale.

I can make an issue for sophisticated profiling and stress testing, perhaps a sub-issue of #934 but focusing on DAE ⇄ BGRAPH round-trip of large block graphs. This can be entirely text-based so a maybe good first issue would be bypassing the instantiation of `BlockGraphs` in `src/tqec/interop/converters.py` (I mention this in the file header). [Rust packages can be particularly good at this](https://github.com/attilio-oliva/converter-buddy). Does this address your concern?

In my mind, the reasonable scale in question for this pipeline is not of the same order of magnitude as the scale expected to stress test `Topologiq`'s ZX graph translation or `tqec` and `tqecd`'s circuit generation. The size of the expected input is bounded above by what a human being would be capable of inspecting and/or hand-compiling. Thus I'm not worried about `.braph` or `.dae` files expressing structures larger than say the block graphs in [Appendix A here](https://arxiv.org/pdf/2505.15917).  

> AFAIK, tools like Piper Draw and Entropica export to DAE; has there been a discussion as to whether bgraph will be a preferred format? (which I think it should be, given that it's a lot easier to parse and possibly less error prone than DAE, which was made for non-quantum use cases)

Yeah it depends on if their endpoint is a DAE file that can be interpreted as a `tqec` block graph. If they want to use `tqec` downstream then it'd be easiest for them to build such a bridge.

KabirDubey:
> I would almost support deprecation, except, indeed, DAE imports are not available on v8. The plugin is the only way to put a blockgraph into v8. It's a very short plugin, tho, so it can also be put in the assets folder.

Okay thanks, will make a new issue to depreciate `tqec_integrations` and move the plug-in to the assets folder.

jbolns:
@KabirDubey 

Gave it a closer look. I think it makes sense. 

A couple clarity checks to be sure I understood before I approve.

Implementation-wise, the core of the modification here is that you're **not** repositioning or re-scaling anything for BGRAPH in/out, and **yes** for DAE. This maintains the "nice" factor of being able to see half cubes in DAE visualisations but removes the complexity of transformation for BGRAPH. Correct? 

By extension, the core value add is that BGRAPH becomes a more literal reflection of the BlockGraph, correct?

jbolns:
> AFAIK, tools like Piper Draw and Entropica export to DAE; has there been a discussion as to whether bgraph will be a preferred format? (which I think it should be, given that it's a lot easier to parse and possibly less error prone than DAE, which was made for non-quantum use cases)

Following on a previous response. Confirmed, the DEV version of Piper Draw supports BGRAPH.
https://piper-draw-dev.walruscomputing.com/

@smburdick

KabirDubey:
> Implementation-wise, the core of the modification here is that you're not repositioning or re-scaling anything for BGRAPH in/out, and yes for DAE. This maintains the "nice" factor of being able to see half cubes in DAE visualisations but removes the complexity of transformation for BGRAPH. Correct?

Yes but I didn't think of this PR as having one single core modification--what's in the PR description is what needs review.

> By extension, the core value add is that BGRAPH becomes a more literal reflection of the BlockGraph, correct?

Yes. I think BGRAPH and BlockGraph being the same is what the file type should aspire to be. This was my interpretation of the discussion set in #864 (not this PR). I think this by far the most stable vision of what a `.bgraph` file is--a text-based version of the `BlockGraph` intermediate representation, roughly analogous design purpose to a `.ll` file. 

> Beyond being implemented as a language, LLVM IR is actually defined in three isomorphic forms: the textual format above, an in-memory data structure inspected and modified by optimizations themselves, and an efficient and dense on-disk binary "bitcode" format. The LLVM Project also provides tools to convert the on-disk format from text to binary: llvm-as assembles the textual .ll file into a .bc file containing the bitcode goop and llvm-dis turns a .bc file into a .ll file.

[source](https://aosabook.org/en/v1/llvm.html)

I'm generally of the opinion that having a text-based IR will greatly speed up testing and development, especially as we progress to dynamic circuits and bigger workloads. Block graphs are typed graphs. The generation of circuits can and should be a separate aspect of development from visualization. Additional complexity, like pipe-length scaled visualization, the parameterized dynamic generation of `Stim` circuits, etc. would be handled at a lower level. Indeed, this expects the IR to be interpretable for this stuff (e.g. we'd need to write conditional logic into the .bgraph text file for conditional cubes), and it doesn't all need to happen now, but hopefully the vision and utility is clear.

KabirDubey:
Made discussion post https://github.com/tqec/tqec/discussions/989 to discuss block graphs as an IR more

jbolns:
Am I delaying this? I lost track. 
I think I'm fine with everything. Any recent changes or is it last-commit ready to go give it a final look before hopefully approving?

KabirDubey:
> Am I delaying this? I lost track. I think I'm fine with everything. Any recent changes or is it last-commit ready to go give it a final look before hopefully approving?

No more commits planned on my end. No rush if you need more time to review. Otherwise please approve.

jbolns:
Yeah, just went over it again and I'm good to go. Mostly, the tests cover for changes introduced, so if the tests clear, then all should be fine. 

Just one question for @HaoTy before clicking the magic button -> All good with the new method introduced in `src/tqec/computation/block_graph.py`?

src/tqec/computation/block_graph.py
```diff
@@ -243,6 +243,20 @@ def add_pipe(
...
+        """Add a pipe between every adjacent pair of cubes that does not already have one.
+
+        Pipe kind is inferred via :py:meth:`~tqec.computation.pipe.Pipe.from_cubes`, which
+        detects Hadamard transitions from face-basis mismatches. Raises
+        :py:class:`~tqec.utils.exceptions.TQECError` if any adjacent pair of cubes has
+        ambiguous or incompatible face bases.
+        """
+        positions = self.occupied_positions
+        for i, p1 in enumerate(positions):
+            for p2 in positions[i + 1 :]:
```
HaoTy, line 256: Seems like a useful method, but the quadratic iteration can be made linear.

jbolns:
Since I like to layer things into short deliverables, I'm up for merging and closing this as soon as the comment above by @HaoTy is addressed. 

I need to note for completeness, however, that #979 might require changing more or less the same lines that are being changed here. 

@afowler @nelimee

src/tqec/computation/block_graph.py
KabirDubey: Addressed [here](https://github.com/tqec/tqec/pull/1008/changes#diff-a6d234df82a92d5536948017eaa5df90a6b9731b1e330fbfe524190b33ccf680R254) in a stacked PR #1008.

KabirDubey:
> Since I like to layer things into short deliverables, I'm up for merging and closing this as soon as the comment above by @HaoTy is addressed.
> 
> I need to note for completeness, however, that #979 might require changing more or less the same lines that are being changed here.
> 
> @afowler @nelimee

Thanks @jbolns , @HaoTy's comment is addressed in #1008, can you approve this for merge?

src/tqec/computation/block_graph.py
jbolns: I confirmed #1008 touches this exact same lines. Additional improvement can be discussed there. Resolving here.

jbolns (approved):
Confirmed the change suggested by @HaoTy is addressed in a related PR.
I don't see a need to check thoroughly again: taking your word that PRs after last review of mine didn't change dis-improved the rest of the PR which was already in line with expectations. If that's not the case, I guess we'll both get in troubles.
Cheers!

PR #980, opened 2026-06-25
docs: fix typos, broken cross-references, and terminology figure template

rturrado:
# Description

A pass over the README and the user guide / architecture pages to fix small but real issues found while reading through the docs. No code or behavior changes: docs only.

- **README.md**: typo fixes: duplicate `for for`, and `facilite` to `facilitate`.
- **docs/architecture.rst**: minor wording (`through` to `via` / `by using`), `hadamard` to `Hadamard`, and inline-code formatting for `tqec`.
- **docs/user_guide/terminology.rst**: corrected the example template in the Template section so the figure (rows 2 to 5 with 5 plaquettes) matches the surface image. Also: `inplace` to `in-place`, duplicate `can` removed, `These condition` to `These conditions`, and added a `_port:` anchor so the Port section can be cross-referenced.
- **docs/user_guide/bgraph.rst**: replaced two broken links with proper Sphinx cross-references (`pipe <./terminology.rst>` to `:ref:Pipe <pipe>`, external `gallery/steane_encoding.html` URL to `:doc:/gallery/steane_encoding`), linked the bare `PORT` mention to the new `_port:` anchor, and added small `<br>` spacers so text doesn't render flush against the `view_as_html()` iframes.

No new dependencies.

Fixes # (no issue)

Please add the `documentation` label.

# How Has This Been Tested?

Docs built locally with sphinx-autobuild and visually inspected in a browser:

```
cd docs && uv run sphinx-autobuild . _build/html
```

Verified:

- The build produces no new warnings.
- The new "Port" link in `bgraph.rst` lands on the Port section in `terminology.rst`.
- The "Steane code example" link in `bgraph.rst` lands on the local gallery page, not the external URL.
- The figure-vs-template mismatch in `terminology.rst` is gone (visual check against the surface image).
- All the lists in `bgraph.rst` are displayed correctly.
- Text that follows the HTML figures (`view_as_html()` iframes) in `bgraph.rst` is now displayed with proper vertical separation.
- Inline `tqec` in `architecture.rst` renders as monospaced code, not plain text.

**Test Configuration**:
* Python version: 3.12.3
* Operating system and system architecture: Ubuntu 24.04 (Linux 6.17, x86_64)

# Checklist:

* &nbsp; [ ] My code follows the style guidelines of this project
* &nbsp; [ ] I have commented my code, particularly in hard-to-understand areas
* &nbsp; [ ] My code and tests pass locally and have at least 80% line coverage
* &nbsp; [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/docs/user_guide/bgraph.rst
+++ b/docs/user_guide/bgraph.rst
@@ -4,7 +4,7 @@ BGRAPH is a work-in-progress file format (`.bgraph`) that can be used to facilit
 
 While only a preliminary minimum baseline in need of improvement, the current BGRAPH specification already enjoys the following advantages:
 
-- Common format both `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce.
+- Common format that `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce.
 - Human-readable (QASM-like) format that is easy to inspect.
 - Clear structure that allows parsing using.
 - Sufficiently complete as to enable read/write of any arbitrary block graph :code:`tqec` can currently handle.
@@ -99,12 +99,21 @@ Contains the information needed to translate each line of this section into a :c
 
 The information in this section is meant to be parsed and must contain, at a minimum:
 
-- ``src``: the ID of the cube at which the pipe starts (the source for the edge represented by the pipe).
-- ``tgt``: the ID of the cube at which the pipe ends (the target for the edge represented by the pipe).
+- ``src``: the ID of one of the pipe's two endpoints.
+- ``tgt``: the ID of the other endpoint.
 - ``kind``: the kind of the pipe (see :ref:`Pipe <pipe>` for all possibilities.).
 
 Each Pipe item should be given as a CSV-separated sequence.
 All separating semi-colons should be included and all fields should be given.
+
+.. note::
+
+    The block graph is **undirected**.
+    The ``src`` and ``tgt`` labels are only a writing convention for the two endpoints of a pipe;
+    swapping them describes the same edge, and no ordering or flow is implied.
+    Pipes represent spacetime volumes between two cubes and are not inherently directional.
+    The one exception is Hadamard pipes, whose *head* and *tail* ends carry a visualisation-only meaning
+    inherited from the DAE importer (which end is the rotated one).
 
 .. admonition:: Example
 
@@ -145,7 +154,6 @@ For instance, it is not allowed to use the word ``source`` as the name of the gr
 .. warning::
 
     This constraint is part of the BGRAPH specification but is not yet enforced by the :code:`tqec` parser.
-    Using a reserved term as a value will currently parse without error, but the result is undefined.
 
 Lastly, while probably obvious due to the nature of lattice surgery and the use of semi-colon-separated-values:
 
--- a/docs/user_guide/bgraph.rst
+++ b/docs/user_guide/bgraph.rst
@@ -1,26 +1,27 @@
 BGRAPH Schema
-========================
-
+=============
 BGRAPH is a work-in-progress file format (`.bgraph`) that can be used to facilitate interoperability between external lattice surgery (LS) tools.
 
 While only a preliminary minimum baseline in need of improvement, the current BGRAPH specification already enjoys the following advantages:
 
-- Common format both `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce
-- Human-readable (QASM-like) format that is easy to inspect
-- Clear structure that allows parsing using
-- Sufficiently complete as to enable read/write of any arbitrary blockgraph :code:`tqec` can currently handle.
+- Common format both `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce.
+- Human-readable (QASM-like) format that is easy to inspect.
+- Clear structure that allows parsing using.
+- Sufficiently complete as to enable read/write of any arbitrary block graph :code:`tqec` can currently handle.
 
-We welcome proposals for improved specifications that expand upon these minimum advantages. In the ideal scenario, we foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
+We welcome proposals for improved specifications that expand upon these minimum advantages.
+In the ideal scenario, we foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
 
 Schema
--------
+------
 A BGRAPH file can be conceived as a QASM-like representation of a lattice surgery.
 
 Any BGRAPH file should be divided into four main sections:
-- HEADER: The title of the document, with information about the source tool used to produce it.
-- METADATA: A short section with meta information on top
-- CUBES: A subsequent section with detailed information about the cubes in the blockgraph
-- PIPES: A final section with detailed information about the pipes in the blockgraph.
+
+- ``Header``: the title of the document, with information about the source tool used to produce it.
+- ``Metadata``: a short section with meta information on top.
+- ``Cubes``: a subsequent section with detailed information about the cubes in the block graph.
+- ``Pipes``: a final section with detailed information about the pipes in the block graph.
 
 An example BGRAPH file is available from :code:`tqec.assets`.
 
@@ -38,14 +39,15 @@ The information in this section is for identification purposes and is not meant 
 
 Metadata
 ~~~~~~~~
-Contains the information needed to produce a blockgraph using the information available in other sections.
+Contains the information needed to produce a block graph using the information available in the other sections.
 
 The information in this section is parseable, but the specific fields are optional:
-- source: The tool used to create the BGRAPH.
-- name of the circuit: used to give the blockgraph a reference name.
-- length of pipes used in the source tool: The length of the pipes between any two adjacent cubes (see note at the end of this section).
 
-Each METADATA item should be given as a CSV-separated pair. If a field is not included, TQEC will assign default values.
+- ``source``: the tool used to create the BGRAPH.
+- ``circuit_name``: used to give the block graph a reference name.
+- ``pipe_length``: the length of the pipes between any two adjacent cubes (see note at the end of this section).
+
+Each Metadata item should be given as a CSV-separated pair. If a field is not included, ``tqec`` will assign default values.
 
 .. admonition:: Example
 
@@ -58,19 +60,23 @@ Each METADATA item should be given as a CSV-separated pair. If a field is not in
 
 .. note::
 
-    Fields in this section are optional. Note, in particular, that if `pipe_length` is not given, the TQEC parser will default to `pipe_length = 0.0`.
+    Fields in this section are optional. Note, in particular, that if `pipe_length` is not given, the ``tqec`` parser will default to `pipe_length = 0.0`.
 
 Cubes
 ~~~~~
-Contains the information needed to translate each line of this section into a :code:`tqec` `cube <./terminology.rst>`.
+Contains the information needed to translate each line of this section into a :code:`tqec` :ref:`Cube <cube>`.
 
 The information in this section is meant to be parsed and must contain, at a minimum:
-- index: the ID of the cube, ideally an integer but alphanumeric IDs also possible.
-- position: The (x, y, z) position of the cube, a tuple of integers.
-- kind: The kind of the cube, as a string (see `Terminology <https://tqec.github.io/tqec/user_guide/terminology.html>`_, as well as the note below).
-- label (optional): An optional annotation that is typically used to denote when a cube is a PORT.
 
-Each CUBE item should be given as a CSV-separated sequence. All separating semi-colons should be included even if the (optional) label field is blank. This helps communicating explicitly to the parser that there is no label (for robustness, the parses *will* fail if an incorrect number of semi-colons is used).
+- ``index``: the ID of the cube, ideally an integer but alphanumeric IDs also possible.
+- ``x``, ``y``, ``z``: the position of the cube, a tuple of integers.
+- ``kind``: the kind of the cube, as a string (see :ref:`Cube <cube>`, as well as the note below).
+
+Optionally, it may also contain a ``label``, an annotation that is typically used to denote when a cube is a :ref:`Port <port>`.
+
+Each Cube item should be given as a CSV-separated sequence.
+All separating semi-colons should be included even if the (optional) label field is blank.
+This helps communicating explicitly to the parser that there is no label (for robustness, the parses *will* fail if an incorrect number of semi-colons is used).
 
 .. admonition:: Example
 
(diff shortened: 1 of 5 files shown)


docs/user_guide/bgraph.rst
```diff
@@ -1,26 +1,27 @@
...
+We welcome proposals for improved specifications that expand upon these minimum advantages.
+In the ideal scenario, we foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
 
 Schema
--------
+------
 A BGRAPH file can be conceived as a QASM-like representation of a lattice surgery.
 
 Any BGRAPH file should be divided into four main sections:
-- HEADER: The title of the document, with information about the source tool used to produce it.
```
smburdick, line 20: These are capitalized in the `.bgraph` schema as-is.

docs/user_guide/bgraph.rst
```diff
@@ -38,14 +39,15 @@ The information in this section is for identification purposes and is not meant
 
 Metadata
 ~~~~~~~~
-Contains the information needed to produce a blockgraph using the information available in other sections.
+Contains the information needed to produce a block graph using the information available in the other sections.
```
smburdick, line 42: I think `BlockGraph` could be the canonical way to refer to it.

docs/user_guide/bgraph.rst
```diff
@@ -58,19 +60,23 @@ Each METADATA item should be given as a CSV-separated pair. If a field is not in
...
-- label (optional): An optional annotation that is typically used to denote when a cube is a PORT.
 
-Each CUBE item should be given as a CSV-separated sequence. All separating semi-colons should be included even if the (optional) label field is blank. This helps communicating explicitly to the parser that there is no label (for robustness, the parses *will* fail if an incorrect number of semi-colons is used).
+- ``index``: the ID of the cube, ideally an integer but alphanumeric IDs also possible.
+- ``x``, ``y``, ``z``: the position of the cube, a tuple of integers.
+- ``kind``: the kind of the cube, as a string (see :ref:`Cube <cube>`, as well as the note below).
+
+Optionally, it may also contain a ``label``, an annotation that is typically used to denote when a cube is a :ref:`Port <port>`.
+
+Each Cube item should be given as a CSV-separated sequence.
```
smburdick, line 77: I prefer the old layout.

docs/user_guide/bgraph.rst
```diff
@@ -82,22 +88,23 @@ Each CUBE item should be given as a CSV-separated sequence. All separating semi-
...
+Contains the information needed to translate each line of this section into a :code:`tqec` :ref:`Pipe <pipe>`.
 
 The information in this section is meant to be parsed and must contain, at a minimum:
-- index: the ID of the cube, typically an integer.
-- u: The ID of the cube at which the pipe starts (the source for the edge represented by the pipe).
-- v: The ID of the cube at which the pipe ends (the target for the edge represented by the pipe).
-- kind: The kind of the pipe (see `Terminology <https://tqec.github.io/tqec/user_guide/terminology.html>`_ for all possibilities.).
 
-Each PIPE item should be given as a CSV-separated sequence. All separating semi-colons should be included and all fields should be given.
+- ``src``: the ID of the cube at which the pipe starts (the source for the edge represented by the pipe).
```
smburdick: I'm actually not 100% sure whether block graph edges are directed or not.

docs/user_guide/bgraph.rst
```diff
@@ -113,24 +120,41 @@ General
...
+
+.. warning::
+
+    This constraint is part of the BGRAPH specification but is not yet enforced by the :code:`tqec` parser.
+    Using a reserved term as a value will currently parse without error, but the result is undefined.
+
+Lastly, while probably obvious due to the nature of lattice surgery and the use of semi-colon-separated-values:
+
+- the strings used as kinds for either cubes or pipes (e.g. ``zxz``, ``xxz``, ``oxz``, ``ozxh``, etc) should not be used for any other purpose.
 - no additional semi-colons can be used anywhere in the field except as separators for the specific fields in each of the lines in metadata, cubes, or pipes sections.
```
smburdick: "semicolons"

docs/user_guide/bgraph.rst
```diff
@@ -173,7 +208,7 @@ In the example below, we print the BGRAPH string. To produce a `.bgraph` file wi
     from tqec.gallery import cnot
     from tqec.utils.enums import Basis
 
-    # Import blockgraph from gallery
+    # Import BlockGraph from gallery
```
smburdick, line 219: See, you used it here.

docs/user_guide/terminology.rst
```diff
@@ -235,14 +232,14 @@ by convention.
...
+   ``3``, ``4``, ``5``, ``8``, ``11`` and ``14`` to the "no plaquette" index ``0``
    and print ``0`` with ``.`` for visual clarity::
 
       .  .  6  .  6  .
-      7  9 10  9 10 11
-      . 10  9 10  9  .
-      7  9 10  9 10 11
-      . 10  9 10  9  .
+      7  9 10  9 10  .
+      . 10  9 10  9 12
```
smburdick, line 240: Shouldn't it still be 11?

docs/user_guide/bgraph.rst
```diff
@@ -1,26 +1,27 @@
...
+=============
 BGRAPH is a work-in-progress file format (`.bgraph`) that can be used to facilitate interoperability between external lattice surgery (LS) tools.
 
 While only a preliminary minimum baseline in need of improvement, the current BGRAPH specification already enjoys the following advantages:
 
-- Common format both `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce
-- Human-readable (QASM-like) format that is easy to inspect
-- Clear structure that allows parsing using
-- Sufficiently complete as to enable read/write of any arbitrary blockgraph :code:`tqec` can currently handle.
+- Common format both `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce.
```
smburdick: "Common format that"

docs/user_guide/bgraph.rst
```diff
@@ -1,26 +1,27 @@
...
+- Human-readable (QASM-like) format that is easy to inspect.
+- Clear structure that allows parsing using.
+- Sufficiently complete as to enable read/write of any arbitrary block graph :code:`tqec` can currently handle.
 
-We welcome proposals for improved specifications that expand upon these minimum advantages. In the ideal scenario, we foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
+We welcome proposals for improved specifications that expand upon these minimum advantages.
+In the ideal scenario, we foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
```
smburdick: Remove "in the ideal scenario"

smburdick (changes requested):
It's definitely better in a lot of places, but I'd run this past @jbolns since he's the original author of the bgraph schema.

jbolns:
100/100 @smburdick recommendations.

docs/user_guide/bgraph.rst
rturrado: Fair enough. Just to explain my change here. First, I noticed `HEADER` wasn't actually a reserved keywork (alike `METADATA`, `CUBES` and `PIPES`). And then my intention was to use a more generic term to refer to each section, e.g., `Header` or `Metadata`, instead of employing the reserved keywords.

docs/user_guide/bgraph.rst
rturrado: OK, I need to explain this change also. First, I wanted to distinguish between a generic block graph, e.g., what could go in a `.bgraph` file, and a `BlockGraph`, which is the class name used in `tqec` to represent a block graph. Second, I thought of `block graph` as a  better term than `blockgraph`. So I tried to follow that idea across the document: just use `BlockGraph` when we are referring to an instance of `tqec`'s `BlockGraph`, otherwise falling back to `block graph`.

docs/user_guide/bgraph.rst
rturrado: There are 2 changes here:
- Changing `position` to `x, y, z`: I just pretended to be more consistent with the lists in the rest of the document where there is a correspondence between the bullets and the different fields.
- The other is taking `label` out of the list. So instead of saying: the minimal set of attributes is a, b, c, and d (optional); we are now saying: the minimal set of attributes is a, b, and c; optionally you also can have d.

docs/user_guide/bgraph.rst
rturrado: But a Pipes section has this header: `PIPES: src;tgt;kind;`. I just changed the bullet list to use the field names you would find in a bgraph, rather than `u` and `v`.

docs/user_guide/bgraph.rst
rturrado: Good catch. I'll add that, thanks.

docs/user_guide/bgraph.rst
rturrado: Yes, because here we are importing an instance of `tqec`'s `BlockGraph` class.

docs/user_guide/terminology.rst
rturrado: I don't think so.

We start from this template:

```
1  5  6  5  6  2
7  9 10  9 10 11
8 10  9 10  9 12
7  9 10  9 10 11
8 10  9 10  9 12
3 13 14 13 14  4
```

And then we say:

> To see the correspondence more clearly, one can map the indices [...] to the "no plaquette" index 0 and print 0 with . for visual clarity:

The change I am proposing here is to map index 11, and not 12, to 0.

Why? Because there are no plaquettes at index 11. If you check the image, rows 2 to 5 (starting at 1) have 5 plaquettes (7 9 10 9 10). Current upstream's template (after the mapping to 0) has 6 plaquettes at rows 2 and 4 (7 9 10 9 10 11), and only 4 plaquettes at rows 3 and 5 (10 9 10 9).

docs/user_guide/bgraph.rst
rturrado: Cool, OK!

docs/user_guide/bgraph.rst
rturrado: Sure!

rturrado:
@smburdick @jbolns Thanks for the reviews! I have answered some of the comments explaining the reason behind some changes, just to give you more info, but last word is on you. I'm happy to take back or modify whatever points you finally consider.

docs/user_guide/bgraph.rst
```diff
@@ -113,24 +120,41 @@ General
...
+
+    .. code-block:: text
+
+        METADATA: attr_name; value;
+        circuit_name; source;
+
+.. warning::
+
+    This constraint is part of the BGRAPH specification but is not yet enforced by the :code:`tqec` parser.
+    Using a reserved term as a value will currently parse without error, but the result is undefined.
```
jbolns: Check `tests/interop/bgraph/read_write_test.py` for examples to ensure the phrasing of this sentence is correct. My recollection is that this first implementation was purposely designed to let you get away with as most errors as possible, which I think means reserved words are flagged so that people won't use a term that *can* but won't necessarily cause a failure or undefined result.

jbolns:
@rturrado . Thanks for looking critically at the matter. Definitely on the positive side, for me. 
I'll defer final choices to @smburdick since it looks he is managing very well.

docs/user_guide/bgraph.rst
rturrado: We can remove that last line. Current implementation won't complain (warn or error) if a reserved word is used anywhere as a value. But docs do warn you about that going against the BGRAPH spec.

docs/user_guide/bgraph.rst
rturrado: If I look at the codebase, `blockgraph` is mainly used at `bgraph/read_write{,_test}.py` and a couple of times at `blockgraph{,_test}.py`. And `block graph` elsewhere I would say.

docs/user_guide/bgraph.rst
```diff
@@ -1,26 +1,27 @@
...
-- Clear structure that allows parsing using
-- Sufficiently complete as to enable read/write of any arbitrary blockgraph :code:`tqec` can currently handle.
+- Common format both `Topologiq <https://github.com/tqec/topologiq>`_ and `TopoLS <https://github.com/tqec/TopoLS>`_ can produce.
+- Human-readable (QASM-like) format that is easy to inspect.
+- Clear structure that allows parsing using.
+- Sufficiently complete as to enable read/write of any arbitrary block graph :code:`tqec` can currently handle.
 
-We welcome proposals for improved specifications that expand upon these minimum advantages. In the ideal scenario, we foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
+We welcome proposals for improved specifications that expand upon these minimum advantages.
+We foresee future versions of BGRAPH being flexible enough as to enable interoperability *at any stage* of the LS process, which the current version cannot yet do.
```
KabirDubey, line 13: Maybe a better question for @jbolns, but I don't understand what is meant by the LS process? `Topologiq` instantiates a `BlockGraph` object, and then every subsequent rewrite or transformation will be over the `BlockGraph` data structure. A `.bgraph` is simply a text-based interpretation of the same data structure, so any gaps in the translation `BlockGraph` to/from `.bgraph` should be explicit GitHub issues.

docs/user_guide/bgraph.rst
KabirDubey: > First, I wanted to distinguish between a generic block graph, e.g., what could go in a .bgraph file, and a BlockGraph, which is the class name used in tqec to represent a block graph.

My vote is that there should be no semantic distinction between the two. They're two presentations of the same graphical data structure: one's text-based, the other is a graph designed for intended for compilation.    

"BlockGraph" is an instance of `tqec`'s `BlockGraph` class, "block graph" is a kind of ZX graph presenting TQEC system operations over a 2D NN architecture in spacetime, and "blockgraph" is a typo.

docs/user_guide/bgraph.rst
jbolns: Fair question @KabirDubey 
I see BGRAPH as transitional because the final goal is a format you can actually use at any point between Topologiq and the final circuit to run in hardware. Hence, I speak of it as related to the full lattice surgery process. 
I agree proposals for improvement should be discussed explicitly.
I don't think this should deter merging this PR, in that it improves the wording that is already there.

docs/user_guide/bgraph.rst
smburdick: The actual [graph type](https://github.com/tqec/tqec/blob/437118ae961fdf6399d158e5828ed948085c8b0f/src/tqec/computation/block_graph.py#L65) is the undirected kind, though. I do think that's worth explicitly documenting, even if the bgraph spec as-is makes it a little ambiguous. That's because edges are just pipes that span spacetime volume, which isn't inherently directed. Though I assume such an aspect may be inferred from the correlation surfaces though.

jbolns: @smburdick is completely correct here.

The confusion traces all the way back to the use of and "head" and "tail" position for pipes in the [original DAE importer](https://github.com/tqec/tqec/blob/main/src/tqec/interop/collada/read_write.py). So, for instance, a Hadamard pipe's kind corresponds to its "lower" end, i.e., the "head", whereas the "tail" is the rotated bit. This makes them sort of directional at least for visualisation purposes.

However, the use of appropriately directed graph terms is, as he suggests, at the very least, ambiguous. 

Maybe worth considering clarifying explanations to match the (u,v) rationale as opposed to using (src, tgt). As long as it's done consistently across the schema, I don't see a downside.

Apologies @rturrado for any additional time this might impose on you. It was entirely my mistake for not clarifying this appropriately in the wording.

docs/user_guide/bgraph.rst
KabirDubey: Agree that this is out of the scope of this PR. Can move to here:  https://github.com/tqec/tqec/discussions/989 

Still not 100% sure if I understand what you mean (describing _everything_ in between Topologiq and hardware executable in one abstraction level?).

docs/user_guide/bgraph.rst
rturrado: I have updated the description of `src` and `tgt` and added a note to explicitely say that graphs, and pipes, have no direction (only a Hadamard pipe having a *head* and a *tail*).

rturrado:
@smburdick @KabirDubey many of the changes should be resolved/outdated now. Can you please re-review? Thanks!

PR #984, opened 2026-06-28
Nondet additions

KabirDubey:
local review


KabirDubey:
whoops I think I did this wrong, my bad

PR #1008, opened 2026-07-25
Support batched and orchestrated circuit production and simulation 

### LINKED ISSUE #967: dae import failing for rotated Y half cube structures

KabirDubey:
This PR supports batch processing of block graphs for `tqec` benchmarking. Here, batch processing means that a block graph object or DAE file specifying `N` disjoint block graphs is emitted as a collection of `N` disjoint block graph objects.

On top of that input layer, this PR adds a two-stage orchestration API in `tqec.orchestration`: a `prepare_batch` stage that splits, validates, and compiles every gadget and writes noiseless circuits plus a `manifest.json`, and a `simulate_batch` stage that reads the manifest and flattens every case into a single `sinter.collect`, writing per-gadget `results.json`. The two stages communicate only through run-directory files, so preparation and simulation can run as separate scheduler jobs.

This PR includes fixes to DAE file import due to @ica574, quoting https://github.com/KabirDubey/tqec/pull/4:

> I narrowed the issue down to a combination of affine transformation decomposition, rotation handling, and lattice-origin reconstruction during DAE importation. I've implemented a fix that imports all eight DAEs without errors, while keeping all other tests running successfully.

This PR therefore closes https://github.com/tqec/tqec/issues/967.

This PR is stacked on https://github.com/tqec/tqec/pull/978 and is based on that PR's head branch, `KabirDubey:main`. That parent already updates `add_pipes_automatically` from quadratic to linear time, as requested in https://github.com/tqec/tqec/pull/978#discussion_r3516844477. That update is a dependency inherited from the parent branch rather than a change this PR introduces; the parent branch will be merged ASAP after approval.

To keep this PR free of any unreleased or git-pinned dependency, SI1000 gate-set transpilation is included only as a `NotImplementedError` stub (I have a working `stimflow`-backed implementation on a local branch), and the `stimflow` git dependency has been dropped. Every other noise model works unchanged.

### Why batch processing?

The most immediate use case is a API for gadget-distance experiments coming soon to `tqecd`. A gadget is a block graph with logical observables which must be fault tolerant. The intended workflow is that a developer produces a single DAE file or block graph containing multiple disjoint block graphs, each containing a questionable block. This PR exposes `tqec` to recognize that input, break it into connected components, and optionally convert the file type of each component. The upcoming `tqecd` gadget-distance interface would then find and match the flows in each gadget according to some experimental configuration, give them to `sinter` for simulation, and feed the results back into `tqecd` gadget testing logs. That `tqecd` integration are an intended workflow that will be in a sibling PR. When that's done, I will stack demos and documentation on top of this branch. In the meantime, I will keep this as a PR draft just to keep things synced.

`orchestration` makes it marginally easier to input and process data with `tqec`. There are ways to make batch processing more broadly relevant; this is just intended to be a foundation. 

Benchmarking whether every block can compile (note that this is slightly different from gadget testing, which checks that a block graph compiles into a circuit with a fault-tolerant threshold) can be done by batch processing the `template.dae` asset as described here:

> One could run [[this batch parsing](https://github.com/KabirDubey/tqec/blob/dae-component-reproduction/_reproduction/inputs/parser.py)](https://github.com/KabirDubey/tqec/blob/dae-component-reproduction/_reproduction/inputs/parser.py) script over the template file and benchmark their tqec system configuration over all computational primitives in a single workflow. Automatic port filling should cover all deterministic fillings of the pipes. I think the port cube would need to get programmatically skipped.

[[source](https://github.com/tqec/tqec/pull/815#issuecomment-4793428204)](https://github.com/tqec/tqec/pull/815#issuecomment-4793428204)

Likewise, one can benchmark compiler performance #934 and data flow by batch processing multiple stressful block graphs @smburdick. 

Beyond benchmarking, this could also be useful for supporting multiprogramming circuit generation. We already support streamed circuits, so I don't think it would be much additional work to support parallel streams on top of this @BluSteve. Could be of interested to those working on upstream compilation @jbolns @ioanamoflic.

### How has this been tested?

I confirmed that @ica574's modifications fixed the error described in #967. @ica574 please comment on this PR and I can add you as a reviewer. I also added functionality tests for block graph partitioning. Because the ordering between automatic pipe insertion and partitioning is non-obvious, I documented the connectivity/grouping contract--that `add_pipes_automatically` defines gadget membership and must run before partitioning, and that the DAE route groups by touching bounding boxes--in the `partition` and batch-module docstrings, with a canonical gap-then-connect-then-split workflow example, and added adversarial tests for both routes. The orchestration layer has its own tests, including for the usage of `sinter.collect` and an import-benchmark over `template.dae` described below. The full test suite passes locally, and `ruff check` / `ruff format` are clean across `src/tqec` and `tests`.

Since the main usage of this API will make parsing failures obvious, I was comfortable using Claude to write a lot of this code. Most of the changes were pair-programmed with Claude Sonnet in an ask-before-editing workflow. I reviewed everything. I checked the generated changes as I gradually increased the set of test gadgets for the Y half-cube. 

# Checklist

*   [x] My code follows the style guidelines of this project
*   [x] I have performed a self-review of my own code
*   [x] I have commented my code, particularly in hard-to-understand areas
*   [x] I have made corresponding changes to the documentation
*   [ ] My changes generate no new errors or warnings

There appear to be ~22,000 warnings .. idt they're from this PR but instead inherited issues like the numpy 2.5 deprecation type stuff.   

*   [x] My code and tests pass locally and have at least 80% line coverage
*   [ ] Any dependent changes have been merged and published in downstream modules
*   [x] All AI-generated changes have been thoroughly human-reviewed and checked


jbolns:
Quite cool! At the very least, I like the idea. 

Question about potential runtime tradeoffs.  Abstract version. My "CPU core usage" panel already blows up with a single BlockGraph, so I imagine something needs to give for batch processing to be possible? Practical version. Is batch processing going to significantly augment testing CI/CD run-times?

> Could be of interested to those working on upstream compilation

Topologiq is never going to give you a disconnected BlockGraph in the same way a DAE comes out when there are disconnected groups of blocks in SketchUp. However, it could give you several BGRAPHs each corresponding to a different part of a computation, i.e., to easily test small parts of your circuit design separately rather than all at once. Would require some changes to BGRAPH loader, though, to enable multiple file consumption.

src/tqec/computation/block_graph.py
```diff
@@ -243,6 +250,24 @@ def add_pipe(
...
+        Runs in time linear in the number of cubes: a cube has at most six lattice
+        neighbours, and only the three positive directions are visited so that each
+        adjacent pair is considered exactly once.
+        """
+        for position in self.occupied_positions:
+            for direction in Direction3D.all_directions():
+                neighbour = position.shift_in_direction(direction, 1)
+                if neighbour in self and not self.has_pipe_between(position, neighbour):
+                    self.add_pipe(position, neighbour)
+
```
jbolns, line 270: Might just need to draw this out but, does this suffice for cubes at the lower extremes of the BlockGraph? 
If it does, then that's that. 
If it doesn't, perhaps `src/tqec/computation/cube.py:75` (`ZXCube::normal_basis`) could be introduced between L265 and L266 to visit positive and negative directions but limit visitations to only the 4 viable directions. Adds one check per position, but is thorough. Additionally, while I'm not sure the maths of an extra check inside the nested loop add up, the extra iteration could potentially be offset with a `continue` if the cube's degree is reached and no more connections are therefore needed for cube.

docs/architecture.rst
```diff
@@ -39,7 +39,7 @@ the order in which the sub-modules will be covered here.
         C[tqec.templates] --> E
         D[tqec.plaquette] --> E
         E --> F[tqec.circuit]
-        F --> G[tqec.utils.noise_models]
+        F --> G[tqec.utils.noise_model]
```
nelimee, line 42: Could be done in another PR as that has nothing to do with the current advertised changes of the PR.

src/tqec/computation/block_graph.py
```diff
@@ -243,6 +250,24 @@ def add_pipe(
...
+        Runs in time linear in the number of cubes: a cube has at most six lattice
+        neighbours, and only the three positive directions are visited so that each
+        adjacent pair is considered exactly once.
+        """
+        for position in self.occupied_positions:
+            for direction in Direction3D.all_directions():
+                neighbour = position.shift_in_direction(direction, 1)
+                if neighbour in self and not self.has_pipe_between(position, neighbour):
+                    self.add_pipe(position, neighbour)
```
nelimee, line 269: Why would anyone need to call such a method? In my opinion, you should construct your blockgraph correctly in the first place. I really don't like this method as:

1. It "addresses" a problem that should not exist in the first place.
2. It encourages users to be lazy and construct their blockgraph with only cubes...
3. ... but by construction this method is very error-prone because it will link **any** cube that is neighbouring with another one...
4. ... and I feel like it will be very hard for a user that only constructs blockgraphs with only cubes to ensure that only the cubes they want linked are actually neighbours.

TLDR: in my opinion, this solves a problem that does not exist and does so by introducing a method that will seem handy to user, but comes with pre-conditions that are actually hard for them to check.

Did I miss a use-case where the above is wrong?

src/tqec/computation/block_graph.py
```diff
@@ -665,6 +690,65 @@ def compose(self, other: BlockGraph, self_port: str, other_port: str) -> BlockGr
...
+        and only if the current pipes connect them. There is no separate grouping signal.
+        The consequences are worth stating explicitly:
+
+        - :py:meth:`add_pipes_automatically` defines membership and must be called *before*
+          partitioning if lattice-adjacent cubes are meant to be grouped. Splitting a
+          node-only graph first yields one component per cube.
+        - Because :py:meth:`add_pipes_automatically` connects *every* 3d-lattice-adjacent
+          compatible pair, it can merge two gadgets that were meant to stay separate but
+          happen to sit next to each other. Once merged, the original boundary cannot be
+          recovered from the graph.
```
nelimee, line 711: This reads as "please use add_pipes_automatically before calling", except if you read very carefully everything. I don't really like it, because of my comment on this method.

src/tqec/computation/open_graph.py
nelimee, line 1: I feel like the change in this file would be better in its own PR as it seems to be only tangentially related to the changes advertised in this (quite long) PR.

src/tqec/interop/collada/read_write.py
nelimee, line 1: Same comment: are these changes really related to the PR, or rather orthogonal and so probably better in their own PR? It feels like this is:
1. adding support for the Y cube
2. Fixing a bug when calling numpy (changing the axis, why?)

but I don't see the direct link with batching.

src/tqec/interop/converters.py
nelimee, line 1: Same comment here: new PR for that, as it is not directly related to batching as far as I can tell.

src/tqec/interop/shared.py
nelimee, line 1: New PR for Y-half cube? Or is that related to batching and I don't see it?

src/tqec/orchestration/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+)
+from tqec.orchestration.schema import (
+    ResolvedLogicalObservable as ResolvedLogicalObservable,
+)
+from tqec.orchestration.schema import (
+    UnitResult as UnitResult,
+)
+from tqec.orchestration.schema import (
+    UnitStatus as UnitStatus,
+)
```
nelimee, line 49: Don't need the `as ...` part when you have a `__all__`, you can remove it.

src/tqec/orchestration/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+]
+
+
+def __getattr__(name: str) -> Any:
+    """Lazily resolve :func:`simulate_batch` so importing the package stays sinter-free."""
+    if name == "simulate_batch":
+        from tqec.orchestration.simulate import simulate_batch  # noqa: PLC0415
+
+        return simulate_batch
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
nelimee, line 76: That's cool, could probably be re-used in other places eventually.

src/tqec/utils/noise_transpilation.py
nelimee, line 1: Probably another new PR.

src/tqec/utils/rotations.py
```diff
@@ -251,7 +251,7 @@ def rotate_on_import(
     kind = rotate_block_kind_by_matrix(kind, rotation_matrix)
 
     # Shift nodes slightly according to rotation
-    translation = FloatPosition3D(*translation_matrix + rotation_matrix.dot(scale_matrix))
+    translation = FloatPosition3D(*translation_matrix)
```
nelimee, line 254: Not sure about that change. Seems to be related to Y cube, but if not, could you explain why this change?

src/tqec/__init__.py
```diff
@@ -3,3 +3,4 @@
 from .computation.block_graph import BlockGraph as BlockGraph
 from .utils import Basis as Basis
 from .utils import NoiseModel as NoiseModel
+from .utils import transpile_to_si1000_gateset as transpile_to_si1000_gateset
```
nelimee, line 6: This is a `NotImplemented` function, don't add it to the public API until it at least has a default implementation.

nelimee (commented):
General comment: this should be split into 3 or 4 different PRs. Keep the changes as small as possible. This PR is too large for a review as it is. Rule of thumb (which I did not respect sometime, I know...) is < 1000 lines of diff, except if you have a real justification.

The new PRs that should likely be opened to reduce this one:
- Y cube support
- Changes to the `sinter`-related API
- noise transpilation

I may have missed some here, feel free to revise the list above if you think things can be grouped in a better way :)

Issue #578, opened 2025-05-08
Fix extended stablizer measurement implementation


nelimee:
## Describe the bug

The extended stablizers implementation from the review:

![Image](https://github.com/user-attachments/assets/2b1b29fc-88b4-4ede-a485-c9ca9b94c303)

have an issue with scheduling: reversed plaquettes clash with the other regular 5-steps stablizer measurement schedules.

I think that happens because reversing the schedule of 2-qubit gates is the same as performing a central symmetry on the plaquettes (i.e., operations on the top-left data-qubit are performed on the bottom-right one for the reversed variant, same for the 3 other corners, and operations on the ancilla are left unchanged), but the reversed implementation in the screen above is not the central symmetric of the forward version.

The current implementation simply takes the forward implementation (the top one) and applies a central symmetry to obtain:

![Image](https://github.com/user-attachments/assets/6e66f627-fca6-4c12-bfa6-6a51c1dd14da)

The problem being that the last `Mx` measurement cannot be "absorbed" in the next plaquette layer anymore, making the extended-stablizers one timestep longer.

## Steps to reproduce the behavior

The drawing below should illustrate the problem with the initial schedule. Schedules for regular (non-extended) stablizers are taken from the review, schedules for the extended stablizers are computed from the implementation above.

![Image](https://github.com/user-attachments/assets/66175364-b205-420f-b0fe-3e6d37e2d247)

## Expected behavior

An extended stablizer plaquette that can be scheduled in both forward and reverse without any schedule clash and that takes the minimum number of timesteps (ideally 6, like the initial proposal).

Note that shrinking down to 6 timesteps *might* not be possible. Any idea is welcomed.

Zhaoyilunnn:
Hi, I can have a try on this issue. Please feel free to take over me if my progress is too slow

nelimee:
> Hi, I can have a try on this issue. (Anybody) Please feel free to take over me if my progress is too slow

Yes go ahead! I wanted to have a try at implementing something, but I would need to test it and I do not really have the time right now and I need to take some to visualise the graph resulting from alternating stabilisers.

I'll assign you, please keep us updated :)

Zhaoyilunnn:
To make sure that I understand this issue correctly,  I'd like to confirm some details

> reversing the schedule of 2-qubit gates is the same as performing a central symmetry on the plaquettes

[...]

PR #1029, opened 2026-08-13
feat: add horizontal extended stabilisers

sdoygb:
## Summary

Implements **horizontal extended stabilisers** (LEFT/RIGHT), completing the existing vertical UP/DOWN support. This is the primitive needed for the x-direction spatial Hadamard gates in #838, as discussed in #839.

## Changes

- `ExtendedPlaquettePosition` gains `LEFT` / `RIGHT` (with `flip()` support)
- `ExtendedPlaquetteDrawer` transposes the vertical geometry inside the drawing domain (T(x, y) = (y/2, 2x)) for horizontal positions: square shapes, weight-three triangles, interaction-order texts, hook-error lines, reset/measurement layers
- `get_horizontal_extended_plaquette` builds the (LEFT, RIGHT) pair from the left/right columns of the RPNG description
- `ExtendedPlaquetteCollection.from_description` / `from_basis` accept `orientation="UP_DOWN" | "LEFT_RIGHT"` (default unchanged, backward compatible)
- 15 new tests — 47 total, all green

## Layout

```
UP (vertical, data on top)         LEFT (horizontal, data on left)
  dl ---- dr                         dl ---- s2
  |        |                         |        |
  |   s1   |                         |   s1   |
  |        |                         |        |
  s2 ---- s2                         dr ---- s2

DOWN (vertical, data on bottom)    RIGHT (horizontal, data on right)
  s2 ---- s2                         s2 ---- dl
  |        |                         |        |
  |   s1   |                         |   s1   |
  |        |                         |        |
  dl ---- dr                         s2 ---- dr
```

## Notable property

The horizontal stabilizer flow is **independent of `is_reversed`**: both directions yield `1 -> XX___XX xor rec[0]` (the 4 data-qubit X product), because the LEFT/RIGHT data-qubit groups each stay within their own column under the schedule flip. The vertical UP/DOWN flow, in contrast, permutes which syndrome qubit is measured. This makes x-direction scheduling structurally simpler than y-direction scheduling.

## Testing

- Stabilizer flows: X/Z bases x reversed/non-reversed x both flow directions (8 assertions)
- Qubit layouts: data-qubits on the left/right columns, shared s2 ancilla
- Drawer positions for all 7 plaquette types; drawing does not raise
- Transposed-geometry checks for square and weight-three shapes
- `ruff check` clean

Issue #839, opened 2026-02-12
Implement new horizontal extended stabilisers in the fixed bulk convention


BSchelpe:
The existing extended stabilisers have vertical long directions. This is needed for pipes in the y-direction. For pipes in the x-direction new extended stabilisers are needed, with the long direction in the x-direction (horizontal). Both rectangular and triangular extended stabilisers are needed.

sdoygb:
Hi, I would like to implement this. I have been studying the existing vertical extended stabilisers (`_make_spatial_cube_arm_memory_plaquette_up/down`, `ExtendedPlaquetteCollection` shapes in `extended_stabilizers.py`) and the X-axis pipe path in `fixed_bulk.py` (`get_memory_vertical_boundary_*`).

Plan: add horizontal (x-long) variants of the rectangular and triangular extended stabilisers, wire them into the vertical-boundary (X-pipe) path, and extend the tests in `tests/compile/specs/library/generators/extended_stabilizers_test.py`.

Open question: would you prefer a separate LEFT/RIGHT position pair in `ExtendedPlaquettePosition`, or a rotation/axis parameter shared with the vertical implementations? Happy to follow the maintainers' preference.

BSchelpe:
Thanks for looking at this! I have no preference a priori.  Generally in the code base, there is some duplication of code, rather than reusing functions, where it makes the code easier to read, but if it just as legible to reuse with a rotation parameter, then that would be good. I will look at your PR to understand your design choices, and then I might have a view!

PR #1038, opened 2026-08-22
Issue578 extended stabilizers

doczenwiry:
# Description

I've implemented the extended stabilizers in a different way to enable a complete meshing as originally described in the review paper. The idea is that the forward direction operates as documented in the review paper and the reverse direction simply flips all the schedules along the vertical axis of the plaquettes. Using the (TL, TR, BL, BR) convention, the schedules are transformed as follows:

- (1, 2, 3, 5) <-> (2, 1, 5, 3)
- (1, 4, 3, 5) <-> (4, 1, 5, 3)
- (2, 3, 4, 5) <-> (3, 2, 5, 4)

This new implementation exists side-by-side with the current implementation and a new convention has been added called "fixed_boundary_flipping".

Partially adresses issue #578.

# How Has This Been Tested?

Originally, by adapting the run-example subcommand to use a spatial junction computation with red/blue tips to generate both cases of the layout (with complete plaquettes or trimmed plaquettes in the corners)

Following my addition of the TQEC simulate subcommand (see PR #1037), I've used the commands

$ tqec simulate -c fixed_boundary -k 2 3 4 5 -s 1e8 -- assets/spatial-junction-x-basis-blue-boundaries.dae
$ tqec simulate -c fixed_boundary_flipping -k 2 3 4 5 -s 1e8 -- assets/spatial-junction-x-basis-blue-boundaries.dae

And compared visually the graphs to see that the new implementation falls in same ballpark as the existing one.

spatial-junction-x-basis-blue-boundaries_fixed_boundary_results_observable_0
spatial-junction-x-basis-blue-boundaries_fixed_boundary_flipping_results_observable_0

This implementation lays out the gates in a way that can be fully meshed to achieve its full potential. However, TQECD is unable to find all the detectors when there is an RX/MX at the same moment which limits my ability to thoroughly simulate this construction. The performance of this new implementation should improve once PR [tqec/tqecd#69](https://github.com/tqec/tqecd/pull/69) is completed. I have manually completed the annotation for distance d=5 and observed a very slight improvement on the performance but not significant enough to be conclusive (see https://github.com/tqec/tqec/issues/578#issuecomment-5326419315).

**Test Configuration**:
* Python version: 3.13
* Operating system and system architecture: MacOS (Apple MacMini M2)

# Checklist:

- The existing tests all run successfully

### DIFF
--- a/src/tqec/compile/convention.py
+++ b/src/tqec/compile/convention.py
@@ -9,7 +9,9 @@ from tqec.compile.observables.fixed_boundary_builder import (
 from tqec.compile.observables.fixed_bulk_builder import FIXED_BULK_OBSERVABLE_BUILDER
 from tqec.compile.specs.library.fixed_boundary import (
     FIXED_BOUNDARY_CUBE_BUILDER,
+    FIXED_BOUNDARY_CUBE_BUILDER_FLIPPING,
     FIXED_BOUNDARY_PIPE_BUILDER,
+    FIXED_BOUNDARY_PIPE_BUILDER_FLIPPING,
 )
 from tqec.compile.specs.library.fixed_bulk import (
     FIXED_BULK_CUBE_BUILDER,
@@ -63,4 +65,19 @@ FIXED_BOUNDARY_CONVENTION = Convention(
     ),
 )
 
-ALL_CONVENTIONS = {conv.name: conv for conv in [FIXED_BULK_CONVENTION, FIXED_BOUNDARY_CONVENTION]}
+FIXED_BOUNDARY_FLIPPING_CONVENTION = Convention(
+    "fixed_boundary_flipping",
+    ConventionTriplet(
+        FIXED_BOUNDARY_CUBE_BUILDER_FLIPPING,
+        FIXED_BOUNDARY_PIPE_BUILDER_FLIPPING,
+        FIXED_BOUNDARY_OBSERVABLE_BUILDER,
+    ),
+)
+ALL_CONVENTIONS = {
+    conv.name: conv
+    for conv in [
+        FIXED_BULK_CONVENTION,
+        FIXED_BOUNDARY_CONVENTION,
+        FIXED_BOUNDARY_FLIPPING_CONVENTION,
+    ]
+}
--- a/src/tqec/compile/specs/library/fixed_boundary.py
+++ b/src/tqec/compile/specs/library/fixed_boundary.py
@@ -110,7 +110,10 @@ def _get_block(
 
 class FixedBoundaryCubeBuilder(CubeBuilder):
     def __init__(
-        self, compiler: PlaquetteCompiler, translator: RPNGTranslator = DefaultRPNGTranslator()
+        self,
+        compiler: PlaquetteCompiler,
+        translator: RPNGTranslator = DefaultRPNGTranslator(),
+        flipping: bool = True,
     ) -> None:
         """Implement the :class:`.CubeBuilder` interface for the fixed boundary convention.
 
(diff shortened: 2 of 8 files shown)


sdoygb:
Hi @doczenwiry, thanks for this PR! I noticed it overlaps with my own open PR #1029 ("feat: add horizontal extended stabilisers"), which implements the LEFT/RIGHT orientations requested in #839. Since both touch `extended_stabilizers.py` and its test file, I wanted to lay out how the two relate and propose a coordination path before either merges.

**They address orthogonal concerns:**

- **#1038** changes the *schedule semantics* for reversed plaquettes: instead of full time-reversal (legacy `(5,3,4,2)`), the flipping schedules `(4,2,5,3)` mirror moments along the vertical axis of the plaquette. This is the fix for the #578 scheduling clash.
- **#1029** changes the *spatial orientation*: it adds LEFT/RIGHT positions on top of the existing UP/DOWN, transposing the vertical geometry into the horizontal domain via `T(x, y) = (y/2, 2x)`, with 15 new tests locking the transposition contract.

**Merge conflicts to expect (regardless of order):**

1. `EXTENDED_PLAQUETTE_SCHEDULES` / `VERTICAL_HOOK_SCHEDULES` / `HORIZONTAL_HOOK_SCHEDULES` go from `dict[bool, tuple]` to the nested `dict[bool, dict[bool, tuple]]` in #1038. My #1029 tests still index the flat structure (`EXTENDED_PLAQUETTE_SCHEDULES[is_reversed]`), so the nesting will need to be threaded through whichever PR lands second.
2. Both PRs extend the plaquette constructor signature: #1038 adds `is_flipping`, while #1029 merges the four `_make_spatial_cube_arm_memory_plaquette_{up,down,left,right}` functions into a single `_make_spatial_cube_arm_memory_plaquette(direction, ...)` with an `Orientation` enum (UP/DOWN/LEFT/RIGHT). The combined signature would naturally be `(direction, is_reversed, is_flipping)`.

**Proposed coordination:**

- Conceptually, flipping is a mirror along the plaquette's *long axis*: vertical for UP/DOWN, horizontal for LEFT/RIGHT. Under the #1029 transposition, a flipping schedule for a horizontal plaquette is exactly the transpose of the vertical one — so the two mechanisms compose cleanly rather than fighting.
- I'm happy to rebase #1029 on top of whichever lands first, and to align the schedule-table structure now so the merge is a no-op rather than a rewrite. If useful, I can prepare a small follow-up on your branch that extends the flipping schedules to the horizontal orientations using the #1029 transposition.

Curious what you think — happy to help with the merge plumbing either way.

doczenwiry:
That is a very thorough and well articulated feedback, thank you !

In order to make sure I understand where our PRs differ, would it be fair to contrast them as follows:
- PR#1029 deals with rotating counter-clockwise the entire layout of plaquettes by pi/2 and thus having the schedules of the plaquettes rotated by pi/2 as part of this
- PR#1038 deals with reflecting the schedules along the vertical axis when the extended plaquettes are stretched vertically

So the separate contributions are rotation by pi/2 from your work and vertical reflection from mine ?

I expect the conflict for the *_SCHEDULES dictionaries to be straightforward either way. Adapting the existing tests when I altered those was semi-automatic.

Your merge of the _make_spatial_cube_arm_memory_plaquette_*(..) functions taking a direction argument seems like the proper way to organise that part of the code to me. I was only reluctant to do it for fear that the logic of a single function dealing with all those cases could become tedious to read and maintain.

I would feel better to stick to the FIFO here; I'll add my changes once yours have been merged. It'll give me the opportunity to expand my changes and test for the horizontal case as that is something my PR ought to deal with in order to be complete :)

Issue #840, opened 2026-02-12
Keep track of hadamard arms in fixed bulk convention


BSchelpe:
Previously the code for spatial junctions applied the 'corner' plaquettes (ie plaquettes in the corner between arms) were iplemented twice - once by the left-right arm, once by the up-down arm. This didn't matter as what the two arms were implementing were the same. Now that an up-down hadamard is implemented (in PR #774 ) this is no longer true: the up-down arm would add a split colour extended stabiliser triangle, whereas the left-right arm would add a 'normal' triangle, because the left-right arm doesn't know that there is a hadamard on the up-down arm. The way #774 gets around this is simply to stop left-right arms adding corner plaquettes. This will no longer be an ok solution once left-right hadamards are implemented. 

I see the cubes keeping track of which of their arms have hadamards (like they currently keep track of the directions) as the only solution. However, I am concerned that this may slow the code down as it will have to load the arms and look at them, which will be more cumbersome than the current code, so alternatives ideas also welcomed.

sdoygb:
Hi — I'd like to take this one on, together with #837 (regular-pipes spatial Hadamard), since they form a single dependency chain and I'm already working in this exact area: my open PR #1029 adds the horizontal (LEFT/RIGHT) extended stabilisers that the Hadamard arms will need.

**Why #840 first**

The corner-plaquette conflict described here is real on main right now: `get_spatial_cube_arm_plaquettes` dispatches on `is_hadamard` to `_get_left_right_spatial_hadamard_cube_arm_plaquettes` vs `_get_up_down_spatial_hadamard_cube_arm_plaquettes` (fixed_bulk.py), and the two arms of a junction currently don't know about each other's Hadamard status — exactly the duplicated-corner problem in the issue. Once left-right Hadamards land (post-#837), the "just drop the left-right corner plaquettes" workaround from #774 will break, so tracking Hadamard arms on the cube is the right structural fix.

**Proposed approach**

- Add a Hadamard-arm record alongside the existing direction tracking on the spatial cube (following the issue's suggestion: cubes keep track of which of their arms have Hadamards, like they already track directions).
- I'd keep the lookup cheap — a per-cube `frozenset`/flag set computed once at construction time, rather than re-reading arms during plaquette generation, to address the performance concern raised in the issue.
[...]

Issue #837, opened 2026-02-12
Implement spatial hadamards on regular pipes in the fixed bulk convention


BSchelpe:
Horizontal and vertical hadamard methods in the fixed bulk convention are not currently implemented. These are needed for hadamard junctions on regular spatial pipes (ie spatial pipes which connect temporal cubes).

It would make sense for this not to be implemented until PR #774 is merged, as it will touch much of the same code, and in fact I believe can copy much of the implementation of the up-down hadamard in #774 , as both require extended stabilisers.

sdoygb:
Hi — I'd like to implement this one. Per #840 (which I'm also taking on, with arm tracking landing first), the dependency is now unlocked: #774 has been merged, so the up-down Hadamard implementation is available as the blueprint the issue anticipated.

**Current state on main (checked today)**

The `FixedBulkConventionGenerator` already exposes the skeletons we need, but most are stubs:

- `get_spatial_vertical_hadamard_raw_template` → `raise NotImplementedError` (fixed_bulk.py:1360)
- `get_spatial_vertical_hadamard_rpng_descriptions` → stub (line 1408)
- `get_spatial_horizontal_hadamard_rpng_descriptions` → stub (line 1511)
- `get_spatial_horizontal_hadamard_raw_template` + `get_spatial_extended_stabiliser_hadamard_*` → implemented
- Also `get_spatial_cube_arm_plaquettes` already dispatches on `is_hadamard` to separate left-right vs up-down Hadamard paths

So this issue is precisely: turn the vertical/horizontal stubs into real implementations, mirroring the #774 pattern.

**Implementation plan**

1. **Reuse #774's structure**: the merged up-down Hadamard (`_get_up_down_spatial_hadamard_cube_arm_plaquettes` and its extended-stabiliser variant) defines the template + schedule pattern; regular pipes are simpler than junctions (no corners), so the raw template can likely be shared with the existing `get_memory_{vertical,horizontal}_boundary_raw_template` and only the RPNG descriptions need Hadamard-specific treatment.
2. **Extended stabilisers**: this is where my PR #1029 comes in — the LEFT/RIGHT (horizontal) extended stabilisers it adds are exactly the primitives needed for the horizontal Hadamard path; I'll rebase #1029 as needed so this builds on it cleanly.
3. **Corner consistency**: build on the arm-tracking from #840 so corner plaquettes (split-colour vs normal triangles) are emitted from the cube's Hadamard-arm record rather than duplicated per arm.
[...]

PR #991, opened 2026-07-04
Refactor spatial hadamard code

BSchelpe:
# Description
This refactors the recently pushed spatial hadamard code, as suggested by @smburdick in #774. 

It closes one half of issue #916.  This does the refactoring. Adding extra tests remains to be done. 

# How Has This Been Tested?

It has been tested by running the hadamard snippet of code given on issue #657.  

# Checklist:

* &nbsp; [x] My code follows the style guidelines of this project
* &nbsp; [x] I have performed a self-review of my own code
* &nbsp; [x] I have commented my code, particularly in hard-to-understand areas
* &nbsp; [N/A ] I have made corresponding changes to the documentation. No changes are needed.
* &nbsp; [x] My changes generate no new errors or warnings
* &nbsp; [ ] My code and tests pass locally and have at least 80% line coverage
* &nbsp; [x] Any dependent changes have been merged and published in downstream modules
* &nbsp; [N/A ] All AI-generated changes have been thoroughly human reviewed and checked. AI has had no part in this code change.

PR #1042, opened 2026-08-26
feat: implement Y-axis regular-pipe spatial Hadamard in fixed bulk (#837)

sdoygb:
## Summary

Implements the **Y-axis (horizontal) regular-pipe spatial Hadamard** in the fixed-bulk convention from #837, by reusing the extended-stabiliser spatial Hadamard introduced in #774.

**Important scope clarification**: the X-axis (vertical) case is **not** part of this PR. As documented in #839, X-direction pipes need *horizontal* (LEFT/RIGHT) extended stabilisers, which land in #1029. Until then the X-axis path stays `NotImplementedError` (the compile test documents this dependency).

## Changes

- `FixedBulkPipeBuilder._get_spatial_regular_pipe_plaquettes_factory`: the `(Direction3D.Y, has_hadamard=True)` branch now routes to `get_spatial_extended_stabiliser_hadamard_plqts` with the no-arm configuration (`NRNR` for `sbb=Z`, `NLNL` for `sbb=X`), resolving to the left/right half-rectangle extended stabilisers on the `QubitHorizontalBorders` template.
- `test_compile_spatial_hadamard_vertical_correlation_surface`: the fixed-bulk Y-direction case now compiles for real (was `NotImplementedError`); the X-direction case keeps `NotImplementedError` with a comment pointing at #1029.
- 3 new generator tests for the extended-stabiliser reuse.

## Known limitation (documented)

The generated circuit is **deterministic** (the detector error model is well-defined, 88 detectors at k=1), but the code distance is reduced to d=2 instead of d=3 due to hook-error orientation in the Hadamard region. Distance optimisation (e.g. alternating reversed schedules, as discussed in #774) is left as follow-up work — this PR focuses on the generator/compilation wiring.

## Context

This is the first half of #837. The X-axis half depends on #1029 (horizontal extended stabilisers); the remaining distance optimisation is independent follow-up. #840 (Hadamard-arm tracking, PR #1041) is the other prerequisite of the full spatial-Hadamard family.

## How Has This Been Tested?

- 3 new generator unit tests (all passing).
- Full compile suite: **352 passed, 0 failed** (`uv run pytest tests/compile -m "not slow"`).
- `ruff check` clean on all touched files.

### DIFF
--- a/src/tqec/compile/specs/library/fixed_bulk.py
+++ b/src/tqec/compile/specs/library/fixed_bulk.py
@@ -14,7 +14,7 @@ from tqec.compile.specs.base import (
     PipeBuilder,
     PipeSpec,
 )
-from tqec.compile.specs.enums import SpatialArms
+from tqec.compile.specs.enums import PipeCubeArmConfig, SpatialArms
 from tqec.compile.specs.library.generators.fixed_bulk import (
     FixedBulkConventionGenerator,
 )
@@ -331,9 +331,18 @@ class FixedBulkPipeBuilder(PipeBuilder):
 
             case Direction3D.Y, True:
                 # Hadamard pipe in the Y direction.
-                top_left_basis = spec.pipe_kind.get_basis_along(Direction3D.X, at_head=True)
-                return lambda r, m: self._generator.get_spatial_horizontal_hadamard_plaquettes(
-                    top_left_basis == Basis.Z, r, m
+                # Reuse the extended-stabiliser spatial Hadamard introduced in
+                # #774: for a regular pipe (no spatial cube at either end) the
+                # arm configuration is NRNR (sbb=Z) or NLNL (sbb=X), which
+                # resolves to the left/right half-rectangle extended
+                # stabilisers on the QubitHorizontalBorders template.
+                sbb = spec.pipe_kind.get_basis_along(Direction3D.X, at_head=True)
+                assert sbb is not None
+                arms_parameter = (
+                    PipeCubeArmConfig.NRNR if sbb == Basis.Z else PipeCubeArmConfig.NLNL
+                )
+                return lambda r, m: self._generator.get_spatial_extended_stabiliser_hadamard_plqts(
+                    sbb, arms_parameter, r, m
                 )
             case _:
                 raise TQECError("Spatial pipes cannot have a direction equal to Direction3D.Z.")
--- a/src/tqec/compile/specs/library/generators/fixed_bulk.py
+++ b/src/tqec/compile/specs/library/generators/fixed_bulk.py
@@ -1273,7 +1273,7 @@ class FixedBulkConventionGenerator:
             raise NotImplementedError(
                 "This spatial boundary basis (neither X nor Z) is not supported."
             )
-        return self.get_spatial_extended_stabiliser_hadamard_plqts(
+        return self.get_spatial_horizontal_hadamard_plaquettes(
             _sbb, arms_parameter, reset, measurement
         )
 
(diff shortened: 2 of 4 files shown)


sdoygb:
Addressed both automated-review notes (greptile P1/P2) — summary for maintainers:

**P1 "Duplicate bottom-corner ownership" — assessed as a false positive.** The 7/8 bottom-corner plaquettes come from `get_spatial_extended_stabiliser_hadamard_plqts`, which already landed on main via #774 (and was refactored in #991); this PR only calls it for the Y-direction pipe. Empirically, the generated circuit compiles cleanly (no qubit-merge conflict), decodes to a well-defined DEM, and `shortest_graphlike_error` returns a concrete logical error of length 2 — none of which would hold if shared data qubits were double-owned and the logical state corrupted. The same class of P1 finding on #1029 was also a false positive (transposition, not wrong boundary).

**P2 "distance regression passes CI unnoticed" — fixed.** The Y-axis branch called `generate_circuit_and_assert` without `expected_distance`, so the known distance reduction (d=2 vs ideal 2k+1=3, from hook-error orientation) was not locked in. New commit `79685b03` asserts `expected_distance=2` on this branch head, so any structural regression (compile failure, distance < 2) now fails CI. Distance optimisation (towards 3) is tracked as follow-up work in the PR description.

src/tqec/compile/specs/library/fixed_bulk.py
```diff
@@ -331,9 +331,18 @@ def _get_spatial_regular_pipe_plaquettes_factory(
 
             case Direction3D.Y, True:
                 # Hadamard pipe in the Y direction.
-                top_left_basis = spec.pipe_kind.get_basis_along(Direction3D.X, at_head=True)
-                return lambda r, m: self._generator.get_spatial_horizontal_hadamard_plaquettes(
```
BSchelpe, line 335: I think it would be easier to read the code structure, if you kept this function self._generator.get_spatial_horizontal_hadamard_plaquettes(), and put your logic with the reuse of #774's spatial hadamard inside it.

BSchelpe:
Could you explain the purpose of the extra tests added to fixed_bulk_test.py?

sdoygb:
Hi @BSchelpe, thank you for the review! I've addressed both points:

**1. Code structure (refactor done in f101f7b):**
You're right — I've refactored so that `get_spatial_horizontal_hadamard_plaquettes()` is now the entry point for Y-axis (horizontal) regular-pipe spatial Hadamards, and the reuse of the #774 extended-stabiliser implementation lives inside it. The method signature changed from `(top_left_is_z_stabilizer, reset, measurement)` to `(spatial_boundary_basis, arms_parameter, reset, measurement)`, and it now delegates to `get_spatial_extended_stabiliser_hadamard_plqts()` internally. The caller in the pipe-plaquette factory now calls `get_spatial_horizontal_hadamard_plaquettes()` instead of reaching directly for the extended-stabiliser variant. The underlying `get_spatial_horizontal_hadamard_rpng_descriptions()` and `get_spatial_extended_stabiliser_hadamard_plqts()` are preserved as-is.

**2. Purpose of the extra tests in `fixed_bulk_test.py`:**
The 3 new generator tests lock in the design decision that the Y-direction regular-pipe spatial Hadamard correctly reuses the #774 extended-stabiliser implementation:
- `test_spatial_horizontal_hadamard_uses_extended_stabilisers` — verifies the returned plaquette collection has 6 entries and that the raw template is `QubitHorizontalBorders` (the Y-axis pipe template), confirming the extended-stabiliser path is exercised.
- `test_spatial_horizontal_hadamard_reuses_nrnr_for_z_sbb` — verifies that for `sbb=Z` the arm configuration maps to `PipeCubeArmConfig.NRNR` (regular pipe, no arms, RIGHT-side mapping for Z basis).
- The third test covers `sbb=X` → `NLNL` (the LEFT-side mirror).

These tests guard against a future refactor accidentally reverting to the old `top_left_is_z_stabilizer`-based implementation or breaking the arm-configuration mapping.

Let me know if you'd like any further adjustments!

PR #1048, opened 2026-08-29
feat: add CubeColor enum and Cube.colour property (#817)

### LINKED ISSUE #817: Add node colour information on BlockGraph

mhaye9545:
Resolves #817. This PR introduces a \CubeColor\ enumeration (\BLUE\, \RED\, \GREEN\, \WHITE\) and adds a \colour\ property to \Cube\ nodes in the BlockGraph, enabling easy access to block colours without relying on string representations.

### DIFF
--- a/src/tqec/computation/cube.py
+++ b/src/tqec/computation/cube.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from enum import Enum
 from dataclasses import astuple, dataclass
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


smburdick:
I think we discussed this a while back and agreed we weren't adding this since it can be inferred from the ZX cube kind.

